PR #1, opened 2015-11-23
use append([]byte, string) special case

alicebob:
From the specs:

> As a special case, copy also accepts a destination argument assignable to type []byte with a source  argument of a string type. This form copies the bytes from the string into the byte slice.
https://golang.org/ref/spec#Appending_and_copying_slices

Doesn't seem faster or slower, but is nice to use, and makes it easier for the compiler to be smart.

### DIFF
--- a/args.go
+++ b/args.go
@@ -46,7 +46,7 @@ func (a *Args) Len() int {
 
 // Parse parses the given string containing query args.
 func (a *Args) Parse(s string) {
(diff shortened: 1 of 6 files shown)


valyala:
Thanks for the suggestion. Already done in the commit 4aac41a9188840d8db4872a613b51992b887aa1a .

valyala:
Btw, append(byte, string..) is 2x faster than the previous version according to BenchmarkAppendBytesStr.

PR #21, opened 2015-12-20
Redirect 307 instead of 302

buaazp:
Hi @valyala 

When I test my fasthttprouter, I want to redirect a 307 request while you return a 302 response. I don't know what you consider about that. Wish your reply.


valyala:
Thanks, @buaazp .

buaazp:
My pleasure.

PR #24, opened 2015-12-24
100-Continue support for forms

celer:
Hi, I was blocked by lack of 100-Continue support for forms. 

You can reproduce this issue by trying to upload a file using curl, like so:

curl -v -v -X POST -F file=@/etc/hosts http://bear.local:8081/upload.php

> POST /upload.php HTTP/1.1
> User-Agent: curl/7.38.0
> Host: bear.local:8081
> Accept: */*
> Content-Length: 381
> Expect: 100-continue
> Content-Type: multipart/form-data; boundary=------------------------807912e865bb9c45
>
< HTTP/1.1 100 Continue
< HTTP/1.1 200 OK

I've added a handler to Server called On100Continue which lets the user override the behavior when the client expects a 100-Continue. 

Essentially you'd be able to override this function to accept or reject requests with 100-continue.

### DIFF
--- a/http.go
+++ b/http.go
@@ -389,8 +389,8 @@ func (resp *Response) resetSkipHeader() {
 // RemoveMultipartFormFiles or Reset must be called after
 // reading multipart/form-data request in order to delete temporarily
 // uploaded files.
-func (req *Request) Read(r *bufio.Reader, ctx *RequestCtx, on100Continue func(req *Request) bool) error {
-	return req.ReadLimitBody(r, 0, ctx, on100Continue)
+func (req *Request) Read(r *bufio.Reader, w io.Writer, on100Continue func(req *Request) bool) error {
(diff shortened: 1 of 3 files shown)


valyala:
@celer , thanks for the pull request. I just landed the commit 3284c3e6711972e6934d122a10bcaec8bf967026 , which is based on your pull request.

I couldn't merge your pull request without modifications, since it changes public API (Request.Read* methods), which may break third-party code, which uses fasthttp.

celer:
Thanks! I sent you an email about my attempt to add middleware to fasthttp:

https://github.com/celer/fasthttp/commit/2fdada68102340da62146b86fb7e6d301be77e58

valyala:
@celer , I've got your email and need more time to think about your use case. Will reply you by email.

PR #30, opened 2016-01-05
Donot modify content-length when if header set

buaazp:
Hi @valyala 

In my cases, when I send `HEAD` requests, fasthttp use body length to set `Content-Length` header.

But the `HEAD` response doesn't has body. So the `Content-Length` is always 0.

I tested in `net/http`, it doesn't modify this header. So, what do you think about?

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -12,14 +12,6 @@ import (
 	"unsafe"
 )
 
-var gmtLocation = func() *time.Location {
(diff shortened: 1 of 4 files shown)


buaazp:
There are some CI fails, I'm fixing.

valyala:
@buaazp , now fasthttp properly sets `Content-Length` for `HEAD` responses.

PR #31, opened 2016-01-05
Donot modify content-length when if header set

buaazp:
Hi @valyala 

In my cases, when I send `HEAD` requests, fasthttp use body length to set `Content-Length` header.

But the `HEAD` response doesn't has body. So the `Content-Length` is always 0.

I tested in `net/http`, it doesn't modify this header. So, what do you think about?

### DIFF
--- a/http.go
+++ b/http.go
@@ -869,7 +869,9 @@ func (resp *Response) Write(w *bufio.Writer) error {
 		return resp.closeBodyStream()
 	}
 
-	resp.Header.SetContentLength(len(resp.body))
+	if resp.Header.contentLength == 0 {
+		resp.Header.SetContentLength(len(resp.body))
+	}
 	if err = resp.Header.Write(w); err != nil {
 		return err
 	}


buaazp:
In `net/http` I found these:

```
	// ContentLength records the length of the associated content.  The
	// value -1 indicates that the length is unknown.  Unless Request.Method
	// is "HEAD", values >= 0 indicate that the given number of bytes may
	// be read from Body.
```

valyala:
@buaazp , thanks for the pull request. I merged slightly modified version - see https://github.com/valyala/fasthttp/commit/dbe26b3541865876bac5fea07e77fa0183c3b1bb .

buaazp:
Quick reply~

There are other issues about `HEAD` method:

`HEAD` requests should return nil body. `net/http` did this while fasthttp doesn't.

It's a hard work to cover all `HEAD` issues.

valyala:
@buaazp , now HEAD responses don't contain body.

buaazp:
👍

PR #32, opened 2016-01-06
changed code to use []byte/string conversions.

sprstnd:
changed code to use []byte/string conversions where applicable. I believe there is a comment within the source referencing this optimization, one of few optimizations that exist within the gc toolchain. first of probably several commits regarding conversions.

this code accomplishes exactly the same goal as the previous code. it is just faster and clearer.

### DIFF
--- a/header.go
+++ b/header.go
@@ -234,7 +234,7 @@ func (h *ResponseHeader) ContentType() []byte {
 
 // SetContentType sets Content-Type header value.
 func (h *ResponseHeader) SetContentType(contentType string) {
-	h.contentType = append(h.contentType[:0], contentType...)
+	h.contentType = []byte(contentType)
 }
 
 // SetContentTypeBytes sets Content-Type header value.
@@ -249,7 +249,7 @@ func (h *ResponseHeader) Server() []byte {
 
 // SetServer sets Server header value.
 func (h *ResponseHeader) SetServer(server string) {
-	h.server = append(h.server[:0], server...)
+	h.server = []byte(server)
 }
 
 // SetServerBytes sets Server header value.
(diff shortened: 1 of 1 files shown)


valyala:
@sprstnd , your approach results in performance degradation on the following server benchmarks on my laptop:

```
$ git checkout master
$ go test -bench=kServerGet.*10KClients -benchmem -benchtime=5s > old.txt
$ git checkout sprstnd-master
$ go test -bench=kServerGet.*10KClients -benchmem -benchtime=5s > new.txt
$ benchcmp old.txt new.txt
benchmark                                       old ns/op     new ns/op     delta
BenchmarkServerGet1ReqPerConn10KClients-4       1031          1266          +22.79%
BenchmarkServerGet2ReqPerConn10KClients-4       771           997           +29.31%
BenchmarkServerGet10ReqPerConn10KClients-4      488           662           +35.66%
BenchmarkServerGet100ReqPerConn10KClients-4     460           601           +30.65%

benchmark                                       old allocs     new allocs     delta
BenchmarkServerGet1ReqPerConn10KClients-4       0              1              +Inf%
BenchmarkServerGet2ReqPerConn10KClients-4       0              1              +Inf%
BenchmarkServerGet10ReqPerConn10KClients-4      0              1              +Inf%
BenchmarkServerGet100ReqPerConn10KClients-4     0              1              +Inf%

benchmark                                       old bytes     new bytes     delta
BenchmarkServerGet1ReqPerConn10KClients-4       0             17            +Inf%
BenchmarkServerGet2ReqPerConn10KClients-4       0             17            +Inf%
BenchmarkServerGet10ReqPerConn10KClients-4      0             36            +Inf%
BenchmarkServerGet100ReqPerConn10KClients-4     4             27            +575.00%
```

Could you compare fasthttp benchmark results on your computer for the origin branch and for your branch and report results here? Benchmarks can be run with `go test -bench=. -benchmem -benchtime=5s`

sprstnd:
the assembly produced by gc has less instructions and a simple benchmark of just the two methods(the old vs the new) shows improvement by 20%. Both methods produce a []byte and fill it. gc is optimized for conversations.

Im sorry, but I feel your benchmarks are flawed. Accept the changes or don't, doesn't bother me a bit. just thought I'd share some insight.

sprstnd:
flawed benchmarks, clueless lead, not worth the time and hassle.

PR #42, opened 2016-01-28
fix fasthttpadaptor to work with http.ServeMux in Go 1.5

zviadm:
I am not sure if this is the best fix for it, however without it the adaptor will just cause the process to crash.

I couldn't find string version of Path() so []byte to string conversion has to happen here.


valyala:
Yeah, I forgot about Request.URL initialization :) Thanks for the fix, @zviadm .

PR #44, opened 2016-01-30
Readme: Suggest use of Go 1.6 for performance

Freeaqingme:
I was doing some benchmarking, comparing fasthttp and Varnish. Initially fasthttp was 5-10% slower than Varnish was (requesting a single url). However, after switching from Go 1.5 to 1.6 it was consistently 5-10% faster than Varnish. I figured a 15-20% performance improvement warrants a mention in the Readme.

Thank you for this amazing piece of work.


valyala:
Thanks, @Freeaqingme !

PR #50, opened 2016-02-09
Fix parsing of fragments

### LINKED ISSUE #48: URI.UpdateBytes does not appear to be working with fragments
I've been writing a simple web server with fasthttp to deal with some simple logic in how we serve our web apps. And one of the things we need to do is to do a redirect with a fragment into our web app. As in:
https://godoc.org/some#fragment?sss=sds

Now the problem is that it appear that the URI parser code does not appear to handle that scenario. And will simply urlencode the fragment marker (#)

Now if I was using URI directly I could work around it by setting the fragment directly. But since I am using the RequestCtx.RedirectBytes() function it is not really an option. And I presume something like this is should work?

Quick demonstration of the issue:
```
package main

import (
	"fmt"

	"github.com/valyala/fasthttp"
)

func main() {
  uri := fasthttp.URI{}
  uri.Parse(nil, []byte("https://godoc.org"))
  fmt.Printf("Initial: %s\n", uri.FullURI())
  uri.Update("/some#fragment?sss=sds")
  fmt.Printf("Updated: %s\n", uri.FullURI())
}
```

kristofferpeterhansel:
Should fix #48. Since it appears that the URI.parse() function didn't deal with fragments unless there also was a query string

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -203,23 +203,41 @@ func (x *URI) parse(host, uri []byte, h *RequestHeader) {
 	lowercaseBytes(x.host)
 
 	b := uri
-	n := bytes.IndexByte(b, '?')
-	if n < 0 {
+	queryIndex := bytes.IndexByte(b, '?')
+	fragmentIndex := bytes.IndexByte(b, '#')
+	// Ignore query in fragment part
+	if fragmentIndex >= 0 && queryIndex > fragmentIndex {
+		queryIndex = -1
+	}
+
+	if queryIndex < 0 && fragmentIndex < 0 {
 		x.pathOriginal = append(x.pathOriginal, b...)
 		x.path = normalizePath(x.path, b)
 		return
 	}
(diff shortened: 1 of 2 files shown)


valyala:
@kristofferpeterhansel , thanks for the patch!

kristofferpeterhansel:
No problem. Was a good learning experience with debugging Go. And finding out how annoying it is that there is no easy way to run just one test ;)

valyala:
> And finding out how annoying it is that there is no easy way to run just one test ;)

Did you try

```
go test -run=<filter_regexp_here>
```

For instance, if you want to run only tests containing `URI` in their names and see test names, run:

```
go test -run=URI -v
```

kristofferpeterhansel:
Hmmm. I was looking for something like that. But I could only find examples of running specific packages.

But yeah. That is much better. Thanks.

PR #56, opened 2016-02-22
Added BodyInflate and WriteInflate to match BodyGunzip and WriteGunzip

sschepens:
- Extracted some logic from `(Respone) BodyInflate` to WriteInflate replicated from WriteGunzip.
- Added `(Request) BodyInflate`


valyala:
Thanks for the patch!

Issue #86, opened 2016-04-27
ServeFile prepends / on windows


henkman:
I get following error when trying to serve a file using ServeFile:
`2016/04/27 11:05:40 0.001 #0000000100000001 - [::1]:8080<->[::1]:50575 - GET http://localhost:8080/C:%5CProjects%5Cgo%5Csrc%5Ctest%5Cindex.html - cannot open file "/C:\\Projects\\go\\src\\test\\index.html": open /C:\Projects\go\src\test\index.html.fasthttp.gz: The filename, directory name, or volume label syntax is incorrect.`
Looks like it always prepends a slash which on windows does not make a lot of sense.
I use ServeFile like this:
`fasthttp.ServeFile(ctx, "index.html")`

valyala:
I have no access to Windows, so cannot fix the issue at the moment.
Will happily merge a pull request for the issue provided it contains the corresponding tests and it doesn't break `fasthttp` on linux.

kataras:
@henkman You maybe want to check [Iris:ServeFile](https://github.com/kataras/iris/blob/development/context_renderer.go#L227) implementation for ServeFile, which is working both linux & windows. Note that iris is build on top of the fasthttp so it's code can help you with this!

ernado:
I have access to windows machine and will try to debug this issue today.

valyala:
Closing the issue, since it should be fixed now

PR #89, opened 2016-05-04
Fix for issue #86

ernado:
Quick fix for leading slashes on Windows.
Not sure about coverage of probable security issues on that platform.

Also moved client tests that only work on unix to separate file with build tags, tests are passing now on Windows.

For further discussion feel free to contact me at golang-ru (@ernado)


ernado:
Switching to ubuntu for fixing unix part.

kataras:
This was my previous atemp to fix it https://github.com/valyala/fasthttp/pull/85 but with domains, it also has problems with it but I closed my pull request because it wasn't good for all cases, you did it better nice!

ernado:
@kataras 
My commit don't change existing logic for non-windows case, so this PR probably don't fix your issue.

Can you give more details on your issue with subdomains? I.e. what you are doing, what you expect and got instead. 
Thanks :)

client_unix_test.go
```diff
@@ -0,0 +1,44 @@
+// +build darwin dragonfly freebsd linux nacl netbsd openbsd solaris
```
valyala: I'd suggest enumerating platforms without unix socket support here, since this list is shorter.

valyala:
Thanks for the contribution!

kataras:
@ernado I already fixed it , I just don't use the ctx.SetPath, and make the redirect manually. I'm checking this repository everyday if someone post about the issue I had I will add pr a complete solution, thanks for your contribution both of you!

PR #90, opened 2016-05-04
Explicitly skip tests for platforms without unix sockets

ernado:
Comment in #89:
> I'd suggest enumerating platforms without unix socket support here, since this list is shorter.

So I just checked how it is done in go sources:
https://sourcegraph.com/github.com/golang/go@43d2a10e2647b2d6ee10a2a7629ee4055dcce740/-/blob/src/net/platform_test.go#L34
```go
case "unix", "unixgram":
	switch runtime.GOOS {
	case "android", "nacl", "plan9", "windows":
		return false
	}
	// iOS does not support unix, unixgram.
	if runtime.GOOS == "darwin" && (runtime.GOARCH == "arm" || runtime.GOARCH == "arm64") {
		return false
	}
```
And moved tests back to `client_tests.go` to use `t.Skip` instead of build tags.
I think it is more ideomatic way.


valyala:
Merged,

but I'd recommend switching from unix sockets to [InmemoryListener](https://godoc.org/github.com/valyala/fasthttp/fasthttputil#InmemoryListener) in these tests, so they may run on all platforms. @ernado , waiting for the corresponding PR... :)

ernado:
@valyala 
Ok, will see what I can do :)

PR #94, opened 2016-05-10
reuseport: close Listener

noxiouz:
If closing of an underlying File returns error, a Listener will not be returned to a user and will not be closed as well.


valyala:
Thanks for the contribution!

PR #102, opened 2016-05-21
Add func (resp *Response) ReleaseBody(size int)

jfesler:
This adds a method to permit explicit freeing of memory.

```go
	request := fasthttp.AcquireRequest()
	response := fasthttp.AcquireResponse()
        // ... stuff ... //

	response.ReleaseBody(1024 * 1024)
	fasthttp.ReleaseResponse(response)
	fasthttp.ReleaseRequest(request)
```
---

Background:

I spent a fair bit of time this morning trying to figure out a memory leak, that wasn't.  This was in a  load generator replaying CDN traffic, with a mix of mostly small objects but the occasional large object.  

I was primarily baffled at the increasing number of objects that were 16M or 32M each, claimed by fasthttp.appendBodyFixedSize.  

```
(pprof) top 1
8.82GB of 9.21GB total (95.72%)
Dropped 218 nodes (cum <= 0.05GB)
Showing top 1 nodes out of 22 (cum >= 8.95GB)
      flat  flat%   sum%        cum   cum%
    8.82GB 95.72% 95.72%     8.95GB 97.11%  github.com/valyala/fasthttp.appendBodyFixedSize
```

Digging into the memory allocations and counts, the bulk of the memory is lost to 16MB and 32MB chunks of memory.  Looking at the code, and rereading the fasthttp's documentation, this is clearly a feature, not a bug.  I understand very much why the code does this.

Yet, it still hurts in my case case.

With the attached PR, a new function is exposed to allow "large" backing stores to get cleared.  I don't think this is a one-size-fits-all thing; I don't think there is a good default value.  It very much depends on what you're doing; and whether you're just *occasionally* dealing with video-sized HTTP segments versus page assets.

Putting this change to use, and setting the cutoff at 1 meg, I saw no reduction in the load generator performance. I did however see my memory footprint much more stable, much more sane.

It might be *nicer* if this could be a client or request setting; but I didn't want to make a huge sweeping change.  This PR is minimal change, and won't change anything for people who don't opt to use it.

---

**PS:**  Thank you for this excellent http client module.  Within a couple hours I had a basic load generator working that could pull 25 gigabit of traffic on a single host, in a single process.  This exceeded my expectations by an order of magnitude.

### DIFF
--- a/http.go
+++ b/http.go
@@ -375,6 +375,16 @@ func (resp *Response) ResetBody() {
 	resp.body = resp.body[:0]
 }
 
+// ReleaseBody retires the response body if it is greater than "size" bytes. 
+// This permits GC to reclaim the large buffer.  If used, must be before
+// ReleaseResponse.
+func (resp *Response) ReleaseBody(size int) {
+	if cap(resp.body) > size {
+		resp.ResetBody()
+		resp.body = []byte{}
+	}
+}
+
 // Body returns request body.
 func (req *Request) Body() []byte {
 	if req.bodyStream != nil {


valyala:
Thanks for the patch, @jfesler !

valyala:
@jfesler , could you try updating to the https://github.com/valyala/fasthttp/commit/1fc1f4cbf5d3502243555dac0fae10f0f9603c62 and temporarily disabling `ReleaseBody` in your code? The commit reduces the amount of memory used by concurrently running `fasthttp` clients, since it obtains response body buffers from the `sync.Pool` only after the client reads response headers. This should increase the probability for GC to collect large response body buffers put in the `sync.Pool`.

Additionally, I'm planning to substitute https://github.com/valyala/fasthttp/blob/master/bytebuffer.go  with a separate `bytebufferpool` package, which should be smart enough to properly handle variously sized byte buffers as in your case. I.e. it shouldn't return oversized buffers to the pool if they occur infrequently.

jfesler:
@valyala - This (1fc1f4c) does indeed keep my process relatively memory size stable.  With ~500 clients grabbing 100k to 8000k at a time, I'm holding steady with about 34 gigs of RSS.  Thank you. 

My next trick (when I get time) is to figure out how to not save the results *at all*.  I don't actually need the content, I just need the GET (not HEAD) results. :-)   I think what I want is the ability to set a writer for the http client response, and to have said writer simply do nothing.

valyala:
@jfesler , `bytebuffer` with anti-fragmentation protection has been extracted into a separate package - see [bytebufferpool](https://github.com/valyala/bytebufferpool). Now `fasthttp` uses it for request / response bodies - see the https://github.com/valyala/fasthttp/commit/8ca66d3de8e812e79f37fcfa1331c9d2b4dc1b79 for details. Could you check whether this helps reducing memory usage in your case when `ReleaseBody` is disabled?

jfesler:
valyala, these changes are *immensely* better in my use case ("simple" load generator).
150 clients running in parallel: 4 minutes in, and I'm at 3.1g ram instead of 30gig. 
500 clients running in parallel: 4 minutes in, 4 gig of ram, instead of 57+gig RSS + swap

Ping me if you're ever in the SF bay area.  Beverages are in order.

valyala:
@jfesler , glad that [bytebufferpool](https://github.com/valyala/bytebufferpool) helped not only me, but in your case too. The package has been rewritten multiple times after my last message here - could you confirm that the latest `bytebufferpool` version still works great in your case?

jfesler:
Alas, I won't be able to do much more with this, at least any time soon :-(

Issue #64, opened 2016-03-10
Reverse Proxy?


caleblloyd:
The golang httputil package has a ReverseProxy that will serve from an http.Request.

Is there any comporable revrese proxy for fasthttp that will serve from a fasthttp.Request?

valyala:
Currently there is no reverse proxy, but it is [on the radar](https://github.com/valyala/fasthttp/blob/master/TODO) - see 'proxy handler' line there.

Currently reverse proxy may be implemented quite easily using [HostClient](https://godoc.org/github.com/valyala/fasthttp#HostClient):
```go
import (
    "github.com/valyala/fasthttp"
)

var proxyClient = &fasthttp.HostClient{
    Addr: "upstream.host:port",
    // set other options here if required - most notably timeouts.
}

func ReverseProxyHandler(ctx *fasthttp.RequestCtx) {
    req := &ctx.Request
    resp := &ctx.Response
    prepareRequest(req)
    if err := proxyClient.Do(req, resp); err != nil {
        ctx.Logger().Printf("error when proxying the request: %s", err)
    }
    postprocessResponse(resp)
}

func prepareRequest(req *fasthttp.Request) {
    // do not proxy "Connection" header.
    req.Header.Del("Connection")
    // strip other unneeded headers.

    // alter other request params before sending them to upstream host
}

func postprocessResponse(resp *fasthttp.Response) {
    // do not proxy "Connection" header
    resp.Header.Del("Connection")

    // strip other unneeded headers

    // alter other response data if needed
}

func main() {
    if err := fasthttp.ListenAndServe(":8080", reverseProxyHandler); err != nil {
        log.Fatalf("error in fasthttp server: %s", err)
    }
}
```

caleblloyd:
Thanks for the example.  To reverse proxy to a Unix Domain Socket, could I simply pass HostClient a Dialer function that makes use of [net.DialUnix](https://golang.org/pkg/net/#DialUnix)?

valyala:
> To reverse proxy to a Unix Domain Socket, could I simply pass HostClient a Dialer function that makes use of net.DialUnix?

Yes

pdcawley:
There's a problem, I'm afraid, `HostClient.Do` follows redirects, and there's no way to turn that off.

valyala:
[...]

PR #106, opened 2016-05-27
[WIP] Add ProxyClient

hnakamur:
See https://github.com/valyala/fasthttp/issues/64 for discussion.

- [x] Create ProxyClient with methods SendRequest, ReadResponseHeader, ReadResponseBody.
- [x] Move ProxyClient to the separate package "proxy".
- [x] Add methods Do and CleanupResponse and remove SendRequest, ReadResponseHeader, ReadResponseBody.
- [ ] Clean up the code.

An example of using a ProxyClient just as a http client is at:
https://github.com/hnakamur/fasthttp/blob/97498c0b12834fe31992796e3af34e0cb0004f1d/proxy/proxy_client_test.go#L39-L65


valyala:
Try implementing `ProxyClient` in a separate package, so you could spot missing functionality in `fasthttp` for such an implementation.

I'll look into the PR on the next week.

hnakamur:
I copied the source from the net/http package and modified to just enough to work it with ProxyClient.
Am I on the right track? If you agree, I'll try to clean up the copied code a little bit more.

valyala:
I thought a separate `ProxyClient` package will be created outside `fasthttp` namespace, which will use public API provided by `fasthttp`. If the current API isn't enough, then just propose API changes to `fasthttp` in order to satisfy such a `ProxyClient` implementation in a separate package.

hnakamur:
OK. I'm closing this PR.

Issue #107, opened 2016-05-27
Question about HTTP DELETE method


tracer0tong:
Hi, I'm just curious why you doesn't support DELETE method? This is a one of useful methods for classic RESTful interfaces. Do you have some reasons for that?
And yet another question. Do you have a method to get information about all request headers (I want to get information from X-Real-IP for my task), possibly I'm just didn't find it.

valyala:
> Hi, I'm just curious why you doesn't support DELETE method?

Actually `fasthttp` should support arbitrary methods including DELETE. Method name may be obtained via [RequestCtx.Request.Header.Method](https://godoc.org/github.com/valyala/fasthttp#RequestHeader.Method). See the test added in [this commit](https://github.com/valyala/fasthttp/commit/ccf67b7c98691668b7658ddfd91751f6f4e48e6f) for details.

> Do you have a method to get information about all request headers (I want to get information from X-Real-IP for my task), possibly I'm just didn't find it.

See [RequestCtx.Request.Header.Peek](https://godoc.org/github.com/valyala/fasthttp#RequestHeader.Peek) and [RequestCtx.Request.Header.VisitAll](https://godoc.org/github.com/valyala/fasthttp#RequestHeader.VisitAll).

tracer0tong:
I've made short pull request and added a IsDelete(). That was a first question. About second one, thank you, that's what I need!

valyala:
Merged pull request

PR #108, opened 2016-05-27
Issue #107

tracer0tong:
I'm just about this short cosmetic fix. From my point of view DELETE popular enough to have own method IsDelete


valyala:
Thanks!

PR #109, opened 2016-05-28
Add missing messages to statusMessages

hnakamur:


valyala:
Thanks! Though there are new HTTP status messages to be added - see https://github.com/golang/go/commit/b9ec0024fbc18dd94eff7240afd82fac6b4d8fdc for details :)

hnakamur:
@valyala Oh I didn't notice that. How about using [net.http.StatusText()](https://golang.org/pkg/net/http/#StatusText) instead?

valyala:
>  How about using net.http.StatusText() instead?

Initially I didn't want introducing net/http dependency for fasthttp. But now this looks OK, since the dependency already exists in [fs.go](https://github.com/valyala/fasthttp/blob/master/fs.go).

PR #117, opened 2016-06-10
reuseport: set FD_CLOSEEXEC for the sockets

noxiouz:
I am not sure if `rumprun` platform suppports it, but others mentioned in `build+` support


valyala:
Thanks, @noxiouz !

PR #118, opened 2016-06-13
Keep Server header on reset

0x616E676572:


valyala:
@0x616E676572 , could you explain the purpose of this PR?

0x616E676572:
Server sends ```Server: fasthttp``` Header for 404/500 even if I've set custom name.
For an example:
```
server := fasthttp.Server{Name: "myserver", Handler: func(c *fasthttp.RequestCtx) {
	switch string(c.Path()) {
	case "/":
		c.SetStatusCode(fasthttp.StatusOK)
	default:
		c.NotFound()
	}
}}
server.ListenAndServe(":8001")
```
Now, if I send a request to / I'll get the response:
```
HTTP/1.1 200 OK
Server: myserver
```
But if I send a request to /somepath I'll get this:
```
HTTP/1.1 404 Not Found
Server: fasthttp
```

Is it expected behaviour?

kataras:
@0x616E676572 are you sure that your solution is the best to solve this?

kataras:
@0x616E676572 are you sure that your pr is the proper solution to solve this?

0x616E676572:
@kataras I believe that this pr fixes the problem I've experienced but of course there might be another cleaner solution. Any suggestions or concerns?

valyala:
@0x616E676572 , did you try customizing `Server.Name`? See [Server struct](https://godoc.org/github.com/valyala/fasthttp#Server) for details.

Also I believe `Server` header may be overriden after the call to `ctx.NotFound`.

kataras:
> Also I believe Server header may be overriden after the call to ctx.NotFound

It can

kataras:
> Also I believe Server header may be overriden after the call to ctx.NotFound

Yes, it does,

@0x616E676572 I think that we  don't need this kind of 'fix' for that, try setting the server's property correctly as @valyala mentioned, Iris does that and has the header: `Server: iris ` on 404 also...
Its a bad idea to clear all headers' values **except one** on the function which its **destiny** is to clear **all** headers' values, I think it's better to create another function which 'fixs' your 'issue' IF any (Sorry for my bad english, but I hope you get the point)

0x616E676572:
> did you try customizing Server.Name? See Server struct for details.

@valyala yes, I did. Please see my example above
```fasthttp.Server{Name: "myserver" ...```
Of course I can write something like this:
```c.Response.Header.Set("Server", "myserver")```
but why do I need to override the header if I've already set the server's name?

@kataras I agree that it looks a bit strange. But please note that Server Name can be customized when you create a server. And in this case it looks a bit strange that the value you've set gets cleared

**Comment by kataras** [comment]
dunno, the decision is on valyala's hands, but on my app the 404 returns
```sh
HTTP/1.1 404 Not Found
Server: iris
```

0x616E676572:
@kataras try to change your code to use ```NotFound``` method from ```fasthttp.RequestCtx``` instead of ```SetStatusCode(404)```  (or ```NotModified``` instead of ```SetStatusCode(304)```) and you'll find out what I'm talking about

kataras:
@0x616E676572 So the proper solution must be at the end of this function, here https://github.com/valyala/fasthttp/blob/master/server.go#L976 , no at the function which clears all header's values, am I right?

valyala:
I think the proper solution would be setting `Server` header to `Server.Name` before sending the response to the wire if the `Server` header wasn't set by the request handler.

Will try creating the corresponding patch tomorrow.

kataras:
Yes if Response has access to the Server instance it will be ok, thanks valyala

0x616E676572:
@valyala ok, thanks!

valyala:
@0x616E676572 , try `fasthttp` tip after the commit https://github.com/valyala/fasthttp/commit/d42167fd04f636e20b005e9934159e95454233c7 .

0x616E676572:
@valyala 👌

PR #124, opened 2016-06-29
Handle TransferEncoding in fasthttpadaptor

gavv:
When incoming http.Request is constructed, "Transfer-Encoding" header is removed, and http.Request.TransferEncoding is set instead. See [`fixTransferEncoding()`](https://golang.org/src/net/http/transfer.go?h=fixTransferEncoding#L432).

This behaviour is now emulated in `fasthttpadaptor`.


valyala:
Thanks, @gavv !

PR #128, opened 2016-07-03
Add test for fsSmallFileReader.WriteTo and fix bug

gavv:
Hi,

The bug raises when io.Writer passed to fsSmallFileReader.WriteTo doesn't support ReadFrom interface.

/link gavv/httpexpect#16

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -593,7 +593,7 @@ func (r *fsSmallFileReader) WriteTo(w io.Writer) (int64, error) {
 	curPos := r.startPos
 	bufv := copyBufPool.Get()
 	buf := bufv.([]byte)
-	for err != nil {
+	for err == nil {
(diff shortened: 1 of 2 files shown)


gavv:
**Update**

This bug was originally caught with outdated `valyala/bytebufferpool` when `Response.Body()` was called for a small file.

However, after [this commit](https://github.com/valyala/bytebufferpool/commit/21a4b2115581cdabcc9cd97b1624d56785fc407e) `ByteBuffer` implements `ReadFrom`, so this bug is not reproduced when calling `Response.Body()` any more.

But it's still reproduced when calling `fsSmallFileReader.WriteTo` directly with a `Writer` that doesn't implement `ReadFrom`.

valyala:
Nice catch, @gavv ! Merged.

Issue #126, opened 2016-07-02
VisitUserValues inside RequestCtx


kataras:
Hello @valyala I wanted to ask you to export this `context's`  field of`  []userDataKV` or make a func something like `VisitUserValues` . I need this in order to improve the iris' view engine's render functions which will communicate with the context's user's values , I think this will be useful for more users at the future, thanks you!

valyala:
Closing the issue, since the corresponding pull request has been merged to the master

PR #129, opened 2016-07-03
Implement RequestCtx.VisitUserValues as requested

kataras:
See this issue: https://github.com/valyala/fasthttp/issues/126.

RequestCtx.userValues are not exported, but many times is useful to get all context's user values, exporting the .userValues is not a good idea because the key/value is request lifetime, with `VisitUserValues` the developer can copy these values and use anywhere he/she wants, test added.


valyala:
Thanks, @kataras !

PR #135, opened 2016-07-05
Fixed bug with quoted multipart form boundaries

ClaytonAnderson:
[RFC 2046](https://tools.ietf.org/html/rfc2046#section-5.1.1) allows for the boundary portion of the Content-Type header to be quoted. This was not being correctly handled. 

MultipartFormBoundary will now check for wrapping quotes, and remove them if present.


valyala:
Thanks, @ClaytonAnderson !

PR #145, opened 2016-07-22
fix DoDeadline side effect when no free connection

nicbaz:
This fix relates to concurrent `DoTimeout/DoDeadline` using a `HostClient`: when there is no more free connection, as the `Request`'s body has been swapped, all subsequent calls will have an empty body.

I originally used `req.CopyTo` instead of swapping the body until I saw 3546c31: I think this solution should have a minimal impact on performance.


nicbaz:
ping @valyala :-]

valyala:
Thanks, @nicbaz !

PR #147, opened 2016-07-22
Asynchronous processing of RequestCtx in a goroutine

### LINKED ISSUE #146: Asynchronous processing of RequestCtx in a goroutine
I am experimenting with using fasthttp to implement an API gateway/reverse proxy.  At the end of a handler, I'd like to continue using the RequestCtx in a goroutine (e.g. for capturing attributes to send to an analytics server) while concurrently sending the response to the client.  The problem with this is that in the current code (serveConn in server.go), the RequestCtx can be immediately used by the next request for the connection so the context data will get overwritten.

I'm proposing a Hold() function for RequestCtx that signals serveConn to acquire a new instance instead of reusing the current ctx.  If Hold() is called by the handler, then it is also responsible for calling a Release() function to return the ctx instance to the pool.  I'm not sure if this is the most elegant or ideal way of handling this, but looking at the code, it seemed straight forward to add this logic.

pkedy:
Fixes #146: Added a held flag in RequestCtx used to signal serveConn to acquire a new instance for the connection so that the current instance can be processed asynchronously.

### DIFF
--- a/server.go
+++ b/server.go
@@ -404,6 +404,8 @@ type RequestCtx struct {
 	timeoutTimer    *time.Timer
 
 	hijackHandler HijackHandler
+
+	held bool
 }
 
 // HijackHandler must process the hijacked connection c.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1602,6 +1621,12 @@ func (s *Server) serveConn(c net.Conn) error {
 			break
 		}
 
+		if ctx.held {
```
valyala: There is a data race here:

1) The ctx.held may be false here, so the ctx is re-used.
2) Then ctx.Hold() is called in a separate goroutine created by request handler.
3) Then the ctx is concurrently used by another request handler and the goroutine from the step 2. Data race detected :)

Why not using existing [RequestCtx.TimeoutError*](https://godoc.org/github.com/valyala/fasthttp#RequestCtx.TimeoutError) mechanism?

Personally, I don't like both approaches - the existing one and the approach from the PR, since they look fragile and racy.

I prefer passing only relevant copies of data to external goroutines from request handler. This automatically limits RequestCtx lifetime by request handler scope and avoids data races.

pkedy: @valyala Agreed!  Closing

Issue #99, opened 2016-05-16
Support for streaming HostClient response body


cezarsa:
Currently using HostClient.Do() will always read the whole request body before returning. Are there plans  to add a way to stream partial body responses, similar to what can be done by reading Response.Body when using net/http?
This is specially useful for long lived connections that keep sending data over some time

valyala:
Response body streaming is nice feature, but currently I'm unsure with the API and the approach for implementing the feature.

Patches and API proposals are welcome.

runner-mei:
Add Request body streaming also？

tuxmonteiro:
+1

aspcartman:
@cezarsa See #150 , is it what you want? :)

cezarsa:
@aspcartman Great work! I think that's one step on the right direction. But I have one question, this implementation would not handle chunked transfer encodings, right?

I think this may be a burden for some use cases as every user would have to manually parse the chunks. Did you consider adding the logic currently inside `readBodyChunked` to the Reader itself?

When I wrote this issue I thought about contributing this in the future but ended up without enough time. Thanks for working on this.

aspcartman:
@cezarsa I'll take a look into this. I'm not familiar with chunked encoding yet. It would help to have a testable url with chunked response body. Do you have any? :)

I'm inclined to make a new special HostClient `Do` method (instead of manual accessor struct), which would take an boolean argument - chunked or not. Twitter's API response header says it's chunked, but I don't see any chunks in the stream, just a raw json bytes flowing.

cezarsa:
@aspcartman Even better than a url with chunked response. Here is an example of a chunked response using net/http you can easily turn into a test case: https://play.golang.org/p/vbkvDEbq9o

As you can see I'm using `net.Dial` instead of `http.Get` because when using the http client directly the Reader returned in `rsp.Body` would already parse the chunks and you wouldn't be able to see them printed.

flawedmatrix:
What's the current status of adding support for streaming request and response bodies with fasthttp?

Jack47:
After digging the source code, I found fasthttp support streaming request and response bodies.
1. stream response to client

```
// this only set a callback(sw), real streaming is done in sw(stream writing) and after RequestHandler() finished so fast http will write data
ctx.SetBodyStreamWriter(sw)
```
2. stream request to server
```
request.SetBodyStreamWriter(sw)
```
3. stream fasthttp response to client (useful in proxy mode)
```
[...]

PR #150, opened 2016-07-28
Proof of Concept: handling huge/endless response body

aspcartman:
Not to be merged yet, PoC only.
It's done by setting a flag `HandlingBodyManually` (new) into `Request`. If it's set, then the `HostClient` disables automatic response body fetch for the response by setting `SkipBody` flag on the response in it's `Do`. Then it sets a `manualBodyReadAccessor` in the response, which is to be used by user to read the data.

As the result, it's impossible to release/close resources used (the reader and the connection) inside the `doNonNilReqResp()` since they are needed to read data afterward, they leak outside.

 `manualBodyReadAccessor` holds references to and is responsible to release:
1. The `buffered reader` that had been used for reading the response header in the first place. It's the data source off which the user is meant to read data. As I get it, there is no way to throw the one away and acquire a new one later: it may have already pulled some parts of data out of the connection and buffered it.
2. The connection (for releasing).
3. ShouldClose flag: should it close the connection or just release one.
4. HostClient to release resources to it.

Usage:

``` go
    req.HandlingBodyManually = true
    _ = fasthttp.Do(req, resp)
    r := resp.ManualBodyReader()
    for {
        // r is io.ReadCloser, pass it around, read, do stuff.
    }
    _ = r.Close()
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -1092,7 +1092,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		}
 	}
 
-	if !req.Header.IsGet() && req.Header.IsHead() {
+	if (!req.Header.IsGet() && req.Header.IsHead()) || req.HandlingBodyManually {
 		resp.SkipBody = true
 	}
 	if c.DisableHeaderNamesNormalizing {
@@ -1108,13 +1108,21 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		}
 		return false, err
 	}
-	c.releaseReader(br)
 
-	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() {
-		c.closeConn(cc)
+	shouldClose := resetConnection || req.ConnectionClose() || resp.ConnectionClose()
+	if req.HandlingBodyManually {
+		resp.manualBodyReader = c.manualBodyReadAccessor(br, cc, shouldClose)
 	} else {
-		c.releaseConn(cc)
+		c.releaseReader(br)
+
+		if shouldClose {
+			c.closeConn(cc)
+		} else {
+			c.releaseConn(cc)
+		}
 	}
+
+
 
 	return false, err
 }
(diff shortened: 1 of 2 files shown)


aspcartman:
Closes #99 on merge, as I see it. 
Tested manually on Twitter's streaming API 
Let's discuss the solution and improve it. As for now it looks more like a hack, then an actual improvement.

Related thoughts:
Also I can't figure out what exactly existing `*Stream` object are for. What `Response.bodyStream` is for? It doesn't do what I thought it should do (e.g. give me a way to read data from response). `Response.Read()` and `Response.ReadLimitBody()` func names are also feel misleading, I spent some time reading sources to understand that they are about reading the `Response` object itself out of the connection. I propose renaming them to `Response.ReadFrom()` and `Response.ReadFromLimitingBody()` respectively, and instead of allocating and giving away `manualBodyReadAccessor`, implement the `io.Reader` or `io.ReadCloser` interface in `Response` for reading incoming body. 
The project architecture is clean: `Response` doesn't hold references to the `clientConn` or `net.Conn`. Any thoughts how to preserve the current architecture and implement the feature?

aspcartman:
I think I came up with a thought: what if to leave `Response` alone and work with `HostClient`: make it `io.ReadCloser`? Or add a new Do method, which returns a similar to `manualBodyReadAccessor` `io.ReadCloser`? Opinions?

valyala:
I'd prefer the last proposal:

```go
func (c *HostClient) DoBodyReader(req *Request, resp *Response) (io.ReadCloser, error)
```

It should fill `resp` with response header leaving empty response body and return response body reader instead. Could you try implementing it?

As of your questions:
> Also I can't figure out what exactly existing *Stream object are for.  What Response.bodyStream is for?

These are hacks for streaming response body from server side.

> I propose renaming them to Response.ReadFrom() and Response.ReadFromLimitingBody() respectively

Agreed that `Response.Read` is a misleading method name. But it cannot be renamed, since it may break existing code.

shalako:
@valyala Are you still open to this PR, if your requests were addressed? We're interested in picking it up and taking it to completion.

Jack47:
We can use response.BodyWriteTo(w) to stream response
https://github.com/valyala/fasthttp/issues/99#issuecomment-369814706

erikdubbelboer:
Closed because of age and merge conflicts. If you are still interested in contributing this please rebase on master and reopen this issue.

Issue #165, opened 2016-08-25
Cannot follow redirect when using Do() but Get() doesn't expose `Request` object for fine-grain control


shicky:
With current implementation, it is not possible to follow redirects and manipulate `Request` object at the same time.

Access to `Request` object is absolutely necessary for setting headers etc.
- Currently `Get(dst []byte, url string)` allows you to follow redirects but it doesn't expose `Request` object.
- `Do(req *Request, resp *Response)`, exposes `Request` object but doesn't support following redirects.
- `doRequestFollowRedirects(req *Request, dst []byte, url string, c clientDoer)` is not exposed.

Proposal:

Implement alternative `DoFollowRedirect` and `doRequestFollowRedirects` that exposes `Request` and `Response` objects.

``` go
// DoFollowRedirect makes http request based on Request
func (c *Client) DoFollowRedirect(req *Request, resp *Response) error {
        return doRequestFollowRedirects(req, resp, c)
}

// doRequestFollowRedirects follows redirects and return final Response
func doRequestFollowRedirects(req *Request, resp *Response, c clientDoer) error {
        // TODO
}
```

valyala:
Generic `DoFollowRedirects` isn't that easy as it may seem.

The following questions immediately arise when implementing this function:

- Which request headers must be preserved and which must be cleared when following redirects.
- Should timeout be passed to `DoFollowRedirects`?

So I think it would be better if redirects' following is implemented in user code either on top of `Do` or `DoTimeout` depending on user needs.

PR #166, opened 2016-08-25
Redirection support for Do(*Request, *Response) for Issue #165

shicky:
Cannot follow redirect when using Do() but Get() doesn't expose `Request` object for fine-grain control.

This pull request adds `DoFollowRedirects(req *Request, resp *Response, maxRedirects int) error` method for making redirecting http request while having `Request` and `Response` objects exposed.

For more details, please visit issue #165.

All tests passed.

### DIFF
--- a/client.go
+++ b/client.go
@@ -833,48 +833,6 @@ func doRequestFollowRedirects(req *Request, dst []byte, url string, c clientDoer
 	return statusCode, body, err
 }
 
-func doRequestFollowRedirects2(req *Request, dst []byte, url string, c clientDoer) (statusCode int, body []byte, err error) {
-	resp := AcquireResponse()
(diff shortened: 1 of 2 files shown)


valyala:
The change has been reverted, since it breaks `Client.Get*()` following redirects between distinct hosts.

`DoFollowRedirects` must be implemented on top of `Do` instead of piggybacking on `doRequestFollowRedirects'.

valyala:
Also, `DoFollowRedirects` opens the following questions:

- Which request headers must be preserved and which must be cleared when following redirects.
- Should timeout be passed to `DoFollowRedirects`?

So I think it would be better if redirects' following is implemented in user code either on top of `Do` or `DoTimeout` depending on user needs.

PR #180, opened 2016-10-05
Adding WriteDeflate[Level] and AppendDeflateBytes[Level] methods

diegossilveira:
Now we have just the following methods that are able to gzip bytes and write it to a given `io.Writer`:

* `WriteGzip(...)`
* `WriteGzipLevel(...)`
* `AppendGzipBytes(...)`
* `AppendGzipBytesLevel(...)`

But the corresponding methods to deal with deflate-compression were missing:

* `WriteDeflate(...)`
* `WriteDeflateLevel(...)`
* `AppendDeflateBytes(...)`
* `AppendDeflateBytesLevel(...)`

### DIFF
--- a/client.go
+++ b/client.go
@@ -91,29 +91,6 @@ func DoDeadline(req *Request, resp *Response, deadline time.Time) error {
 	return defaultClient.DoDeadline(req, resp, deadline)
 }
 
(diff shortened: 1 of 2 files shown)


diegossilveira:
See #183

PR #183, opened 2016-10-07
Adding support to case-insensitive cookie attributes

diegossilveira:
Accordingly to [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.2.2) - _HTTP State Management Mechanism_, `Set-Cookie` header includes cookie attributes that must be treated case-insensitively. `fasthttp` `Parse()` and `ParseBytes()` methods work only with lower case attributes, and this PR fixes that.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -275,7 +275,7 @@ func (c *Cookie) ParseBytes(src []byte) error {
 		if len(kv.key) == 0 && len(kv.value) == 0 {
 			continue
 		}
-		switch string(kv.key) {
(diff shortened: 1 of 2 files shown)


diegossilveira:
Hello @valyala! What do you think about this PR?

erikdubbelboer:
Thank you for the pull request, it's a good change. But, using `strings.ToLower` allocates a new string on the heap. This is something we try to avoid with fasthttp.

I have made a pull request for the exact same change that doesn't allocate here: https://github.com/valyala/fasthttp/pull/397

PR #186, opened 2016-10-09
Deprecate ByteBuffer and use bytebufferpool.ByteBuffer

bogem:


valyala:
@bogem , unfortunately I cannot merge this PR, since it may break existing third-party code using old `ByteBuffer`.

PR #187, opened 2016-10-12
Update server.go

alphacentory:
ran into ctx.c being nil in a another package. Not sure if the issue lies in their code or this code base. This fixed it though, and all tests passed.

### DIFF
--- a/server.go
+++ b/server.go
@@ -825,6 +825,11 @@ func (ctx *RequestCtx) IsHead() bool {
 //
 // Always returns non-nil result.
 func (ctx *RequestCtx) RemoteAddr() net.Addr {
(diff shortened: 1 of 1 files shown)


valyala:
`ctx.c` may become nil only after `releaseCtx()` is called. This means that either `fasthttp` improperly releases `ctx` or user-defined `RequestHandler` access `ctx` after returning.

Currently this PR just hides the origin of the issue.

@BafFlynN , could you provide panic stack traces, so I could guess what is really happening here?

BafFlynN:
Yes I will. I need to recreate the issue. Sorry have been ill the last week.

PR #195, opened 2016-10-25
Adding errors Request/Response reads

fmstephe:
Currently in order to read from a request or response we needed to read
the entire body into a buffer and then read from that buffer. This
approach does not return any errors that were encountered while reading
from the network. Instead errors are written into the Request/Response
body and can be read from there.

This makes it very hard to reliably determine if an error has occurred.
I have added two very basic functions which do return an explicit error.

One implements current behaviour, read body into buffer and return that,
but returns an explicit error. The second exposes the bodyStream field
as an io.Reader allowing client code to stream data with minimal memory
consumption, as well as exposing errors.

### DIFF
--- a/http.go
+++ b/http.go
@@ -292,6 +292,28 @@ func (resp *Response) Body() []byte {
 	return resp.bodyBytes()
 }
 
+// Body returns response body.
+func (resp *Response) BodyWithErr() ([]byte, error) {
+	var err error
+	if resp.bodyStream != nil {
+		bodyBuf := resp.bodyBuffer()
+		bodyBuf.Reset()
+		_, err = copyZeroAlloc(bodyBuf, resp.bodyStream)
+		resp.closeBodyStream()
+		if err != nil {
+			bodyBuf.SetString(err.Error())
+		}
+	}
+	return resp.bodyBytes(), err
+}
+
+func (resp *Response) StreamBody() io.Reader {
+	if resp.bodyStream != nil {
+		return resp.bodyStream
+	}
+	return bytes.NewReader(resp.bodyBytes())
+}
+
 func (resp *Response) bodyBytes() []byte {
 	if resp.body == nil {
 		return nil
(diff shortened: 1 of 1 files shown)


kataras:
Dear @fmstephe 

Could you, please, write some tests for those ?

Thanks,
kataras

fmstephe:
This PR is in a very basic state. It contains no tests and isn't yet fit for merging. I would like to gauge your interest in this change.

If you are interested, I'll write tests and look more carefully at the implementation.

fmstephe:
@kataras, there are two methods in the PR. I would like to find out which interfaces people are most attracted to before I invest in writing tests.

The main property I would like to have are access to explicit errors when reading from the body of a Request/Response.

kataras:
I don't have access to approve or decline this PR, but without tests it will not be merged.

> It contains no tests and isn't yet fit for merging. I would like to gauge your interest in this change.

I'm ok with this change but I think the `StreamBody()` is already exists with other name, can you lookup and if needed remove this function ?

fmstephe:
@kataras I couldn't find a StreamBody() equivalent. If it exists then I am satisfied and don't need any changes. I'll have another careful look. If anyone can point to these functions I would be very grateful.

kataras:
Yes, you have right the `StreamBody` doesn't exists but I use this method(on a web framework which I maintain):

```go
// StreamReader sets response body stream and, optionally body size.
//
// If bodySize is >= 0, then the bodyStream must provide exactly bodySize bytes
// before returning io.EOF.
//
// If bodySize < 0, then bodyStream is read until io.EOF.
//
// bodyStream.Close() is called after finishing reading all body data
// if it implements io.Closer.
//
// See also StreamReader.
func (ctx *Context) StreamReader(bodyStream io.Reader, bodySize int) {
	ctx.RequestCtx.Response.SetBodyStream(bodyStream, bodySize)
}

```

fmstephe:
@kataras thanks for your pointer, I had misunderstood the way the StreamReader method was intended to be used. So in this scenario you would call the `SetBodyStream(...)` method before you start reading and then you would expect reads to be streamed to the bodyStream arg from then on?

I think this satisfies my needs, I can stream the response body and get to see any errors that occur along the way.

Thanks.

kataras:
Exactly :) Happy to helped, thanks for your interest, fasthttp wants nice contributors like you

PR #206, opened 2016-11-22
Fixes #157 - https-detection

### LINKED ISSUE #157: Wrong redirect with relative path for https
On iris 3/4, the iris.Redirect("/home") redirects to http:// instead of https:// when the webserver is using https. This behavior is different from net/http. I think it's better to be the same as net/http, and it's very convenient for developers. That is, when scheme is not specified in the path, take the scheme of the current webserver automatically.

The iris author analyzed it and figured out it's a fasthttp bug. Please see https://github.com/kataras/iris/issues/355 for reference.

Thanks.

EtienneBruines:
Also added a feature which checks for the X-Forwarded-Proto header being 'https'


EtienneBruines:
It seemed that the tests that were in place, were invalid. (Correct me if I'm wrong.)

They were expecting the function to return `http` when it was a `https` request. No idea why, but I'm assuming it was a workaround. Since it now returns `https` when a `https` request was done, we're also expecting it to do so in the tests. 

Hence the second commit.

server.go
```diff
@@ -1483,6 +1483,8 @@ func (s *Server) serveConn(c net.Conn) error {
 			}
 		}
 
+		ctx.Request.isTLS = isTLS
```
erikdubbelboer: Why set it here if you are overwriting it a couple lines down?

server.go
```diff
@@ -1525,6 +1527,14 @@ func (s *Server) serveConn(c net.Conn) error {
 			}
 		}
 
+		if !ctx.Request.isTLS {
+			if forwarded := ctx.Request.Header.PeekBytes(strForwarded); len(forwarded) > 0 {
+				ctx.Request.isTLS = bytes.Contains(forwarded, strProtoHTTPS)
+			} else {
+				ctx.Request.isTLS = bytes.Contains(ctx.Request.Header.PeekBytes(strXForwardedProto), strHTTPS)
+			}
+		}
```
erikdubbelboer: How about redirects from an https server to a http url? Shouldn't the check be both ways?

server.go
erikdubbelboer: Why set it here if you are overwriting it a couple lines down?

PR #218, opened 2017-01-17
Added `Door` router

claygod:
I have a fairly quick router `Bxog` (https://github.com/claygod/Bxog). Because of him, I made a special version (called `Door` https://github.com/claygod/door), oriented to work with `Fast HTTP server`, adding references to this router in the description.


erikdubbelboer:
If you still want this to be merged please rebase it on master and call it a `router`. A `multiplexer` is something completely different (serving multiple channels over the same connection). Thanks.

PR #219, opened 2017-01-19
Don't try to compress 204 or 304

erikdubbelboer:
These responses can not have a body so trying to compress them and adding the Content-Encoding header is useless.


erikdubbelboer:
The failing travis check has nothing to do with this pull request:
```
client_test.go:1155: cannot listen "127.0.0.1:57794": listen tcp 127.0.0.1:57794: bind: address already in use
```

erikdubbelboer:
Any feedback?

erikdubbelboer:
@valyala I think this is quite a no-brainer to merge?

erikdubbelboer:
I see there is a minimum body size now so this pull request is not needed anymore.

PR #224, opened 2017-01-21
Add IsConnect method

guotie:
Add IsConnect method


erikdubbelboer:
Was fixed in https://github.com/valyala/fasthttp/commit/a84dc2dc0e695384e3ecaebba0a4598524366445#diff-34c6b408d72845d076d47126c29948d1

PR #230, opened 2017-02-14
Do case insensitive comparison of cookie fields

### LINKED ISSUE #181: Cookie parse not working with upper case attributes
Cookie's `Parse()` method only works with lower case attribute keys. If you try parsing this cookie string `Name=MyCookie; Path=/blog; Domain=site.net; Secure; Expires=Thu, 18 Dec 2013 12:00:00 UTC"` you will see the failure.

Accordingly to RFC-6265 (HTTP State Management Mechanism)[https://tools.ietf.org/html/rfc6265#section-4.1.1], these keys begins with a capital letter.

Perelandric:
The case insensitive comparison implemented in this PR avoids any new
allocation and provides improved performance over the current
case sensitive comparison.

The approach taken was to quickly narrow the possibilities down to a
single value by comparing only the first character in the key. It
then performs a full case insensitive comparison using a special
function that compares two []byte. The function is meant to only be
used to compare words that have only letters. Other uses could result
in false positives, though it would be unlikely and probably harmless.

Micro benchmarking showed this approach performing in roughly half the
time per operation over the current case sensitive implementation;
16.4 ns/op vs 29.4 ns/op.

Fixes #181 

---

*EDIT: Test failure is the Go 1.6 build only, and doesn't appear related to this PR.*

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -256,6 +256,22 @@ func (c *Cookie) Parse(src string) error {
 	return c.ParseBytes(c.buf)
 }
 
+// Case insensitive equality comparison of two []byte. Assumes only
+// letters need to be matched.
+func alphaOnlyCIEqual(a, b []byte) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for i := 0; i < len(a); i++ {
+		if a[i]|0x20 != b[i]|0x20 {
+			return false
+		}
+	}
+
+	return true
+}
+
 // ParseBytes parses Set-Cookie header.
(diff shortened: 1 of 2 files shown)


diegossilveira:
Also created #183 to fix that!

erikdubbelboer:
@diegossilveira your `bytes.ToLower` solution allocates which makes it less optimal I guess.

Perelandric:
@diegossilveira Unfortunately, that approach turned out to be several times slower per operation. I can only guess that that's the reason it wasn't merged.

PR #231, opened 2017-02-14
Use random ports for tests

erikdubbelboer:
This should stop travis-ci from constantly failing with `"bind: address already in use"`.


erikdubbelboer:
@valyala it's a simple fix that will prevent many pull requests from getting failing builds. Should be a no-brainer to merge?

PR #232, opened 2017-02-14
Allow elision of Server header in response

### LINKED ISSUE #221: Allow removing of Server header
Right now it's impossible to no send the `Server` header. Setting it to `""` will just fallback to `Server.Name` and setting that to `""` will just fallback to `"fasthttp"`.

I have looked into it but adding code to make `Header.SetServer("")` not send the `Server` header is not easy as methods like [`ctx.NotModified()`](https://github.com/valyala/fasthttp/blob/master/server.go#L1005) also set the header to an empty value.

Perelandric:
Allow elision of Server header in response

This PR adds an option to the Server called `ElideServerHeader` that
allows a user to indicate that neither the Server's `Name` field nor
the `defaultServerName` should be included in the Server's responses
as a `Server` header by default.

Now when `ResponseHeader.AppendBytes` is found to have an empty `server`
field, it simply skips the `Server` header. Previously that method
would write the `defaultServerName` when no value was found. The only
code paths that took advantage of that were ones originating from
`writeErrorResponse`, which now handles setting the server name
directly.

Fixes #221

---

Note:

- This retains the ability to call `ResponseHeader.SetServerBytes()` with a desired value. The elision is only with respect to the initial "global" values.
- As a side-effect of this change, even if this option is not set, you'll be able to eliminate the `Server` header by explicitly setting an empty string during a request.
- With the previous behavior, methods that call `ResponseHeader.Reset()` would cause the `defaultServerName` to be substituted for whatever value had been given. Now this will cause the `Server` header to not be sent, which IMO is preferable.

### DIFF
--- a/header.go
+++ b/header.go
@@ -1425,10 +1425,9 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 	dst = append(dst, statusLine(statusCode)...)
 
 	server := h.Server()
-	if len(server) == 0 {
-		server = defaultServerName
+	if len(server) != 0 {
+		dst = appendHeaderLine(dst, strServer, server)
 	}
-	dst = appendHeaderLine(dst, strServer, server)
 	dst = appendHeaderLine(dst, strDate, serverDate.Load().([]byte))
 
 	// Append Content-Type only for non-zero responses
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
I would say `Elide` is a bit of a weird verb to use. Better would be `OmitServerHeader` or `NoDefaultServerHeader`.

Perelandric:
@erikdubbelboer  I think I like `NoDefaultServerHeader` since it more fully describes the feature. I'll change it today, and we can always change it again if needed.

PR #234, opened 2017-02-21
Make SetMethod slice to zero before appending

Perelandric:
SetMethod appended directly to the `h.method` slice without setting its
length to zero, so multiple calls would create an unexpected value.


valyala:
Thanks for the fix!

PR #240, opened 2017-02-26
http: Use full URI in request line

jinpan:
This resolves issue 239 and makes fasthttp more compliant with rfc 2616.

### DIFF
--- a/http.go
+++ b/http.go
@@ -1088,7 +1088,7 @@ func (req *Request) Write(w *bufio.Writer) error {
 			return errRequestHostRequired
 		}
 		req.Header.SetHostBytes(host)
-		req.Header.SetRequestURIBytes(uri.RequestURI())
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
I don't think this change should be made without proper research. All web clients I know always send relative paths for the Request-URI. I have no idea how many servers actually follow RFC 2616 and allow a full URI here. I'm afraid this change might break for some people.

From RFC 2616:
```
The most common form of Request-URI is that used to identify a resource on an origin server or gateway. In this case the absolute path of the URI MUST be transmitted...
```
But also:
```
The absoluteURI form is REQUIRED when the request is being made to a proxy.
```

So I guess the best option would be some field to specify that the request is going to be made though a proxy. How does `net/http` handle this?

erikdubbelboer:
Closed because of age and concerns.

PR #243, opened 2017-03-14
Add pprofhandler, fix #235

### LINKED ISSUE #235: Integrate pprof
Is there any easy way to integrate pprof with fasthttp ? Instead of using fasthttpadapter and pointing to handler functions provided by pprof

erikdubbelboer:
Similar to [expvarhandler](https://github.com/valyala/fasthttp/tree/fc109d6887b5edb43510d924d14d735f3975fb51/expvarhandler) but for [net/http/pprof](https://golang.org/pkg/net/http/pprof/).

### DIFF
--- a/pprofhandler/pprof.go
+++ b/pprofhandler/pprof.go
@@ -1,0 +1,35 @@
+package pprofhandler
+
+import (
+	"net/http/pprof"
+	"strings"
+
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttpadaptor"
+)
+
+var (
+	cmdline = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Cmdline)
+	profile = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Profile)
+	symbol  = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Symbol)
+	trace   = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Trace)
+	index   = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Index)
+)
+
+// PprofHandler serves server runtime profiling data in the format expected by the pprof visualization tool.
+//
+// See https://golang.org/pkg/net/http/pprof/ for details.
+func PprofHandler(ctx *fasthttp.RequestCtx) {
+	ctx.Response.Header.Set("Content-Type", "text/html")
+	if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/cmdline") {
+		cmdline(ctx)
+	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/profile") {
+		profile(ctx)
+	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/symbol") {
+		symbol(ctx)
+	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/trace") {
+		trace(ctx)
+	} else {
+		index(ctx)
+	}
+}


codesenberg:
Hmm... I wonder if creating a new function on each request is really good idea. Maybe, it would be better to do something like:
```go
var (
	Cmdline = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Cmdline)
	Profile = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Profile)
	Symbol = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Symbol)
	Trace = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Trace)
	Index = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Index)
)

// PprofHandler serves server runtime profiling data in the format expected by the pprof visualization tool.
//
// See https://golang.org/pkg/net/http/pprof/ for details.
func PprofHandler(ctx *fasthttp.RequestCtx) {
	ctx.Response.Header.Set("Content-Type", "text/html")
	if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/cmdline") {
		Cmdline(ctx)
	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/profile") {
		Profile(ctx)
	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/symbol") {
		Symbol(ctx)
	} else if strings.HasPrefix(string(ctx.Path()), "/debug/pprof/trace") {
		Trace(ctx)
	} else {
		Index(ctx)
	}
}
```
Or, even better, rewrite each handler handler to use `fasthttp` by hand. A

erikdubbelboer:
@codesenberg you are right, creating a new function each request is really not needed. I changed this now.

I would say rewriting each handler by hand is not good. It would only introduce more code that will have to be maintained. Also the pprof handlers aren't really performance sensitive so the little overhead of `fasthttpadaptor` doesn't matter.

codesenberg:
> I would say rewriting each handler by hand is not good. It would only introduce more code that will have to be maintained.

@erikdubbelboer, well, true and to my amusement there are no tests for `net/http/pprof` in [golang repo](https://github.com/golang/go/tree/master/src/net/http/pprof).
Take note that `Cmdline`, `Profile`, etc. are exported variables and they need comments.
Also, I just realised that those new variables, unlike original functions they wrap, could be modified from outside the package, so, maybe, we should make them private and expose functions that call them instead? Like this:
```go
var (
	cmdline = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Cmdline)
	profile = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Profile)
	symbol = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Symbol)
	trace = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Trace)
	index = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Index)
)

// Cmdline responds with the running program's
// command line, with arguments separated by NUL bytes.
func Cmdline(ctx *fasthttp.RequestCtx) {
	cmdline(ctx)
}
// Same thing for other handlers...
```

erikdubbelboer:
Do the functions really have to be exposed. Maybe it's better to just keep them private. Most people will only want to use `PprofHandler`. If they want anything else they can easily write the one fasthttpadaptor line themselves?

codesenberg:
@erikdubbelboer, it's hard to tell. I'd expose them anyway to save users from a few keystrokes, but let's be honest this PR probably won't get merged.

erikdubbelboer:
I made them internal now, I think that's the best solution for now.

You mean this won't get merged because @valyala doesn't seem to care about this package anymore?

codesenberg:
@erikdubbelboer, yeah, something like that. Maybe, he's just too busy at @Vertamedia, hard to tell.

DuckerMan:
So, how use pprof with fasthttp?

Can You provide an example?

kirillDanshin:
@DuckerMan you should use this as you do with your own request handler.
Here you have two options: serve pprof on its own port or use your favorite router to serve this handler on app's port. I recommend to use dedicated port.

If you need, I can send an example code later, but I think you got the idea :)

DuckerMan:
So, i should import "net/http/pprof" and create dedicated port by fasthttp?

kirillDanshin:
@DuckerMan nope. instead of `net/http/pprof` use `github.com/valyala/fasthttp/pprof.PprofHandler()`

DuckerMan:
@kirillDanshin where i should call `github.com/valyala/fasthttp/pprofhandler.PprofHandler()` ?

kirillDanshin:
@DuckerMan 
```go
package main

import (
	"github.com/valyala/fasthttp"
	"github.com/valyala/fasthttp/pprofhandler"
)

func main() {
	go runPprof()

	fasthttp.ListenAndServe(":1337", func(ctx *fasthttp.RequestCtx) {
		ctx.WriteString("Hi")
	})
}

func runPprof() {
	fasthttp.ListenAndServe(":7777", pprofhandler.PprofHandler)
}
```

DuckerMan:
@kirillDanshin thanks!

PR #248, opened 2017-04-06
Fix hex2intTable out of range bug

erikdubbelboer:
`hex2intTable` only contains 254 entries while it should be 255.
Sending for example `Transfer-Encoding: \xff` causes an out of range panic.
Because fasthttp never uses defer-recover and runs in its own goroutines, the user of this package can do nothing to defend against this bug.

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -311,15 +311,17 @@ func hexCharUpper(c byte) byte {
 }
 
 var hex2intTable = func() []byte {
-	b := make([]byte, 255)
-	for i := byte(0); i < 255; i++ {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Why hasn't this been merged yet? This is a major bug. I can crash any webserver using fasthttp!

erikdubbelboer:
Finally fixed in https://github.com/valyala/fasthttp/commit/e5f51c11919d4f66400334047b897ef0a94c6f3c
Only exposed every user of fasthttp for more than half a year...

PR #250, opened 2017-04-14
Add ';' to unescaped characters

siemin93:
I need to make ctx.Redirect() to doubleclick links. Example link is:

[https://ad.doubleclick.net/bla/bla/A1.B1;param1=val1;param2=val2;redir=?www.destination.com](url)

ctx.Redirect function was encoding ; to %3B and redirect were failing.


erikdubbelboer:
Looking at the stdlib I guess all these characters don't need to be encoded: https://github.com/golang/go/blob/a4c18f063b6659079ca2848ca217a0587dabc001/src/net/url/url.go#L128

erikdubbelboer:
The only reason the test is failing is because https://github.com/valyala/fasthttp/pull/231 still isn't merged yet :(

erikdubbelboer:
Thanks.

PR #259, opened 2017-05-08
Fix client.go.

cklxmu:
Fix the bug that not setting read deadline in a new connection caused by old lastReadDeadlineTime.
You have checked the cc.lastReadDeadlineTime in doNonNilReqResp(), it's good to improve performance. But, you should reset all the fields of clientConn when you release it. Or the new connection may use the lastReadDeadlineTime of the old connection to decide if it should set the read deadline.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1281,7 +1281,7 @@ func acquireClientConn(conn net.Conn) *clientConn {
 }
 
 func releaseClientConn(cc *clientConn) {
-	cc.c = nil
+	*cc = clientConn{}
 	clientConnPool.Put(cc)
 }
 


erikdubbelboer:
This is actually a really big bug that I never noticed before (both the bug and the pull request).

`lastReadDeadlineTime` is indeed only set after it has been used:
https://github.com/valyala/fasthttp/blob/d3181922e1b29ed270229c9feb9ff44e7f7b51cb/client.go#L1112-L1117

@cklxmu can you maybe rebase on master and add this test to your pull request. After that this should definitely be merged!
```go
func TestClientReadTimeout(t *testing.T) {

  timeout := false
  s := &Server{
      if timeout {
        time.Sleep(time.Minute)
      } else {
        timeout = true
      }
    },
    Logger: &customLogger{}, // Don't print closed pipe errors.
  } 
  go s.Serve(ln)
      
  c := &HostClient{
    ReadTimeout:               time.Second * 4,
    MaxIdemponentCallAttempts: 1,
      return ln.Dial()
    },
  } 
  
  req := AcquireRequest()
  res := AcquireResponse()
  
  req.SetRequestURI("http://localhost")
  
  // Setting Connection: Close will make the connection be
  // returned to the pool.
  req.SetConnectionClose()
      
  if err := c.Do(req, res); err != nil {
    t.Fatal(err)
  } 
  
  ReleaseRequest(req)
  ReleaseResponse(res)
    
  done := make(chan struct{})
  go func() {
    req := AcquireRequest()
    res := AcquireResponse()
  
    req.SetRequestURI("http://localhost")
    req.SetConnectionClose()
    
    c.Do(req, res)
    
    ReleaseRequest(req)
    ReleaseResponse(res)
    close(done)
  }()

  select {
  case <-done:
    // This shouldn't take longer than the timeout times the number of requests it is going to try to do.
    // Give it 2 seconds extra seconds just to be sure.
  case <-time.After(c.ReadTimeout*time.Duration(c.MaxIdemponentCallAttempts) + time.Second*2):
    t.Fatal("Client.ReadTimeout didn't work")
  }
}
```

PR #269, opened 2017-06-09
Fix premature connection closing.

ayang64:
Using a coarse timer causes the server to send Connection: close early.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1054,7 +1054,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		panic("BUG: resp cannot be nil")
 	}
 
-	atomic.StoreUint32(&c.lastUseTime, uint32(CoarseTimeNow().Unix()-startTimeUnix))
+	atomic.StoreUint32(&c.lastUseTime, uint32(time.Now().Unix()-startTimeUnix))
 
 	// Free up resources occupied by response before sending the request,
 	// so the GC may reclaim these resources (e.g. response body).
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
@ayang64 In what case does this cause the connection to close prematurely? Do you depend on certain connection timeouts of less than a second?

ayang64:
Yes -- we use fasthttp in a very time-critical application that requires sub-second timeouts.

erikdubbelboer:
I have completely removed this coarse timer from my fork. The 10ns (yes nano seconds) speed-up per call is really not worth the added complexity. See: https://github.com/erikdubbelboer/fasthttp/commit/033e10c58522144bd570ff3aa45fbd2f3df917db

ayang64:
Ya -- I ran a similar benchmark:

https://github.com/ayang64/fasttime

I agree that it isn't worth it.  It seems like an abuse of the time.Time type.  Perhaps if we want to provide 1s grangularity we should expose a type that only allows you to speak about times in terms of seconds (like simply an int32 or int64).  That way users might not make the mistake of talking about time in any other units.

erikdubbelboer:
@ayang64 I guess you should close this pull request seeing as you are adding commits to it now that make no sense for it.

ayang64:
@valyala -- hi Aliaksandr, can you please comment on this PR?

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/commit/f2ddaffc31d94ffb7ac4c0f9d19adf2fbc96bd71

PR #277, opened 2017-07-07
Update README.md in third-party using fasthttp

danjou-a:
Delete Iris : deprecated,  the project has been acquired by a Dubai-based startup.


smares:
#255

Issue #278, opened 2017-07-09
normalizeHeader optimization


kirillDanshin:
Hi @valyala 

Since I working on [gramework](https://github.com/gramework/gramework), I found NormalizeHeader in my cpu top20 and found it slow. Here's original benchmark result:

```
BenchmarkNormalizeHeaderKeyCommonCase-4         30000000                46.5 ns/op
BenchmarkNormalizeHeaderKeyLowercase-4          30000000                44.4 ns/op
BenchmarkNormalizeHeaderKeyUppercase-4          30000000                45.5 ns/op
```

Here's optimized result:

```
BenchmarkNormalizeHeaderKeyCommonCase-4         50000000                37.3 ns/op
BenchmarkNormalizeHeaderKeyLowercase-4          50000000                35.9 ns/op
BenchmarkNormalizeHeaderKeyUppercase-4          50000000                37.7 ns/op
```

Working on a PR now.

erikdubbelboer:
What where you doing when NormalizeHeader was in your top20? As you can see the function takes only a couple of nano seconds. Compared to all other things going on this should always be a minuscule amount of time. Unless you are sending huge (multiple megabytes) headers in all upper case I suppose.

kirillDanshin:
@erikdubbelboer basically I'm handling many requests with many {lower,upper}cased headers.
some time ago I think that issue in gramework itself and spent alot of time trying to find where exactly I spent the time and CPU, but I found issue here. with patched version the issue is solved for me.

erikdubbelboer:
I don't see how a 10ns difference per header per request can make a difference. Lets say you have 100 headers (way to many) per request, that would still only mean 1 microsecond (not millisecond) less per request. Unless you need to handle millions of requests per second without doing any processing this change isn't going to do much. As soon as you do any processing it's going to be so much slower that a 1 microsecond change isn't going to change a thing.

I would say this is such a minuscule optimization that it's not worth having to maintain `fastlen` which might break with other Go releases. Also before v2.0 since it's something internal. For example the internal representation of `map` [changed in 1.6](https://github.com/golang/go/commits/master/src/runtime/hashmap.go).

kirillDanshin:
@erikdubbelboer note that only breaking change may be it's struct ordering change, that is way too difficult to implement. There is no reason to change .len ordering.
Next. Note that 10ns difference is the result on my laptop. On my servers I have more than 60% difference. Also, I know that if I need that patch, someone else can also need it.
[...]

PR #279, opened 2017-07-09
normalizeHeader optimization; helps #278

kirillDanshin:
helps #278

Signed-off-by: Kirill Danshin <k@guava.by>

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -344,16 +344,14 @@ func hexbyte2int(c byte) int {
 const toLower = 'a' - 'A'
 
 func uppercaseByte(p *byte) {
-	c := *p
-	if c >= 'a' && c <= 'z' {
-		*p = c - toLower
+	if *p >= 'a' && *p <= 'z' {
+		*p = *p - toLower
 	}
 }
 
 func lowercaseByte(p *byte) {
-	c := *p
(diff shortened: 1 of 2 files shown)


bytesconv.go
```diff
@@ -363,6 +361,21 @@ func lowercaseBytes(b []byte) {
...
+// implementation will change, that seems to be impossible
+// without releasing Go2
+//
+// here we taking slice pointer, converting it to unsafe.Pointer
+// instruction, then converting it to a pointer to array of three
+// unsafe.Pointers, reading the middle one (1st index) as a uintptr,
+// then converting it to int. cause we don't have a real unsafe.Pointer
+// conversions at runtime (there is no real unsafe.Pointer type),
+// it's cheap and fast
+func fastlen(b []byte) int {
```
valyala: I doubt this is good idea :) The latest versions of Go should already optimize `len([]byte)` the same way.

valyala:
Closing the PR, since it has been already used in https://github.com/valyala/fasthttp/commit/952171f61b8dfaf89a2bbf0083fb7e79231e9f13 .

bytesconv.go
kirillDanshin: But it's not, I've tested it on latest versions and fastlen still faster than len()

PR #281, opened 2017-07-17
Make maxAttempts in client.go configurable via HostClient.MaxAttempts

ayang64:
We use fasthttp in a way that if one connection fails for any reason, we'd prefer not to attempt again.  Our time constraints make it necessary to only attempt once.

This patch makes the client max attempts configurable via the HostClient.MaxAttempts value.

If MaxAttempts is less than or equal to 0, then the client uses the default of 5 max attempts.

### DIFF
--- a/client.go
+++ b/client.go
@@ -511,6 +511,12 @@ type HostClient struct {
 	// DefaultMaxConnsPerHost is used if not set.
 	MaxConns int
 
+	// Maximum number of attempts to make to to connect.
+	//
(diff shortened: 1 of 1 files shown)


ayang64:
@valyala -- hi Aliaksandr, can you please comment on this PR?

erikdubbelboer:
https://github.com/valyala/fasthttp/pull/317 is preferred over this as it contains better documentation and a constant instead of just `5` in the code.

PR #283, opened 2017-07-20
Added function: GetContentType to get type of file 

themester:
I know that this function uses memory. But is more efficiently than use "if strings.Compare(...) {...} else if strings.Compare(...){..}" or saving to a file and searching


erikdubbelboer:
If you build the map once outside of the function it shouldn't allocate when you call it.

PR #291, opened 2017-08-28
Do not log ErrSmallBuffer for request headers unless LogAllErrors is set

oschwald:
fasthttp returns a 431 error when the request
headers are too large. Most other HTTP servers do not log an error when
the request header exceeds the limit. When serving HTTP requests
directly to browsers, it is normal to occasionally hit the limit due to
broken browsers or broken sites linking to your resource. These issues
are generally outside of your control in the same way the already
ignored errors are.

Note that ErrSmallBuffer _should_ always be logged for responses as this
should always be under your control.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -17,7 +17,26 @@ import (
 	"github.com/valyala/fasthttp/fasthttputil"
 )
 
-func TestServerErrSmallBuffer(t *testing.T) {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Already fixed in https://github.com/valyala/fasthttp/commit/fbf4b2d9cd9a9bb84baf187775bf90cab15a3728. Thanks.

PR #298, opened 2017-09-28
Added GetPostArgs which returns Args and error

Tehtehteh:
If content-type is set to 'application/json' `PostArgs` couldn't parse args because of that. New method supposed to return `args` and `error` in this case, signaling about wrong Content-Type.

### DIFF
--- a/http.go
+++ b/http.go
@@ -657,21 +657,36 @@ func (req *Request) PostArgs() *Args {
 	return &req.postArgs
 }
 
-func (req *Request) parsePostArgs() {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
1) Always run `go fmt` on your code.
2) I feel like this is deceiving and not a good idea. It makes it seem like `GetPostArgs` returns an error in case there is one, but in reality it only returns an error if `Content-Type` is not `x-www-form-urlencoded` and ignores all other errors in the actual post data (such as encoding errors). Besides the `Content-Type` check is something people can easily do themselves.

erikdubbelboer:
Closed due to mentioned concerns.

PR #302, opened 2017-10-06
Allow overwrite of ContentType by caller of RequestCtx.ServeFile

### LINKED ISSUE #301: RequestCtx.SendFile Content-Type cannot be overwritten
I have a use case where I have my own route and want an query argument (`&dl=true`) to raw download te file instead of viewing by setting the content type always to `application/octet-stream`. Currently this is not possible and I have looked through the fs.go code, or I'm missing something.

xor-gate:
Fixes #301, but CPU and I/O is burned to guess the content type even when the application sets content type (due to the guessing is at the lowest level).

Now this is possible without auto-setted content type to `text/plain`:

```
func README(ctx *fasthttp.RequestCtx) {
	ctx.SetContentType("application/octet-stream")
	ctx.SendFile("./README.md")
}
```

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -808,7 +808,9 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 			}
 		}
 	}
-	ctx.SetContentType(ff.contentType)
+	if len(ctx.Response.Header.ContentType()) == 0 {
+		ctx.SetContentType(ff.contentType)
+	}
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks.

The burned CPU/IO is only once per file (since it's cached). Trying to remove this would require a significant amount of changes that is probably not worth it.

xor-gate:
Good 👍

PR #315, opened 2017-11-28
Parsing Origin From the Request Header

marcsantiago:
Exposing the Parsing and Exposing The Origin header.

### DIFF
--- a/header.go
+++ b/header.go
@@ -65,6 +65,7 @@ type RequestHeader struct {
 	host        []byte
 	contentType []byte
 	userAgent   []byte
+	origin      []byte
 
 	h     []argsKV
 	bufKV argsKV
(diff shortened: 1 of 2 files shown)


themester:
valyala fasthttp seems to be abandoned. If you want to continue developing this library you can do a fork of https://github.com/erikdubbelboer/fasthttp and push your ideas in there.

Thank you

erikdubbelboer:
What would be the advantage over `RequestHeader.Peek("Origin")`? I think the other headers that have their own function such as UserAgent are a design mistake as well but are impossible to remove now without breaking backwards compatibility.

marcsantiago:
I come the from the Adtech world. Where cors headers are important pieces
of information to pass to demand. By grabbing the origin the access control
can easily set.  As for the advantage over Peak, there isn't much but it is
nicer to use. Just like ctx.UserAgent() .

Azareal:
`valyala fasthttp seems to be abandoned.` He made a commit a couple of days ago, maybe he's busy or something.

illotum:
@Azareal this is not unusual. @valyala pushes couple commits a month, I assume, without ever opening github, for about a year now.

kirillDanshin:
@dgrr @erikdubbelboer any thoughts on this?

marcsantiago:
Agreed for the most part, though adding, I'll go ahead and close this one
up, I forgot I opened this, it's been a while. Thanks!

kirillDanshin:
@erikdubbelboer @marcsantiago if we going to keep fasthttp API as simple as possible, we can use gramework as a place for shortcuts like this
Closing PR, but thanks for your work :)

PR #317, opened 2017-11-30
Configuration for automatic retry for idempotent calls

NexoMichael:
The fix allows to configure retries attempts count for idempotent calls made by HostClient.
Motivation: a lot of API's do not treat PUT/DELETE/HEAD HTTP requests as idempotent that's why retrying of this calls can lead to serious behaviour issues.

### DIFF
--- a/client.go
+++ b/client.go
@@ -443,6 +443,9 @@ const DefaultMaxConnsPerHost = 512
 // connection is closed.
 const DefaultMaxIdleConnDuration = 10 * time.Second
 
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
@kirillDanshin does this really need any tests? The code is so simple that I vote to just merge it.

kirillDanshin:
thanks for mergind. as I said in t.me/gramework, I've done with this PR's tests

PR #327, opened 2018-02-14
Fix readme: update 3rd-party library list

kirillDanshin:
Remove iris, add 3rd-party lib types, sort by type

Iris doesn't use fasthttp anymore.
Also, in this list it's unclear which library does what, so I added library types. For developers that needs to use software only under certain licenses, I've also added latest licenses for each library.
Last, I sorted the list by types: you can see frameworks and then routers.

also, note that lu is currently was not updated for a year, but I didn't remove the lib because it may be just a _really_ stable and reliable project and just doesn't require any updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,11 +555,10 @@ uintBuf := fasthttp.AppendUint(nil, 1234)
   There are no plans to add request routing into fasthttp.
   Use third-party routers and web frameworks with fasthttp support:
 
-    * [Iris](https://github.com/kataras/iris)
-    * [fasthttp-routing](https://github.com/qiangxue/fasthttp-routing)
-    * [fasthttprouter](https://github.com/buaazp/fasthttprouter)
-    * [gramework](https://github.com/gramework/gramework)
-    * [lu](https://github.com/vincentLiuxiang/lu)
+    * [gramework](https://github.com/gramework/gramework) (framework, Apache 2.0)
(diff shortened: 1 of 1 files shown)


kataras:
Good moring guys,

[Iris.v5](http://gopkg.in/kataras/iris.v5) is the most complete fasthttp web framework, removing it will cause loses to the developer's  research about how fasthttp works and how they can benefit from it and also what is its downsides (such as like of http/2, at least for now)

kataras:
Good moring guys,

[Iris.v5](http://gopkg.in/kataras/iris.v5) is the most complete fasthttp web framework, removing it will cause loses to the developer's  research about how fasthttp works and how they can benefit from it and also what is its downsides (such as lack of http/2, at least for now)

Solution: rename https://github.com/kataras/iris to https://github.com/kataras/iris/tree/v5

kataras:
Good morning,

[Iris.v5](http://gopkg.in/kataras/iris.v5) is the most complete fasthttp web framework, it's working with the latest fasthttp - fasthttp didn't change any of its APIs neither added new call that may be useful to be inserted into iris.v5, therefore removing it will cause loses to the developer's  research about how fasthttp works and how they can benefit from it and also what is its downsides (such as lack of http/2, at least for now)

Solution: rename https://github.com/kataras/iris to https://github.com/kataras/iris/tree/v5 instead.

Thanks,

Gerasimos Maropoulos,  Author of Iris.

PR #356, opened 2018-07-07
Use form headers in WriteMultipartForm

tommy351:
[CreateFormFile](https://golang.org/pkg/mime/multipart/#Writer.CreateFormFile) always set `Content-Type` header to `application/octet-stream`.
Use [CreatePart](https://golang.org/pkg/mime/multipart/#Writer.CreatePart) instead so all headers in the multipart form can be used.

### DIFF
--- a/http.go
+++ b/http.go
@@ -744,7 +744,7 @@ func WriteMultipartForm(w io.Writer, f *multipart.Form, boundary string) error {
 	// marshal files
 	for k, fvv := range f.File {
 		for _, fv := range fvv {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Looks like a good fix to me even though it breaks backwards compatibility. People who didn't provide a correct form the `Content-Disposition` and `Content-Type` headers will run into issues where these aren't automatically set anymore.

erikdubbelboer:
Travis CI seems to have had an error there and that's why it failed.

Like I said before, looks good to me 👍

PR #364, opened 2018-08-17
Added PathNotFound handle func for handling 404.

erikdubbelboer:
See https://github.com/erikdubbelboer/fasthttp/pull/10

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -241,6 +241,14 @@ type FS struct {
 	// By default request path is not modified.
 	PathRewrite PathRewriteFunc
 
+	// PathNotFound fires when file is not found in filesystem
+	// this functions tries to replace "Cannot open requested path"
+	// server response giving to the programmer the control of server flow.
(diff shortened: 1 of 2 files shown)


fs.go
```diff
@@ -726,7 +736,12 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 			}
 		} else if err != nil {
 			ctx.Logger().Printf("cannot open file %q: %s", filePath, err)
-			ctx.Error("Cannot open requested path", StatusNotFound)
+			if h.pathNotFound == nil {
+				ctx.Error("Cannot open requested path", StatusNotFound)
+			} else {
+				h.pathNotFound(ctx)
+				ctx.SetStatusCode(StatusNotFound)
```
kirillDanshin: pathNotFound may be used to serve SPA index page, so we should not overwrite status code if it was already set

PR #369, opened 2018-08-17
handle 't' and 'true' as bool in QueryArgs

erikdubbelboer:
See: https://github.com/erikdubbelboer/fasthttp/pull/46

### DIFF
--- a/args.go
+++ b/args.go
@@ -287,11 +287,14 @@ func (a *Args) GetUfloatOrZero(key string) float64 {
 
 // GetBool returns boolean value for the given key.
 //
-// true is returned for '1', 'y' and 'yes' values,
+// true is returned for "1", "t", "T", "true", "TRUE", "True", "y", "yes", "Y", "YES", "Yes",
 // otherwise false is returned.
 func (a *Args) GetBool(key string) bool {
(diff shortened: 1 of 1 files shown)


args.go
```diff
@@ -287,11 +288,11 @@ func (a *Args) GetUfloatOrZero(key string) float64 {
 
 // GetBool returns boolean value for the given key.
 //
-// true is returned for '1', 'y' and 'yes' values,
+// true is returned for '1', 'y', 'yes', 'true', 't' and also their UPPERCASE analogs values,
 // otherwise false is returned.
 func (a *Args) GetBool(key string) bool {
-	switch string(a.Peek(key)) {
-	case "1", "y", "yes":
+	switch strings.ToLower(string(a.Peek(key))) {
```
kirillDanshin: please get rid of ToLower and just add UPPERCASE analogs in the `case`. we should omit unnecessary allocations

erikdubbelboer: I didn't even notice the original pull request had this when I merged it into my fork :(
I removed the possible allocation and added a bunch of fixed cases including all the cases supported by `strings.ParseBool`.

PR #370, opened 2018-08-17
Add Server.Shutdown

erikdubbelboer:
I'm still not sure if we should keep idle connections in a slice so we can close them on shutdown. It would prevent the issue where the server only shuts down after `ReadTimeout` seconds have passed. What do you think @kirillDanshin?

### DIFF
--- a/server.go
+++ b/server.go
@@ -280,6 +280,12 @@ type Server struct {
 	writerPool     sync.Pool
 	hijackConnPool sync.Pool
 	bytePool       sync.Pool
+
+	// We need to know our listener so we can close it in Shutdown().
+	ln net.Listener
+
+	wg   sync.WaitGroup
+	stop int32
 }
 
 // TimeoutHandler creates RequestHandler, which returns StatusRequestTimeout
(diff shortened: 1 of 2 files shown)


savsgio:
Hi @kirillDanshin,
I want to know, ¿When do you think to merge this PR?

kirillDanshin:
hi @savsgio
I currently testing Erik's pull requests, so stay tuned for upcoming updates. Thanks for your question :)

server.go
```diff
@@ -1473,6 +1512,8 @@ func nextConnID() uint64 {
 const DefaultMaxRequestBodySize = 4 * 1024 * 1024
 
 func (s *Server) serveConn(c net.Conn) error {
+	defer s.wg.Done()
```
kirillDanshin: defer is too heavy. it simplify code, but it costs too much in this case. please, no defer on hot paths.

erikdubbelboer: 1) `defer` used to be heavy in previous versions of Go where it would heap allocate for the closure. In recent versions of Go (after 1.8 IIRC) the compiler does analysis and stack allocates the space so it has only little overhead left: https://gist.github.com/erikdubbelboer/90ca1b7045edbae9a263b267812af550 (only about 43ns per call)
2) `serveConn` is only called once per connection which means this only adds 43ns of overhead per connection, not per request. I think the 43ns is worth the code complexity of having to add the statement in each return path and having all of fasthttp lock up in case of a panic.

PR #371, opened 2018-08-17
Fix supported redirect status codes

erikdubbelboer:
Support the same redirect status codes as the golang standard library
does: https://github.com/golang/go/blob/15095be9fbe726d4a3ef43b68d2fbc83e6484ded/src/net/http/client.go#L419


erikdubbelboer:
The travis error here is because this pull request isn't merged yet: https://github.com/valyala/fasthttp/pull/380

PR #374, opened 2018-08-17
Configuration for automatic retry for idempotent calls

erikdubbelboer:
The fix allows to configure retries attempts count for idempotent calls made by HostClient.
Motivation: a lot of API's do not treat PUT/DELETE/HEAD HTTP requests as idempotent that's why retrying of this calls can lead to serious behaviour issues.

See: https://github.com/erikdubbelboer/fasthttp/pull/7

### DIFF
--- a/client.go
+++ b/client.go
@@ -443,6 +443,9 @@ const DefaultMaxConnsPerHost = 512
 // connection is closed.
 const DefaultMaxIdleConnDuration = 10 * time.Second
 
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Sorry I didn't see https://github.com/valyala/fasthttp/pull/317 was still open. Seeing as they are the same I'm closing this one.

PR #376, opened 2018-08-17
Added support for multiple TLS domains

erikdubbelboer:
See: https://github.com/erikdubbelboer/fasthttp/pull/24

### DIFF
--- a/examples/multidomain/Makefile
+++ b/examples/multidomain/Makefile
@@ -1,0 +1,6 @@
+writer: clean
+	go get -u github.com/valyala/fasthttp
+	go build
+
+clean:
+	rm -f multidomain
--- a/examples/multidomain/multidomain.go
+++ b/examples/multidomain/multidomain.go
@@ -1,0 +1,118 @@
+package main
+
+import (
+	"crypto/rand"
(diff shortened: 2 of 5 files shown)


examples/multidomain/Makefile
```diff
@@ -0,0 +1,6 @@
+writer: clean
+	go get -u github.com/erikdubbelboer/fasthttp
```
erikdubbelboer: Probably should change this 😃

examples/multidomain/multidomain.go
```diff
@@ -0,0 +1,118 @@
...
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/erikdubbelboer/fasthttp"
```
erikdubbelboer: Change this.

examples/multidomain/Makefile
kirillDanshin: indeed :rofl:

Issue #271, opened 2017-06-16
Coarse time causes race condition to time.Local change


vearutop:
https://github.com/valyala/fasthttp/blob/master/coarseTime.go#L22
```
	t := time.Now().Truncate(time.Second)
```
When this line runs in a goroutine it reads `time.Local` with `time.Now()`. Because goroutine starts from `init()` external application has a very limited control over public and exported `time.Local`.

This simple code would trigger race condition
```
func main() {
	time.Local = time.UTC
        // continue app initialization
}
```

erikdubbelboer:
It is not clear why @valyala introduced this coarse time. Benchmarks on
different systems show that the speedup is no where enough to justify
the added code complexity.

See: https://github.com/erikdubbelboer/fasthttp/commit/033e10c58522144bd570ff3aa45fbd2f3df917db (this is a fork I maintain until @valyala starts maintaining this project again)

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/pull/379

kirillDanshin:
fixed, but useful note: do **not** change time.Location. This variable should be the system's timezone, not your service's one. you should probably change system timezone instead of overwriting global variables.

Issue #261, opened 2017-05-26
random fail response from server


smthpickboy:
I'm using fasthttp + fasthttprouter.

fasthttp version: b154429ca951971beae8427b1a918ebce58f64af
fasthttprouter version: ade4e2031af3aed7fffd241084aad80a58faf421

After I update fasthttp from some old version to the current version, I get random failure response from server.

curl: (52) Empty reply from server
curl: (56) Failure when receiving data from the peer

fasthttprouter hasn't been updated for quite some time.

smthpickboy:
I confirmed it's fasthttprouter's problem (maybe incompatible with Go 1.8?)

smthpickboy:
Sorry. After further testing, I found that it's caused by ReadTimeout/WriteTimeout of fasthttp.Server

	func main() {
		m := func(ctx *fasthttp.RequestCtx) {
			switch string(ctx.Path()) {
			case "/foo":
				foo(ctx)
			default:
				ctx.Error("not found", fasthttp.StatusNotFound)
			}
		}

		s := fasthttp.Server{
			Handler: m,
			ReadTimeout:  time.Second + 500*time.Millisecond,
			//WriteTimeout: time.Second,
			MaxRequestBodySize: 2 * 1024,
		}

		fmt.Println("testing with ReadTimeout 1.5s")
		err := s.ListenAndServe("127.0.0.1:8000")
		if err != nil {
			fmt.Println(err)
		}
	}

	func foo(ctx *fasthttp.RequestCtx) {
		ctx.WriteString("Success")
	}


If I set ReadTimeout to 1.5 second on my test machine, It's no problem. But if set to less than 1.5 sec, such as 1 sec, there would be some random fail response.

Is there any additional config that I should do to use ReadTimeout/WriteTimeout of fasthttp.Server?

algorithm-buaa:
```
func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error) {
	if req == nil {
		panic("BUG: req cannot be nil")
	}
	if resp == nil {
		panic("BUG: resp cannot be nil")
	}

	atomic.StoreUint32(&c.lastUseTime, uint32(CoarseTimeNow().Unix()-startTimeUnix))

	// Free up resources occupied by response before sending the request,
	// so the GC may reclaim these resources (e.g. response body).
	resp.Reset()

	cc, err := c.acquireConn()
	if err != nil {
		return false, err
	}
	conn := cc.c

[...]

PR #379, opened 2018-08-17
Remove CoarseTime

erikdubbelboer:
It is not clear why @valyala introduced this coarse time. Benchmarks on different systems show that the speedup is no where enough to justify the added code complexity and bugs it seems to have introduced.
```
Mac:
BenchmarkCoarseTimeNow-8   	2000000000	         2.49 ns/op
0 B/op	       0 allocs/op
BenchmarkTimeNow-8         	500000000	         3.14 ns/op
0 B/op	       0 allocs/op

Ubuntu:
BenchmarkCoarseTimeNow-4   	300000000	         6.74 ns/op
0 B/op	       0 allocs/op
BenchmarkTimeNow-4         	100000000	        15.9 ns/op
0 B/op	       0 allocs/op
```
This reverts commit https://github.com/valyala/fasthttp/commit/6309f42188ecb28ccf8ac58442739cdb43d75d9e and https://github.com/valyala/fasthttp/commit/32c72cde80f0c591604f825586d6a4bbbb39d9c5.

See: https://github.com/valyala/fasthttp/issues/271, https://github.com/valyala/fasthttp/pull/269 and https://github.com/valyala/fasthttp/issues/261.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1041,7 +1041,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		panic("BUG: resp cannot be nil")
 	}
 
-	atomic.StoreUint32(&c.lastUseTime, uint32(CoarseTimeNow().Unix()-startTimeUnix))
+	atomic.StoreUint32(&c.lastUseTime, uint32(time.Now().Unix()-startTimeUnix))
 
 	// Free up resources occupied by response before sending the request,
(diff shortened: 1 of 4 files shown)


kirillDanshin:
We should leave public method for now, but warn about method deprecation and remove init()

erikdubbelboer:
Ok looks good 👍

PR #384, opened 2018-08-17
Do case insensitive comparison of cookie fields

erikdubbelboer:
See: https://github.com/valyala/fasthttp/pull/230


cookie_test.go
```diff
@@ -176,6 +176,7 @@ func TestCookieParse(t *testing.T) {
 	testCookieParse(t, `foo="bar"`, "foo=bar")
 	testCookieParse(t, `"foo"=bar`, `"foo"=bar`)
 	testCookieParse(t, "foo=bar; domain=aaa.com; path=/foo/bar", "foo=bar; domain=aaa.com; path=/foo/bar")
+	testCookieParse(t, "foo=bar; Domain=aaa.com; PATH=/foo/bar", "foo=bar; domain=aaa.com; path=/foo/bar")
```
erikdubbelboer: @kirillDanshin this is a simple test to see if it works by using `Domain=`. I guess I could add tests for all the other attributes as well but that might be a bit overkill?

erikdubbelboer:
Superseded by https://github.com/valyala/fasthttp/pull/410

PR #397, opened 2018-08-29
Case-insensitively cookie attributes

erikdubbelboer:
Accordingly to RFC 6265 - HTTP State Management Mechanism, Set-Cookie header includes cookie attributes that must be treated case-insensitively.

This fixes https://github.com/valyala/fasthttp/pull/183 without causing any additional heap allocations by calling `strings.ToLower`.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -272,27 +272,40 @@ func (c *Cookie) ParseBytes(src []byte) error {
 	c.value = append(c.value[:0], kv.value...)
 
 	for s.next(kv) {
-		if len(kv.key) == 0 && len(kv.value) == 0 {
-			continue
-		}
-		switch string(kv.key) {
-		case "expires":
-			v := b2s(kv.value)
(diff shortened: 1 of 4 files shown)


cookie.go
```diff
@@ -272,27 +272,40 @@ func (c *Cookie) ParseBytes(src []byte) error {
...
-		case "path":
-			c.path = append(c.path[:0], kv.value...)
-		case "":
-			switch string(kv.value) {
-			case "HttpOnly":
-				c.httpOnly = true
-			case "secure":
-				c.secure = true
+		} else {
+			switch kv.key[0] {
```
erikdubbelboer: Using this switch is slightly faster than just `if (compareLowerCaseA) { } else if (compareLowerCaseA) { } ...`.

erikdubbelboer:
I need to sleep more, I already made a completely different pull request for this 😄 https://github.com/valyala/fasthttp/pull/384

Issue #205, opened 2016-11-20
Is that possible to add support for net.ConnState?


velikaww:
I hope there are callbacks for net.ConnState as net/http has, is that possible?
If fine, I'd to do that and raise a PR later.
Thx!

lday0321:
We also need this feature

Jamesxql:
Monitoring connection status changes is a very common requirement.

In `net/http` package, there are five states with client connection:
* StateNew
* StateActive
* StateIdle
* StateHijacked
* StateClosed

I want to be able to register a callback function that is called when a client connection changes state on the `fasthttp.Server` struct. 

Can you @velikaww  submit a PR to complete this function?

tsingson:
https://github.com/valyala/fasthttp/blob/master/expvarhandler/expvar.go 

you want this ?

erikdubbelboer:
A PR for this would be welcome! I'm not sure when I'll have time to implement this myself.

PR #401, opened 2018-08-30
add support for http.ConnState

gotoxu:
issue #205

### DIFF
--- a/server.go
+++ b/server.go
@@ -1545,11 +1545,13 @@ func (s *Server) ServeConn(c net.Conn) error {
 
 	if err != errHijacked {
 		err1 := c.Close()
+		s.setState(c, StateClosed)
 		if err == nil {
 			err = err1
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Do we really want to import all of `net/http` just for `http.ConnState`? It's a big dependency, it add some extra memory usage from global variables and slows down compile time. On the other hand it allows code that uses `http.ConnState` to be reused.

[Server.ServeConn](https://github.com/valyala/fasthttp/blob/master/server.go#L1515) should also call the callbacks.

kirillDanshin:
@erikdubbelboer I'd like to vendor this type

gotoxu:
Sorry, I didn't notice that` fasthttp` doesn't reference `http` package at all. If so, I tend to maintain our own ConnState.

I will change it

PR #402, opened 2018-08-31
Added Server.nextProtos

dgrr:
This implementation allows user to handle crypto/tls.Config.NextProtos to use their own handlers for the negotiated TLS protos like HTTP/2.
Workerpool where changed to adapt WorkerFunc to another conns server with ServeConn type and contains the Server structure which is used to check configured protos.

### DIFF
--- a/server.go
+++ b/server.go
@@ -134,6 +134,9 @@ func ListenAndServeTLSEmbed(addr string, certData, keyData []byte, handler Reque
 // must be limited.
 type RequestHandler func(ctx *RequestCtx)
 
+// ServeHandler must process tls.Config.NextProto negotiated requests.
+type ServeHandler func(c net.Conn) error
+
 // Server implements HTTP server.
 //
 // Default Server settings should satisfy the majority of Server users.
@@ -311,7 +314,8 @@ type Server struct {
 	// By default standard logger from log package is used.
 	Logger Logger
 
-	tlsConfig *tls.Config
+	tlsConfig  *tls.Config
+	nextProtos map[string]ServeHandler
 
 	concurrency      uint32
 	concurrencyCh    chan struct{}
@@ -1242,6 +1246,26 @@ func (ctx *RequestCtx) TimeoutErrorWithResponse(resp *Response) {
 	ctx.timeoutResponse = respCopy
 }
 
+// NextProto adds nph to be processed when key is negotiated when TLS conection is established.
+//
+// This function can only be called before the server is started.
+func (s *Server) NextProto(key string, nph ServeHandler) {
+	if s.nextProtos == nil {
+		s.nextProtos = make(map[string]ServeHandler)
+	}
+	s.configTLS()
+	s.tlsConfig.NextProtos = append(s.tlsConfig.NextProtos, key)
+	s.nextProtos[key] = nph
+}
+
+func (s *Server) hasNextProto(c net.Conn) (proto string) {
+	tlsConn, ok := c.(connTLSer)
+	if ok {
+		proto = tlsConn.ConnectionState().NegotiatedProtocol
+	}
+	return
+}
+
 // tcpKeepAliveListener sets TCP keep-alive timeouts on accepted
 // connections. It's used by ListenAndServe, ListenAndServeTLS and
 // ListenAndServeTLSEmbed so dead TCP connections (e.g. closing laptop mid-download)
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -1187,6 +1191,26 @@ func (ctx *RequestCtx) TimeoutErrorWithResponse(resp *Response) {
 	ctx.timeoutResponse = respCopy
 }
 
+// NextProto adds nph to be processed when key is negotiated when TLS conection is stablished.
+func (s *Server) NextProto(key string, nph ServeHandler) {
```
kirillDanshin: Established

server.go
```diff
@@ -1187,6 +1191,26 @@ func (ctx *RequestCtx) TimeoutErrorWithResponse(resp *Response) {
...
+	if s.nextProtos == nil {
+		s.nextProtos = make(map[string]ServeHandler)
+	}
+	if s.tlsConfig == nil {
+		s.tlsConfig = &tls.Config{}
+	}
+	s.tlsConfig.NextProtos = append(s.tlsConfig.NextProtos, key)
+	s.nextProtos[key] = nph
```
kirillDanshin: Maps and slices are not a threadsafe types. We should either document that this func is not concurrency-safe, or add a mutex here

dgrr: @erikdubbelboer yes. It must be called before *Serve functions.

workerpool.go
```diff
@@ -202,12 +204,20 @@ func (wp *workerPool) workerFunc(ch *workerChan) {
...
 		if c == nil {
 			break
 		}
+		handler = wp.WorkerFunc
+		if proto := wp.s.hasNextProto(c); proto != "" {
+			h, ok := wp.s.nextProtos[proto]
+			if ok {
+				handler = h
+			}
+		}
```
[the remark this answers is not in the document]
dgrr: Yes. It can be put before the `defer s.wg.Done()` as a checking function.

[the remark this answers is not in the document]
erikdubbelboer: I would prefer that.

server.go
```diff
@@ -1187,6 +1191,28 @@ func (ctx *RequestCtx) TimeoutErrorWithResponse(resp *Response) {
...
 
+// NextProto adds nph to be processed when key is negotiated when TLS conection is established.
+//
+// This function MUST NOT be executed by different goroutines.
+func (s *Server) NextProto(key string, nph ServeHandler) {
+	if s.nextProtos == nil {
+		s.nextProtos = make(map[string]ServeHandler)
+	}
+	if s.tlsConfig == nil {
+		s.tlsConfig = &tls.Config{}
```
erikdubbelboer: Please add `PreferServerCipherSuites: true,` here. Otherwise calling `NextProto` before calling `AppendCert` causes it to be `false`. Better yet, add a function that initializes `tlsConfig` and use that function in [`AppendCert`](https://github.com/valyala/fasthttp/blob/b8ce34c94a0b8c3ad3a83c7bab4197373172476d/server.go#L1337) and [`AppendCertEmbed`](https://github.com/valyala/fasthttp/blob/b8ce34c94a0b8c3ad3a83c7bab4197373172476d/server.go#L1361) as well so we always initialize it the same way and avoid possible future bugs where things get added in one place but not in another.

erikdubbelboer:
@kirillDanshin I think this can be merged since all requested changes have been done.

erikdubbelboer:
@dgrr can you please resolve the merge conflict so we can merge this?

server.go
```diff
@@ -1701,7 +1720,14 @@ func nextConnID() uint64 {
 const DefaultMaxRequestBodySize = 4 * 1024 * 1024
 
 func (s *Server) serveConn(c net.Conn) error {
-	defer atomic.AddInt32(&s.open, -1)
+  defer atomic.AddInt32(&s.open, -1)
```
erikdubbelboer: Sorry to be pedantic but please `go fmt`.

erikdubbelboer:
There are still merge conflicts. You can not just `Merge branch 'master' into master`. On your fork. Rebasing doesn't work that way. Assuming the original remote is called `upstream` you should run `git pull --rebase upstream master` and it should do a proper rebase for you where you have to fix the conflicts and then `git rebase --continue` to finish the rebase.

erikdubbelboer:
@dgrr I hope you don't mind that I force pushed to your fork to fix the merge conflicts seeing as it seems you don't know how to do this.

Everything looks good now so I'm going to merge this once the Travis tests are done.

dgrr:
Thank you @erikdubbelboer 
I don't know how to do that and I am developing http2 in another branch and I don't want to merge my http2 progress with master branch or another.

Issue #161, opened 2016-08-23
http.Transport.Dial adapter (for SOCKS5 proxy support)


shicky:
It is more common to find SOCKS5 proxies being used instead of HTTP proxies.

SOCKS5 proxies are currently supported by `golang.org/x/net/proxy`

It would be nice to have some kind of `http.Transport.Dial` adapter for `fasthttp.DialFunc`.

For example, take a look at Socks5 Dialer adapter I implemented.

``` go
func FasthttpSocksDialer(proxyAddr string) func(addr string) (net.Conn, error) {
     return func(addr string) (net.Conn, error) {
         dialer, err := proxy.SOCKS5("tcp", proxyAddr, nil, proxy.Direct)
         if err != nil {
             log.Fatalf("error connecting to socks5 proxy")
         }   
         return dialer.Dial("tcp", addr)
     }   
 }

c := &fasthttp.Client{
    Dial: FasthttpSocksDialer("localhost:9050"),
}
```

Is there a reason why `fasthttp.DialFunc` is not compatible with `http.Transport.Dial`?

bernardolm:
+1

themester:
Why `fasthttp.DialFunc` should be compatible with `http.Transport.Dial`?

erikdubbelboer:
See https://github.com/valyala/fasthttp/pull/405

PR #405, opened 2018-09-01
Add SOCKS5 dialer

erikdubbelboer:
See: https://github.com/valyala/fasthttp/issues/161


erikdubbelboer:
I have put this in a separate package so people importing fasthttp aren't required to have `golang.org/x/net/proxy` as well.

PR #407, opened 2018-09-07
Allows for empty content type by default. Fixes #214

### LINKED ISSUE #214: Content-Type Transparency
Currently, if no `Content-Type` is present in the response header, fasthttp sets a default `Content-Type` to `text/plain; charset=utf-8`:  https://github.com/valyala/fasthttp/blob/ef299a95e2a898859c8652374b33cc7553b1532e/header.go#L270

We would like to provide a transparent proxy and do not want to add (or default) the `Content-Type`, if it is not present. Is there any way around this or would you be interested in a PR that adjusts this behavior?

davidbyttow:


erikdubbelboer:
A minor problem I see is that `ResponseHeader.ContentType()` is a public API as well that is changing behavior now. There might be people who are using `ResponseHeader` directly in which case the behavior will change. But seeing as `ResponseHeader.ContentType()` never contained any documentation about it returning a default I guess it's not really breaking any contract.
Most people will be using the server which doesn't change with this. So I guess I approve.

davidbyttow:
Yeah I thought about that too and reasoned the same thing. 

One alternative would be to change the logic in `ResponseHeader` to instead use `if contentType == nil` rather than `if len(contentType) == 0` and then add `DefaultContentType` to `Server` which defaults to nil and only calls `ResponseHeader.SetContentType` if it's non-nil.

davidbyttow:
Here's an alternative, but while fewer changes, it may be a little bit more brittle/implicit. https://github.com/valyala/fasthttp/pull/407

davidbyttow:
@erikdubbelboer should we merge this?

erikdubbelboer:

kirillDanshin:
I think we should

server.go
```diff
@@ -1702,6 +1713,9 @@ func (s *Server) serveConn(c net.Conn) error {
 		if serverName != nil {
 			ctx.Response.Header.SetServerBytes(serverName)
 		}
+		if contentType != nil {
+			ctx.Response.Header.SetContentTypeBytes(contentType)
```
erikdubbelboer: Doesn't this always overwrite the header no matter if the user set it first or not?

davidbyttow: Well, isn't this called *before* the Handler? So it's just setting prior to user code.

davidbyttow:
OK, I changed this by threading through the flag.

davidbyttow:
FYI @erikdubbelboer ^

server_test.go
```diff
@@ -1456,7 +1461,7 @@ func TestCompressHandler(t *testing.T) {
 	hh(&ctx)
 	s = ctx.Response.String()
 	br = bufio.NewReader(bytes.NewBufferString(s))
-	if err := resp.Read(br); err != nil {
+	if err = resp.Read(br); err != nil {
```
erikdubbelboer: Sorry if I'm being pedantic but please revert this change. It's always better to have variables scoped as tightly as possible.

server_test.go
```diff
@@ -1479,7 +1484,7 @@ func TestCompressHandler(t *testing.T) {
 	h(&ctx)
 	s = ctx.Response.String()
 	br = bufio.NewReader(bytes.NewBufferString(s))
-	if err := resp.Read(br); err != nil {
+	if err = resp.Read(br); err != nil {
```
erikdubbelboer: Same as above.

server_test.go
davidbyttow: The reason this was done is because it's actually a golinter warning about shadowing.

davidbyttow:
Should be good to go @erikdubbelboer

header.go
```diff
@@ -652,6 +653,7 @@ func (h *ResponseHeader) DisableNormalizing() {
 // Reset clears response header.
 func (h *ResponseHeader) Reset() {
 	h.disableNormalizing = false
+	h.noDefaultContentType = false
```
erikdubbelboer: I'm wondering now if this should be done here or in `resetSkipNormalize()`. I don't think it matters much but maybe doing it in `resetSkipNormalize()` is cleaner as everything is reset there.

erikdubbelboer:
@kirillDanshin you have any thoughts about this or can I merge it?

davidbyttow:
fyi @kirillDanshin ^

kirillDanshin:
@erikdubbelboer @davidbyttow ok, let's merge it. sorry for the delay.

PR #408, opened 2018-09-07
Adds ability to override defaultContentType for responses

davidbyttow:
Note, this is just an alternative to https://github.com/valyala/fasthttp/pull/407 for academic purposes

### DIFF
--- a/server.go
+++ b/server.go
@@ -276,6 +276,8 @@ type Server struct {
 	// value is explicitly provided during a request.
 	NoDefaultServerHeader bool
 
+	DefaultContentType []byte
+
 	// ConnState specifies an optional callback function that is
 	// called when a client connection changes state. See the
 	// ConnState type and associated constants for details.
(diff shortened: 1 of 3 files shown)


server.go
```diff
@@ -1702,6 +1705,9 @@ func (s *Server) serveConn(c net.Conn) error {
 		if serverName != nil {
 			ctx.Response.Header.SetServerBytes(serverName)
 		}
+		if contentType != nil {
+			ctx.Response.Header.SetContentTypeBytes(contentType)
```
erikdubbelboer: But now it will always overwrite the user set content type with the default content type?

erikdubbelboer:
There is one big issue with this solution. Now parsing a response will always allocate a new slice on the heap, even when reusing the response struct as `contentType` is set to `nil` instead of an empty slice reusing the same memory.

davidbyttow:
Going to close this, as I'm not really a fan of the approach.

PR #412, opened 2018-09-11
Adds support for max-age cookie value. Fixes #184

### LINKED ISSUE #184: Cookies with max-age
I was looking at the [RFC 6265](https://tools.ietf.org/html/rfc6265#section-5.3) (item 3 exactly) and there's specifying that cookie attribute max-age has priority over expires. 
fasthttp does not handle the cookie max-age attribute, but it would be nice to have the behaviour described on RFC.

davidbyttow:
Issue: #184 

Note, because `max-age` takes precedence over expires, it only supports `max-age` when encoded/decoded. However, it does not allow the client to set `max-age` at the API and prefers `SetExpire` instead.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"errors"
 	"io"
-	"strconv"
 	"sync"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


davidbyttow:
Addressed feedback @erikdubbelboer, made the API simple and only removes expires if max-age is set.

PR #414, opened 2018-09-13
Remove redundant else statements

cristaloleg:


erikdubbelboer:
I does exactly the same but I guess it's a bit cleaner this way. Thanks.

PR #415, opened 2018-09-13
Support Go modules

cristaloleg:


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/valyala/fasthttp
+
+require (
+	github.com/klauspost/compress v1.4.0
+	github.com/klauspost/cpuid v0.0.0-20180405133222-e7e905edc00e // indirect
+	github.com/valyala/bytebufferpool v1.0.0
```
kirillDanshin: are you sure this vendoring would not break backward compatibility? folks can't use types from our vendored packages as types for unvendored packages

kirillDanshin: > Currently we do not export any vendored types. 

this exactly what we do. https://github.com/valyala/fasthttp/blob/master/bytebuffer.go#L16

cristaloleg: Ouch. Completely missed that ;(

cristaloleg: From the other side, currently we're exposing same version, same as in `go.mod`. So no breaking changes at the moment.

kirillDanshin: @cristaloleg that's not compatible. if we vendor bytebufferpool, the type will be `github.com/valyala/fasthttp/vendor/github.com/valyala/bytebufferpool`. other packages can expect `github.com/valyala/bytebufferpool`, but it cannot be fixed by just replacing import path, 'cause you can't import vendored packages. unfortunately, directly used packages incompatible with vendored ones.

cristaloleg: There is no vendor directory in the PR :)

kirillDanshin: @cristaloleg go get with go mod will create vendor locally.

erikdubbelboer: `fasthttp.ByteBuffer` has been [deprecated for more than 2 years](https://github.com/valyala/fasthttp/commit/a35f26a704bb8fdf20f7cbf9b80049b039595b46). I suggest we remove this code completely to make it possible to add go mod support.

What we could do is tag a version before we remove this that people could continue using if they are too lazy to update their code. And then tag a new version after with the code removed and go mod support added.

cristaloleg: Yes, making a long-awaited release will help us a lot 👍 
Should we discuss it in a new issue?

erikdubbelboer: @kirillDanshin Do you agree with removing `fasthttp.ByteBuffer`?

kirillDanshin: @erikdubbelboer  I doubt that we will not break anyone's software.
I think we should notify subscribed users about removing published type and give two to three more days to recheck their software

erikdubbelboer: `fasthttp.ByteBuffer` has been removed so I'm merging this.

PR #424, opened 2018-09-20
Enable TCP KeepAlive

athurg:
To fix HTTPCode_ELB_5XX_Count problem on AWS ApplicationELB with app in docker.
https://github.com/moby/moby/issues/31208

Copy from golang http package
https://github.com/golang/go/blob/master/src/net/http/server.go

### DIFF
--- a/README.md
+++ b/README.md
@@ -550,7 +550,7 @@ uintBuf := fasthttp.AppendUint(nil, 1234)
   
   **NOTE**: Go 1.9.7 is the oldest tested version. We recommend you to update as soon as you can. As of 1.11.3 we will drop 1.9.x support.
 
-* *Please provide real benchmark data and sever information*
+* *Please provide real benchmark data and server information*
 
   See [this issue](https://github.com/valyala/fasthttp/issues/4).
 


server.go
```diff
@@ -1204,15 +1204,38 @@ func (ctx *RequestCtx) TimeoutErrorWithResponse(resp *Response) {
...
+	*net.TCPListener
+}
+
+func (ln tcpKeepAliveListener) Accept() (net.Conn, error) {
+	tc, err := ln.AcceptTCP()
+	if err != nil {
+		return nil, err
+	}
+	tc.SetKeepAlive(true)
+	tc.SetKeepAlivePeriod(3 * time.Minute)
```
kirillDanshin: I would prefer to have an ability to change keepalive period

athurg: OK, I agree with that , and it is necessary. I'll add a new method on it.

erikdubbelboer: @athurg just add an option to the `Server` struct.

athurg:
Closed, reopen with a specified branch in #427

PR #430, opened 2018-10-05
Fix client example

nikandfor:
Am I right that there should be like this?

### DIFF
--- a/client_example_test.go
+++ b/client_example_test.go
@@ -24,7 +24,7 @@ func ExampleHostClient() {
 	useResponseBody(body)
 
 	// Fetch foobar page via local proxy. Reuse body buffer.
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Its not technically required because before reading the body [the body buffer is reset](https://github.com/valyala/fasthttp/blob/63ea2d380ff30e6f09dc8c39205363c7c064bc31/http.go#L946) which [does the same thing](https://github.com/valyala/bytebufferpool/blob/cdfbe9377474227bb42120c1e22fd4433e7f69bf/bytebuffer.go#L110).

That's why [the doc says](https://godoc.org/github.com/valyala/fasthttp#Client.Get):
```
The contents of dst will be replaced by the body and returned, if the dst is too small a new slice will be allocated. 
```

PR #449, opened 2018-10-30
Make sure to always release resources

amezghal:
Make sure to always release resources in the event of timeout

### DIFF
--- a/client.go
+++ b/client.go
@@ -961,12 +961,6 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	// usually continue execution on the host.
 	go func() {
 		ch <- c.Do(reqCopy, respCopy)
-	}()
-
-	tc := acquireTimer(timeout)
-	var err error
-	select {
-	case err = <-ch:
 		if resp != nil {
 			respCopy.copyToSkipBody(resp)
 			swapResponseBody(resp, respCopy)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -961,12 +961,6 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	// usually continue execution on the host.
 	go func() {
 		ch <- c.Do(reqCopy, respCopy)
```
erikdubbelboer: Now you have a race condition where `clientDoDeadline` returns and the calling code probably uses `resp` but the code below is also still copying the response into `resp`.
To fix this you should save the return value in a variable and put it into the channel at the end of this function.

amezghal: i will try to fix this if possible

amezghal: what about this:

```
cleanup := false
	go func() {
		ch <- c.Do(reqCopy, respCopy)
		if cleanup{
			if resp != nil {
				respCopy.copyToSkipBody(resp)
				swapResponseBody(resp, respCopy)
			}
			swapRequestBody(reqCopy, req)
			ReleaseResponse(respCopy)
			ReleaseRequest(reqCopy)
			errorChPool.Put(chv)
		}
	}()

	tc := acquireTimer(timeout)
	var err error
	select {
	case err = <-ch:
		if resp != nil {
			respCopy.copyToSkipBody(resp)
			swapResponseBody(resp, respCopy)
		}
		swapRequestBody(reqCopy, req)
		ReleaseResponse(respCopy)
		ReleaseRequest(reqCopy)
		errorChPool.Put(chv)
	case <-tc.C:
		cleanup = true
		err = ErrTimeout
	}
	releaseTimer(tc)
```

erikdubbelboer: Its much easier to do:
```go
go func() {
	err := c.Do(reqCopy, respCopy)

	// cleanup code here

	ch <- err
}()
```

amezghal: you will cleanup twice, unless you remove the cleanup code from the 
`case err = <-ch:`
but this looks like a valid cleanup, what do you think ?

PR #450, opened 2018-10-30
Cleanup resources

amezghal:
cleanup resources in case of timeout errors

### DIFF
--- a/client.go
+++ b/client.go
@@ -959,8 +959,27 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	// Without this 'hack' the load on slow host could exceed MaxConns*
 	// concurrent requests, since timed out requests on client side
 	// usually continue execution on the host.
+	var cleanup int32
+	atomic.StoreInt32(&cleanup, 0)
+
 	go func() {
 		ch <- c.Do(reqCopy, respCopy)
+		if atomic.LoadInt32(&cleanup) == 1 {
+			if resp != nil {
+				respCopy.copyToSkipBody(resp)
+				swapResponseBody(resp, respCopy)
+			}
+			swapRequestBody(reqCopy, req)
+			ReleaseResponse(respCopy)
+			ReleaseRequest(reqCopy)
+			errorChPool.Put(chv)
+		}
+	}()
+
+	tc := acquireTimer(timeout)
+	var err error
+	select {
+	case err = <-ch:
 		if resp != nil {
 			respCopy.copyToSkipBody(resp)
 			swapResponseBody(resp, respCopy)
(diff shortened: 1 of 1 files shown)


amezghal:
I have a working solution, but it require to set resp to nil,

so in case of timeout i will set resp to nil (since it's thread safe),
and once c.Do completes i will check if resp is nil then i will cleanup resources, and i will not notify the channel, otherwise i will notify the channel and keep the code as is.
```
go func() {
	errDo := c.Do(reqCopy, respCopy)
	if resp == nil {
		ReleaseResponse(respCopy)
		ReleaseRequest(reqCopy)
		errorChPool.Put(chv)
	} else {
		ch <- errDo
	}
}()

tc := acquireTimer(timeout)
var err error
select {
case err = <-ch:
	if resp != nil {
		respCopy.copyToSkipBody(resp)
		swapResponseBody(resp, respCopy)
	}
	swapRequestBody(reqCopy, req)
	ReleaseResponse(respCopy)
	ReleaseResponse(respNotif)
	ReleaseRequest(reqCopy)
	errorChPool.Put(chv)
case <-tc.C:
	resp = nil
	err = ErrTimeout
}
```

amezghal:
This is the final code

```
        var cleanup int32
	atomic.StoreInt32(&cleanup, 0)

	go func() {
		errDo := c.Do(reqCopy, respCopy)
		if atomic.LoadInt32(&cleanup) == 1 {
			ReleaseResponse(respCopy)
			ReleaseRequest(reqCopy)
			errorChPool.Put(chv)
		} else {
			ch <- errDo
		}
	}()

	tc := acquireTimer(timeout)
	var err error
	select {
	case err = <-ch:
		if resp != nil {
			respCopy.copyToSkipBody(resp)
			swapResponseBody(resp, respCopy)
		}
		swapRequestBody(reqCopy, req)
		ReleaseResponse(respCopy)
		ReleaseRequest(reqCopy)
		errorChPool.Put(chv)
	case <-tc.C:
		atomic.StoreInt32(&cleanup, 1)
		err = ErrTimeout
	}
	releaseTimer(tc)
```

erikdubbelboer:
That seems good to me. I would change one last thing. Since `cleanup` is set before the goroutine is started you don't need any atomics yet. So I would change the first 2 lines to `cleanup := int32(0)`.

erikdubbelboer:
Another minor thing, you can put the `errorChPool.Put(chv)` after the `atomic.StoreInt32(&cleanup, 1)` to clean it up a little bit faster as it isn't needed anymore after that.

amezghal:
Great, then i will submit a new pull request,
Meanwhile a was able to change the code, in order to close the underlying net.Conn, 
if the request timeouted i don't see why we do need to keep the connection running in the background,

i have already run this in our production code.

what do you think ?

erikdubbelboer:
You mean closing the connection on timeout? I think keeping it is better as it might be reused in the future.

erikdubbelboer:
@amezghal can you add the last changes to your pull request so we can merge it?

amezghal:
Hi, i have submitted  the pull request,
i didn't use atomic as suggested, waiting if all checks successed,

On other hand for our needs, we did an other refactoring in order to close net.Conn, when it's timeouted, and we see better performances.

PR #452, opened 2018-11-01
Added method (*Server).GetCurrenctConcurrency()

xaionaro:
Hello.

We need to monitor how many incoming concurrent do we have at a moment. So we need a method like this one. It would be nice if you merge it :)


xaionaro:
@erikdubbelboer 
Sure :)

erikdubbelboer:
Thanks

Issue #453, opened 2018-11-05
About connTLSer and NextProto


dgrr:
Hello everyone,

I have question... There are any reason to use connTLSer instead *tls.Conn? Look at this two pieces of code:
https://github.com/valyala/fasthttp/blob/8f4c5d5d1dbb772a29f059645f0f03a4556c1dad/server.go#L1270-L1276

https://github.com/dgrr/fasthttp/blob/nextproto/server.go#L1270-1276

In that case it is required to use *tls.Conn to solve the NextProto problem.

dgrr:
[Fixed](https://github.com/valyala/fasthttp/commit/3fb1313e335481ed94a34b8a8e480ce056245e5b)

PR #454, opened 2018-11-05
Fixed NextProto error caused by non-Handshake call.

dgrr:
As you can see here https://golang.org/src/crypto/tls/conn.go#L1041
crypto/tls.Conn does not perform a TLS handshake until Read or Write functions are called.
Then the solution is to call crypto/tls.Conn.Handshake() to check Client values.

Solves https://github.com/valyala/fasthttp/issues/453

### DIFF
--- a/server.go
+++ b/server.go
@@ -1267,9 +1267,10 @@ func (s *Server) NextProto(key string, nph ServeHandler) {
 	s.nextProtos[key] = nph
 }
 
-func (s *Server) hasNextProto(c net.Conn) (proto string) {
-	tlsConn, ok := c.(connTLSer)
+func (s *Server) hasNextProto(c net.Conn) (proto string, err error) {
+	tlsConn, ok := c.(*tls.Conn)
 	if ok {
+		err = tlsConn.Handshake()
 		proto = tlsConn.ConnectionState().NegotiatedProtocol
 	}
 	return
@@ -1747,11 +1748,14 @@ const DefaultMaxRequestBodySize = 4 * 1024 * 1024
 func (s *Server) serveConn(c net.Conn) error {
 	defer atomic.AddInt32(&s.open, -1)
 
-	if proto := s.hasNextProto(c); proto != "" {
+	if proto, err := s.hasNextProto(c); err != nil {
+		return err
+	} else {
 		handler, ok := s.nextProtos[proto]
 		if ok {
 			return handler(c)
 		}
+		// TODO: else continue?
 	}
 
 	var serverName []byte
--- a/server.go
+++ b/server.go
@@ -546,6 +546,7 @@ func (ctx *RequestCtx) VisitUserValues(visitor func([]byte, interface{})) {
 }
 
 type connTLSer interface {
+	Handshake() error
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
So the problem is that `tlsConn.ConnectionState().NegotiatedProtocol` isn't available yet because there was no handshake yet?

Can we then also add something to return a proper response if someone tries to connect to a https endpoint with a http connection? In that case the `Handshake()` will probably return an error.

dgrr:
In the case where client connects to a https endpoint with http connection the tlsConn.Handshake won't be executed. So in that case the connection will continue being processing by fasthttp. I can add a method to detect HTTP to HTTPS connection and treat that as an error.

server.go
```diff
@@ -1747,11 +1749,14 @@ const DefaultMaxRequestBodySize = 4 * 1024 * 1024
...
 
-	if proto := s.hasNextProto(c); proto != "" {
+	if proto, err := s.hasNextProto(c); err != nil {
+		return err
+	} else {
 		handler, ok := s.nextProtos[proto]
 		if ok {
 			return handler(c)
 		}
+		// TODO: else continue?
```
erikdubbelboer: You can remove this comment, it just needs to continue, no TODO necessary.

server.go
```diff
@@ -1267,9 +1268,10 @@ func (s *Server) NextProto(key string, nph ServeHandler) {
 	s.nextProtos[key] = nph
 }
 
-func (s *Server) hasNextProto(c net.Conn) (proto string) {
+func (s *Server) hasNextProto(c net.Conn) (proto string, err error) {
 	tlsConn, ok := c.(connTLSer)
 	if ok {
+		err = tlsConn.Handshake()
 		proto = tlsConn.ConnectionState().NegotiatedProtocol
```
erikdubbelboer: It's weird to do the `ConnectionState()` call after `Handshake()` already might have returned an error. Maybe rewrite the function to:
```go
func (s *Server) getNextProto(c net.Conn) (string, error) {
	if tlsConn, ok := c.(connTLSer); ok {
		if err := tlsConn.Handshake(); err != nil {
			return "", err
		} else 
			return tlsConn.ConnectionState().NegotiatedProtocol, nil
		}
	} else {
		return "", nil
	}
}
```
I also renamed the function to `getNextProto` as it gets the proto, it doesn't check if it has any as the previous name implied. I would expect a function called `hasNextProto` to return a `bool`.

dgrr: I want to keep the named parameters. See [this article](https://blog.minio.io/golang-internals-part-2-nice-benefits-of-named-return-values-1e95305c8687). So a possible implementation of this function could be:
```go
func (s *Server) getNextProto(c net.Conn) (proto string, err error) {
  tlsConn, ok := c.(connTLSer)
  if ok {
    err = tlsConn.Handshake()
    if err == nil {
      proto = tlsConn.ConnectionState().NegotiatedProtocol
    }
  }
  return
}
```

erikdubbelboer: Yes that looks good 👍

erikdubbelboer:
Thanks.

Issue #457, opened 2018-11-09
Allow to return user-specific responses even on fasthttp errors


derElektrobesen:
If we have a limit for the request body and this limit exceeded, fasthttp server returns default error:
```
$ curl -gvv 'http://127.0.0.1:9998/' -H 'Content-Length: 127890' -X POST
* About to connect() to 127.0.0.1 port 9998 (#0)
*   Trying 127.0.0.1...
* Connected to 127.0.0.1 (127.0.0.1) port 9998 (#0)
> POST /batch HTTP/1.1
> User-Agent: curl/7.29.0
> Host: 127.0.0.1:9998
> Accept: */*
> Content-Length: 127890
>
< HTTP/1.1 400 Bad Request
< Server: fasthttp
< Date: Fri, 09 Nov 2018 09:55:39 GMT
< Content-Type: text/plain; charset=utf-8
< Content-Length: 26
< Connection: close
<
* Closing connection 0
Error when parsing request
```

We need to redeclare this response: content type should be in JSON format: there are a lot of old mobile applications (for older platforms which are not supported but used by many users) which could crash when unexpected response comes.

erikdubbelboer:
This is now be possible since https://github.com/valyala/fasthttp/commit/e771b6fe43f02a3abec94adf6100ea4243bd12d6

derElektrobesen:
Thanks ^^

PR #458, opened 2018-11-09
#457: allow to rewrite system error response

derElektrobesen:


erikdubbelboer:
I think this is a feature other people have requested before. I would like to see some changes to make this more usable before I can accept it.

I think it would be useful to add some documentation on what kind of errors the user can expect as argument to the `ErrorHandler` function. I think it's error that can be returned by Go while reading from a socket such as `io.EOF` or`io.ErrUnexpectedEOF`. It can be `ErrBodyTooLarge`.
But it can also be `errGetOnly`, so I think its best if you change [`errGetOnly`](https://github.com/valyala/fasthttp/blob/8f4c5d5d1dbb772a29f059645f0f03a4556c1dad/http.go#L850) to be exported.
Can you also turn [this](https://github.com/valyala/fasthttp/blob/8f4c5d5d1dbb772a29f059645f0f03a4556c1dad/http.go#L1674) into a properly exported error and add it to the list.
Then I think only [these](https://github.com/valyala/fasthttp/blob/8f4c5d5d1dbb772a29f059645f0f03a4556c1dad/http.go#L1691-L1707) errors remain. I think its best if you turned these into a custom error struct that is exported so users can do a type test for it. Since they contain formatted messages we can't just have an exported error like the others. The custom error struct only has to contain an unexported string and a `Error() string` function that returns that string.

derElektrobesen:
Thanks for review.

All fixes are done

http.go
```diff
@@ -1652,6 +1654,11 @@ func appendBodyFixedSize(r *bufio.Reader, dst []byte, n int) ([]byte, error) {
 	}
 }
 
+// ErrBrokenChunks is returned when server got broken chunked body (Transfer-Encoding: chunked).
```
erikdubbelboer: Can you change this to:
```
ErrBrokenChunks is returned when server receives a broken chunked body (Transfer-Encoding: chunked).
```

http.go
```diff
@@ -847,7 +847,9 @@ func (req *Request) Read(r *bufio.Reader) error {
 
 const defaultMaxInMemoryFileSize = 16 * 1024 * 1024
 
-var errGetOnly = errors.New("non-GET request received")
+// ErrGetOnly is returned when server expects only GET requests,
+// but some other type of request came (Server.GetOnly option is enabled).
```
erikdubbelboer: Can you change `enabled` to `true`.

server.go
```diff
@@ -152,6 +152,17 @@ type Server struct {
 	// Handler for processing incoming requests.
 	Handler RequestHandler
 
+	// ErrorHandler for returning an error response in user-defined way.
```
erikdubbelboer: Can you change this to:
```
ErrorHandler for returning a response in case of an error while receiving or parsing the request.
```

server.go
```diff
@@ -152,6 +152,17 @@ type Server struct {
...
+	//
+	// Following errors will be captured with this handler:
```
erikdubbelboer: Since I'm not sure if these are really all the errors (I think more `io.*` errors are possible it's probably better to change this to:
```
The following is a non-exhaustive list of errors that can be expected as argument:
```

http.go
```diff
@@ -1652,6 +1654,11 @@ func appendBodyFixedSize(r *bufio.Reader, dst []byte, n int) ([]byte, error) {
 	}
 }
 
+// ErrBrokenChunks is returned when server got broken chunked body (Transfer-Encoding: chunked).
+type ErrBrokenChunks struct {
```
erikdubbelboer: Sorry for being pedantic but can you change this to `ErrBrokenChunk`. The error is only about one chunk being broken, not multiple.

PR #464, opened 2018-11-15
cleanup resources

amezghal:


client.go
```diff
@@ -959,8 +959,16 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	// Without this 'hack' the load on slow host could exceed MaxConns*
 	// concurrent requests, since timed out requests on client side
 	// usually continue execution on the host.
+
+	cleanup := false
 	go func() {
-		ch <- c.Do(reqCopy, respCopy)
+		errDo := c.Do(reqCopy, respCopy)
+		if cleanup {
```
erikdubbelboer: Without `sync/atomic` this will result in a race condition where one goroutine is reading `cleanup` and the other is possibly at the same time writing to it (if the timeout occurs at the same time as `c.Do` returning).

client.go
```diff
@@ -959,8 +959,17 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	// Without this 'hack' the load on slow host could exceed MaxConns*
 	// concurrent requests, since timed out requests on client side
 	// usually continue execution on the host.
+
+	var cleanup int32
+	atomic.StoreInt32(&cleanup, 0)
```
erikdubbelboer: Since the other gorountine hasn't started here and the default value of an `int32` is `0` you don't need this line at all.

client.go
```diff
@@ -976,6 +985,8 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 		ReleaseRequest(reqCopy)
 		errorChPool.Put(chv)
 	case <-tc.C:
+		atomic.StoreInt32(&cleanup, 1)
+		errorChPool.Put(chv)
```
erikdubbelboer: I just figured out that this line can cause a lot of problems. What if the request had an error just before the timeout. In that case the request goroutine will write to `ch` but the timeout case will put `ch` back into the pool. So now we have a channel with a value already in there in the pool. This will then cause the next request to get the channel from the pool to immediately hit the case with the previous error.
The solution is to move this line to line `970` so we only reuse the channel if we know a cleanup is needed and we know we haven't written to it yet.

amezghal: but the second case will never be executed, 
only one case will be executed ??

erikdubbelboer: Why wouldn't the second case be executed? Since its two different goroutines its definitely possible that one is just entering the timeout case without having changed cleanup yet, and the other goroutine is just after the cleanup check but hasn't put the value into the channel yet.

amezghal: Ok, I've already moved the channel pool release to the upper coroutine as suggested

client.go
erikdubbelboer: @amezghal if you just remove this line then it's perfect and I can merge it.

amezghal:
Ok, it's done

PR #474, opened 2018-11-19
Disable sleep when concurrency limits exceeded.

kshvakov:


erikdubbelboer:
Hmm I never noticed this sleep. Judging from the comment it is based on the assumption that multiple processes are listening on the same port handling new connections. I think for 99% of the fasthttp users this isn't the case so I think it might be better to turn the condition around and disable the sleep by default. @kirillDanshin what do you think?

xaionaro:
This PR is not actual anymore. See: https://github.com/valyala/fasthttp/pull/485

Issue #470, opened 2018-11-18
Keep the read and write buffers assigned to the connection, unless the `ReduceMemoryUsage` is configured


cipriancraciun:
I've looked into the connection handler code, and saw that the request and response buffers are put back into the pool immediately after each request.

However this leads to a performance overhead on busy keep-alive connections.

----

Thus, given that the buffer sizes and the connection keep alive time and maximum requests can be explicitly configured, an idea would be to keep these buffers attached to the connection, unless the `ReduceMemoryUsage` configuration is enabled.

The following patch tries to solve this:

  https://github.com/cipriancraciun/go-fasthttp/commit/03ca6ec36bb5d32144d842f03ecc9b4f1b71cdf7

----

Please note that last chunk at line 1949 it also introduces a `bw.Flush`.  I think that the lack of `bw.Flush` is either a left-over from when `fasthttp` used to support pipelining or is perhaps for the use-case of connection hijacking.  However given I don't have a good "mental-map" of all the states, the core developers might properly address this last chunk of the patch.

erikdubbelboer:
Looks good to me on first look. Are you already using this for busy keep-alive connections somewhere?

cipriancraciun:
Yes, but not in production, I've used it only in benchmarks with up to 4096 concurrent keep-alive connections, and it works fine.

However this change might be disruptive for those that have configured large buffers and expect non-busy keep-alive connections.  Therefore perhaps another configuration knob (disabled by default) might be better?

erikdubbelboer:
I'm testing this now with production traffic.

I don't think we need to add another flag for this. I don't think there are many people who use huge buffers with keep-alive connections that are inactive for most of the time. People that want low memory usage probably already use `ReduceMemoryUsage` anyways.

cipriancraciun:
I've created pull request #476 for this if accepted.

erikdubbelboer:
I merged the pull request.

PR #476, opened 2018-11-19
Update server to not release read and write buffers if `ReduceMemoryUsage` is not set

cipriancraciun:
See issue #470.


erikdubbelboer:
Thanks.

PR #485, opened 2018-12-06
Added method (*Server).GetOpenConnectionsCount() and option DisableSleepWhenConcurrencyLimitsExceeded

xaionaro:
The method `(*Server).GetOpenConnectionsCount()` is required for our monitoring system

The option `DisableSleepWhenConcurrencyLimitsExceeded` is required in our high-load application. This `Sleep()` caused some problems in our infrastructure when servers were overloaded so we'd like to be able to disable it if possible :)

### DIFF
--- a/server.go
+++ b/server.go
@@ -16,10 +16,6 @@ import (
 	"time"
 )
 
-const (
-	defaultSleepDurationWhenConcurrenyLimitExceeded = time.Millisecond * 100
-)
-
 var errNoCertOrKeyProvided = errors.New("Cert or key has not provided")
 
 var (
@@ -307,9 +303,8 @@ type Server struct {
 	DisableHeaderNamesNormalizing bool
 
 	// SleepWhenConcurrencyLimitsExceeded is a duration to be slept of if
-	// the concurrency limit in exceeded (default [when is 0]: 100ms;
-	// any negative value to disable the sleeping and accept new connections
-	// immidiatelly).
+	// the concurrency limit in exceeded (default [when is 0]: don't sleep
+	// and accept new connections immidiatelly).
 	SleepWhenConcurrencyLimitsExceeded time.Duration
 
 	// NoDefaultServerHeader, when set to true, causes the default Server header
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -1566,7 +1570,9 @@ func (s *Server) Serve(ln net.Listener) error {
 			//
 			// There is a hope other servers didn't reach their
 			// concurrency limits yet :)
-			time.Sleep(100 * time.Millisecond)
+			if !s.DisableSleepWhenConcurrencyLimitsExceeded {
+				time.Sleep(100 * time.Millisecond)
```
erikdubbelboer: Can you switch this around and make the option `SleepWhenConcurrencyLimitsExceeded` and make it a `time.Duration` of how long to sleep (with 0 obviously no sleep at all). I think the default of not sleeping at all is better and making it possible to configure how long to sleep is also better.

server.go
```diff
@@ -1741,6 +1747,13 @@ func (s *Server) GetCurrentConcurrency() uint32 {
 	return atomic.LoadUint32(&s.concurrency)
 }
 
+// GetOpenConnectionsCount returns a number of opened connections.
+//
+// This function is intended be used by monitoring systems
+func (s *Server) GetOpenConnectionsCount() int32 {
+	return atomic.LoadInt32(&s.open) - 1
```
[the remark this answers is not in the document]
xaionaro: Because of line: https://github.com/valyala/fasthttp/blob/3c7913fc29526508ce2e8724810cf1699826c1a4/server.go#L1542

So if I don't do `- 1` then I get a wrong value. I see `2` if there's only one incoming open connection.

server.go
xaionaro: Fixed.

But we cannot do `0` as "no sleep at all" here. Server is initialized directly (without any factory or whatever) so I cannot set a default value for `SleepWhenConcurrencyLimitsExceeded`. So `0` will be the default value. And I think we don't want to change the default behavior. So I made `0` to be overrided by a real default value `100ms`. And to disable the sleeping it's required to set any negative value.

erikdubbelboer: That's what I mean, I want to change the default behavior as for most people it doesn't make sense. The default behavior now only makes sense if you have multiple processes listening on the same port. Most people don't even know how to set this up and won't be using a setup like this.

xaionaro: Ah, nice. Fixed :)

erikdubbelboer:
Thanks.

xaionaro:
Thank you.

PR #488, opened 2018-12-12
Support SameSite cookie attribute

winduptoy:
This is an implementation of the `SameSite` attribute for cookies. It is inspired by go v1.11's `net/http` implementation of the SameSite attribute, seen here: https://golang.org/src/net/http/cookie.go?s=4244:4276#L214

This PR addresses issue #489.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -18,6 +18,14 @@ var (
 	CookieExpireUnlimited = zeroTime
 )
 
+type CookieSameSite int
+const (
+	CookieSameSiteDisabled CookieSameSite = iota
+	CookieSameSiteDefaultMode
+	CookieSameSiteLaxMode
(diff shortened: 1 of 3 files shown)


cookie.go
```diff
@@ -18,6 +18,14 @@ var (
 	CookieExpireUnlimited = zeroTime
 )
 
+type CookieSameSite int
```
erikdubbelboer: Can you add a comment like:
```go
// See https://tools.ietf.org/html/draft-ietf-httpbis-cookie-same-site-00 for details.
```

erikdubbelboer:
Very nice, just one small addition requested and then it can be merged.

erikdubbelboer:
Thanks.

PR #490, opened 2018-12-13
Add RequestHeader.VisitAllInOrder

### LINKED ISSUE #486: Feature request: get request headers in the same order they arrived
It would be great if RequestHeader.VisitAll method could pass request headers to 'f' in the same order they originally arrived.

erikdubbelboer:
Fixes #486

I'm actually wondering if we should just change `VisitAll` to use this code instead. The current code for `VisitAll` was written before `rawHeaders` was added and now goes though all kinds of hoops to put things like the cookie header back together. My benchmark below shows that VisitAllInOrder is only slightly slower. @kirillDanshin what do you think?

```
BenchmarkVisitAllInOrder-8   	10000000	       141 ns/op	       0 B/op	       0 allocs/op
BenchmarkVisitAll-8          	30000000	      44.5 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/header.go
+++ b/header.go
@@ -778,6 +778,8 @@ func (h *RequestHeader) VisitAllCookie(f func(key, value []byte)) {
 //
 // f must not retain references to key and/or value after returning.
 // Copy key and/or value contents before returning if you need retaining them.
+//
+// To get the headers in order they were received use VisitAllInOrder.
 func (h *RequestHeader) VisitAll(f func(key, value []byte)) {
 	h.parseRawHeaders()
 	host := h.Host()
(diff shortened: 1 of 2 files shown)


kirillDanshin:
@erikdubbelboer I think we should not change VisitAll to use this code instead. Almost 100ns is a significant performance overhead for me, and I don't require data to be in order

erikdubbelboer:
@kirillDanshin What are you doing that 100ns makes a difference? If you're doing `VisitAll` 1000 times per request it would still only be a difference of 0.1 millisecond.

kirillDanshin:
@erikdubbelboer
I'm working on a crypto exchange. In our case, 100ns in one place, another 100ns in another and so on, you know. If you need, you can use ordered Visitor, if you don't — you can use faster and unordered Visitor. Why should we force anybody to use an implementation that about 3.15 times slower?

erikdubbelboer:
I guess fasthttp is all about speed and keeping everything as fast as possible should be out objective. But sometimes when the difference is really small (like in this case) I wonder if keeping the code simpler and more maintainable is more important.

kirillDanshin:
@erikdubbelboer you see, for that project using an implementation that more than three times slower is noticeable when we are talking about 44.5ns vs. 141ns. of course, 1ns vs. 3ns isn't so much, but in this particular case, it's almost a hundred ns difference that may be optional.

illotum:
I feel that if you hunt nanoseconds you might be better switching to soft realtime platforms. Erlang, Rust, there's plenty to choose from.

That is, I'm on Erik's side here.

kirillDanshin:
@illotum golang is a better choice for projects like this. it's a huge [off] topic, but in brief:
- rust has a memory owner pattern, that doesn't fit for projects like this
- erlang has memory management issues and is slower generally speaking.

anyway, I think you missed my main point in previous messages: there's no excuse when you are slowing down and changing the behavior of a method just because of a feature request that created three years after the original method introduced.

Issue #495, opened 2018-12-16
panic when writing on closed hijacked connection.


dgrr:
Hello,

I was having troubles with fastws concurrent connections receiving panics when a goroutine closes the descriptor and another goroutine writes in connection after closing this descriptor. This case causes the following panic:
```
github.com/valyala/fasthttp.(*hijackConn).Write(0xc000094ae0, 0xc000018230, 0x2, 0xe, 0xc000134100, 0xc00014c580, 0xc000046b28)
        <autogenerated>:1 +0x2e
github.com/dgrr/fastws.(*Frame).WriteTo(0xc000134100, 0x7fd7d96870e8, 0xc000094ae0, 0x7fd7d96870e8, 0xc000094ae0, 0x0)
        /home/dgrr/.go/src/github.com/dgrr/fastws/frame.go:362 +0xb4
github.com/dgrr/fastws.(*Conn).sendClose(0xc00008c400, 0x6e03e8, 0x0, 0x0, 0x0, 0xc00006e780, 0xc000024000)
        /home/dgrr/.go/src/github.com/dgrr/fastws/conn.go:314 +0x166
github.com/dgrr/fastws.(*Conn).CloseString(0xc00008c400, 0x0, 0x0, 0x0, 0x0)
        /home/dgrr/.go/src/github.com/dgrr/fastws/conn.go:358 +0x169
github.com/dgrr/fastws.(*Conn).Close(0xc00008c400, 0x4569a6, 0x0)
        /home/dgrr/.go/src/github.com/dgrr/fastws/conn.go:335 +0x33
```

Debugging the code I found that the problem is at [this line](https://github.com/valyala/fasthttp/blob/master/server.go#L2111). If `hjc.Conn = nil` is deleted the received error will be the following instead of a panic:
```
 error sending: write tcp4 127.0.0.1:9999->127.0.0.1:52234: use of closed network connection
```

This will be better than a panic with the following recover functions... I think that happens the same with the following line `hjc.r = nil`.

As a solution I propose to delete this two lines.

erikdubbelboer:
I'm afraid deleting that line (as in your pull request) will cause even more issues. After that line the struct is released to a pool for reuse so the next hijacked connection will overwrite `hjc.Conn`.
A hijacked connection handler should only return when it is done with the connection. After the handler returns it shouldn't keep any references to the `hijackConn` struct.

If you have a separate goroutine writing I suggest having a way to wait for that goroutine to stop before returning from the handler.

[...]

PR #496, opened 2018-12-16
Fix https://github.com/valyala/fasthttp/issues/495

dgrr:


erikdubbelboer:
See: https://github.com/valyala/fasthttp/issues/495#issuecomment-448183350

I guess the only way to fix this in fasthttp is if we don't use a pool here. But that would mean at least one allocation for each hijacked connection.

Issue #499, opened 2018-12-19
[Feature Request] Allow queue size of workers modifiable


vijayviji:
Currently the queue size of the worker (the channel size) is fixed as 1 or 0 (0 in case of single processor). However, in certain cases, we want to go with a model of single worker to single CPU which consistently processes tasks. In that case, the worker queue size has to be higher.

fasthttp.Server should take a parameter to modify this.

PR #500, opened 2018-12-19
Allow queue size (channel capacity) of workers modifiable 

vijayviji:
This change is required in certain cases such as in the model, where there will be 1 acceptor and (n-1) workers, n being the no. of CPUs. In this model, the idea is to make workers CPU bound, and also not let acceptor suffer. Nevertheless, by default, it will behave the old way.

I'm creating a PR, addressing this issue https://github.com/valyala/fasthttp/issues/499

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea/
 tags
 *.pprof
 *.fasthttp.gz
--- a/client_example_test.go
+++ b/client_example_test.go
@@ -3,7 +3,7 @@ package fasthttp_test
 import (
 	"log"
 
-	"github.com/valyala/fasthttp"
+	"github.com/vijayviji/fasthttp"
 )
 
 func ExampleHostClient() {
--- a/client_test.go
+++ b/client_test.go
@@ -15,7 +15,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/valyala/fasthttp/fasthttputil"
+	"github.com/vijayviji/fasthttp/fasthttputil"
 )
 
 func TestClientPostArgs(t *testing.T) {
--- a/client_timing_test.go
+++ b/client_timing_test.go
@@ -13,7 +13,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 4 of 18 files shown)


erikdubbelboer:
I don't think the channel size does what you think it does. Even if the channel size is bigger than 1 it will never put more connection in the channel. The idea behind the channel size of 1 is that the goroutine that accepts new connections doesn't have to wait for the worker to read from the channel. Instead it can put the connection in the channel and continue with the next connection. Each worker has his own channel and a worker is only assigned one connection at a time. Therefor making the channel size larger than 1 doesn't make sense.

Normally the channel size should always be 1 but when there is only 1 CPU putting it in the channel and having to wait for the worker goroutine to schedule to get it out is slower than having a blocking channel that makes the accept goroutine go to sleep immediately so the CPU can immediately switch to the worker goroutine.

I'm guessing what you want to do is change [`Server.Concurrency`](https://godoc.org/github.com/valyala/fasthttp#Server.Concurrency)?

vijayviji:
If I rightly understand, Server.Concurrency creates 'n' number of workers. This change is not for increasing number of workers. This is to increase the queue size of the workers.

In my model, I want to one acceptor accept the connections very fast (during spiky loads) and hand out the connection to workers in a queue. The worker at its pace can take connections from the queue and then read and do processing. I don't understand why this won't work.

The issue with creating more number of workers than the available CPUs is that the workers are contending with the acceptor for CPU and acceptor is dropping connections. We're receiving so many short-lived connections in a brief span of time, and this would help.

erikdubbelboer:
The problem is that the `workerPool.ready` slice is used to assign workers that aren't doing anything a connection. Once a worker is assigned a connection it gets removed from the `ready` slice. So it will never get more than 1 connection put into its channel. What you want requires a lot more change. You'll need to do something like only removing a worker from `ready` when its channel is completely full and putting it back in once it handles at least one connection and has room again.

vijayviji:
I see. Thanks for this. You can reject this change.

PR #502, opened 2018-12-26
fix args empty string be changed to boolen

xuecai:
args is a=&b=1,  when use args.String() ,  it be changed to a&b=1

### DIFF
--- a/args.go
+++ b/args.go
@@ -45,9 +45,9 @@ type Args struct {
 }
 
 type argsKV struct {
-	key   	[]byte
-	value  	[]byte
-	noValue	bool
+	key     []byte
+	value   []byte
+	noValue bool
 }
 
 // Reset clears query args.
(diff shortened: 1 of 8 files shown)


erikdubbelboer:
Do you have an example of a program or library that interprets those 2 as different values for `a`? I thought they mean the same?

xuecai:
maybe this is nignx+lua's question.
nignx+lua code like this
```
local get = ngx.req.get_uri_args()

ngx.say("a1 is " .. type(get['a']))
ngx.say(get['a'])
ngx.say("b1 is " .. type(get['b']))
ngx.say(get['b'])
```
request is http://127.0.0.1/a.lua?a&b=
result is 
```
a1 is boolean
true
b1 is string
```

in nginx + php-fpm is ok
```
<?php
var_dump($_GET);exit;
```
request is http://127.0.0.1/a.lua?a=&b=1&c&d=1
result is 
```
array(4) { ["a"]=> string(0) "" ["b"]=> string(1) "1" ["c"]=> string(0) "" ["d"]=> string(1) "1" }
```

btw, maybe nochange is better? 
source is a&b=, args.String()  return a&b= is better than a&b ?

xuecai:
i complite it in new pull  request.
please review, tks !

erikdubbelboer:
Can you also look into the tests that are failing now.

erikdubbelboer:
Looks good so far. I think it would also be very useful to add a function to `Args` called `SetNoValue` that just adds the key and sets `noValue` to true. Can you add that (and a proper comment explaining what the function does).

args.go
```diff
@@ -147,48 +155,77 @@ func (a *Args) DelBytes(key []byte) {
...
 func (a *Args) AddBytesKV(key, value []byte) {
-	a.args = appendArg(a.args, b2s(key), b2s(value))
+	a.args = appendArg(a.args, b2s(key), b2s(value), ArgsHasValue)
+}
+
+// AddNoValue adds only 'key' argument.
+//
+// Multiple values for the same key may be added.
+// Different with Add empty string, like key1=&key1=
+// Only key in argumemt, like key1&key1
```
erikdubbelboer: I think this would be a better comment:
```go
// AddNoValue adds only 'key' as argument without the '='.
//
// Multiple values for the same key may be added.
```

args.go
```diff
@@ -147,48 +155,77 @@ func (a *Args) DelBytes(key []byte) {
...
+func (a *Args) AddNoValue(key string) {
+	a.args = appendArg(a.args, key, "", ArgsNoValue)
+}
+
+// AddBytesKNoValue adds 'key' argument.
```
erikdubbelboer: Same here:
```go
// AddBytesKNoValue adds only 'key' as argument without the '='.
```

args.go
```diff
@@ -147,48 +155,77 @@ func (a *Args) DelBytes(key []byte) {
...
 // SetBytesKV sets 'key=value' argument.
 func (a *Args) SetBytesKV(key, value []byte) {
-	a.args = setArgBytes(a.args, key, value)
+	a.args = setArgBytes(a.args, key, value, ArgsHasValue)
+}
+
+// SetNoValue sets only 'key' argument.
+// 
+// Different with Set empty string, like key1=&key2=
+// Only key in argumemt, like key1&key2
```
erikdubbelboer: Same here:
```go
// SetNoValue sets only 'key' as argument without the '='.
```

args.go
```diff
@@ -147,48 +155,77 @@ func (a *Args) DelBytes(key []byte) {
...
+func (a *Args) SetNoValue(key string) {
+	a.args = setArg(a.args, key, "", ArgsNoValue)
+}
+
+// SetBytesKNoValue sets 'key' argument.
```
erikdubbelboer: And here:
```go
// SetBytesKNoValue sets only 'key' as argument without the '='.
```

args.go
```diff
@@ -322,7 +359,12 @@ func copyArgs(dst, src []argsKV) []argsKV {
 		dstKV := &dst[i]
 		srcKV := &src[i]
 		dstKV.key = append(dstKV.key[:0], srcKV.key...)
-		dstKV.value = append(dstKV.value[:0], srcKV.value...)
+		if ArgsNoValue == srcKV.noValue {
```
erikdubbelboer: I would prefer `srcKV.noValue == ArgsNoValue` which reads easier and is more in line with the rest of the codebase.

args.go
```diff
@@ -345,31 +387,41 @@ func delAllArgs(args []argsKV, key string) []argsKV {
...
 }
 
-func setArg(h []argsKV, key, value string) []argsKV {
+func setArg(h []argsKV, key, value string, noValue bool) []argsKV {
 	n := len(h)
 	for i := 0; i < n; i++ {
 		kv := &h[i]
 		if key == string(kv.key) {
-			kv.value = append(kv.value[:0], value...)
+			if ArgsNoValue == noValue {
```
erikdubbelboer: I would prefer `noValue == ArgsNoValue` which reads easier and is more in line with the rest of the codebase.

args.go
```diff
@@ -345,31 +387,41 @@ func delAllArgs(args []argsKV, key string) []argsKV {
...
+	return appendArg(h, b2s(key), b2s(value), noValue)
 }
 
-func appendArg(args []argsKV, key, value string) []argsKV {
+func appendArg(args []argsKV, key, value string, noValue bool) []argsKV {
 	var kv *argsKV
 	args, kv = allocArg(args)
 	kv.key = append(kv.key[:0], key...)
-	kv.value = append(kv.value[:0], value...)
+	if ArgsNoValue == noValue {
```
erikdubbelboer: I would prefer `noValue == ArgsNoValue` which reads easier and is more in line with the rest of the codebase.

args.go
```diff
@@ -114,9 +120,11 @@ func (a *Args) AppendBytes(dst []byte) []byte {
 	for i, n := 0, len(a.args); i < n; i++ {
 		kv := &a.args[i]
 		dst = AppendQuotedArg(dst, kv.key)
-		if len(kv.value) > 0 {
+		if ArgsHasValue == kv.noValue {
```
erikdubbelboer: Sorry one last change, can you swap this one around as well.

xuecai: done :)

erikdubbelboer:
Looks good. What do you think @kirillDanshin ?

args.go
```diff
@@ -9,6 +9,11 @@ import (
 	"github.com/valyala/bytebufferpool"
 )
 
+const (
+	ArgsNoValue = true
```
kirillDanshin: why may we need to introduce such variables in the public API?

erikdubbelboer: Good point, I think they should be made internal only.

erikdubbelboer:
Thanks!

PR #505, opened 2018-12-29
add method Sort in args

xuecai:
sorted args string is useful in cache key。

### DIFF
--- a/args.go
+++ b/args.go
@@ -9,6 +9,11 @@ import (
 	"github.com/valyala/bytebufferpool"
 )
 
+const (
+	argsNoValue = true
+	argsHasValue = false
+)
+
 // AcquireArgs returns an empty Args object from the pool.
 //
 // The returned Args may be returned to the pool with ReleaseArgs
@@ -45,8 +50,9 @@ type Args struct {
 }
 
 type argsKV struct {
-	key   []byte
-	value []byte
+	key     []byte
+	value   []byte
+	noValue bool
 }
 
 // Reset clears query args.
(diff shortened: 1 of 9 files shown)


xuecai:
in your meas , it use like this ?
```
args := fasthttp.AcquireArgs()
// do something 

args.Sort() // just AppendBytesSort, it call sort.SliceStable and AppendBytes
cachekey := args.QueryString()  // call AppendBytes 
```

AppendBytesSort  do sort Args.args and rewrite Args.buf call AppendBytes
String and QueryString also call AppendBytes.
AppendBytes called twice is necessary ?

erikdubbelboer:
Yes that is exactly what I mean. `Sort` should only do the sorting and have no argument or return value. That wat it's clear that you're changing the data by calling `Sort`. Right now it's not clear that you're changing the underlying data by calling `AppendBytesSort` for example.

xuecai:
done , only Sort method add.
btw, Sort cost 3 allocs/op.

args.go
```diff
@@ -115,6 +116,17 @@ func (a *Args) QueryString() []byte {
 	return a.buf
 }
 
+// Sort change args order which ASCII sorted by key, value.
```
erikdubbelboer: Can you please change this to:
```go
// Sort sorts Args lexicographically by key and then value.
```

xuecai: Args.args is not public , so maybe can't have the sort function.
I need time to think about it.

args.go
```diff
@@ -115,6 +116,17 @@ func (a *Args) QueryString() []byte {
 	return a.buf
 }
 
+// Sort sorts Args lexicographically by key and then value.
```
erikdubbelboer: Good but of course now the comment should be
```go
// Sort sorts Args by key and then value using 'f' as comparison function.
//
// For example args.Sort(bytes.Compare)
```

erikdubbelboer:
Looks good. Any comments @kirillDanshin ?

xuecai:
@kirillDanshin   any comments?  my parterner want to use this feature . tks.

PR #507, opened 2019-01-04
Make the max Dial concurrency limit configurable

### LINKED ISSUE #478: maxDialConcurrency = 1000
Hi, I use fasthttp to serve as a reverse-proxy.I occasionaly found there were many long durtions requests which shouldn't be so long.I read the fasthttp code and made some logs.Now I am sure that  maxDialConcurrency (1000) is too small for me .

How does the 1000 come from ?Can I change it ?

erikdubbelboer:
Fixes #478

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -33,7 +33,7 @@ import (
 //     * foo.bar:80
 //     * aaa.com:8080
 func Dial(addr string) (net.Conn, error) {
-	return getDialer(DefaultDialTimeout, false)(addr)
+	return defaultDialer.Dial(addr)
 }
 
 // DialTimeout dials the given TCP addr using tcp4 using the given timeout.
@@ -58,7 +58,7 @@ func Dial(addr string) (net.Conn, error) {
 //     * foo.bar:80
 //     * aaa.com:8080
 func DialTimeout(addr string, timeout time.Duration) (net.Conn, error) {
-	return getDialer(timeout, false)(addr)
+	return defaultDialer.DialTimeout(addr, timeout)
 }
 
 // DialDualStack dials the given TCP addr using both tcp4 and tcp6.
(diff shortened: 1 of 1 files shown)


tcpdialer.go
```diff
@@ -160,11 +160,25 @@ type tcpDialer struct {
...
+
+// SetMaxDialConcurrency sets the maximum number of concurrent Dails
+// that can be performed. Setting this to 0 means unlimited.
+//
+// The initial max concurrency is set to 1000.
+//
+// WARNING: This function needs to be called before any Dials are
+// performed. Calling this function after calling any of the Dial
+// functions might not have any effect and can cause a race condition.
+func SetMaxDialConcurrency(concurrency int) {
```
valyala: It would be better to add ability to override default dial concurrency per each dialer. Global setting for all the dialers smells not very good.

erikdubbelboer: The problem is that there is no such thing as a dialer. The only thing exposed by the api is the `Dial*` functions. Giving those an option and be backwards compatible means we have to introduce 4 new functions with options for all 4 `Dial*` functions.

What we could do is implement a `NewDialer` function that returns a `Dialer` interface that exposes a `Dial` function and some other functions to set options (such as `SetDualStack`, `SetTimeout` and `SetConcurrency`). @valyala @kirillDanshin If you guys think this is a good idea I'll update this pull request.

kirillDanshin: @erikdubbelboer I think it's a good idea to add such dialer and expose defaultDialer.Dial* methods as a Dial* API. In such way, we can allow us to add an unlimited number of options to the dialer without having to pay for interface calls.

valyala: Agreed with @kirillDanshin on exposing `Dialer`

erikdubbelboer:
@valyala @kirillDanshin please check my latest changes.

tcpdialer.go
```diff
@@ -160,41 +137,145 @@ type tcpDialer struct {
...
+//
+// For instance, per-host counters and/or limits may be implemented
+// by such wrappers.
+//
+// The addr passed to the function must contain port. Example addr values:
+//
+//     * foobar.baz:443
+//     * foo.bar:80
+//     * aaa.com:8080
+func (d *TcpDialer) Dial(addr string) (net.Conn, error) {
```
kirillDanshin: Oops, didn't notice — I think we should use TCPDialer instead of TcpDialer. That's a common abbreviation.

erikdubbelboer: You are right, I have fixed it.

erikdubbelboer:
@kirillDanshin @valyala any input?

erikdubbelboer:
@kirillDanshin @valyala no? Cause then I'll merge it.

PR #521, opened 2019-01-30
all: fix typo on comments

shuLhan:
While at it wrap comments with long lines.

### DIFF
--- a/header.go
+++ b/header.go
@@ -73,7 +73,8 @@ type RequestHeader struct {
 
 	rawHeaders []byte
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Sorry but I don't think these lines are too long and I'd rather not pollute the git history too much with commits like this.

shuLhan:
@erikdubbelboer The main point is in the typo actually, the line wrap is syntactic sugar.

erikdubbelboer:
Sorry I missed that, it's merged now. Thanks.

PR #522, opened 2019-01-30
all: pre-allocated slice with possible known size

shuLhan:
This fix is based on suggestion of "prealloc" static analysis tool [1].

Per note of the tool's author suggestion, its recommended to allocate the
slice length/capability, if we known their possible size.  This is to
minimize "append" to re-allocate the slice underlying array.

[1] https://github.com/alexkohler/prealloc

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -943,7 +943,7 @@ func (h *fsHandler) createDirIndex(base *URI, dirPath string, mustCompress bool)
 	}
 
 	fm := make(map[string]os.FileInfo, len(fileinfos))
-	var filenames []string
+	filenames := make([]string, 0, len(fileinfos))
 	for _, fi := range fileinfos {
 		name := fi.Name()
 		if strings.HasSuffix(name, h.compressedFileSuffix) {
--- a/cookie_test.go
+++ b/cookie_test.go
@@ -98,7 +98,7 @@ func TestCookieSameSite(t *testing.T) {
 	if !strings.Contains(s, "; SameSite") {
 		t.Fatalf("missing SameSite flag in cookie %q", s)
 	}
-	
+
 	if err := c.Parse("foo=bar; samesite=lax"); err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
@@ -109,7 +109,7 @@ func TestCookieSameSite(t *testing.T) {
 	if !strings.Contains(s, "; SameSite=Lax") {
 		t.Fatalf("missing SameSite flag in cookie %q", s)
 	}
-	
+
 	if err := c.Parse("foo=bar; samesite=strict"); err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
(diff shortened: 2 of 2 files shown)


erikdubbelboer:
Thanks.

fs.go
```diff
@@ -943,7 +943,7 @@ func (h *fsHandler) createDirIndex(base *URI, dirPath string, mustCompress bool)
 	}
 
 	fm := make(map[string]os.FileInfo, len(fileinfos))
-	var filenames []string
+	filenames := make([]string, 0, len(fileinfos))
```
kirillDanshin: @erikdubbelboer this change may use more memory than needed. I don't think this is a good idea.
Note that we don't append file if it was compressed.

erikdubbelboer: @kirillDanshin slightly more memory yes. But it's not memory we need to optimize for, its the number of allocations and that is definitely less with this change.

shuLhan: @kirillDanshin  The following examples may give an illustration on how much underlying array in slice growth (re-allocated),

```
package main

import (
	"fmt"
)

func main() {
	var s []int
	for x := 0; x < 5; x++ {
		fmt.Printf("len:%d cap:%d\n", len(s), cap(s))
		s = append(s, x)
	}
}
```

In my environment the slice `s` re-allocated 3 times.

```
len:0 cap:0
len:1 cap:2
len:2 cap:2
len:3 cap:4
len:4 cap:4
```

kirillDanshin: @erikdubbelboer @shuLhan let's assume that we have 512 files in an upload directory shard.
That gives us 512 elements of minimum capacity.
With this patch, it will pre-allocate 1024 elements for this case.
So this will use 16384 bytes instead of 8192 minimum.

We may need to pre-allocate some space, but I don't think that using the length of this array is a good idea.

shuLhan: @kirillDanshin  I am sorry, I am not following your math.  If we have 512 files how come it will pre-allocate 1024 elements? Even if we pre-allocate 1024 the total slice size would be 12 + 1024*8 = 8204.  Where is the 16384 come from?

The main idea is to minimize the number of memory re-allocated (call to malloc and copying previous array memory to new allocated memory).  Does it will consume more memory on initialize make? Maybe, depends home many compressed files in user's dirPath.  Does it require less process? Yes.

erikdubbelboer: With 512 elements in total you'll get 9 allocations (`2,4,8,16,32,64,128,265,512`) with the old code but only 1 with the new one. In total those 9 allocations will allocate almost the same (`2+4+8+16+32+64+128+265+512 = 1022`) as if we preallocate 1024 elements at once. So in every way this is an improvement.

kirillDanshin: @shuLhan 512 raw files + 512 gzipped *.fasthttp.gz.
`unsafe.Sizeof("z")` = `uintptr(16)`. `cap(s) * int(unsafe.Sizeof("z")` = 16384, where `cap(s)` = 1024.

I agree with @erikdubbelboer: I forgot to calculate all reallocations. But let's pre-allocate `len(fileinfos) / 2` here.
In that way, we still preallocate 512, but filter-out `*.fasthttp.gz` files and saving 8192 bytes just like that.

PR #523, opened 2019-01-30
Various changes regarding code readibility

shuLhan:
Changes:

07f0c88 (Shulhan, 2 hours ago)
   all: simplify variable initialization

   If we assign the variable after declaring it, we can simplify it using
   ":=" operator or "= value". The reader can still known the type of variable
   from the struct name or variable type before assignment operator.

7e00ec6 (Shulhan, 2 hours ago)
   args: simplify if-condition check on boolean value

cf6c51e (Shulhan, 2 hours ago)
   http_test: simplify strings.Index with strings.Contains

   Both have the same O(n), but strings.Contains more readable on
   if-condition.

8439088 (Shulhan, 2 hours ago)
   all: replace string(bytes.Buffer.Bytes()) with bytes.Buffer.String()

   Although its only occured on test files, it may be worth to simplified it.

5f1f929 (Shulhan, 2 hours ago)
   all: use sort.Strings when applicable

   Basically, sort.Strings is the shortcut of Sort(StringSlice(a)) but its
   more readable.

### DIFF
--- a/args.go
+++ b/args.go
@@ -134,7 +134,7 @@ func (a *Args) AppendBytes(dst []byte) []byte {
 	for i, n := 0, len(a.args); i < n; i++ {
 		kv := &a.args[i]
 		dst = AppendQuotedArg(dst, kv.key)
-		if kv.noValue == argsHasValue {
+		if !kv.noValue {
 			dst = append(dst, '=')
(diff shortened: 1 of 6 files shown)


erikdubbelboer:
Thanks.

PR #524, opened 2019-01-31
fix the gracefully shutdown

wilhelmliao:
The Ctrl+C will interrupt the server process immediately without wait active connections finish.
I change the Shutdown() make it perfectly.

The following sample can help to reproduce the scenario.
```go
package main

import (
	"log"
	"net"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/valyala/fasthttp"
)

var address = ":38080"

func connStateCallback(conn net.Conn, state fasthttp.ConnState) {
	log.Printf("conn: %+v\n", conn)
	log.Printf("state: %+v\n", state)
}

func main() {
	handler := processRequest

	server := &fasthttp.Server{
		Handler:                       handler,
		Name:                          "Demo Service",
		DisableKeepalive:              true,
		DisableHeaderNamesNormalizing: true,
		ConnState:                     connStateCallback,
	}

	signalChan := make(chan os.Signal, 1)
	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
	//register for interupt (Ctrl+C) and SIGTERM (docker)
	go func() {
		<-signalChan
		log.Printf("server.Shutdown()")
		server.Shutdown()
	}()

	log.Printf("%s listening on address %s\n", server.Name, address)
	if err := server.ListenAndServe(address); err != nil {
		log.Fatalf("Error in ListenAndServe: %s\n", err)
	}
}

func processRequest(ctx *fasthttp.RequestCtx) {
	ctx.Conn()
	time.Sleep(time.Second * 10)
	ctx.Success("text/plain", []byte("123"))
	log.Printf("connection '%+v' finshed\n", ctx.Conn())
	return
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -1599,25 +1599,29 @@ func (s *Server) Shutdown() error {
 	atomic.StoreInt32(&s.stop, 1)
 	defer atomic.StoreInt32(&s.stop, 0)
 
-	if s.ln == nil {
-		return nil
-	}
-
-	if err := s.ln.Close(); err != nil {
-		return err
-	}
-
 	// Closing the listener will make Serve() call Stop on the worker pool.
 	// Setting .stop to 1 will make serveConn() break out of its loop.
 	// Now we just have to wait until all workers are done.
 	for {
-		if open := atomic.LoadInt32(&s.open); open == 0 {
+		// Checking the .open to ensure all active connections had finished. 
+		// NOTE: The .open keeping the number of the active connections which
+		// include the server connection itself. That is, the .open will be 
+		// 1 + <active client connections> on a runnig server. 
+		if open := atomic.LoadInt32(&s.open); open <= 1 {
 			break
 		}
 		// This is not an optimal solution but using a sync.WaitGroup
 		// here causes data races as it's hard to prevent Add() to be called
 		// while Wait() is waiting.
 		time.Sleep(time.Millisecond * 100)
+	}
+
+	if s.ln == nil {
+		return nil
+	}
+
+	if err := s.ln.Close(); err != nil {
+		return err
 	}
 
 	s.ln = nil


erikdubbelboer:
The way it currently works is the exact same way as [`net/http.Server.Shutdown()`](https://golang.org/pkg/net/http/#Server.Shutdown) works. From that documentation:
```
When Shutdown is called, Serve, ListenAndServe, and ListenAndServeTLS immediately
return ErrServerClosed. Make sure the program doesn't exit and waits instead for
Shutdown to return. 
```
fasthttp [has the exact same line in the documentation for `Shutdown()`](https://godoc.org/github.com/valyala/fasthttp#Server.Shutdown)!

wilhelmliao:
Is there way to shutdown a server process without interrupting critical running tasks?

For example, I want load data from database A and write the data to database B and database C in a single request handler. I want the server.Shutdown() will wait for the operation finished.

erikdubbelboer:
Yes you just have to switch it around. You do your `ListenAndServe` in a goroutine and `Shutdown` in your main goroutine:
```go
package main

import (
	"log"
	"net"
	"os"
	"os/signal"
	"syscall"
	"time"

	"github.com/valyala/fasthttp"
)

var address = ":38080"

func connStateCallback(conn net.Conn, state fasthttp.ConnState) {
	log.Printf("conn: %+v\n", conn)
	log.Printf("state: %+v\n", state)
}

func main() {
	handler := processRequest

	server := &fasthttp.Server{
		Handler:                       handler,
		Name:                          "Demo Service",
		DisableKeepalive:              true,
		DisableHeaderNamesNormalizing: true,
		ConnState:                     connStateCallback,
	}

	go func() {
		log.Printf("%s listening on address %s\n", server.Name, address)
		if err := server.ListenAndServe(address); err != nil {
			log.Fatalf("Error in ListenAndServe: %s\n", err)
		}
	}()

	signalChan := make(chan os.Signal, 1)
	signal.Notify(signalChan, os.Interrupt, syscall.SIGTERM)
	//register for interupt (Ctrl+C) and SIGTERM (docker)
	<-signalChan
	log.Printf("server.Shutdown()")
	server.Shutdown()
}

func processRequest(ctx *fasthttp.RequestCtx) {
	ctx.Conn()
	time.Sleep(time.Second * 10)
	ctx.Success("text/plain", []byte("123"))
	log.Printf("connection '%+v' finshed\n", ctx.Conn())
	return
}
```

wilhelmliao:
Thank you, @erikdubbelboer ! It work well :)

PR #526, opened 2019-02-03
Cookiejar implementation.

dgrr:
Hello,

This commit tries to implement a basic CookieJar structure to handle Cookies easily without using external packages.

### DIFF
--- a/cookiejar.go
+++ b/cookiejar.go
@@ -13,32 +13,11 @@ type CookieJar struct {
 	hostCookies map[string][]*Cookie
 }
 
-func (cj *CookieJar) init() {
-	cj.m.Lock()
-	{
-		if cj.hostCookies == nil {
-			cj.hostCookies = make(map[string][]*Cookie)
-		}
-	}
-	cj.m.Unlock()
-}
-
-func copyCookies(cookies []*Cookie) (cs []*Cookie) {
-	// TODO: Try to delete the allocations
-	cs = make([]*Cookie, 0, len(cookies))
-	for _, cookie := range cookies {
-		c := AcquireCookie()
-		c.CopyTo(cookie)
-		cs = append(cs, c)
-	}
-	return
-}
-
 // Get returns the cookies stored from a specific domain.
 //
 // If there were no cookies related with host returned slice will be nil.
 //
-// returned cookies can be released safely.
+// CookieJar keeps a copy of the cookies, so the returned cookies can be released safely.
 func (cj *CookieJar) Get(uri *URI) (cookies []*Cookie) {
 	if uri != nil {
 		cookies = cj.get(uri.Host(), uri.Path())
@@ -47,7 +26,9 @@ func (cj *CookieJar) Get(uri *URI) (cookies []*Cookie) {
 }
 
 func (cj *CookieJar) get(host, path []byte) (rcs []*Cookie) {
-	cj.init()
+	if cj.hostCookies == nil {
+		return
+	}
 
 	var (
 		err     error
@@ -59,12 +40,8 @@ func (cj *CookieJar) get(host, path []byte) (rcs []*Cookie) {
 	if err != nil {
 		hostStr = b2s(host)
 	}
-	cj.m.Lock()
-	{
-		// get cookies deleting expired ones
-		cookies = cj.getCookies(hostStr)
-	}
-	cj.m.Unlock()
+	// get cookies deleting expired ones
+	cookies = cj.getCookies(hostStr)
 
 	rcs = make([]*Cookie, 0, len(cookies))
 	for i := 0; i < len(cookies); i++ {
@@ -79,25 +56,29 @@ func (cj *CookieJar) get(host, path []byte) (rcs []*Cookie) {
 }
 
 // getCookies returns a cookie slice releasing expired cookies
-func (cj *CookieJar) getCookies(hostStr string) []*Cookie {
-	var (
+func (cj *CookieJar) getCookies(hostStr string) (cookies []*Cookie) {
+	cj.m.Lock()
+	{
 		cookies = cj.hostCookies[hostStr]
-		t       = time.Now()
-		n       = len(cookies)
-	)
-	for i := 0; i < len(cookies); i++ {
-		c := cookies[i]
-		if !c.Expire().Equal(CookieExpireUnlimited) && c.Expire().Before(t) { // cookie expired
-			cookies = append(cookies[:i], cookies[i+1:]...)
-			ReleaseCookie(c)
-			i--
+		var (
+			t = time.Now()
+			n = len(cookies)
+		)
+		for i := 0; i < len(cookies); i++ {
+			c := cookies[i]
+			if !c.Expire().Equal(CookieExpireUnlimited) && c.Expire().Before(t) { // cookie expired
+				cookies = append(cookies[:i], cookies[i+1:]...)
+				ReleaseCookie(c)
+				i--
+			}
+		}
+		// has any cookie been deleted?
+		if n > len(cookies) {
+			cj.hostCookies[hostStr] = cookies
 		}
 	}
-	// has any cookie been deleted?
-	if n > len(cookies) {
-		cj.hostCookies[hostStr] = cookies
-	}
-	return cookies
+	cj.m.Unlock()
+	return
 }
 
 // Set sets cookies for a specific host.
@@ -105,6 +86,8 @@ func (cj *CookieJar) getCookies(hostStr string) []*Cookie {
 // The host is get from uri.Host().
 //
 // If the cookie key already exists it will be replaced by the new cookie value.
+//
+// CookieJar keeps a copy of the cookies, so the parsed cookies can be released safely.
 func (cj *CookieJar) Set(uri *URI, cookies ...*Cookie) {
 	if uri != nil {
 		cj.set(uri.Host(), cookies...)
@@ -114,15 +97,18 @@ func (cj *CookieJar) Set(uri *URI, cookies ...*Cookie) {
 // SetByHost sets cookies for a specific host.
 //
 // If the cookie key already exists it will be replaced by the new cookie value.
+//
+// CookieJar keeps a copy of the cookies, so the parsed cookies can be released safely.
 func (cj *CookieJar) SetByHost(host []byte, cookies ...*Cookie) {
 	cj.set(host, cookies...)
 }
 
 func (cj *CookieJar) set(host []byte, cookies ...*Cookie) {
-	cj.init()
-
 	cj.m.Lock()
 	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
 		hostStr := b2s(host)
 		hcs, ok := cj.hostCookies[hostStr]
 		if !ok {
@@ -132,13 +118,11 @@ func (cj *CookieJar) set(host []byte, cookies ...*Cookie) {
 		}
 		for _, cookie := range cookies {
 			c := searchCookieByKeyAndPath(cookie.Key(), cookie.Path(), hcs)
-			if c != nil {
-				c.ParseBytes(cookie.Cookie())
-			} else {
+			if c == nil {
 				c = AcquireCookie()
-				c.CopyTo(cookie)
 				hcs = append(hcs, c)
 			}
+			c.CopyTo(cookie)
 		}
 		cj.hostCookies[hostStr] = hcs
 	}
@@ -177,11 +161,12 @@ func (cj *CookieJar) dumpTo(req *Request) {
 }
 
 func (cj *CookieJar) getFrom(host, path []byte, res *Response) {
-	cj.init()
-
 	hostStr := b2s(host)
 	cj.m.Lock()
 	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
 		cookies, ok := cj.hostCookies[hostStr]
 		if !ok {
 			// If the key does not exists in the map then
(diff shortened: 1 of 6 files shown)


cookiejar.go
```diff
@@ -0,0 +1,80 @@
...
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
+		cookies, ok := cj.hostCookies[host]
+		if !ok {
+			cookies = make([]*Cookie, 0)
+		}
+		res.Header.VisitAllCookie(func(key, value []byte) {
+			cookie := &Cookie{}
```
erikdubbelboer: Use [AcquireCookie](https://godoc.org/github.com/valyala/fasthttp#AcquireCookie) here?

cookiejar.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Get returns the cookies stored from a specific domain.
+//
+// If there were no cookies related with host returned slice will be nil.
+func (cj *CookieJar) Get(host string) (cookies []*Cookie) {
+	cj.m.RLock()
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
```
erikdubbelboer: Why not just return nil here?

cookiejar.go
```diff
@@ -0,0 +1,80 @@
...
+		cj.hostCookies[host] = hc
+	}
+	cj.m.Unlock()
+}
+
+func (cj *CookieJar) dumpTo(host string, req *Request) {
+	cj.m.Lock()
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
```
erikdubbelboer: Why not just `return` here?

cookiejar.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (cj *CookieJar) getFrom(host string, res *Response) {
+	cj.m.Lock()
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
+		cookies, ok := cj.hostCookies[host]
+		if !ok {
+			cookies = make([]*Cookie, 0)
```
erikdubbelboer: You don't have to `make` an empty slice here. `append` to an empty slice is perfectly fine.

cookiejar.go
```diff
@@ -0,0 +1,80 @@
...
+		}
+		res.Header.VisitAllCookie(func(key, value []byte) {
+			cookie := &Cookie{}
+			cookie.SetKeyBytes(key)
+			if !res.Header.Cookie(cookie) {
+				// TODO: error?
+				cookie.SetValueBytes(value)
+			}
```
erikdubbelboer: Instead of this `if` just do `cookie.ParseBytes(value)`.

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
+	}
+	cj.m.Unlock()
+}
+
+func hasPort(host []byte) bool {
+	return bytes.IndexByte(host, ':') > 0
```
erikdubbelboer: This is a bit tricky since an IPv6 IP as host would also contain `:`.

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+	if hasPort(host) {
+		hostStr, _, err = net.SplitHostPort(hostStr)
+	}
+	if err == nil {
+		cj.m.Lock()
+		{
+			// get cookies deleting expired ones
+			cookies := cj.getCookies(hostStr)
+			if len(cookies) > 0 {
+				rcs = copyCookies(cookies) // make a copy
```
erikdubbelboer: Why are you making a copy here?

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+// getCookies returns a cookie slice releasing expired cookies
+func (cj *CookieJar) getCookies(hostStr string) []*Cookie {
+	var (
+		cookies = cj.hostCookies[hostStr]
+		t       = time.Now()
+		n       = len(cookies)
+	)
+	for i := 0; i < len(cookies); i++ {
+		c := cookies[i]
+		if !c.expire.IsZero() && t.Sub(c.expire) > 0 { // expired
```
erikdubbelboer: This should just be `if c.expire.After(t) {`.

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+		{
+			// get cookies deleting expired ones
+			cookies := cj.getCookies(hostStr)
+			if len(cookies) > 0 {
+				rcs = copyCookies(cookies) // make a copy
+				for i := 0; i < len(rcs); i++ {
+					cookie := rcs[i]
+					if len(path) > 1 && len(cookie.path) > 1 && !bytes.HasPrefix(cookie.path, path) {
+						rcs = append(rcs[:i], rcs[i+1:]...)
+						ReleaseCookie(cookie)
```
erikdubbelboer: You shouldn't release the cookie here!

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+	for _, cookie := range cookies {
+		req.Header.SetCookieBytesKV(cookie.Key(), cookie.Value())
+	}
+}
+
+func (cj *CookieJar) getFrom(host, path []byte, res *Response) {
+	cj.init()
+
+	hostStr := b2s(host)
+	cookies, ok := cj.hostCookies[hostStr]
```
erikdubbelboer: This should probably be protected by the mutex as well?

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+		// we must make a copy for the key.
+		hostStr = string(host)
+	}
+	res.Header.VisitAllCookie(func(key, value []byte) {
+		c := searchCookieByKeyAndPath(key, path, cookies)
+		if c == nil {
+			c = AcquireCookie()
+			cookies = append(cookies, c)
+		}
+		c.ParseBytes(value)
```
erikdubbelboer: If the cookie being set has an expire time in the past it's probably better to delete it here instead of just keeping it around.

cookiejar.go
```diff
@@ -0,0 +1,213 @@
...
+	})
+	cj.hostCookies[hostStr] = cookies
+}
+
+func searchCookieByKeyAndPath(key, path []byte, cookies []*Cookie) (cookie *Cookie) {
+	for _, c := range cookies {
+		if bytes.Equal(key, c.Key()) {
+			if len(path) <= 1 || bytes.HasPrefix(c.Path(), path) {
+				cookie = c
+				break
```
erikdubbelboer: Could just replace these 2 lines with `return c`.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+// returned cookies can be released safely.
+func (cj *CookieJar) Get(uri *URI) (cookies []*Cookie) {
+	if uri != nil {
+		cookies = cj.get(uri.Host(), uri.Path())
+	}
+	return
+}
+
+func (cj *CookieJar) get(host, path []byte) (rcs []*Cookie) {
+	cj.init()
```
erikdubbelboer: I still don't like the fact that `cj.hostCookies` gets set here while it's a get method. I would prefer it would just return `nil` if `cj.hostCookies` is `nil`.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+	hostStr, _, err = net.SplitHostPort(hostStr)
+	if err != nil {
+		hostStr = b2s(host)
+	}
+	cj.m.Lock()
+	{
+		// get cookies deleting expired ones
+		cookies = cj.getCookies(hostStr)
+	}
+	cj.m.Unlock()
```
erikdubbelboer: Its's better if `cj.getCookies` does the locking itself.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+	"time"
+)
+
+// CookieJar manages cookie storage
+type CookieJar struct {
+	m           sync.Mutex
+	hostCookies map[string][]*Cookie
+}
+
+func (cj *CookieJar) init() {
```
erikdubbelboer: I would completely remove this method and inline it in the 2 places where it's really needed.

erikdubbelboer: That way you also have less locking and unlocking as the method can just do it once instead of having it done multiple times.

cookiejar.go
```diff
@@ -0,0 +1,146 @@
...
+	cj.m.Lock()
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
+		}
+		hcs := cj.hostCookies[host]
+		for _, cookie := range cookies {
+			c := searchCookieByKey(cookie.Key(), hcs)
+			if c != nil {
+				c.ParseBytes(cookie.Cookie())
```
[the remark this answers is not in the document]
erikdubbelboer: Why was this marked resolved?

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+		cj.hostCookies[hostStr] = cookies
+	}
+	return cookies
+}
+
+// Set sets cookies for a specific host.
+//
+// The host is get from uri.Host().
+//
+// If the cookie key already exists it will be replaced by the new cookie value.
```
erikdubbelboer: You should add a comment that makes clear the CookieJar will make a copy of the cookie to keep and it's up to the caller to release it's own cookie.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+func (cj *CookieJar) Set(uri *URI, cookies ...*Cookie) {
+	if uri != nil {
+		cj.set(uri.Host(), cookies...)
+	}
+}
+
+// SetByHost sets cookies for a specific host.
+//
+// If the cookie key already exists it will be replaced by the new cookie value.
```
erikdubbelboer: Same comment as above here.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+// This function prevents extra allocations by making repeated cookies
+// not being duplicated.
+func (cj *CookieJar) SetKeyValue(host, key, value string) {
+	cj.SetKeyValueBytes(host, s2b(key), s2b(value))
+}
+
+// SetKeyValueBytes sets a cookie by key and value for a specific host.
+//
+// This function prevents extra allocations by making repeated cookies
+// not being duplicated.
```
erikdubbelboer: Needs similar comment as above.

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+		t := time.Now()
+		res.Header.VisitAllCookie(func(key, value []byte) {
+			created := false
+			c := searchCookieByKeyAndPath(key, path, cookies)
+			if c == nil {
+				c = AcquireCookie()
+				created = true
+			}
+			c.ParseBytes(value)
+			if c.Expire().IsZero() || c.Expire().After(t) { // cookie expired
```
erikdubbelboer:
```go
if c.Expire().Equal(CookieExpireUnlimited) || c.Expire().After(t) { // Not expired.
```

cookiejar.go
```diff
@@ -0,0 +1,221 @@
...
+			// we must make a copy for the key.
+			hostStr = string(host)
+		}
+		t := time.Now()
+		res.Header.VisitAllCookie(func(key, value []byte) {
+			created := false
+			c := searchCookieByKeyAndPath(key, path, cookies)
+			if c == nil {
+				c = AcquireCookie()
+				created = true
```
erikdubbelboer: Better do the expired check here first so you don't acquire and release a cookie immediately if it didn't exist already.

dgrr: I don't see any other way to do this.
If I do the expiration checking at cookiejar.go:197 first I need to call ParseBytes and the code will be repeated in the case `c == nil` and `c != nil`. I prefer to keep it like this unless you see any other way to do that.

cookiejar.go
```diff
@@ -13,32 +13,11 @@ type CookieJar struct {
 	hostCookies map[string][]*Cookie
```
xuecai: maybe hostCookies like this better for reuse memory
```golang
type hc struct {
        key []byte
        value []*Cookie
}
ype CookieJar struct {
    m           sync.Mutex
    hostCookies []hc
}
```

dgrr: I don't think we need to reuse memory in the CookieJar. We are already reusing cookies and I think it's enough. Also, this scheme will speed down the search by host. What do you think?

erikdubbelboer: Yes I don't think this will have any impact on allocations for a long running process and it will only complicate the code and make having many hosts slower.

cookiejar.go
```diff
@@ -0,0 +1,205 @@
...
+		req.Header.SetCookieBytesKV(cookie.Key(), cookie.Value())
+	}
+}
+
+func (cj *CookieJar) getFrom(host, path []byte, res *Response) {
+	hostStr := b2s(host)
+	cj.m.Lock()
+	{
+		if cj.hostCookies == nil {
+			cj.hostCookies = make(map[string][]*Cookie)
```
erikdubbelboer: Here you can just do `return`. No need to allocate anything.

erikdubbelboer: Just make sure you do `defer cj.m.Unlock()` instead of doing this at the end of the function.

I myself always prefer to `defer` `Unlock`s as its much easier to follow the code. Also `defer` is basically free these days so no worry about performance.

cookiejar.go
erikdubbelboer: Yes you are right, I forgot that you have to somehow parse the cookie first to get the expiration date. Doing that in a stack allocate cookie value would be possible for example but it would result in messier code.

cookiejar.go
xuecai: Yes, you are right.

efectn:
Any progress?

erg43hergeg:
bump

gaby:
@erikdubbelboer This can be close. I believe fasthttp supports cookiejar now days.

a-c-rodriguez:
@gaby - I can't find a cookiejar impl in fasthttp.  Are you referring to an external project?

gaby:
@a-c-rodriguez You are right, it's not in Fasthttp, but in Fiber.

Someone can work off this PR and create a new one

a-c-rodriguez:
> @a-c-rodriguez You are right, it's not in Fasthttp, but in Fiber.
> 
> Someone can work off this PR and create a new one
@gaby - Thanks for the quick reply!

image

Using this version of Fiber, how do I access the cookiejar?  My goal is to get a cookie and check to see if it's expired.

PR #531, opened 2019-02-05
An atomic approach for HostClient.MaxConns

xaionaro:
We need to dynamically change a `MaxConns` value. So this patch is about safe atomic reading from `MaxConns`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -544,6 +544,8 @@ type HostClient struct {
 	// Maximum number of connections which may be established to all hosts
 	// listed in Addr.
 	//
+	// It is safe to change this value using atomic.StoreUint64()
+	//
 	// DefaultMaxConnsPerHost is used if not set.
 	MaxConns uint64
 
--- a/client.go
+++ b/client.go
@@ -413,7 +413,7 @@ func (c *Client) Do(req *Request, resp *Response) error {
 			DialDualStack:                 c.DialDualStack,
 			IsTLS:                         isTLS,
 			TLSConfig:                     c.TLSConfig,
-			MaxConns:                      c.MaxConnsPerHost,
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks.

xaionaro:
Thank you!

xuecai:
is this maybe uint32 better for it can run on i386 system not support x64？

xaionaro:
> is this maybe uint32 better for it can run on i386 system not support x64？

Sounds reasonable.

@erikdubbelboer 
Should I prepare a fix?

erikdubbelboer:
@xuecai yes for atomic operations it should definitely be a `uint32`. See: https://golang.org/pkg/sync/atomic/#pkg-note-BUG

erikdubbelboer:
@xaionaro I'm sorry I made a mistake merging this pull request (looks like I wasn't thinking).
I have fixed it in: https://github.com/valyala/fasthttp/commit/e722d81dc9dccd94abcc3026236a11207a930b3c

One problem was that we can't just change `HostClient.MaxConns` from an `int` to a `uint64`. This breaks everyone's code that sets it using something else than a constant.
Another problem was indeed the atomic operations on an `uint64`.

xaionaro:
@erikdubbelboer I see. It's quite bad for us. How can we safely change the MaxConns then?

erikdubbelboer:
@xaionaro that's why I introduced the `HostClient.SetMaxConns(newMaxConns)` function.

xaionaro:
> @xaionaro that's why I introduced the HostClient.SetMaxConns(newMaxConns) function.

Ah. Did not notice. OK. Thanks. Sorry if I caused any problems :(

Issue #529, opened 2019-02-04
SkipBody = true ignored in response if client.DoTimeout is used


enchantner:
I have a little bit unusual case, but I need to perform GET requests ignoring the body of the response, like with HEAD requests. Fasthttp provides this by exposing SkipBody parameter (https://github.com/valyala/fasthttp/blob/master/http.go#L75), which I set manually after doing `resp := fasthttp.AcquireResponse()`.

The problem is, when using `client.DoTimeout(req, resp, timeout)`, this approach doesn't work and body is still read, because fasthttp makes an empty copy of `Response` object totally ignoring the original one: https://github.com/valyala/fasthttp/blob/master/client.go#L984 . It calls `copyToSkipBody` only for `Request`, but not for `Response` (https://github.com/valyala/fasthttp/blob/65955d6208bfd6efe330d8c37b18d6402cb1c91b/http.go#L626), and there is no easy workaround for this.

kirillDanshin:
closed by #532

PR #532, opened 2019-02-05
SkipBody is propagated to Response copy

enchantner:
Addressing https://github.com/valyala/fasthttp/issues/529 , shouldn't break anything.


xuecai:
@kirillDanshin @enchantner  
resp.Reset() used in func doNonNilReqResp in client.go, it will reset SkipBody which seted by Custom.

erikdubbelboer:
What do you mean by "Custom"?

PR #538, opened 2019-02-12
Add ResponseErrorHandler for errors while writing the response

gallir:
The handler is called if there was an error writing to the TCP connection.

Business-critical operations require to store the final result sent to the
client. For cloud providers's load balancer the problem is more subtle,
the final result sent to the customer may be different from the application
so it's important to log that the response was properly read (or discarded)
by the LB.

### DIFF
--- a/server.go
+++ b/server.go
@@ -166,6 +166,13 @@ type Server struct {
 	//   * ErrBrokenChunks
 	ErrorHandler func(ctx *RequestCtx, err error)
 
+	// ResponseErrorHandler is called in case of an error while writing the response
+	//
+	// Useful if you want to log details of write failures for compliance and
+	// business critical requests
+	// The handler doesn't need to modify the response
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -166,6 +166,13 @@ type Server struct {
 	//   * ErrBrokenChunks
 	ErrorHandler func(ctx *RequestCtx, err error)
 
+	// ResponseErrorHandler is called in case of an error while writing the response
+	//
+	// Useful if you want to log details of write failures for compliance and
+	// business critical requests
+	// The handler doesn't need to modify the response
```
erikdubbelboer: Can you turn this into a big fat warning. I think if you do things to modify the response it will end up very badly!

erikdubbelboer:
@gallir have you had time to look at the change I requested? Your pull request also has a conflict in `server.go`, can you please fix this.

Issue #541, opened 2019-02-14
if using fasthttp.RequestCtx as context.Context, it will panic during server shutdown


peczenyj:
If I use `*fasthttp.RequestCtx` as `context.Context`

during a `server.Shutdown()`

It will panic because we expect the context.Err() will be different than nil when the context id Done.

More information:
https://github.com/valyala/fasthttp/commit/51532b9517b3a01b442e0d58bcfd7fffbf183c68#commitcomment-32317617

related to issue #459

peczenyj:
example of panic by composing the context with another context.
I trap signal 15 to perform a `server.Shutdown()` so to emulate this we must start an http request ( via curl )

```go
func handler(ctx *fasthttp.RequestCtx) {
  		logrus.Infof("start timeout countdown")
		c, cancel := context.WithTimeout(ctx, 5*time.Second)
		defer cancel()
		<-c.Done()
		logrus.Infof("start timeout done: %v", c.Err())
}
```

peczenyj:
Of course, I forget update `TestShutdownDone` to check if ctx.Err() is not nil...

Arnold1:
hi is it allowed to use ctx in context.WithTimeout and can i write to ctx after c.Done?
```
c, cancel := context.WithTimeout(ctx, 5*time.Second)
defer cancel()
<-c.Done()
ctx.Write("test")
```

erikdubbelboer:
1) You can write to `ctx` as long as the handler that the `ctx` came from hasn't returned yet.
1) Your code looks fine yes.
1) There is always a slight overhead yes. Contexts tend to generate garbage which is also means longer garbage collection pauses.
1) You don't need to use any mutex in the example you gave.

Arnold1:
how to avoid using context.WithTimeout to get less gc?

erikdubbelboer:
You can't avoid garbage when you use context. Then you need to build your own timeout mechanism.

Arnold1:
my process actually talks to another server and waits for response.... and when the outside does an timeout i want the sending to be stopped... how to achieve that?
```
func process(respChan chan<- ResultData) {
    httpReq := fasthttp.AcquireRequest()
    fhttpReq.Header.Add("Content-Type", "application/json")
    fhttpReq.Header.SetMethod("POST")
    fhttpReq.SetRequestURI(url + "/test")
    fhttpReq.SetBody(marshalledReq)
    fresp := fasthttp.AcquireResponse()
[...]

PR #542, opened 2019-02-14
update server.go

peczenyj:
fix issue #541

### DIFF
--- a/server.go
+++ b/server.go
@@ -2,6 +2,7 @@ package fasthttp
 
 import (
 	"bufio"
+	"context"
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -2328,7 +2329,12 @@ func (ctx *RequestCtx) Done() <-chan struct{} {
 // This method always returns nil and is only present to make
 // RequestCtx implement the context interface.
```
erikdubbelboer: Can you please update the function documentation to match.

peczenyj:
plase see #544

PR #544, opened 2019-02-15
fix context Err() issue

peczenyj:
add unit text
improve documentation

fix issue #541
better version of PR #542

### DIFF
--- a/server.go
+++ b/server.go
@@ -2,6 +2,7 @@ package fasthttp
 
 import (
 	"bufio"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -2322,13 +2323,15 @@ func (ctx *RequestCtx) Done() <-chan struct{} {
 // successive calls to Err return the same error.
 // If Done is not yet closed, Err returns nil.
 // If Done is closed, Err returns a non-nil error explaining why:
-// Canceled if the context was canceled
+// Canceled if the context was canceled ( via server Shutdown )
```
kirillDanshin:
```suggestion
// Canceled if the context was canceled (via server Shutdown)
```

server_test.go
```diff
@@ -2857,6 +2857,43 @@ func TestShutdownDone(t *testing.T) {
 	verifyResponse(t, br, StatusOK, "aaa/bbb", "real response")
 }
 
+func TestShutdownErr(t *testing.T) {
+	ln := fasthttputil.NewInmemoryListener()
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			// This will panic, but I was not able to intercept with recover()
```
kirillDanshin: why?

peczenyj: The origin of the panic is here: `/usr/local/go/src/context/context.go:259`

```go
241 // propagateCancel arranges for child to be canceled when parent is.
242 func propagateCancel(parent Context, child canceler) {
...
259     go func() {
260       select {
261       case <-parent.Done():
262         child.cancel(false, parent.Err())
263       case <-child.Done():
264       }
265     }()
266   }
267 }
```
this in some point will call:

```go
346 // cancel closes c.done, cancels each of c's children, and, if
347 // removeFromParent is true, removes c from its parent's children.
348 func (c *cancelCtx) cancel(removeFromParent bool, err error) {
349   if err == nil {
350     panic("context: internal error: missing cancel error")
351   }
```

but since it happens in a goroutine started on `propagateCancel`, I had no success in add a `defer` with `recover` the panic.

PR #545, opened 2019-02-17
add tests for copyto

xuecai:
when write test RequestCopyTO , i find rawHeadersCopy nocopy in ReqeustHeader's CopyTo. i add it.
is it a buf for no need copy?


erikdubbelboer:
Good catch!

PR #556, opened 2019-03-21
update LBClients : add & remove HostClient

goith:
when using etcd in the k8s, addr may be changed, register or unregister. in this case, we need to update LBClients.

### DIFF
--- a/lbclient.go
+++ b/lbclient.go
@@ -98,6 +98,18 @@ func (cc *LBClient) init() {
 	// hammering from a cluster of identical LBClients.
 	cc.nextIdx = uint32(time.Now().UnixNano())
 }
+// clients update: add & remove
+func (cc *LBClient) AddClients(c *HostClient) int {
+    cc.cs = append(cc.cs, &lbClient{
+        c:           c,
+        healthCheck: cc.HealthCheck,
+    })
+    return len(cc.cs)
+}
+func (cc *LBClient) RemoveClients(index int) int {                                                                                                           
+    cc.cs = append(cc.cs[:index], cc.cs[index+1:]...)
+    return len(cc.cs)
+}
 
 func (cc *LBClient) get() *lbClient {
 	cc.once.Do(cc.init)
--- a/server.go
+++ b/server.go
@@ -1869,7 +1869,7 @@ func (s *Server) serveConn(c net.Conn) error {
 		if err != nil {
 			if err == io.EOF {
 				err = nil
-			} else if connRequestNum > 1 && err == errNothingRead {
+			} else if connRequestNum >= 1 && err == errNothingRead {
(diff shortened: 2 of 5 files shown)


server.go
```diff
@@ -1869,7 +1869,7 @@ func (s *Server) serveConn(c net.Conn) error {
 		if err != nil {
 			if err == io.EOF {
 				err = nil
-			} else if connRequestNum > 1 && err == errNothingRead {
+			} else if connRequestNum >= 1 && err == errNothingRead {
```
erikdubbelboer: Why change this?

lbclient.go
```diff
@@ -98,6 +98,18 @@ func (cc *LBClient) init() {
 	// hammering from a cluster of identical LBClients.
 	cc.nextIdx = uint32(time.Now().UnixNano())
 }
+// clients update: add & remove
+func (cc *LBClient) AddClients(c *HostClient) int {
```
erikdubbelboer: Since it's only adding one the method should be called `AddClient`.

lbclient.go
```diff
@@ -98,6 +98,18 @@ func (cc *LBClient) init() {
...
+    cc.cs = append(cc.cs, &lbClient{
+        c:           c,
+        healthCheck: cc.HealthCheck,
+    })
```
erikdubbelboer: This is not thread safe. `cc.cs` can be accessed from multiple goroutines. This will require a `sync.Mutex`.

lbclient.go
```diff
@@ -98,6 +98,18 @@ func (cc *LBClient) init() {
...
+    return len(cc.cs)
+}
+func (cc *LBClient) RemoveClients(index int) int {                                                                                                           
+    cc.cs = append(cc.cs[:index], cc.cs[index+1:]...)
+    return len(cc.cs)
+}
```
erikdubbelboer: I don't really like removing by index. When you remove one the indexes change so all other `RemoveClients` won't do what you might think they do.

Would it be possible to remove by `Addr` for example where you loop over the HostClients and remove the one with the matching Addr. Or maybe add a callback to `RemoveClient` like `func(*HostClient) bool` so the user can decide which ones to remove?

lbclient.go
kirillDanshin: sync.Mutex will kill the performance, we should think of alternative solutions first

erikdubbelboer: No it won't. A slice get guarded by a Lock and Unlock only takes around 22 nanoseconds. Any network operation is going to be multiple magnitudes of orders slower. Unless you're doing hundreds of thousands of Lock/Unlock operations per second optimizing a Mutex makes no sense at all.

Fenny:
I think this PR can be closed since the author did not respond to the requested changes and the current state would impact the performance.

Issue #491, opened 2018-12-14
the master branch's code is ok ?


goith:
1、today  i  go get  new  fasthttp  code,  i found  when the browse  send a  request, somtimes it recevie a error message "Error when parsing request"

2、why use os.Command().Output() int my handler,  the bin will output "child exited" ?

goith:
`} else if connRequestNum > 1 && err == errNothingRead { `

this line  ' connRequestNum > 1' ,  change to " connRequestNum >= 1" ？

erikdubbelboer:
The code is ok. That line doesn't need any change, it handles the cases of an idle connection being closed from the client side.

1. We need more code and describe exactly what you are doing to diagnose this issue.
2. I don't understand. Can you show the code you are using?

erikdubbelboer:
@goith without seeing the code it's really hard to tell what is going wrong. If you don't want other people to see the code you can also email it to erik@dubbelboer.com so only I can see.

goith:
```
// Create custom server.
    s := &fasthttp.Server{
        Handler:     h,
        ReadTimeout: 5 * time.Second,
    }

    if err := s.ListenAndServe(*addr); err != nil {
        log.Fatalf("Error in ListenAndServe: %s", err)
    }
```

stderr:

> 2018/12/25 18:52:03 error when serving connection "10.135.54.224:9080"<->"10.129.157.9:58242": read timeout with nothing read

above is example,  i  found it's  the ReadTimeout' setting.

erikdubbelboer:
Can you show the code of your `h` function? What browser are you using?

goith:
@erikdubbelboer  just using "examples/helloworldserver/helloworldserver.go"

erikdubbelboer:
When running the following code I don't see any problems with the latest versions of Firefox or Chrome. What browser are you using and can you maybe share your whole code? You're being to vague about everything to be able to find the issue.

```go
package main
  
import (
  "fmt"
  "time"

  "github.com/valyala/fasthttp"
)

func h(ctx *fasthttp.RequestCtx) {
  fmt.Fprintf(ctx, "Hello, world!")
}

func main() {
  s := &fasthttp.Server{
    Handler:     h,
    ReadTimeout: 5 * time.Second,
  }

  if err := s.ListenAndServe(":8080"); err != nil {
    panic(err)
  }
}
```

ErgoZUpdates:
> @erikdubbelboer
[...]

PR #564, opened 2019-04-15
Timing fixes

### LINKED ISSUE #563: What's the difference among many timeouts
There're many timeout related configurations, but some of those comments are a little bit vague.

```
	// Maximum duration for reading the full request (including body).
	//
	// This also limits the maximum duration for idle keep-alive
	// connections.
	//
	// By default request read timeout is unlimited.
	ReadTimeout time.Duration

	// Maximum duration for writing the full response (including body).
	//
	// By default response write timeout is unlimited.
	WriteTimeout time.Duration
```

* For ReadTimeout, when it starts timing? Does it mean that a keep-alive connection can be used no more than ReadTimeout, no matter how many times it has been used? What's the difference between ReadTimeout and MaxKeepaliveDuration?
* WriteTimeout, when it starts timing?
* What should I set if I expect an incoming request is processed no more than X seconds,  and will get a error if timeout?

erikdubbelboer:
- Improved timing code to be much cleaner
- Add IdleTimeout
- Remove obsolete optimization (was fixed in Go 2 years ago: https://github.com/golang/go/issues/15133#issuecomment-271571395)

Fixes #563 and potentially #491

### DIFF
--- a/server.go
+++ b/server.go
@@ -199,18 +199,25 @@ type Server struct {
 	// Default buffer size is used if not set.
 	WriteBufferSize int
 
-	// Maximum duration for reading the full request (including body).
-	//
-	// This also limits the maximum duration for idle keep-alive
-	// connections.
+	// ReadTimeout is the amount of time allowed to read
+	// the full request including body. The connection's read
+	// deadline is reset when the connection opens, or for
+	// keep-alive connections after the first byte has been read.
 	//
 	// By default request read timeout is unlimited.
 	ReadTimeout time.Duration
 
-	// Maximum duration for writing the full response (including body).
+	// WriteTimeout is the maximum duration before timing out
+	// writes of the response. It is reset after the request handler
+	// has returned.
 	//
 	// By default response write timeout is unlimited.
 	WriteTimeout time.Duration
+
+	// IdleTimeout is the maximum amount of time to wait for the
+	// next request when keep-alives are enabled. If IdleTimeout
+	// is zero, the value of ReadTimeout is used.
+	IdleTimeout time.Duration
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
@kirillDanshin what do you think about this? I have been testing this in a high RPS environment for some days now and it performs really well.

server.go
```diff
@@ -199,19 +199,26 @@ type Server struct {
 	// Default buffer size is used if not set.
 	WriteBufferSize int
 
-	// Maximum duration for reading the full request (including body).
-	//
-	// This also limits the maximum duration for idle keep-alive
-	// connections.
+	// ReadHeaderTimeout is the amount of time allowed to read
```
kirillDanshin: ReadTimeout

server.go
```diff
@@ -226,15 +233,9 @@ type Server struct {
...
-	// Maximum keep-alive connection lifetime.
-	//
-	// The server closes keep-alive connection after its' lifetime
-	// expiration.
-	//
-	// See also ReadTimeout for limiting the duration of idle keep-alive
-	// connections.
-	//
-	// By default keep-alive connection lifetime is unlimited.
+	// Deprecated!
```
kirillDanshin: https://rakyll.org/deprecated/

server.go
erikdubbelboer: Good catch.

server.go
erikdubbelboer: I didn't know that!

erikdubbelboer:
@kirillDanshin any thoughts on this? Otherwise I'll merge it as it seems to be good now.

PR #567, opened 2019-04-30
Added methods constants

toby3d:
It can little bit reduce compile time and help avoid typos.


methods.go
```diff
@@ -0,0 +1,14 @@
+package fasthttp
+
+// HTTP methods were stolen from net/http.
```
erikdubbelboer: Please change to `copied from` :)

toby3d: Fixed :ok_hand:

erikdubbelboer:
Thanks

kirillDanshin:
We need to add Go license, e.g. in `/third-party_licenses/golang`, and apply this license to the new file

erikdubbelboer:
Or we just remove that one line. This is such basic code and doesn't affect any functionality. I don't think anyone will care.

PR #571, opened 2019-05-14
Fix comment.

imxyb:
Signed-off-by: Xiao YongBiao <xyb4638@gmail.com>

### DIFF
--- a/server.go
+++ b/server.go
@@ -215,7 +215,7 @@ type Server struct {
 	WriteTimeout time.Duration
 
 	// IdleTimeout is the maximum amount of time to wait for the
-	// next request when keep-alives are enabled. If IdleTimeout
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -215,7 +215,7 @@ type Server struct {
 	WriteTimeout time.Duration
 
 	// IdleTimeout is the maximum amount of time to wait for the
-	// next request when keep-alives are enabled. If IdleTimeout
+	// next request when keep-alive are enabled. If IdleTimeout
```
erikdubbelboer: If you're going to fix it then please change it to `next request when keep-alive is enabled.`. Thanks.

imxyb: Done.

erikdubbelboer:
Thanks.

PR #575, opened 2019-05-18
Option for immediate header flush

Bobochka:
In case of body stream if first body bytes generation takes long this adds an option of "faster response".

Will add tests if general idea is ok.

### DIFF
--- a/http.go
+++ b/http.go
@@ -62,7 +62,8 @@ type Response struct {
 	// Response header
 	//
 	// Copying Header by value is forbidden. Use pointer to Header instead.
-	Header ResponseHeader
+	Header               ResponseHeader
+	immediateHeaderFlush bool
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
I'm wondering if this should be two functions of if it should just be an exported field named `Response.FlushHeaders bool`. What do you think @kirillDanshin ?

Bobochka:
@erikdubbelboer @kirillDanshin Exported field and added tests

Bobochka:
@kirillDanshin any action from my side? We're already running a fork in prod that always flushes before body.

erikdubbelboer:
Thanks. We don't need any more testing seeing as this doesn't change anything if you don't set the value to true. And even then, just calling `w.Flush()` is not something complicated.

PR #579, opened 2019-05-27
Added headers constants

toby3d:
Used headers lists from [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers)
P.S.: Not included obsolete or deprecated headers.


erikdubbelboer:
I like it!

But on the one hand this will greatly expand our API surface that we will have to maintain, on the other hand these are only constants so it's pretty safe and won't require any changes in the foreseeable future.

kirillDanshin:
Thanks!

PR #581, opened 2019-05-28
Support SameSite value "None" cookie attribute

RemiSirdata:
As you may have heard Google recently announced changes coming to Chrome with the release of version 76 in July 2019. This change will directly impact the handling of 3rd party cookies.

For Chrome, set the cookie attribute “SameSite = None”, so that your cookie will be sent in the third party context. Leaving the attribute of “SameSite” unspecified would result in your cookie not being sent in case the user enables the “same-site-by-default-cookies” setting in their browser.
Set "SameSite = None"  is required is this case alongside "Secure"

You can find more detailed information here: 
https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
https://blog.chromium.org/2019/05/improving-privacy-and-security-on-web.html

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -398,7 +398,7 @@ func (c *Cookie) ParseBytes(src []byte) error {
 						if caseInsensitiveCompare(strCookieSameSiteStrict, kv.value) {
 							c.sameSite = CookieSameSiteStrictMode
 						}
-					case 'n': // "strict"
+					case 'n': // "none"
 						if caseInsensitiveCompare(strCookieSameSiteNone, kv.value) {
 							c.sameSite = CookieSameSiteNoneMode
 						}
--- a/cookie.go
+++ b/cookie.go
@@ -31,6 +31,9 @@ const (
 	CookieSameSiteLaxMode
 	// CookieSameSiteStrictMode sets the SameSite flag with the "Strict" parameter
 	CookieSameSiteStrictMode
+	// CookieSameSiteStrictMode sets the SameSite flag with the "None" parameter
+	// see https://tools.ietf.org/html/draft-west-cookie-incrementalism-00
+	CookieSameSiteNoneMode
 )
 
 // AcquireCookie returns an empty Cookie object from the pool.
@@ -119,8 +122,12 @@ func (c *Cookie) SameSite() CookieSameSite {
 }
 
 // SetSameSite sets the cookie's SameSite flag to the given value.
+// set value CookieSameSiteNoneMode will set Secure to true also to avoid browser rejection
 func (c *Cookie) SetSameSite(mode CookieSameSite) {
 	c.sameSite = mode
+	if mode == CookieSameSiteNoneMode {
+		c.SetSecure(true)
+	}
 }
 
 // Path returns cookie path.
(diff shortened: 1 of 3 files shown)


cookie.go
```diff
@@ -119,8 +122,12 @@ func (c *Cookie) SameSite() CookieSameSite {
 }
 
 // SetSameSite sets the cookie's SameSite flag to the given value.
+// set value CookieSameSiteNoneMode will set Secure to true also to avoid browser rejection
 func (c *Cookie) SetSameSite(mode CookieSameSite) {
 	c.sameSite = mode
+	if mode == CookieSameSiteNoneMode {
+		c.SetSecure(true)
+	}
```
erikdubbelboer: I don't think we should do this by default. This might cause cookies to work if the server is HTTP only. Requiring Secure for SameSiteNone is a separate flag in Chrome that isn't enabled by default.

cookie.go
```diff
@@ -31,6 +31,9 @@ const (
 	CookieSameSiteLaxMode
 	// CookieSameSiteStrictMode sets the SameSite flag with the "Strict" parameter
 	CookieSameSiteStrictMode
+	// CookieSameSiteStrictMode sets the SameSite flag with the "None" parameter
```
erikdubbelboer: Should be `CookieSameSiteNoneMode`.

cookie.go
RemiSirdata: In the specifications same-site with value "None" must have Secure parameter otherwise the cookie is ignored (https://tools.ietf.org/html/draft-west-cookie-incrementalism-00)
I hesitate between call SetSecure(true) in SetSameSite and AppendBytes() 

cookie.go L280
`if c.secure {`
to
`if c.secure || c.sameSite == CookieSameSiteNoneMode {`

I think we should force the value because the specification make it clear and developers could not be aware.

RemiSirdata:
Mike West (Chrome security team) indicate default value to Lax have been suggest for 77 (September) for not approved yet, Secure bit either.

erikdubbelboer:
Lets wait until Chrome changes this before we do.

kirillDanshin:
Gecko going to default to Lax as well (w3ctag/design-reviews#373, [mozilla.dev.platform list]( https://groups.google.com/forum/#!msg/mozilla.dev.platform/nx2uP0CzA9k/BNVPWDHsAQAJ))

I'd like to @-mention one of primary contacts here: @mikewest, maybe the author of this idea got some comments or ideas here.

httpwg/http-extensions#788

cookie.go
peczenyj: My two cents.
https://tools.ietf.org/html/draft-west-cookie-incrementalism-00#section-3.2 the specification is clear. What we can consider is create an extra flag in `fasthttp` configuration to `refuse non secure SameSite None` cookies. This will affect how we parse cookies, but I am not sure how it should affect writing cookies. It is really easy to force a non secure cookie with SameSite None.

Perhaps we can include a `Valid() error` method on cookie, who can verify the inner state of the cookie, like parse but to itself.

it is better than just refuse `SetSecure(false)` if `SameSite=None` IMHO

cookie.go
```diff
@@ -386,6 +398,10 @@ func (c *Cookie) ParseBytes(src []byte) error {
 						if caseInsensitiveCompare(strCookieSameSiteStrict, kv.value) {
 							c.sameSite = CookieSameSiteStrictMode
 						}
+					case 'n': // "strict"
```
erikdubbelboer: The comment should say `"none"`.

RemiSirdata: My bad, fixed

erikdubbelboer:
Thanks.

PR #583, opened 2019-05-28
add cookie SameSite=None support fix #582

### LINKED ISSUE #582: cookie should support cookie option SameSite=None
Reason

> Chrome 76 introduces a new `same-site-by-default-cookies` flag. Setting this flag will shift the default behaviour of `Set-Cookie` to apply `SameSite=Lax` if no other `SameSite` value is provided.

https://web.dev/samesite-cookies-explained/
https://blog.chromium.org/2019/05/improving-privacy-and-security-on-web.html

I'd like to set `SameSite=None` and ignore chrome's default behaviour

peczenyj:
this will fix issue #582

to force set a cookie with `SameSite=None`


erikdubbelboer:
We are already having a discussion and pull request for this here: https://github.com/valyala/fasthttp/pull/581

PR #613, opened 2019-07-19
improve fasthttpadaptor

wencan:


erikdubbelboer:
I'm not sure what I should think of the usage of `b2s`. Normally it's safe for a `http.Handler` to keep references to any data from the request (method, uri, headers). The way you use `b2s` makes it faster but it also breaks this compatibility. People who want speed shouldn't be using fasthttpadaptor at all. On the other hand maybe some people are stuck with a library that only supports `http.Handler`.

kirillDanshin:
@erikdubbelboer I don't like that b2s changes, that could break something horribly.

wencan:
@erikdubbelboer @kirillDanshin 
I have used string instead of b2s

fasthttpadaptor/adaptor.go
```diff
@@ -48,95 +51,125 @@ func NewFastHTTPHandlerFunc(h http.HandlerFunc) fasthttp.RequestHandler {
...
-			return
+		reqBody := ctx.PostBody()
+		if reqBody != nil {
+			r.Body = ioutil.NopCloser(bytes.NewReader(reqBody))
+			r.ContentLength = int64(len(reqBody))
 		}
-		r.URL = rURL
+		r.Host = string(ctx.Host())
+		r.RemoteAddr = ctx.RemoteAddr().String()
+		r.RequestURI = string(ctx.RequestURI())
```
erikdubbelboer: You're doing `string(ctx.RequestURI())` above as well, might as well combine them for one less allocation.

wencan: Hello.
I added a stringsBuilder to combine multiple bytes into one string.
I also added BenchmarkAdaptor.

erikdubbelboer: Interesting concept the `stringsBuilder` but I'm afraid I don't see the use here? You're just putting everything into one big buffer. That's not what I meant when I said `combine them for one less allocation`. I meant doing `url.ParseRequestURI(r.RequestURI)` like you already changed. How does `stringsBuilder` improve anything?

wencan: I use the stringsBuilder  to  in the initHTTPRequest function convert the method, uri, host, header to a string. It only needs to allocate memory once.
I think stringbuilder is very important if the header is large.

erikdubbelboer:
Can you put `stringsBuilder` in a separate package. Maybe also add a `sync.Pool` for the `lengths` slice. And I would like to see some tests to make sure this doesn't break in the future. I don't like the added complexity of having to read strings from the builder in exactly the same order you write them in. So that needs to be tested well.

wencan:
Your proposal is very good. I have also considered those.
But I don't know which package to put it in. Do you have any good advice?
The stringsBuilder methods looks very bad. I must improve the stringsBuilder methods, if I need to implement the test code.

erikdubbelboer:
I would be ok with `fasthttp/stringsbuilder`.

wencan:
I can add stringsBuilder to [bytespool](https://github.com/wencan/bytespool).
But fasthttp has been import bytebufferpool.

erikdubbelboer:
What's the difference between https://godoc.org/github.com/wencan/bytespool and https://godoc.org/github.com/valyala/bytebufferpool ? I'd rather have the code as module inside fasthttp instead of having another dependency.

wencan:
How about move stringsBuilder to fasthttp/fasthttputil?
I think we don't should create too many internal packages.

erikdubbelboer:
I don't have control over `fasthttp/fasthttputil`. Fasthttp doesn't have too many packages so adding one more is fine. I'd rather keep things internal so things can be updated with one pull request instead of many on different repos.

wencan:
Sorry, I can't understand what you mean.
Can I add stringsBuilder to fasthttp/fasthttputil?

erikdubbelboer:
Sorry my bad, I though you meant a different package with `fasthttp/fasthttputil`. Yes you can add it to this directory.

wencan:
I have re-pushed it.

fasthttputil/stringslices.go
```diff
@@ -0,0 +1,107 @@
...
+
+	lastError error
+
+	str string
+}
+
+// WriteBytes writes the byte slice of a string.
+func (slices *StringSlices) WriteBytes(bytes []byte) error {
+	if slices.lastError != nil {
+		return nil
```
erikdubbelboer: Shouldn't it return `slices.lastError` here?

fasthttputil/stringslices.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
```
erikdubbelboer: You should prevent new calles to `WriteBytes` after `NextStringSlice` has been called.

fasthttputil/stringslices.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// ReleaseStringSlices returns slices acquired via AcquireStringSlices to pool.
+func ReleaseStringSlices(slices *StringSlices) {
+	slices.Reset()
+	stringSlicesPool.Put(slices)
+}
+
+// StringSlices is used to convert byte slices to string slices,
+// it only needs to allocate memory once.
```
erikdubbelboer: Can you explain here why it only allocates once and why this is useful?

fasthttpadaptor/adaptor_timing_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"fmt"
+	"net"
+	"net/http"
+	"testing"
+
+	"github.com/valyala/bytebufferpool"
+	"github.com/valyala/fasthttp"
+)
+
+func BenchmarkAdaptor(b *testing.B) {
```
erikdubbelboer: Have you tried running this benchmark before and after your changes to see the improvements?

fasthttputil/stringslices.go
erikdubbelboer: And you should definitely explain here or below that each `WriteBytes` and `NextStringSlice` call must match up as `NextStringSlice` returns things in the same order as `WriteBytes` has written them.

I'm wondering if it would be possible to tag each string. So it would be `s.WriteBytes("method", methodBytes)` and `s.GetString("method")`. This might be hard to do without allocations I guess. It would require `lengths` to be a `map[string]int`.

fasthttputil/stringslices.go
```diff
@@ -0,0 +1,107 @@
...
+		slices.buffer = bytebufferpool.Get()
+	}
+	len, lastError := slices.buffer.Write(bytes)
+	slices.lengths = append(slices.lengths, len)
+	return lastError
+}
+
+// NextStringSlice returns next string slice.
+// it returns empty string if StringSlices have no more string slices.
+func (slices *StringSlices) NextStringSlice() (string, bool) {
```
erikdubbelboer: Why the second return value? You seem to be ignoring it everywhere and you already have `Remain()`.

fasthttputil/stringslices.go
```diff
@@ -0,0 +1,107 @@
...
+
+// Remain returns the number of remaining readable string slices.
+func (slices *StringSlices) Remain() int {
+	return len(slices.lengths) - slices.pos
+}
+
+// LastError Return the last error of slices
+func (slices *StringSlices) LastError() error {
+	return slices.lastError
+}
```
erikdubbelboer: This method isn't really useful if `WriteBytes` just always returns this error.

Fenny:
I think this PR can be closed since the author did not respond to the requested changes and the current state would impact the performance.

PR #614, opened 2019-07-20
Add support for user:pass in URLs

### LINKED ISSUE #609: Can we support user:pass in RequestURIs?
Passing a URL struct would be nice... should not be getting this error:

`(*net.AddrError)(0xc0001ea000)(address user:pass@host:1234: too many colons in address)`

erikdubbelboer:
Fixes #609

### DIFF
--- a/http.go
+++ b/http.go
@@ -3,6 +3,7 @@ package fasthttp
 import (
 	"bufio"
 	"bytes"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


http.go
```diff
@@ -1148,6 +1149,24 @@ func (req *Request) Write(w *bufio.Writer) error {
...
+			if tl > cap(req.Header.bufKV.value) {
+				req.Header.bufKV.value = make([]byte, 0, tl)
+			}
+			buf := req.Header.bufKV.value[:0]
+			buf = append(buf, uri.username...)
+			buf = append(buf, strColon...)
+			buf = append(buf, uri.password...)
+			buf = buf[:tl]
+			base64.StdEncoding.Encode(buf[nl:], buf[:nl])
+			req.Header.SetBytesKV(strAuthorization, buf[nl:])
```
erikdubbelboer: I agree this is all a bit dirty to use the scratch buffer of the `req.Header`. But the alternative would be to have another `sync.Pool` for `[]byte` just for this.

dgrr: https://go-review.googlesource.com/c/go/+/37639/1/src/encoding/base64/base64.go

erikdubbelboer: That would change these lines:
```go
buf = buf[:tl]
base64.StdEncoding.Encode(buf[nl:], buf[:nl])
```
into:
```go
buf = base64.StdEncoding.AppendEncode(buf, buf[:nl])
```
but it wouldn't help avoiding with the allocation.

paween1980:
It's not work. My test URL is https://user1:pass1@sub.domain.com.
It's not set authorize Header correctly.

dgrr:
@paween1980 maybe it's what you pointed out [here](https://gist.github.com/dgrr/2e41b058db3ead93c161feb9472254a7#gistcomment-3039092)?

erikdubbelboer:
@paween1980 what doesn't work? Please share your code. It seems to work fine here: https://play.golang.org/p/M-I-i8VC1Kd

erikdubbelboer:
@paween1980 I see now, thanks for telling us! It should be fixed in https://github.com/valyala/fasthttp/commit/1d6a7e049393256c15158690904adc053338cf74

PR #620, opened 2019-08-01
ServerTrace

pencil:
This adds a mechanism to trace events that occur within the Server. I
added hooks for some common use cases that I could think of but it can
easily be expanded to other things as people see fit. The pattern is
heavily inspired by
[`net/http/httptrace`](https://golang.org/pkg/net/http/httptrace/#ClientTrace).

The most common use case for this will be logging and metrics: it allows
centrally logging the status code that was sent to clients, bytes
transferred in the process, errors that might have occurred during the
transfer, etc.

fasthttp already has some limited tracing capability for connection
state which I now marked as deprecated since the new tracing mechanism
encompasses all the previous hooks.

A side note regarding backward compatibility: I ended up adding an
additional int64 return value for many of the existing Write methods.
This was necessary in order to provide bytes written to the respective
trace hook - which I think is a very valuable data point for logging and
stats that might get generated. This however breaks backward
compatibility but it is trivial to update as the additional return
argument can simply be ignored. It also allowed for the removal of the
now unused statsWriter code.


http.go
```diff
@@ -1139,12 +1105,12 @@ func (req *Request) onlyMultipartForm() bool {
 // Write doesn't flush request to w for performance reasons.
 //
 // See also WriteTo.
-func (req *Request) Write(w *bufio.Writer) error {
+func (req *Request) Write(w *bufio.Writer) (nn int64, err error) {
```
erikdubbelboer: Yeah I wish we could just change our API like that. I really wanted to do this. But this simple change would break a lot of our users code. So I'm afraid we can't do this 😞 

Maybe if we would do a fasthttp 2 and change the import path we could do this. But not for now.

So please revert all these API breaking changes. I suggest maybe just not returning the amount of bytes in your trace functions. `net/http/httptrace` doesn't seem to have this either.

tsingson: it's cool feature. 
and yes, agree to @erikdubbelboer , don't change the API

pencil: Thanks for the feedback. I was hoping it would be OK since it is just an additional return value and it is really useful to have for logging. I can see why you would not want to break the API though and will update the PR accordingly 👍

pencil: I updated the PR to remove the backwards incompatible changes.

server.go
```diff
@@ -2218,9 +2252,7 @@ func writeResponse(ctx *RequestCtx, w *bufio.Writer) error {
 	if ctx.timeoutResponse != nil {
 		panic("BUG: cannot write timed out response")
 	}
-	err := ctx.Response.Write(w)
-	ctx.Response.Reset()
```
erikdubbelboer: I see you moved the `Response.Reset()` to the callers of `writeResponse` but you didn't do this for all callers causing some bugs.

pencil: Thanks, updated.

pencil:
With the fasthttp model, the handler loses control over the response handling as soon as it returns. It is impossible to log canonically wether or not a response was successfully served.

This is especially daunting when using a `BodyStreamWriter` that slowly writes data over time as it is common for clients to disconnect half-way through the process. The `workerPool` provides some limited logging capabilities but it can't be customized to include more context from the original request. The existing logging mechanism also is not useful for generating operational metrics (e.g. through StatsD or CloudWatch).

pencil:
Here is an example usage that logs the response with request context metadata using Uber's zap library. Additionally metrics to count status codes as well as timing information are emitted using the [cactus statsd client](https://github.com/cactus/go-statsd-client):

```
fasthttp.ServerTrace{
	GotRequest: func(ctx *fasthttp.RequestCtx) {
		ctx.SetUserValue(ctxKeyRequestStart, time.Now())
	},
	WroteResponse: func(ctx *fasthttp.RequestCtx, length int64, err error) {
		start := ctx.UserValue(ctxKeyRequestStart)
		var dur time.Duration
		if start != nil {
			if t, ok := start.(time.Time); ok {
				dur = time.Since(t)
			}
		}
		logger.Info("response served",
			zap.Error(err),
			zap.String("addr_local", ctx.LocalAddr().String()),
			zap.String("addr_remote", ctx.RemoteAddr().String()),
			zap.ByteString("path", ctx.Path()),
			zap.Int("status", ctx.Response.StatusCode()),
			zap.Duration("duration", dur),
		)
		statsClient.Inc(fmt.Sprintf("status.%d", ctx.Response.StatusCode()), 1.0, 0.1)
		statsClient.TimingDuration("request", dur, 0.1)
	},
}
```

server.go
```diff
@@ -332,6 +335,7 @@ type Server struct {
 	// ConnState specifies an optional callback function that is
 	// called when a client connection changes state. See the
 	// ConnState type and associated constants for details.
+	// Deprecated: Use <Trace> instead.
```
kirillDanshin: if we want to deprecate this in favor of ServerTrace, we should provide an identical feature. I don't feel like
```go
s.trace := &fasthttp. ServerTrace{
	GotConn:	connStateChanged,
	ClosedConn:	connStateChanged,
	ActivatedConn:	connStateChanged,
	IdledConn:	connStateChanged,
	HijackedConn:	connStateChanged,
}
```

instead of

```go
s.ConnState = connStateChanged
```

server.go
```diff
@@ -167,6 +167,9 @@ type Server struct {
 	//   * ErrBrokenChunks
 	ErrorHandler func(ctx *RequestCtx, err error)
 
+	// Trace is invoked for various events generated by this server.
+	Trace ServerTrace
```
kirillDanshin: I think this field should have more docs; and obviously we can't say that a struct can be invoked

server.go
```diff
@@ -1859,11 +1876,13 @@ func (s *Server) serveConn(c net.Conn) error {
 	ctx := s.acquireCtx(c)
 	ctx.connTime = connTime
 	isTLS := ctx.IsTLS()
+	if s.Trace.AcquiredContext != nil {
+		s.Trace.AcquiredContext(ctx)
```
kirillDanshin: looking at all of these hooks, I've got some questions:

- what will happen with a server, if s.Trace.AcquiredContext will block for 10 minutes under load?
- how much this will cost (in performance terms) to a person, who will not use this tracing?
- are we sure that we're not introducing unnecessary hooks?

server.go
pencil: Having separate hooks for the various state transitions simplifies the code that the user of `fasthttp` has to write.

In your example, `connStateChanged` might look like this:
```
func handleNewConnection(conn net.Conn) {
	logger.Info("new connection established", zap.String("remote_addr", conn.RemoteAddr().String()))
	// do other things
}

func handleClosedConnection(conn net.Conn) {
	logger.Info("connection closed", zap.String("remote_addr", conn.RemoteAddr().String()))
	// do other things
}

func connStateChanged(conn net.Conn, state ConnState) {
	switch state {
	case fasthttp.StateNew:
		handleNewConnection(conn)
	case fasthttp.StateClosed:
		handleClosedConnection(conn)
	// ...
	}
}
```

Where as using the new `ServerTrace` implementation, this code will become:
```
func handleNewConnection(conn net.Conn) {
	logger.Info("new connection established", zap.String("remote_addr", conn.RemoteAddr().String()))
	// do other things
}

func handleClosedConnection(conn net.Conn) {
	logger.Info("connection closed", zap.String("remote_addr", conn.RemoteAddr().String()))
	// do other things
}

s.Trace = fasthttp.ServerTrace{
	GotConn: handleNewConnection,
	ClosedConn: handleClosedConnection,
	// ...
}
```

As a user of `fasthttp` I personally would prefer that an improvement as I no longer have to deal with differentiating the various states.

server.go
pencil: I felt like most of the documentation would be redundant with what I included in the documentation for `ServerTrace` itself. I'm happy to expand here though if you have concrete suggestions.

server.go
pencil: >what will happen with a server, if s.Trace.AcquiredContext will block for 10 minutes under load?

If a user of the tracing functionality writes blocking code, it will absolutely affect the overall performance of the server. I can highlight in the documentation that all hooks are processed synchronously in a blocking manner.

>how much this will cost (in performance terms) to a person, who will not use this tracing?

For use cases that do not utilize any kind of tracing this introduces a few `s.Trace.<hook> != nil` checks. I can try to benchmark this but a simple comparison on a pointer should be negligible in terms of performance.

>are we sure that we're not introducing unnecessary hooks?

 I only included hooks for things that I started using in my production use case and the ones required to deprecate `ConnState`. If anything, there is probably room for additional hooks but I will leave that up to other contributors as it is hard to foresee other use cases.

erikdubbelboer:
The only benefit I see is that you get some more hooks and some hooks get more than just a `net.Conn` as argument. I'm not sure if that is worth having to maintain two API's over.

I would be fine with this only if we can really decide to deprecate ConnState and remove it completely within a couple of months. So we would need to for example print something to stdout [at this point if `ConnState != nil`](https://github.com/valyala/fasthttp/blob/d4833f60cbf32443bf047bf39f13f450862d355e/server.go#L1574). But I'm really not sure if this is a good idea.

Maybe we should just add more information to ConnState?

PR #628, opened 2019-08-12
Correctly handle `NoDefaultContentType` without setting an `Content-Type` value

cipriancraciun:
At the moment if one configures `NoDefaultContentType` and does not configure any `Content-Type` header (through any of the appropriate header methods), `fasthttp` outputs an invalid header with an empty value, i.e. `Content-Type: `

This small patch fixes this case.

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -1111,6 +1111,19 @@ func TestRequestHeaderCopyTo(t *testing.T) {
 	}
 }
 
+func TestResponseContentTypeNoDefaultNotEmpty(t *testing.T) {
+	var h ResponseHeader
+
+	h.noDefaultContentType = true
+	h.SetContentLength(5)
+
+	headers := h.String()
+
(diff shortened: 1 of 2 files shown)


cipriancraciun:
I've added a simple test.

I've tested both the reported behaviour, and the fact that my patch fixes that situation.

header_test.go
```diff
@@ -1111,6 +1111,19 @@ func TestRequestHeaderCopyTo(t *testing.T) {
 	}
 }
 
+func TestResponseContentTypeNoDefaultNotEmpty(t *testing.T) {
+	var h ResponseHeader
+
+	h.noDefaultContentType = true
+	h.SetContentLength (5)
```
erikdubbelboer: Please `go fmt` the code.

cipriancraciun:
:)  Sorry.  I'm using a completely different formatting "scheme" for my own projects.  :)

erikdubbelboer:
Thanks.

PR #629, opened 2019-08-12
Add support for `NoDefaultDate` option that excludes the sending of the `Date` header

cipriancraciun:
In certain situations (perhaps for security considerations, or perhaps performance, or perhaps fitting the whole response in a single network packet) one might want **not to send the `Date` header**.

This simple patch adds that option.

---

If one sets both `NoDefaultServerName`, `NoDefaultContentType`, and `NoDefaultDate`, then the simplest reply will contain just the HTTP status line and the `Content-Length` header.

For example:
```
HTTP/1.1 200 OK
Content-Length: 13

hello world!
```

### DIFF
--- a/header.go
+++ b/header.go
@@ -24,6 +24,7 @@ type ResponseHeader struct {
 	noHTTP11             bool
 	connectionClose      bool
 	noDefaultContentType bool
+	noDefaultDate        bool
 
 	statusCode         int
 	contentLength      int
@@ -643,6 +644,7 @@ func (h *ResponseHeader) DisableNormalizing() {
 func (h *ResponseHeader) Reset() {
 	h.disableNormalizing = false
 	h.noDefaultContentType = false
+	h.noDefaultDate = false
 	h.resetSkipNormalize()
 }
 
@@ -696,6 +698,7 @@ func (h *ResponseHeader) CopyTo(dst *ResponseHeader) {
 	dst.noHTTP11 = h.noHTTP11
 	dst.connectionClose = h.connectionClose
 	dst.noDefaultContentType = h.noDefaultContentType
+	dst.noDefaultDate = h.noDefaultDate
 
 	dst.statusCode = h.statusCode
 	dst.contentLength = h.contentLength
@@ -1493,7 +1496,9 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 	if len(server) != 0 {
 		dst = appendHeaderLine(dst, strServer, server)
 	}
-	dst = appendHeaderLine(dst, strDate, serverDate.Load().([]byte))
+	if !h.noDefaultDate {
+		dst = appendHeaderLine(dst, strDate, serverDate.Load().([]byte))
+	}
 
 	// Append Content-Type only for non-zero responses
 	// or if it is explicitly set.
@@ -1511,7 +1516,7 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 
 	for i, n := 0, len(h.h); i < n; i++ {
 		kv := &h.h[i]
-		if !bytes.Equal(kv.key, strDate) {
+		if h.noDefaultDate || !bytes.Equal(kv.key, strDate) {
 			dst = appendHeaderLine(dst, kv.key, kv.value)
 		}
 	}
--- a/server.go
+++ b/server.go
@@ -329,6 +329,13 @@ type Server struct {
 	// set to true, the Content-Type will not be present.
 	NoDefaultContentType bool
 
+	// NoDefaultDateHeader, when set to true, causes the default Date
+	// header to be excluded from the Response.
+	//
+	// The default Date header value is the current date value. When
+	// set to true, the Date will not be present.
+	NoDefaultDate bool
+
 	// ConnState specifies an optional callback function that is
 	// called when a client connection changes state. See the
 	// ConnState type and associated constants for details.
@@ -1910,6 +1917,7 @@ func (s *Server) serveConn(c net.Conn) error {
 
 		ctx.Request.isTLS = isTLS
 		ctx.Response.Header.noDefaultContentType = s.NoDefaultContentType
+		ctx.Response.Header.noDefaultDate = s.NoDefaultDate
 
 		if err == nil {
 			if s.ReadTimeout > 0 {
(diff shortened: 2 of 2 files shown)


header.go
```diff
@@ -1508,9 +1513,7 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 
 	for i, n := 0, len(h.h); i < n; i++ {
 		kv := &h.h[i]
-		if !bytes.Equal(kv.key, strDate) {
-			dst = appendHeaderLine(dst, kv.key, kv.value)
-		}
+		dst = appendHeaderLine(dst, kv.key, kv.value)
```
erikdubbelboer: I don't think these lines should be changed. Before setting your own `Date` header would always get ignored. With this change it will be included which can cause duplicate `Date` headers.
You could do `if noDefaultDate || !bytes.Equal(kv.key, strDate) {` if you think people should be able to set their own `Date` header if they set `NoDefaultDate` to `true`.

cipriancraciun: The `Date` header will not get duplicated because:
* if `server.NoDefaultDate` is set, then the server won't add the `Date` header itself, as per:
https://github.com/valyala/fasthttp/pull/629/files#diff-5e68413c67a51bcc7bf6e42ec37bde13L1496-R1501
* if one insists on adding a `Date` header, one can use the `headers.Add("Date", "...")` method, which bypasses the `SetCanonical` method;

The patch I've submitted allows both behaviours:
* either not sending at all the `Date` header, by just setting `server.NoDefaultDate = true`;
* either sending a custom `Date` header by setting again `server.NoDefaultDate = true`, but also calling `response.Headers.Add("Date", "...")`;

After looking through `header.go`, the only way to set the `Date` header is either through `SetCanonical` (which ignores it) or through `Add` (which will allow it to be set);  (and through their alternate forms that in the end call these two variants).

Thus the question is how often one uses the `Add` method;  moreover how often is it used to set the `Date` header.

erikdubbelboer: I'm afraid that's not true. A very common case for fashttp is to use it as a proxy, see: https://gist.github.com/erikdubbelboer/a7c55d8a6dd467a3840e35352ccf180c

With your patch this now outputs:
```
HTTP/1.1 200 OK
Server: fasthttp
Content-Length: 0
Date: Mon, 12 Aug 2019 19:51:35 GMT
Date: Mon, 12 Aug 2019 19:51:35 GMT
```

cipriancraciun: OK, then if I use as suggested `if noDefaultDate || !bytes.Equal(kv.key, strDate) { appendHeader(...) }`, then by default the "current" behavior is retained, while by setting `NoDefaultDate = true` will disable both setting the `Date` header by default, but also will not filter the `Date` header anymore.

Would this be OK?

erikdubbelboer: That sounds good yes.

cipriancraciun: I've amended my last commit in-line with your suggestion.

kirillDanshin:
@cipriancraciun I would appreciate if you could provide some details on your original idea behind.

speaking of performance, I don't see how OpAtomicLoadPtr (which is basically MOVQ on most target systems) brings performance down.
speaking of security, I don't see any threats either, due to the fact that time is rounded to the seconds.

RFC 7231 (7.1.1.2) states very clearly that while we can omit Date in 1xx/5xx, we must send it with other status codes as long as we have a clock at all

cipriancraciun:
> @cipriancraciun I would appreciate if you could provide some details on your original idea behind.
> 
> speaking of performance, I don't see how OpAtomicLoadPtr (which is basically MOVQ on most target systems) brings performance down.

Indeed, out of the box, just eliminating a pointer reference and appending some bytes doesn't achieve much performance gains.  (Although in my own benchmarks the performance, for a dummy server, grew from 130K requests/second to 135K, i.e. 3%.)


However, this feature (of not setting the date header) goes hand-in-hand with an additional set of changes (that I haven't yet "perfected" for publishing), that take `fasthttp` a step further:  just like a few weeks ago I made a pull request for adding `SetBodyRaw`, with which one can give `fasthttp.Response` a `[]byte` to be used as a body, without any additional copies, this time I wanted to implement the same for response headers, namely complete control over what gets sent as part of the HTTP response "metadata", including status line and headers.

For example in my own patch I've added `hRaw`, which if present is the actual header representation, that bypasses any other logic inside `ResponseHeader`:
https://github.com/cipriancraciun/go-fasthttp/blob/d5d95a7316eaad8fd79f0d030c24cf55ec95ec52/header.go#L1518-L1521

Similarly, as an alternative, I've added support for passing `[]byte` slices that, when concatenated, form a correct chain of headers, that are appended at the end:
https://github.com/cipriancraciun/go-fasthttp/blob/d5d95a7316eaad8fd79f0d030c24cf55ec95ec52/header.go#L1560-L1562

Thus the `NoDefaultDate` goes hand-in-hand with those future patches.


> speaking of security, I don't see any threats either, due to the fact that time is rounded to the seconds.

Perhaps there is no direct threat to `fasthttp`, but perhaps many (not that well written libraries) use the current timestamp for all sorts of "security" related purposes.  Therefore knowing that the server is well in the future, or perhaps well in the past, one can try to use that information and pass "old" sniffed data.

(From a security perspective I would say, the least amount of data "sent", the safer it is; 
 although not sending the `Date` header, if it is so uncommon, might actually increase the likelihood of correctly "guessing" that `fasthttp` with this option set is used behind the scenes...)


> RFC 7231 (7.1.1.2) states very clearly that while we can omit Date in 1xx/5xx, we must send it with other status codes as long as we have a clock at all

The full quote of that section is:
https://tools.ietf.org/html/rfc7231#section-7.1.1.2

>  An origin server MUST NOT send a Date header field if it does not
>  have a clock capable of providing a reasonable approximation of the
>  current instance in Coordinated Universal Time.  An origin server MAY
>  send a Date header field if the response is in the 1xx
>  (Informational) or 5xx (Server Error) class of status codes.  An
>  origin server MUST send a Date header field in all other cases.

Therefore my interpretation of that statement is:  **if and only if the server knows its current date**, does it has to send that header;  therefore clients can presume that if the server doesn't send the `Date` header, the server doesn't know its own current time.

----

My bottom line is this:  I like `fasthttp` because it has a lot of flexibility, never second-guesses the user, and the code is clean enough that I can tailor it to my needs.

Therefore, the only "non-configurable", and "non-mandatory" option with regard to responses was the `Date` header, which I thought ought to have an option just like the `NoServerHeader` and `NoDefaultContentType`.

cipriancraciun:
(I've rebased the patch upon the latest master, and squeezed everything into a single commit.)

kirillDanshin:
@erikdubbelboer yes, being able to turn it off may be a good idea.

Although, I think we need some tests before merge

phuslu:
Hi @all, any progress of this PR ? 

We also need  `NoDefaultDate` option and I indeed to send another PR then I found this....

cipriancraciun:
@phuslu The ball was left in my court to add some tests as described by Erik, however I didn't manage to get around to those.

Could you perhaps add a few tests cases?  (They would be useful both for your own patch, but also for mine.)  (If not perhaps I'll take a stab at these tests next week.)

phuslu:
Hi @cipriancraciun I draft a commit with tests at https://github.com/phuslu/fasthttp/commit/792efdbf37c401e551bc2eeb7ebf9ace2a261d1e

Would please have a look then cherry-pick ?

phuslu:
> I agree that it's a good thing if this is configurable.
Can not agree more. We use fasthttp in a huge traffic environment. turn off `Date` header (also `Content-Type`, `Server`) will significantly save our cost.

phuslu:
No offense, but I sent another & same PR with tests in  https://github.com/valyala/fasthttp/pull/758 。 I'm happy to someone or me close it if this PR merged.

cipriancraciun:
@phuslu NP, any of the pull requests if accepted should be functionally equivalent.

I'm glad if any of them gets merged.  :)

Issue #630, opened 2019-08-12
Does fasthttp.Client has TLSHandshakeTimeout or smth like that internally?


mikefaraponov:

mikefaraponov:
Take a look at tls.DialWithDialer, I think that is should be handled by fasthttp.HostClient

mikefaraponov:
probably smth like that
```go

func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *tls.Config) (net.Conn, error) {
	if dial == nil {
		if dialDualStack {
			dial = DialDualStack
		} else {
			dial = Dial
		}
		addr = addMissingPort(addr, isTLS)
	}
	conn, err := dial(addr)
	if err != nil {
		return nil, err
	}
	if conn == nil {
		panic("BUG: DialFunc returned (nil, nil)")
	}
	if isTLS {
		timeout := 5 * time.Second

		var errChannel chan error

		errChannel = make(chan error, 2)
		time.AfterFunc(timeout, func() {
			errChannel <- timeoutError{}
		})

		colonPos := strings.LastIndex(addr, ":")
		if colonPos == -1 {
			colonPos = len(addr)
		}
		hostname := addr[:colonPos]
		// If no ServerName is set, infer the ServerName
		// from the hostname we're connecting to.
		if config.ServerName == "" {
			// Make a copy to avoid polluting argument or default.
			c := tlsConfig.Clone()
			c.ServerName = hostname
			tlsConfig = c
		}

		conn := tls.Client(rawConn, tlsConfig)

		if timeout == 0 {
			err = conn.Handshake()
		} else {
			go func() {
				errChannel <- conn.Handshake()
			}()

			err = <-errChannel
		}

		if err != nil {
			rawConn.Close()
			return nil, err
		}
	}
	return conn, nil
}
```

mikefaraponov:
Smth like https://github.com/prometheus/prometheus/issues/5394

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/pull/631

PR #631, opened 2019-08-12
Add ability to set timeout for handshake

mikefaraponov:
Relates to https://github.com/valyala/fasthttp/issues/630

### DIFF
--- a/client.go
+++ b/client.go
@@ -1638,8 +1638,10 @@ func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *
 			err = conn.Handshake()
 		} else {
 			go func() {
-				errChannel <- conn.Handshake()
+				err := conn.Handshake()
 				timer.Stop()
+				errChannel <- err
+
 			}()
 			err = <-errChannel
 		}
--- a/client.go
+++ b/client.go
@@ -245,8 +245,6 @@ type Client struct {
 	//     * cONTENT-lenGTH -> Content-Length
 	DisableHeaderNamesNormalizing bool
 
-	TLSHandshakeTimeout time.Duration
-
 	mLock sync.Mutex
 	m     map[string]*HostClient
 	ms    map[string]*HostClient
@@ -424,7 +422,6 @@ func (c *Client) Do(req *Request, resp *Response) error {
 			WriteTimeout:                  c.WriteTimeout,
 			MaxResponseBodySize:           c.MaxResponseBodySize,
 			DisableHeaderNamesNormalizing: c.DisableHeaderNamesNormalizing,
-			TLSHandshakeTimeout:           c.TLSHandshakeTimeout,
 		}
 		m[string(host)] = hc
 		if len(m) == 1 {
@@ -615,8 +612,6 @@ type HostClient struct {
 	//     * content-type -> Content-Type
 	//     * cONTENT-lenGTH -> Content-Length
 	DisableHeaderNamesNormalizing bool
-
-	TLSHandshakeTimeout time.Duration
 
 	clientName  atomic.Value
 	lastUseTime uint32
@@ -1542,7 +1537,7 @@ func (c *HostClient) dialHostHard() (conn net.Conn, err error) {
 	for n > 0 {
 		addr := c.nextAddr()
 		tlsConfig := c.cachedTLSConfig(addr)
-		conn, err = dialAddr(addr, c.Dial, c.DialDualStack, c.IsTLS, tlsConfig, c.TLSHandshakeTimeout)
+		conn, err = dialAddr(addr, c.Dial, c.DialDualStack, c.IsTLS, tlsConfig, c.WriteTimeout)
 		if err == nil {
 			return conn, nil
 		}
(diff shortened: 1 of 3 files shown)


mikefaraponov:
Unfortunately, looks like SetWriteDeadline will not result in connection closed after timeout if server accepts connection and not closing it

mikefaraponov:
Probably, another way to fix it SetReadDeadline earlier than SetWriteDeadline

erikdubbelboer:
Looks like you are right. I mean more a test with code like this:
```go
package main

import (
	"net"
	"time"

	"github.com/valyala/fasthttp"
)

func main() {
	ln, err := net.Listen("tcp4", "127.0.0.1:1337")
	if err != nil {
		panic(err)
	}
	defer ln.Close()

	go func() {
		for {
			_, err := ln.Accept()
			if err != nil {
				panic(err)
			}

			println("connection accepted")
		}
	}()

	res := fasthttp.AcquireResponse()

	req.SetRequestURIBytes([]byte("https://127.0.0.1:1337/"))

	client := fasthttp.Client{
		WriteTimeout: time.Second,
		ReadTimeout:  time.Second,
	}

	if err := client.Do(req, res); err != nil {
		panic(err)
	}

	println("this should happen in 1 second, but it doesn't")
}
```
If you can turn this into a test case that will fail before your pull request but properly timeout after your pull request that would be great.

I think I'm in favor of setting the TLS timeout automatically to the same value as `WriteTimeout`. Since this time is included in trying to write the request.

mikefaraponov:
I think it's a good idea.

client_test.go
```diff
@@ -1735,3 +1735,50 @@ func startEchoServerExt(t *testing.T, network, addr string, isTLS bool) *testEch
 		t:  t,
 	}
 }
+
+func newLocalListener(t testing.TB) net.Listener {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		ln, err = net.Listen("tcp6", "[::1]:0")
```
erikdubbelboer: Is there a specific reason why you added `tcp6` as fallback here? We don't do this anywhere else. It should either be done everywhere or nowhere.

client.go
```diff
@@ -1577,17 +1589,65 @@ func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *
...
 	}
 	if isTLS {
-		conn = tls.Client(conn, tlsConfig)
+		var (
+			errChannel chan error
+			timer      *time.Timer
+		)
+
+		if timeout != 0 {
+			errChannel = make(chan error, 2)
```
erikdubbelboer: Can you use something like `errorChPool` here to keep the number of allocation in `dialAddr` to a minimum.

client.go
```diff
@@ -1577,17 +1589,65 @@ func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *
...
+			timer = time.AfterFunc(timeout, func() {
+				errChannel <- timeoutError{}
+			})
```
erikdubbelboer: Same here, you can use `AcquireTimer()`.

client.go
mikefaraponov: But I am not sure it's possible in the way it is done with chanel of capacity 1, channel with capacity of 2 should be drained first to be reused

client.go
mikefaraponov: Done

client.go
mikefaraponov: Done

client_test.go
mikefaraponov: Done

mikefaraponov: It's a code from golang tls package

client.go
mikefaraponov: time.AfterFunc runs goroutine only in case when timer fired, with AcquireTimer we need to create additional goroutine anyway

client_test.go
```diff
@@ -1735,3 +1735,48 @@ func startEchoServerExt(t *testing.T, network, addr string, isTLS bool) *testEch
...
 	}
 }
+
+func newLocalListener(t testing.TB) net.Listener {
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err)
+	}
+	return ln
+}
```
erikdubbelboer: No need for a separate function for this, just inline it in the test.

client.go
```diff
@@ -1568,7 +1568,44 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
 	return cfg
 }
 
-func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *tls.Config) (net.Conn, error) {
+var ErrTLSHandshakeTimeout = errors.New("tls handshake timed out")
+
+var timeoutErrorChPool sync.Pool
+
+func tlsClientHandshake(rawConn net.Conn, tlsConfig *tls.Config, timeout time.Duration) (net.Conn, error) {
+	tc := AcquireTimer(timeout)
```
erikdubbelboer: By default `WriteTimeout` will be `0` which means this timer will stop immediately.

client.go
```diff
@@ -1568,7 +1568,44 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
...
+		chv = make(chan error)
+	}
+	ch = chv.(chan error)
+	defer timeoutErrorChPool.Put(chv)
+
+	conn := tls.Client(rawConn, tlsConfig)
+
+	go func() {
+		err := conn.Handshake()
+		ReleaseTimer(tc)
```
erikdubbelboer: Releasing the timer here causes a race condition where the timer might be reused while the code below is still listening on it's channel. It needs to be released after the `select {}` below. So best is to just `defer` release it on top.

mikefaraponov: Yes, indeed

client_test.go
mikefaraponov: Done

erikdubbelboer:
Thanks!

PR #635, opened 2019-08-18
Added dial control for reuseport

moredure:
This changes based on https://github.com/libp2p/go-reuseport


mikefaraponov:
These functions allows to use reuseport on dial side of connection (custom dialer for fasthttp, etc.)

examples/client/client.go
```diff
@@ -0,0 +1,28 @@
+// +build go1.11
+
+package main
+
+import (
+	"fmt"
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/reuseport"
```
erikdubbelboer: Please put the above to imports in a separate block like everywhere else ([for example](https://github.com/valyala/fasthttp/blob/master/compress.go)).

Fenny:
Do we still want to continue with this PR? If so, we just have to change the import blocks?

moredure:
@Fenny no need

PR #636, opened 2019-08-18
Added ability to specify local address

mikefaraponov:


tcpdialer.go
```diff
@@ -129,6 +129,8 @@ type TCPDialer struct {
 	// Changes made after the first Dial will not affect anything.
 	Concurrency int
 
+	LocalAddr *net.TCPAddr
```
erikdubbelboer: Can you please add a comment to explain this option.

kirillDanshin: @erikdubbelboer 
this would give an option to chose from which IP we will dial later. When local address is nil, net.DialTCP will choose local address automatically. This option is to choose a public IP, which is useful, for example, on systems with multiple public IPs.

@mikefaraponov please document this option

erikdubbelboer: @kirillDanshin I know what it does 😄 if I didn't I would be a lousy maintainer. I mean he should add a comment as documentation.

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/commit/11e8301d6c52f89fd80eaabe0639eb6851052ad7

PR #638, opened 2019-08-19
Add stub/wrapper for reuseport.Listen on Windows platforms

andrewheberle:
This PR adds a "stub" Listen function for Windows only that always returns (*reuseport.ErrNoReusePort) so the same code can be used on Windows and non-Windows platforms.

So in this case the following code will work on Windows as well as Linux/BSD platforms

```go
// create server
srv := fasthttp.Server{
	Handler:      requestHandler,
	ReadTimeout:  5 * time.Second,
	WriteTimeout: 10 * time.Second,
	IdleTimeout:  120 * time.Second,
}

// try using SO_REUSEPORT
nl, err := reuseport.Listen(addrFamily, listenAddress)
if err != nil {
	if _, noreuseport := err.(*reuseport.ErrNoReusePort); noreuseport {
		// fall back to plain net.Listen
		nl, err = net.Listen(addrFamily, listenAddress)
		if err != nil {
			fmt.Printf("Error: %s\n", err)
			os.Exit(1)
		}
	} else {
		fmt.Printf("Error: %s\n", err)
		os.Exit(1)
	}
}

// run server
if err := srv.Serve(nl); err != nil {
	fmt.Printf("Error: %s\n", err)
}
```


reuseport/reuseport_test.go
```diff
@@ -1,3 +1,5 @@
+// +build linux darwin dragonfly freebsd netbsd openbsd rumprun
```
erikdubbelboer:
```suggestion
// +build !windows
```

reuseport/reuseport_windows.go
```diff
@@ -0,0 +1,27 @@
+// Package reuseport provides TCP net.Listener with SO_REUSEPORT support.
+//
+// SO_REUSEPORT allows linear scaling server performance on multi-CPU servers.
+// See https://www.nginx.com/blog/socket-sharding-nginx-release-1-9-1/ for more details :)
+//
+// The package is based on https://github.com/kavu/go_reuseport .
```
erikdubbelboer: Please remove this whole comment. It should only be present in one file.

andrewheberle: Done

reuseport/reuseport_windows.go
```diff
@@ -0,0 +1,21 @@
+package reuseport
+
+import (
+	"fmt"
+	"net"
+)
+
+// ErrNoReusePort is returned if the OS doesn't support SO_REUSEPORT.
+type ErrNoReusePort struct {
```
erikdubbelboer: Can you maybe put this error and its function into `reuseport/error.go` so we don't have duplicate code. (So you also remove these from `reuseport/reuseport.go`.

andrewheberle: Thats been split out into a seperate file.

andrewheberle:
> Are you 100% sure this is required. This pull request here says it's not needed in Go 1.12 anymore: [valyala/tcplisten#2](https://github.com/valyala/tcplisten/pull/2)

I saw that however I don't know what that comment is based on as the Go 1.12 release notes don't mention any new support/changes in "net" that could make this work now and my Windows Go 1.12 install fails the same way as noted in that PR.

reuseport/reuseport.go
```diff
@@ -9,23 +9,12 @@
 package reuseport
```
erikdubbelboer: Can you change the top line of this file to `// +build !windows` as well please.

reuseport/reuseport_windows_test.go
```diff
@@ -0,0 +1,16 @@
+package reuseport
```
erikdubbelboer: Looks like this file will also need a `// +build windows` comment. Seems like Go only recognizes `_window.go`. See: https://github.com/golang/go/blob/66ff373911a87140319a6550e4e1b2c6043b1329/src/internal/syscall/windows/exec_windows_test.go#L5

reuseport/reuseport.go
andrewheberle: Done

reuseport/reuseport_windows_test.go
andrewheberle: also done !

erikdubbelboer:
Thanks!

PR #640, opened 2019-08-22
Use ipv6-compliant methodology for joining host port

froodian:
Hello! First, thank you for maintaining this great library!

I stumbled recently upon this description of `fmt.Sprintf("%s:%d", addr, port)` as an anti-pattern: https://github.com/golang/go/issues/28308 - it seems to apply here.

### DIFF
--- a/client.go
+++ b/client.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1653,7 +1653,7 @@ func addMissingPort(addr string, isTLS bool) string {
 	if isTLS {
 		port = 443
 	}
-	return fmt.Sprintf("%s:%d", addr, port)
+	return net.JoinHostPort(addr, port)
```
kirillDanshin: net.JoinHostPort(addr, strconv.Itoa(port))

froodian: Sorry, yes, thank you - once again, having run tests this time.

Issue #471, opened 2018-11-18
[question/enhancement] Why do we not `Recover` from `panic` handlers?


cipriancraciun:
Currently a handler that `panic`'s will bring down the whole server, thus impacting other outstanding connections and handlers.

Thus I asked myself why doesn't `fasthttp` protect its connection handler routine with a `Recover`, thus dropping only that connection?  Is it by design?

---

Now (answering and debating with myself) :) I can see both sides of the argument:

* if we don't handle panics, then any small hidden bug would take the whole server down, and if that bug can somehow be easily triggered by the requester, then we end up with a very efficient denial-of-service attack;  (even if we use a process supervision, if the server keeps going down, the supervisor will start delaying the restart;)

* however if we do handle panics and just log them, then a major failure in our code would get unnoticed by the supervisor which will prevent a restart that might (temporarily) solve the issue;

---

As such perhaps a "combined" approach can be taken:
* handle the panics and log them;
* count the panics per client IP, and as in the case of requests limit, apply an enforcement;  (this will cover an attacker that reliably triggers such a condition;)
* count the panics per whole server, and apply an enforcement;  (this will cover a system-wide failure that might be solved by a restart;)

The limits could be either absolute (i.e. this many requests) or percentage (of all served requests).  In case of global percentage limit, there should be a small "burst" that would cover any initial temporary failures caused by "initialization" dependencies or failures.

dgrr:
Probably [this](https://github.com/erikdubbelboer/fasthttp/issues/31#issuecomment-364623249) will be the better option.
> Returning an error instead of a panic would [print the error to the logger](https://github.com/erikdubbelboer/fasthttp/blob/32c72cde80f0c591604f825586d6a4bbbb39d9c5/workerpool.go#L210-L217) but keep trying to reuse the connection which isn't going to work.

erikdubbelboer:
We could add `recover()` but I'm not sure if it's really up to `fasthttp` to handle this case for the user. `net/http` does recover from panics but that's because it has a special mechanism there where a panic aborts the connection. `fasthttp` doesn't have anything like this (at the moment).

Having users add the `recover()` to the top of their handler gives the user the choice of how to log the error and how to handle it instead of having `fasthttp` decide this for the user.

cipriancraciun:
Fair enough, however this behavior (i.e. crashing the entire server on any `panic`) is quite different from what one would expect.  Therefore might I propose a change in the documentation that clearly states this?

See pull request #475.

[...]

PR #646, opened 2019-09-03
Opt in PanicHandler

edoshor:
After recover() band-aids were removed ([commit](https://github.com/valyala/fasthttp/commit/886e5411604884629c566961ea8ed2cec074e4b1)). Users were left with no possibility to act on a panic in a goroutine started inside fasthttp code. They can recover inside their own code. However, that might turn to be repetitive and cumbersome as they have to it on every handler, callback, custom interface implementation (ReadCloser) etc...

The discussion on [this issue](https://github.com/valyala/fasthttp/issues/471) suggests that panic handling is the responsibility of the user. This PR allow the user to provide an opt-in callback which is called on a non nil return value from `recover()`. Users are encouraged to re-panic in their implementation to allow the process to crash completely and not leave fasthttp internals in some unexpected state.

Main use cases are error reporting and proper cleanup of resources upon an unexpected process termination.

**Note:** Currently, not all places creating goroutines are handled. I can add them later on if the general idea is accepted by the maintainers.

### DIFF
--- a/workerpool.go
+++ b/workerpool.go
@@ -213,13 +213,13 @@ func (wp *workerPool) workerFunc(ch *workerChan) {
 
 	if wp.PanicHandler != nil {
 		defer func() {
+			wp.workerDone()
 			if r := recover(); r != nil {
 				if c != nil {
 					c.Close()
 				}
 				wp.PanicHandler(r)
 			}
-			wp.workerDone()
 		}()
 	}
 
--- a/workerpool.go
+++ b/workerpool.go
@@ -22,7 +22,7 @@ type workerPool struct {
 
 	LogAllErrors bool
 
-	RecoverHandler func(r interface{})
+	PanicHandler func(r interface{})
 
 	MaxIdleWorkerDuration time.Duration
 
@@ -211,13 +211,13 @@ func (wp *workerPool) workerDone() {
 func (wp *workerPool) workerFunc(ch *workerChan) {
 	var c net.Conn
 
-	if wp.RecoverHandler != nil {
+	if wp.PanicHandler != nil {
 		defer func() {
 			if r := recover(); r != nil {
 				if c != nil {
 					c.Close()
 				}
-				wp.RecoverHandler(r)
+				wp.PanicHandler(r)
 			}
 			wp.workerDone()
 		}()
@@ -251,7 +251,7 @@ func (wp *workerPool) workerFunc(ch *workerChan) {
 		}
 	}
 
-	if wp.RecoverHandler == nil {
+	if wp.PanicHandler == nil {
 		wp.workerDone()
 	}
 }
--- a/workerpool.go
+++ b/workerpool.go
@@ -22,6 +22,8 @@ type workerPool struct {
 
 	LogAllErrors bool
 
+	RecoverHandler func(r interface{})
+
 	MaxIdleWorkerDuration time.Duration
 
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
I'm still not in favor of merging this. Have you ever had a panic caused by fasthttp crash your process? All panics I have seen came from the user code in the handler. It shouldn't be up to fasthttp to always catch this for you. There is nothing wrong with adding a defer recover at the top of your handler if you want to catch these panics.

What do you think @kirillDanshin?

edoshor:
@erikdubbelboer I truly understand the lack of motivation for merging this. Having the user take responsibility is the correct way.

Our handlers are just like you recommended, i.e. defer recover at the top. However, we had panics inside a ReadCloser Close method (body stream). This method is invoked outside the scope of the handler so we can't recover there. We can recover inside the Close method but it feels cumbersome, repetitive and error prone. Moreover, these ReadClosers may be outside our control *(think some 3rd party package).

This opt-in callback would ease our job controlling the teardown of the process. Implementing one global function once, instead of constantly checking all flows and states beforehand.

erikdubbelboer:
I'm wondering if `Response.SetBodyStream` and `Response.SetBodyStreamWriter` are the only functions that have this issue or if there are other things that I'm missing. (I sometimes which we didn't have such a HUGE api surface to maintain).

We already have [`Server.ErrorHandler`](https://godoc.org/github.com/valyala/fasthttp#Server.ErrorHandler), so I'm wondering if we should just use this instead of adding yet another property to Server. Of course we would then instead have to introduce a new error type that wraps the value returned by `panic()`. So in the end it would expose the same amount of new API to maintain.

workerpool.go
```diff
@@ -200,9 +202,27 @@ func (wp *workerPool) release(ch *workerChan) bool {
 	return true
 }
 
+func (wp *workerPool) workerDone() {
+	wp.lock.Lock()
+	wp.workersCount--
+	wp.lock.Unlock()
+}
+
 func (wp *workerPool) workerFunc(ch *workerChan) {
```
kirillDanshin: I don't really like that this change will silently affect users with PanicHandler == nil as well.

`if wp.PanicHandler != nil` for == nil would at least execute redundant `JZ`, and the `if wp.PanicHandler == nil` below implies something like

```asm
JNZ FUNCRET;
JMP WORKER_DONE;
FUNCRET:
```

and that's all while the user doesn't even use this feature.

while I understand what (*wp).workerDone() is easier to read in the context of this feature, I'm not sure that this feature is really required, or at least that this implementation is optimal for both users who want the PanicHandler and who don't.

Again, I use fasthttp for several years now and I still didn't find a case when fasthttp would panic.

I'd like to hear some more thoughts about this PR.

erikdubbelboer: I also think it's better if we just handle panics from `Response.SetBodyStream` and `Response.SetBodyStreamWriter` and use `s.logger().Printf("...` and return without writing anything more in the response as we don't know what's written already.

For people who want more control it's not that hard to wrap the reader and handle these panics themselves.

This way we don't expose any extra API again.

edoshor: I agree with the above. 

Moreover, https://github.com/valyala/fasthttp/pull/687 is definitely a step in the right direction for us. So closing this one.

edoshor:
Closing as not sure it's the right way to go.

PR #647, opened 2019-09-04
Propagating custom SkipBody value to allow skipping body for non-HEAD requests

enchantner:
As a next step from https://github.com/valyala/fasthttp/pull/532, as in comments @xuecai mentioned an issue with it being overwritten by .Reset() and future check.


erikdubbelboer:
Thanks.

PR #649, opened 2019-09-09
Added option to disable path normalization

dfirsht:


erikdubbelboer:
Can you explain why you need this/why this is useful? Also can you add some tests that show this works?

dfirsht:
Added a test to show path normalization being disabled.

This is helpful for proxy servers that wish to preserve the original path when forwarding requests. With this option, the proxy server will not make any assumptions on how the destination server expects paths to be formatted and instead, can let that server do its own normalization.

http.go
```diff
@@ -49,6 +49,13 @@ type Request struct {
 
 	// To detect scheme changes in redirects
 	schemaUpdate bool
+
+	// Path values are sent as-is without normalization
+	// if this option is set.
+	//
+	// By default path values are normalized, i.e.
+	// extra slashes are removed, special characters are encoded.
+	useOriginalPath bool
```
erikdubbelboer: You added this field to `Request` and you set it in `SetUseOriginalPath` but it is never read?

uri.go
```diff
@@ -47,6 +47,7 @@ type URI struct {
 
 	queryArgs       Args
 	parsedQueryArgs bool
+	useOriginalPath bool
```
erikdubbelboer: Please add support for this field to `URI.CopyTo()` and`URI.Reset()`.

http.go
dfirsht: Removed

uri.go
dfirsht: Added

http.go
```diff
@@ -155,6 +155,15 @@ func (req *Request) SetConnectionClose() {
...
 }
 
+// SetUseOriginalPath sets the useOriginalPath flag.
+// If this flag is set to true, path values are sent as-is without normalization
+//
+// By default path values are normalized, i.e.
+// extra slashes are removed, special characters are encoded.
+func (req *Request) SetUseOriginalPath(useOriginalPath bool) {
+	req.uri.useOriginalPath = useOriginalPath
+}
```
erikdubbelboer: I'm wondering if this function should be called `DisablePathNormalizing`. Just like the `DisableHeaderNamesNormalizing` that some of our APIs have.

I'm also wondering why this is a method of `Request`? Why not of `URI`? So you will have to do `Request.URI().DisablePathNormalizing()`?

dfirsht: Sure, that sounds like a better name. Will change.

It doesn't matter to me whether the method is on `Request` or `URI`, I'll change it to be `URI` only.

dfirsht:
To more closely align with options like `DisableHeaderNamesNormalizing`, I made the `DisablePathNormalizing` bool an exported member. If you prefer though, I can add back the method instead.

erikdubbelboer:
Thanks!

PR #653, opened 2019-09-16
Use table in appendQuotedPath and AppendQuotedArg

ZYunH:
The character comparison is expensive, use table is better. 

A simple benchmark for this change:
> func BenchmarkAppendQuotedArg(b *testing.B) {
	dst := make([]byte, 0, 100)
	src := []byte("foo=@bar&aaa=bbb")
	for i := 0; i < b.N; i++ {
		AppendQuotedArg(dst, src)
	}
}

BenchmarkAppendQuotedArg-12             50000000                24.2 ns/op             0 B/op          0 allocs/op
BenchmarkAppendQuotedArgNew-12            100000000               20.1 ns/op             0 B/op          0 allocs/op

The latter is the new one. And it can be inferred that the longer the input string is, the new one will be much more faster.

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -393,6 +393,80 @@ func s2b(s string) (b []byte) {
 	return b
 }
 
+var quotedArgTable = func() [256]bool {
+	// See http://www.w3.org/TR/html5/forms.html#form-submission-algorithm
+	var a [256]bool
+
+	// '0' ~ '9'
+	for i := int('0'); i <= int('9'); i++ {
+		a[i] = true
+	}
+
+	// 'a' ~ 'z'
+	for i := int('a'); i <= int('z'); i++ {
+		a[i] = true
+	}
+
+	// 'A' ~ 'Z'
+	for i := int('A'); i <= int('Z'); i++ {
+		a[i] = true
+	}
+
+	a[int('*')] = true
+	a[int('-')] = true
+	a[int('.')] = true
+	a[int('_')] = true
+
+	return a
+}()
+
+var quotedPathTable = func() [256]bool {
+	// From the spec: http://tools.ietf.org/html/rfc3986#section-3.3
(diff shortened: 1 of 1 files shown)


ZYunH:
@erikdubbelboer Hi i can't get any useful informations in the failed test, it may some problems in test suites?

erikdubbelboer:
@ZYunH the test failure is unrelated to your pull request.

erikdubbelboer:
The latest commit in master should fix the test issue. So please rebase on master.

ZYunH:
It looks like we don't implement correctly before? the quotedArgTable should be ALPHA / DIGIT / "-" / "." / "_" / "~". A simple test in Go official implementation

> 	u := url.Values{}
	u.Add("test","x.x*")
	print(u.Encode())

We got "test=x.x%2A", but the implementation in fasthttp, it will become "test=x.x*". We should also modify the test code since it is incorrect.

ZYunH:
Mark our previous tests for quotedArg

key -> "f.o,1:2/4"
value -> "~`!@#$%^&*()_-=+\\|/[]{};:'\"<>,./?"

Go implementation
> "f.o%2C1%3A2%2F4=~%60%21%40%23%24%25%5E%26%2A%28%29_-%3D%2B%5C%7C%2F%5B%5D%7B%7D%3B%3A%27%22%3C%3E%2C.%2F%3F"
Fasthttp
> "f.o%2C1%3A2%2F4=%7E%60%21%40%23%24%25%5E%26*%28%29_-%3D%2B%5C%7C%2F%5B%5D%7B%7D%3B%3A%27%22%3C%3E%2C.%2F%3F"

I think we can fix it in this pull request

ZYunH:
pls see the new pull request in https://github.com/valyala/fasthttp/pull/654

PR #657, opened 2019-09-19
Replace hexCharUpeer with upperhex

ZYunH:
In this case, indexed read of the table is faster than calculate the result

(benchmark result)
BenchmarkHexString-8    2000000000               0.65 ns/op
BenchmarkHexCharUpeer-8        500000000                3.77 ns/op

(benchmark codes)
```
var r byte

func BenchmarkHexString(b *testing.B) {
	testSlice := make([]byte, 0, b.N)
	for i := 0; i < b.N; i++ {
		testSlice = append(testSlice, uint8(rand.Uint32()))
	}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		r = "0123456789ABCDEF"[testSlice[i]>> 4]
	}
}

func BenchmarkHexCharUpeer(b *testing.B) {
	testSlice := make([]byte, 0, b.N)
	for i := 0; i < b.N; i++ {
		testSlice = append(testSlice, uint8(rand.Uint32()))
	}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		r = hexCharUpeer(testSlice[i] >> 4)
	}
}
```

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -315,13 +315,6 @@ func int2hexbyte(n int) byte {
 	return 'a' + byte(n) - 10
 }
 
-func hexCharUpper(c byte) byte {
-	if c < 10 {
-		return '0' + c
-	}
-	return c - 10 + 'A'
-}
-
 var hex2intTable = func() []byte {
 	b := make([]byte, 256)
 	for i := 0; i < 256; i++ {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks!

PR #658, opened 2019-09-19
Replace int2hexbyte with lowerhex

ZYunH:
int2hexbyte can be replaced with lowerhex, and i'm wandering maybe we could use only one HEX format, since the constant upperhex is already exists.( If use upperhex only, we need to modify our test instead)

Also replace 15 with 0xf in take the 4 LSBs in byte, maybe 0xf looks more readable in this case?

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -296,7 +296,7 @@ func writeHexInt(w *bufio.Writer, n int) error {
 	buf := v.([]byte)
 	i := len(buf) - 1
 	for {
-		buf[i] = int2hexbyte(n & 0xf)
+		buf[i] = upperhex[n&0xf]
 		n >>= 4
 		if n == 0 {
 			break
@@ -306,13 +306,6 @@ func writeHexInt(w *bufio.Writer, n int) error {
 	_, err := w.Write(buf[i:])
 	hexIntBufPool.Put(v)
 	return err
-}
-
-func int2hexbyte(n int) byte {
-	if n < 10 {
-		return '0' + byte(n)
-	}
-	return 'a' + byte(n) - 10
 }
 
 var hex2intTable = func() []byte {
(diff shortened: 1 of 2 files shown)


bytesconv.go
```diff
@@ -296,7 +296,7 @@ func writeHexInt(w *bufio.Writer, n int) error {
 	buf := v.([]byte)
 	i := len(buf) - 1
 	for {
-		buf[i] = int2hexbyte(n & 0xf)
+		buf[i] = lowerhex[n&0xf]
```
kirillDanshin: I'm actually concerned about this change. do we actually tested all cases, and will it actually work assuming a wide range of use cases?

ZYunH: As i can see, int2hexbyte is actually used inside our project, and only in this line. It's used to convert a uint8 (<16, since we & 0xf) to a hex literal, so in this case it is equal to a string which contain all these hex string. A similar PR we can see in #657

kirillDanshin: @ZYunH yeah, I understand what we're doing here, I just trying to understand if `n&0xf` can exceed `lowerhex` bounds. previously, there was just a function that didn't use any strings indexes

ZYunH: I'm just not sure in which situation n&0xf can exceed lowerhex if n is always a uint8 and less than 16 ( In hex and byte conversion),  or should we add some comments to it ?

ZYunH: May be we can take the practice of official implementation in https://github.com/golang/go/blob/c226f6432d465ad9e2d21962353ba86834a2afcb/src/strconv/quote.go#L15 ?

ZYunH: I'm not a English native speaker, excuse me if there is something offend you : )

kirillDanshin: I've tried to somehow break this thing and it's just works, so LGTM :)

> I'm not a English native speaker, excuse me if there is something offend you : )

I'm also not a native speaker, and I don't know how you would offend me in those messages :)

PR #663, opened 2019-09-26
Use 'go generate' in bytesconv.go

ZYunH:
There are lots of tables in bytesconv.go, maybe we could use 'go generate' to generate these tables in a new file, called 'bytesconv_table.go'. In this way we can reduce init-time and also speed up our conversion, since the table is a constant which defined in compile-time.(feel free to modify my words if there is any syntax error)


bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+	"io/ioutil"
+	"log"
+)
+
+const (
+	toLower = 'a' - 'A'
+)
+
+func main() {
+	var hex2intTable = func() [256]byte {
```
erikdubbelboer: Can you change these to `:=` declarations like `hex2intTable := ...`. I prefer those, thanks!

bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+	w := new(bytes.Buffer)
+	w.WriteString(pre)
+	for _, k := range tables {
+		v := m[k]
+		tmp := make([]byte, 256)
+		for i := 0; i < 256; i++ {
+			tmp[i] = v[i]
+		}
+		fmt.Fprintf(w, table, k, string(tmp))
+	}
```
erikdubbelboer: I would just replace the whole above code with:
```go
fmt.Fprintf(w, "const hex2intTable = %q\n", hex2intTable)
fmt.Fprintf(w, "const toLowerTable = %q\n", toLowerTable)
fmt.Fprintf(w, "const toUpperTable = %q\n", toUpperTable)
fmt.Fprintf(w, "const quotedArgShouldEscapeTable = %q\n", quotedArgShouldEscapeTable)
fmt.Fprintf(w, "const quotedPathShouldEscapeTable = %q\n", quotedPathShouldEscapeTable)
```

bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+
+	out, err := format.Source(w.Bytes())
+	if err != nil {
+		log.Fatal(err)
+	}
```
erikdubbelboer: This step isn't really needed. The generated source is so simple and easy to get right that formatting it shouldn't change it anymore.

bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+
+	if err := ioutil.WriteFile("bytesconv_table.go", out, 0660); err != nil {
+		log.Fatal(err)
+	}
+
```
erikdubbelboer: Remove empty line.

bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+const pre = `// Code generated by go run bytesconv_table_gen.go; DO NOT EDIT.
+// See bytesconv_table_gen.go for more information about these tables.
+
+package fasthttp
+
+`
```
erikdubbelboer: Can you move this constant to the top of the file, I prefer to have them there. Thanks.

bytesconv_table_gen.go
```diff
@@ -0,0 +1,149 @@
...
+
+const table = `const %s = %#v
+`
```
erikdubbelboer: This one can then be removed.

ZYunH:
Thanks for review! all done

erikdubbelboer:
Thanks!

PR #666, opened 2019-09-30
Refactoring to reuse response status codes

Vendin:
Refactoring to reuse response status codes.
Status codes are needed, where we ourselves implement a cycle for all redirects.
To understand where the last redirect is.

### DIFF
--- a/client.go
+++ b/client.go
@@ -861,7 +861,7 @@ func getRedirectURL(baseURL string, location []byte) string {
 
 // StatusCodeIsRedirect checks the response code.
 //
-// If this is a redirect code then it returns true, otherwise false.
+// StatusCodeIsRedirect returns true if the status code indicates a redirect.
 func StatusCodeIsRedirect(statusCode int) bool {
 	return statusCode == StatusMovedPermanently ||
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -864,6 +859,14 @@ func getRedirectURL(baseURL string, location []byte) string {
 	return redirectURL
 }
 
+func IsFinishStatusCode(statusCode int) bool {
```
erikdubbelboer: Can you add some documentation to this function. Also a better name might be `StatusCodeIsRedirect`?

Vendin: Done. Check please.

client.go
```diff
@@ -864,6 +859,17 @@ func getRedirectURL(baseURL string, location []byte) string {
 	return redirectURL
 }
 
+// StatusCodeIsRedirect checks the response code.
+//
+// If this is a redirect code then it returns true, otherwise false.
```
erikdubbelboer: Sorry but could you please change it to:
```
StatusCodeIsRedirect returns true if the status code indicates a redirect.
```

erikdubbelboer:
Thanks!

PR #668, opened 2019-10-08
support wait conn on busy

king526:
support wait conn on busy ,it is aim at avoid make too much conns which i think will make performance problem. thanks  for  reviewing  and i had make same change.
```
func (c *HostClient) releaseConn(cc *clientConn) {
	c.connsLock.Lock()
	c.conns = append(c.conns, cc)
	c.connsLock.Unlock()
	select {
	case c.connNotify <- struct{}{}:
	default:
	}
}
```
  i do not think  lost notify will cause problem. waiting is no guarantee on continuous heavy requests.it is crop traffic on a time bucket (timeout) not for make requests ordered .

### DIFF
--- a/client.go
+++ b/client.go
@@ -992,6 +992,7 @@ func clientDoDeadline(req *Request, resp *Response, deadline time.Time, c client
 	reqCopy := AcquireRequest()
 	req.copyToSkipBody(reqCopy)
 	swapRequestBody(req, reqCopy)
+	reqCopy.deadline = deadline
 	respCopy := AcquireResponse()
 	if resp != nil {
 		// Not calling resp.copyToSkipBody(respCopy) here to avoid
@@ -1076,7 +1077,12 @@ func (c *HostClient) Do(req *Request, resp *Response) error {
 		maxAttempts = DefaultMaxIdemponentCallAttempts
 	}
 	attempts := 0
-
+	if req.deadline.IsZero() {
+		timeout := c.ReadTimeout + c.WriteTimeout
+		if timeout != 0 {
+			req.deadline = time.Now().Add(timeout)
+		}
+	}
 	atomic.AddInt32(&c.pendingRequests, 1)
 	for {
 		retry, err = c.do(req, resp)
@@ -1166,7 +1172,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		req.SetConnectionClose()
 	}
 
-	cc, err := c.acquireConn()
+	cc, err := c.acquireConn(req.deadline)
 	if err != nil {
 		return false, err
 	}
@@ -1275,7 +1281,7 @@ func (c *HostClient) SetMaxConns(newMaxConns int) {
 	c.connsLock.Unlock()
 }
 
-func (c *HostClient) acquireConn() (*clientConn, error) {
+func (c *HostClient) acquireConn(deadline time.Time) (*clientConn, error) {
 	var cc *clientConn
 	createConn := false
 	startCleaner := false
@@ -1311,11 +1317,19 @@ func (c *HostClient) acquireConn() (*clientConn, error) {
 		return cc, nil
 	}
 	if !createConn {
-		if c.PreferWaitConn {
-			<-c.connNotify
-			return c.acquireConn()
+		if !c.PreferWaitConn {
+			return nil, ErrNoFreeConns
 		}
-		return nil, ErrNoFreeConns
+		timeout := time.Hour
+		if !deadline.IsZero() {
+			timeout = -time.Since(deadline)
+		}
+		select {
+		case <-c.connNotify:
+			return c.acquireConn(deadline)
+		case <-time.After(timeout):
+			return nil, ErrTimeout
+		}
 	}
 
 	if startCleaner {
--- a/client.go
+++ b/client.go
@@ -89,7 +89,7 @@ func DoTimeout(req *Request, resp *Response, timeout time.Duration) error {
 // the given deadline.
 //
 // ErrNoFreeConns is returned if all DefaultMaxConnsPerHost connections
-// to the requested host are busy.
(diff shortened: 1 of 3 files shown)


king526:
on my first version.i just wait util get a new conns which no use time.After. i think it's efficient but not sure if it is rational.

king526:
i think that ci has bug?

client.go
```diff
@@ -1069,7 +1077,12 @@ func (c *HostClient) Do(req *Request, resp *Response) error {
 		maxAttempts = DefaultMaxIdemponentCallAttempts
 	}
 	attempts := 0
-
+	if req.deadline.IsZero() {
+		timeout := c.ReadTimeout + c.WriteTimeout
+		if timeout != 0 {
+			req.deadline = time.Now().Add(timeout)
```
erikdubbelboer: I really don't like having the `deadline` field added to `Request` just for this feature. What is the reason this can't be calculated in `doNonNilReqResp` where it is needed? I would argue that each time we retry because of `MaxIdemponentCallAttempts` should have a new timeout/deadline.

client.go
```diff
@@ -1301,7 +1317,19 @@ func (c *HostClient) acquireConn() (*clientConn, error) {
 		return cc, nil
 	}
 	if !createConn {
-		return nil, ErrNoFreeConns
+		if !c.PreferWaitConn {
+			return nil, ErrNoFreeConns
+		}
+		timeout := time.Hour
+		if !deadline.IsZero() {
+			timeout = -time.Since(deadline)
```
erikdubbelboer: Lets not use an arbitrary timeout here of an hour. If `deadline` is zero it should just wait forever I guess. So in this if should be:
```go
<-c.connNotify:
return c.acquireConn(deadline)
```
and the other select in the `else`.

king526: thanks for you advise. i have ajust the implementation. when call clientDoDeadline , limit get conn with the deadline to avoid send request to server stil on getConn timeout. 
it's almost like my first version which no use time ticker.there is a little improve is that check timeout on every wakeup, to avoid make too much goroutines blocking.

king526:
new commit is submited

erikdubbelboer:
Sorry I'm currently working on some changes that will make our travis-ci build more stable. Just ignore it for now. I'll have time this weekend to look at your pull request!

client.go
```diff
@@ -1301,7 +1323,11 @@ func (c *HostClient) acquireConn() (*clientConn, error) {
 		return cc, nil
 	}
 	if !createConn {
-		return nil, ErrNoFreeConns
+		if !c.PreferWaitConn {
+			return nil, ErrNoFreeConns
+		}
+		<-c.connNotify
```
erikdubbelboer: I don't really like this implementation. This way if a conn doesn't free up before the deadline we still have to wait for one to free up before we continue here. So we might be waiting much longer than our timeout.

http.go
```diff
@@ -49,6 +50,7 @@ type Request struct {
 
 	// To detect scheme changes in redirects
 	schemaUpdate bool
+	deadline     time.Time
```
erikdubbelboer: I would prefer that instead of adding this field to ʼRequestʼ you add an extra parameter to ʼdoNonNilReqRespʼ etc.

client_test.go
```diff
@@ -1813,3 +1813,112 @@ func TestClientTLSHandshakeTimeout(t *testing.T) {
...
+	addr := "http://" + listener.Addr().String()
+	var (
+		serverGot int32
+		timeout   int32
+		wg        sync.WaitGroup
+	)
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			atomic.AddInt32(&serverGot, 1)
+			time.Sleep(time.Second * 3)
```
erikdubbelboer: Nice test. But can it also work with a bit shorter times? Maybe tens of milliseconds instead of seconds. That way our test suite doesn't take that long.

client_test.go
```diff
@@ -1813,3 +1813,112 @@ func TestClientTLSHandshakeTimeout(t *testing.T) {
...
+	time.Sleep(time.Millisecond)
+	waitClient := &Client{
+		PreferWaitConn:  true,
+		MaxConnsPerHost: 1,
+	}
+	for i := 0; i < 5; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for j := 0; j < 1; j++ {
```
erikdubbelboer: This loop does only one iteration so it can be removed.

client_test.go
```diff
@@ -1813,3 +1813,112 @@ func TestClientTLSHandshakeTimeout(t *testing.T) {
...
+				req := &Request{}
+				req.SetRequestURI(addr)
+				err2 := waitClient.DoTimeout(req, nil, time.Second*10)
```
erikdubbelboer:
```suggestion
				err := waitClient.DoTimeout(req, nil, time.Second*10)
```

client_test.go
king526: ci should this bag

client.go
king526: i will reconsider the implementation, there can be a better choose. since timer is inefficiency, i made this choose because it will not blocking api ,since we certainly use DoTimeout rather than Do. in our ad serving,5120 conns  is set but sometime it still get a lot of no free conns error at monment.i do think it's necessary to take a action.

client_test.go
erikdubbelboer: I don't understand? People making changes to fasthttp will run `go test` locally and having to wait an extra 3 seconds for this is very annoying. The test should work just as fine if this was `time.Millisecond * 30` and below `time.Millisecond * 100`.

client.go
erikdubbelboer: You could just `select` between `c.connNotify` and a timer that signals at `deadline` (using `AcquireTimer`).

erikdubbelboer:
Superseded by https://github.com/valyala/fasthttp/pull/764

PR #672, opened 2019-10-10
Allow client to disable path normalization

froodian:
See the test for context - basically I have a URL like http://google.com/encoded/Y%2BY%2FY%3D/stuff, but when I use fasthttp, due to path normalization the actual request that gets made is to http://google.com/encoded/Y+Y/Y=/stuff.  On some servers (e.g. fasthttp's server), these URLs are semantically equivalent, but on some servers (e.g. those google urls) they are not. If you click through those URLs you can see the difference in the 404 page that google returns. https://stackoverflow.com/questions/1957115/is-a-slash-equivalent-to-an-encoded-slash-2f-in-the-path-portion-of-a has some more context on the server-side situation as well.

This wires https://github.com/valyala/fasthttp/pull/649 through the client layer to solve this issue.

### DIFF
--- a/client.go
+++ b/client.go
@@ -246,6 +246,15 @@ type Client struct {
 	//     * cONTENT-lenGTH -> Content-Length
 	DisableHeaderNamesNormalizing bool
 
+	// Path values are sent as-is without normalization
+	//
+	// Disabled path normalization may be useful for proxying incoming requests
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #674, opened 2019-10-13
fix 664

### LINKED ISSUE #664: may be data race in pipeConn.readNextByteBuffer
https://github.com/valyala/fasthttp/blob/b1ca43febc44fadacbe003d3760dfa3e8395bf7c/fasthttputil/pipeconns.go#L169

There is three references of c.readDeadlineCh , but only two of them are protected by the locker.
At least this will make the code reader ask whether there is a data race question.
So fix it or a comment why not fix it should be added.

rogercarter1:
fix #664

### DIFF
--- a/fasthttputil/pipeconns.go
+++ b/fasthttputil/pipeconns.go
@@ -166,7 +166,9 @@ func (c *pipeConn) readNextByteBuffer(mayBlock bool) error {
 		select {
 		case c.b = <-c.rCh:
 		case <-readDeadlineCh:
+			c.readDeadlineChLock.Lock()
 			c.readDeadlineCh = closedDeadlineCh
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Sorry for my slow reply. I was still thinking about what to do with this. The issue is that this still doesn't make pipeConn safe for concurrent use at all. There are many places that still read and write members without a mutex. So adding this one lock/unlock doesn't really do much.

I'm wondering if we should try to make it completely safe by adding a lot more lock/unlock calls. Or if we should just keep things the way they are and document that it isn't safe.

erikdubbelboer:
Thanks. I'm also adding a commit to state that pipeConn is not safe for concurrent use.

PR #675, opened 2019-10-15
Request head Host been overwrite when set a custom URI Host value

ipandtcp:
I use req.Header.Read(bufio.NewReader(buf)) to get request, and i need sent this request to a special host, so i call req.SetHost(SPECIAL_ADDR), but then, request head Host value has been change too, i think maybe there is a typo.

### DIFF
--- a/http.go
+++ b/http.go
@@ -1141,7 +1141,7 @@ func (req *Request) onlyMultipartForm() bool {
 //
 // See also WriteTo.
 func (req *Request) Write(w *bufio.Writer) error {
-	if len(req.Header.Host()) == 0 || req.parsedURI {
+	if len(req.Header.Host()) == 0 && req.parsedURI {
 		uri := req.URI()
 		host := uri.Host()
 		if len(host) == 0 {


erikdubbelboer:
I'm not sure I understand. Can you add a test case that fails before your change and passes after? Thanks.

ipandtcp:
### Hi there, Please pardon me express not clear.
#### The origin request header:
```
GET /ping 1.1
Host: 10.20.136.225
User-Agent: curl/7.47.0
Accept: */*
```
I'm trying proxy that request to destination "127.0.0.1:123" by HTTP, so I write some code like this:
```
buf := bytes.NewBuffer(THE_ORIGIN_REQ_BYTES)
req := &fasthttp.Request{}
 err = req.Header.Read(bufio.NewReader(buf))

// I just want set url host, change the destination
// And don't want change header's Host
req.SetHost("127.0.0.1:123")
res := &fasthttp.Response{}
err = fasthttp.DoTimeout(req, res, TIMEOUT)
```

#### The destination receives a request like that:
```
GET /ping 1.1
Host: 127.0.0.1:123
User-Agent: curl/7.47.0
Accept: */*
```

erikdubbelboer:
You can use `req.SetRequestURI` for that (see: https://play.golang.org/p/PFlH-acYdG1) instead of making a pull request that breaks a whole bunch of other cases as the tests show you.

ipandtcp:
Erik, I'm afraid to tell you, the play you‘re provide did not work, the request still send to 10.20.136.225:80, and got response code 400.

erikdubbelboer:
You are right, I'm afraid that this is impossible to do with `Client` at the moment.
I suggest you use [`HostClient`](https://godoc.org/github.com/valyala/fasthttp#HostClient) instead where you can specify the address it should connect to independently from the request.

ipandtcp:
No, I'm use defaultClient from [fasthttp.DoTimeout](https://godoc.org/github.com/valyala/fasthttp#DoTimeout).
The golang net/http package can do that by setting ` req.URL.Host`, but you known, they are f**king slowly.
And fasthttp can do that too by merging this commit, we're product using this commit code and work will, But I don't know why the codes make breaks of the whole bunch of other cases as the tests. 
Maybe I commit not really fit for all situations so the Test was failed.
But fasthttp how to deal with some situation as We need, I see #676 have some problem.

erikdubbelboer:
One important thing that fails with this pull request is:
```
var req Request
req.Header.SetRequestURI("http://google.com/foo/bar?baz=aaa")
```
This completely fails to work.

Another thing is that http to https redirects seem to fail.

I suggest you change this pull request to add a method that does the same as `SetRequestURI` but keeps `parsedURI` as `true`. That way you can change the Host headers and URI without triggering it being overwritten. Or another option would be to add an option to Request to disable this automatic updating of the headers. Any ways it needs to be something backwards compatible so it doesn't break fasthttp for other users.

erikdubbelboer:
Or you can take the code from this pull request https://github.com/valyala/fasthttp/pull/676 and update it with a test that makes sure it doesn't change/break again in the future.

PR #676, opened 2019-10-16
fix: cannot use custom host in header

hackbeex:
fix: cannot use custom host in header, usually used in proxy,
the same as "proxy_set_header Host $host" in nginx
see: https://tools.ietf.org/html/rfc2616#page-128


erikdubbelboer:
Can you please add a test that fails before your change and passes after. Thanks.

Fenny:
@hackbeex, could you add a test case to this PR so we can close/merge it?

hackbeex:
I probably don't need this commit, I can do something similar with 'header X-Forwarded-Host'. 
So I close it.

PR #677, opened 2019-10-17
fix: when multipartform  no data,FormValue do not need panic

alexDango:
no need panic.

### DIFF
--- a/http.go
+++ b/http.go
@@ -830,7 +830,7 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 	// in multipart/form-data requests.
 
 	if size <= 0 {
-		panic(fmt.Sprintf("BUG: form size must be greater than 0. Given %d", size))
+		return nil, fmt.Errorf("form size must be greater than 0. Given %d", size)
 	}
 	lr := io.LimitReader(r, int64(size))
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
I don't think this can ever happen? Can you add a test that panics before your change? Thanks.

alexDango:
sorry, I made a wrong description
what happened was :

func TestClientPostNothingInMultipartFormValue(t *testing.T) {
	t.Parallel()

	realPanic := ""
	s := &Server{
			defer func() {
				if r := recover(); r != nil{
					realPanic = fmt.Sprintf("%v", r)
				}
			}()
			ctx.FormValue("a")
		},
	}
	go s.Serve(ln)

	c := &Client{
			return ln.Dial()
		},
	}
	req := AcquireRequest()
	res := AcquireResponse()

	req.SetRequestURI("http://example.com")
	//this is use postman to start a request, use form-data
	req.Header.SetContentType("multipart/form-data;boundary=--------------------------530872294864144413907775")
	err := c.Do(req, res)
	if err != nil {
		t.Fatal(err)
	}
	if realPanic != "BUG: form size must be greater than 0. Given 0" {
		t.Fatal("realPanic is: ",realPanic)
	}
}

erikdubbelboer:
Thanks.

PR #678, opened 2019-10-21
fasthttputil: add errInmemoryListenerClosed

ZYunH:
There are three identical errors in inmemory_listener.go, maybe we could use errInmemoryListenerClosed to replace them.

### DIFF
--- a/fasthttputil/inmemory_listener.go
+++ b/fasthttputil/inmemory_listener.go
@@ -6,7 +6,7 @@ import (
 	"sync"
 )
 
-var errInmemoryListenerClosed = errors.New("InmemoryListener is already closed: use of closed network connection")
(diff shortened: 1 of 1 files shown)


fasthttputil/inmemory_listener.go
```diff
@@ -1,11 +1,13 @@
 package fasthttputil
 
 import (
-	"fmt"
+	"errors"
 	"net"
 	"sync"
 )
 
+var errInmemoryListenerClosed = errors.New("InmemoryListener is already closed: use of closed network connection")
```
erikdubbelboer:
```suggestion
var ErrInmemoryListenerClosed = errors.New("InmemoryListener is already closed: use of closed network connection")
```
Lets expose the type so people can check for it.

ZYunH: Sure, done!

erikdubbelboer:
Thanks!

Issue #660, opened 2019-09-24
4xx and Unexpected EOF


mwpolcik:
Hey everyone, I started to use fasthttp in one of my projects and I am seeing quite often this error:
```
error when serving connection "172.17.0.2:80"<->"10.192.10.31:48432": unexpected EOF
```
To give you more context:
1. Application is sitting behind AWS Application load balancer with idle timeout set to 60s. 
2. App is sometimes returning 4xx codes when `unexpected EOF` occurs.
3. Server config:
```
func getServer(handlerFunc func(ctx *fasthttp.RequestCtx)) *fasthttp.Server {
	return &fasthttp.Server{
		Handler:            cors(handlerFunc),
		ReadTimeout:        10 * time.Second,
		WriteTimeout:       10 * time.Second,
		IdleTimeout:        15 * time.Second,
		TCPKeepalive:       true,
		TCPKeepalivePeriod: 30 * time.Second,
		LogAllErrors:       false,
		DisableKeepalive:   false,
		Name:               "my-server",
	}
}
```
4. Handler logic on each endpoint is quite similar:
```
type processArgs struct {
	Body          []byte
	UserAgent     []byte
	XForwardedFor []byte
}

func newProcessArgs(ctx *fasthttp.RequestCtx) *processArgs {
	args := &processArgs{}
	args.Body = append(args.Body, ctx.PostBody()...)
	args.UserAgent = append(args.UserAgent, ctx.UserAgent()...)
	args.XForwardedFor = append(args.XForwardedFor, ctx.Request.Header.Peek(fasthttp.HeaderXForwardedFor)...)
	return args
}

// Handler handles session request
func (h Handler) Handler(ctx *fasthttp.RequestCtx) {
	args := newProcessArgs(ctx)
	go h.process(args)
}
```
I am reading what I need from request, start go routine to process data and return right away.

I was trying different configurations for server, but I am still getting `unexpected EOF` and I am not sure what I am doing wrong here. Any help appreciated.

dgrr:
The handler is leaving just after you start the goroutine. I don't know what do you mean with `return right away`, but if your `h.process` call expects to send data to the user you must wait for the goroutine to finish. Anyway, that should not cause the `unexpected EOF` error. The error should be caused because the request is malformed or something like that.

Maybe you can try by adding a proxy in between AWS and fasthttp to log the requests.

mwpolcik:
[...]

PR #682, opened 2019-10-27
Requests with incomplete bodies no longer cause log noise

kevburnsjr:
**Steps to Reproduce**

1. Run [hello world example](https://github.com/valyala/fasthttp/blob/master/examples/helloworldserver/helloworldserver.go) 
2. Send request with timeout where the value of the `content-length` request header exceeds length of the request body.
```
curl 'http://localhost:8080/' -X "POST" -H 'content-length: 5' -d "123" --max-time 1
```

**Actual Behavior:**

- Server logs unexpected EOF.

```
2019/10/26 18:11:57 error when serving connection "127.0.0.1:8080"<->"127.0.0.1:58492": unexpected EOF
```

**Expected Behavior:**

- Server logs no error (matching behavior of net/http).

---

See #660

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -3147,7 +3147,7 @@ func TestMaxWriteTimeoutPerRequest(t *testing.T) {
 	}
 }
 
-func TestIncompleteBody(t *testing.T) {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #683, opened 2019-10-27
Update "TechEmpower Benchmark" from round 12 to round 18

panjf2000:
As title.


savsgio:
I think so it's better to update with last round (18)

panjf2000:
Make sense and sorry I miss round 18, now it is the latest round.

panjf2000:
@erikdubbelboer @kirillDanshin 
PTAL

erikdubbelboer:
Thanks.

PR #687, opened 2019-11-03
Recover from panic in body write

Bobochka:
Possibly a successor of https://github.com/valyala/fasthttp/pull/646.
Went as per @erikdubbelboer suggestion (if I understood it correctly). If the general idea is acceptable, will add tests.
Thanks.


server.go
```diff
@@ -2219,6 +2219,11 @@ func writeResponse(ctx *RequestCtx, w *bufio.Writer) error {
 		panic("BUG: cannot write timed out response")
 	}
 	err := ctx.Response.Write(w)
+	if err != nil {
+		if panicErr, ok := err.(*ErrBodyStreamWritePanic); ok {
+			ctx.logger.Printf("panic while writing response: %s", panicErr.Error())
```
erikdubbelboer: I don't think its necessary to log this here. The error will go all the way up to [workerpool.go#L230](https://github.com/valyala/fasthttp/blob/87c4546c7b1268a1c29c46e8f54def6f0e7596eb/workerpool.go#L230) where it will be logged anyways.

Bobochka:
@erikdubbelboer done.

Unfortunately this doesn't cover `SetBodyStreamWriter` as writer invocation happens in another goroutine.

erikdubbelboer:
I think we can catch the `SetBodyStreamWriter` as well if we do a similar trick as [`io.PipeWriter.CloseWithError()`](https://golang.org/pkg/io/#PipeWriter.CloseWithError). But this will be a bit complicated as [`fasthttputil.NewPipeConns()`](https://github.com/valyala/fasthttp/blob/e5f51c11919d4f66400334047b897ef0a94c6f3c/stream.go#L29) deals with `net.Conn` and not some interface we can just add `CloseWithError()` to.

So lets save that for another pull request.

Issue #123, opened 2016-06-27
Add support for multi-line header values


valyala:
Currently `fasthttp` doesn't support multi-line header values. See https://github.com/kataras/iris/issues/190 for details.

Probably, multi-line header values must be supported or properly handled by `fasthttp`.

erikdubbelboer:
I don't think we should change the `Peek` signature. What we can do is add another method called `PeekMulti(key string) [][]byte`. One issue is that it will always result in a heap allocation for the `[][]byte`. We could avoid that by making it `PeekMulti(key string, reuse [][]byte) [][]byte` where `reuse` gets reused for the return value.

kirillDanshin:
Of course, we will not change existing API. +1 on *Multi, if we'll find anyone that still needs that feature 😀

Fenny:
I think this issue can be closed https://github.com/valyala/fasthttp/pull/688

PR #688, opened 2019-11-04
add extra checking of header buffer, to support multi line header value

tedli:
#123

### DIFF
--- a/header.go
+++ b/header.go
@@ -2022,9 +2022,24 @@ type headerScanner struct {
 	hLen int
 
 	disableNormalizing bool
+
+	// by checking whether the next line contains a colon or not to tell
+	// it's a header entry or a multi line value of current header entry.
+	// the side effect of this operation is that we know the index of the
+	// next colon and new line, so this can be used during next iteration,
+	// instead of find them again.
+	nextColon   int
+	nextNewLine int
+
+	initialized bool
 }
 
 func (s *headerScanner) next() bool {
+	if !s.initialized {
+		s.nextColon = -1
+		s.nextNewLine = -1
+		s.initialized = true
+	}
 	bLen := len(s.b)
 	if bLen >= 2 && s.b[0] == '\r' && s.b[1] == '\n' {
 		s.b = s.b[2:]
@@ -2036,7 +2051,13 @@ func (s *headerScanner) next() bool {
 		s.hLen++
 		return false
 	}
-	n := bytes.IndexByte(s.b, ':')
+	var n int
+	if s.nextColon >= 0 {
+		n = s.nextColon
+		s.nextColon = -1
+	} else {
+		n = bytes.IndexByte(s.b, ':')
+	}
 	if n < 0 {
 		s.err = errNeedMore
 		return false
@@ -2046,14 +2067,48 @@ func (s *headerScanner) next() bool {
 	n++
 	for len(s.b) > n && s.b[n] == ' ' {
 		n++
+		// the newline index is a relative index, and lines below trimed `s.b` by `n`,
+		// so the relative newline index also shifted forward. it's safe to decrease
+		// to a minus value, it means it's invalid, and will find the newline again.
+		s.nextNewLine--
 	}
 	s.hLen += n
 	s.b = s.b[n:]
-	n = bytes.IndexByte(s.b, '\n')
+	if s.nextNewLine >= 0 {
+		n = s.nextNewLine
+		s.nextNewLine = -1
+	} else {
+		n = bytes.IndexByte(s.b, '\n')
+	}
 	if n < 0 {
 		s.err = errNeedMore
 		return false
 	}
+	isMultiLineValue := false
+	for {
+		if n+1 >= len(s.b) {
+			break
+		}
+		d := bytes.IndexByte(s.b[n+1:], '\n')
+		if d <= 0 {
+			break
+		} else if d == 1 && s.b[n+1] == '\r' {
+			break
+		}
+		e := n + d + 1
+		if c := bytes.IndexByte(s.b[n+1:e], ':'); c >= 0 {
+			s.nextColon = c
+			s.nextNewLine = d - c - 1
+			break
+		}
+		isMultiLineValue = true
+		n = e
+	}
+	if n >= len(s.b) {
+		s.err = errNeedMore
+		return false
+	}
+	oldB := s.b
 	s.value = s.b[:n]
 	s.hLen += n + 1
 	s.b = s.b[n+1:]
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thank you for the pull request.
I'm afraid there is one issue. For multi line headers the `\r\n` that splits up the value should be removed.
See: https://play.golang.org/p/Kpmruh6fghm which is your test but then using `net/http`.
Can you please fix this?

tedli:
@erikdubbelboer 
Hi,
I'm glad to working on this. But since we need to modify the original value, it's means we have to make a copy of a range of the original buffer, which will introduce some overhead.
If you have some ideas or thought about this, please share with me.

header.go
```diff
@@ -2133,6 +2178,35 @@ func getHeaderKeyBytes(kv *argsKV, key string, disableNormalizing bool) []byte {
 	return kv.key
 }
 
+func normalizeHeaderValue(buffer []byte) (value []byte) {
+	length := len(buffer)
+	if length <= 0 {
+		value = []byte{}
+		return
+	}
+	value = make([]byte, length)
```
erikdubbelboer: This function now allocates in the hot path which we really can't have.

I think the best way to fix this is to rewrite the input (`headerScanner.b`) to remove the `\r\n` within the header value. This means future parses of the header buffer will still work. But this does require some modifications to your code. Instead of calling `normalizeHeaderValue` you will need to move everything within `headerScanner.b` back by 2 characters instead.

You can add this test to `allocation_test.go` to make sure it works:
```go
func TestAllocationHeaderScanner(t *testing.T) {
  t.Parallel()

  rawHeaders := []byte("HTTP/1.1 200 OK\r\n" +
    "Date: Mon, 04 Nov 2019 03:08:05 GMT\r\n" +
    "Transfer-Encoding: chunked\r\n" +
    "Content-Type: multipart/related; type=\"application/xop+xml\";\r\n" +
    " boundary=\"----=_Part_16_1220614102.1572836885552\"\r\n" +
    "\r\n" +
    "body\r\n")

  n := testing.AllocsPerRun(100, func() {
    var s headerScanner
    s.b = rawHeaders
    for s.next() {
    }
  })

  if n != 0 {
    t.Fatalf("expected 0 allocations, got %f", n)
  }
}
```

Another test you can add to make sure things keep working is a test that calls `RequestHeader.VisitAllInOrder()` twice and see if the headers on the second call are still correct.

header.go
```diff
@@ -2022,9 +2022,19 @@ type headerScanner struct {
 	hLen int
 
 	disableNormalizing bool
+
+	nextColon   int
+	nextNewLine int
```
erikdubbelboer: Can you add some documentation on what these values are used for exactly?

header_timing_test.go
```diff
@@ -26,6 +26,72 @@ func (r *benchReadBuf) Read(p []byte) (int, error) {
...
+}
+
+func BenchmarkHeaderMultiLineValueOld(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		hs := new(oldHeaderScanner)
+		hs.b = webLogicResponse[17:]
+		for hs.next() {
+		}
+	}
+}
```
erikdubbelboer: I don't think we should keep this old code here just for a benchmark comparison. If you can run the benchmark locally and add the result as a comment to this pull request it's good enough. Then you can remove this benchmark.

tedli:
@erikdubbelboer 
Hi,
Thanks for reply. I'm working on this, but it's little complex, and will take some time.
Your point is that modify the request or response buffer in place, rather than alloc new space for multi line value. Am I got you wrong?
If so, we have to move the rest headers that haven't been parsed yet, and the body forward. If only one or two multi line value headers I think it's OK, but when several such header exists, we may move data many times.
And if just move content forward, it's not that harmless, but what if we have to move backward. I mean the `net/http` formatted the header value, like `one=1;two=2` will be parsed into `one=1; two=2`, there is a `space` inserted, in this circumstances we need move backward, and we need extra space, what the buffer may don't have.

erikdubbelboer:
I think if every time you get a newline within a header value you just skip that byte and copy everything after it back by one byte it should work.

tedli:
@erikdubbelboer 
Hi,
I've made new changes according to your suggestion.

erikdubbelboer:
@tedli sorry for the slow reply. This is a big change with lots of edge cases and I haven't found the time to really review this yet. I'll try to do it this weekend.

header.go
```diff
@@ -2046,14 +2067,41 @@ func (s *headerScanner) next() bool {
 	n++
 	for len(s.b) > n && s.b[n] == ' ' {
 		n++
+		s.nextNewLine--
```
erikdubbelboer: I'm not sure why this is needed here. Could you add a comment to explain why?

header_test.go
```diff
@@ -11,6 +11,40 @@ import (
...
+		"002e\r\n" +
+		"\r\n" +
+		"------=_Part_16_1220614102.1572836885552--\r\n" +
+		"\r\n" +
+		"0000\r\n" +
+		"\r\n")
```
erikdubbelboer: Can you simplify this test a bit? It looks very complicated because the multipart body contains headers as well. But these are completely irrelevant to the test. Just have one header split on 2 lines and another header split on 3 lines to show both those cases work. And maybe a normal header before and after to show that those aren't affected (so all in 1 request).

erikdubbelboer:
@tedli not sure if you noticed but the Travis CI build failed because your code panics on certain inputs: https://play.golang.org/p/jJTe0p9Hl9D

tedli:
@erikdubbelboer 
Thanks for notice. I will dig into it.
Only one question: Is this case a valid case, or just need extra checking to make it robust when getting invalid data?

erikdubbelboer:
This is just some random input generated by https://fuzzit.dev/ but it still shouldn't cause a panic of course. It needs to be robust against invalid data yes.

header_test.go
```diff
@@ -11,6 +11,31 @@ import (
...
+	}
+	gotContentType := header.Peek("Content-Type")
+	if string(gotContentType) != expectContentType {
+		t.Fatalf("unexpected content-type: %q. Expecting %q", gotContentType, expectContentType)
+	}
+	gotMultiLine := header.Peek("Multi-Line")
+	if string(gotMultiLine) != expectMultiLine {
+		t.Fatalf("unexpected multi-line: %q. Expecting %q", gotMultiLine, expectMultiLine)
+	}
+}
```
erikdubbelboer: Can you add one more test here, something like https://play.golang.org/p/Q8IJwhKbqHI which compares the header values against how `net/http` parses them.

The request you have in this test doesn't seem to be valid according to `net/http`: https://play.golang.org/p/wDchgwOABAg

erikdubbelboer:
Thanks!

erikdubbelboer:
@tedli I have made some more improvements, can you please check them? https://github.com/valyala/fasthttp/pull/708

tedli:
Hi, @erikdubbelboer 
By coding this multi-lined header one, I gained a deeper understanding about http headers. So I'm glade to work on #708 .
But currently I'm on an urgent sprint, it may be a week or two later.

PR #689, opened 2019-11-06
Ability to pass custom DNS resolver to TCPDialer

enchantner:
Right now default TCPDialer is forced to use default net.LookupIP(), which always uses default mechanism, like looking through "/etc/resolv.conf" on *nix. This request allows specifying a custom resolver that may be used for more flexible querying.

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -1,6 +1,7 @@
 package fasthttp
 
 import (
+	"context"
 	"errors"
 	"net"
 	"strconv"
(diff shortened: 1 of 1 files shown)


tcpdialer.go
```diff
@@ -422,20 +436,25 @@ func resolveTCPAddrs(addr string, dualStack bool) ([]net.TCPAddr, error) {
 		return nil, err
 	}
 
-	ips, err := net.LookupIP(host)
+	if resolver == nil {
+		resolver = &net.Resolver{}
```
erikdubbelboer:
```suggestion
		resolver = net.DefaultResolver
```
`net.DefaultResolver` is the default resolver `net.LookupIP` used. This way we don't break anything for people who would overwrite `net.DefaultResolver` instead (which of course is bad practice).

erikdubbelboer:
Thanks!

PR #691, opened 2019-11-08
fix Request CopyTo 

niqinge:
The 'CopyTo' method does not copy multipartForm


erikdubbelboer:
Can you please remove the change to `README.md` and add a test case that fails before this change and is a success afterwards. Thanks!

Fenny:
@niqinge could you please respond to this PR so we can close/merge it, thank you 👍

PR #694, opened 2019-11-15
Add Windows support to travisCI

panjf2000:
As title.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ dist: bionic
 os:
   - linux
   - osx
+  - windows
+
 go:
   - tip
   - 1.13.x


erikdubbelboer:
Looks like windows builds on TravisCI are still experimental. The page says `Windows builds are in early access stage.` and I see at least one test failing with `The command "travis_setup_go" failed and exited with 86 during .`. Other tests are getting timeouts which normally never happen. I think it's probably better to wait for windows on TravisCI to be more stable before we merge this.

panjf2000:
@erikdubbelboer 
Make sense, I just add Windows support of TravisCI in my own repo before and I also discover some issues about it, I agree to suspend the PR until formal support of Windows.

PR #695, opened 2019-11-15
Add benchmarks with other http frameworks

panjf2000:
As title.


erikdubbelboer:
Thanks for the pull request. But I think it contains too much unrelated information. fasthttp is just an HTTP library. This benchmark compares routers and frameworks which are all unrelated to fasthttp performance. fasthttp can only really be benchmarked against net/http as that's the only other HTTP implementation in Go (as far as I know).

PR #699, opened 2019-11-19
Custom LookupIPAddr function

enchantner:
Continuing the work started at https://github.com/valyala/fasthttp/pull/689/files . 

Turns out, specifying custom resolver doesn't allow to avoid resolving completely (as `Dial()` has to return a connection to a DNS server), so in this way it becomes possible to reimplement `LookupIPAddr()` function to specify resolved name manually or get it from some place other than actual DNS query.

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -443,10 +443,9 @@ func resolveTCPAddrs(addr string, dualStack bool, resolver Resolver) ([]net.TCPA
 	if resolver == nil {
 		resolver = net.DefaultResolver
 	}
-
-	var ipaddrs []net.IPAddr
+	
 	ctx := context.Background()
-	ipaddrs, err = resolver.LookupIPAddr(ctx, host)
+	ipaddrs, err := resolver.LookupIPAddr(ctx, host)
 	if err != nil {
 		return nil, err
 	}
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -120,7 +120,9 @@ var (
 	defaultDialer = &TCPDialer{Concurrency: 1000}
 )
 
-type LookupIPAddrFunc func(ctx context.Context, host string) ([]net.IPAddr, error)
+type Resolver interface {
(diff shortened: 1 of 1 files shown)


enchantner:
Won't it better to use `LookupIPAddr` instead of `LookupAddr`? Afaik, `LookupAddr` only deals with IPv4 addresses, but with `LookupIPAddr` we're getting both. Anyway, here's a commit.

tcpdialer.go
```diff
@@ -440,8 +444,9 @@ func resolveTCPAddrs(addr string, dualStack bool, resolver *net.Resolver) ([]net
 		resolver = net.DefaultResolver
 	}
 
+	var ipaddrs []net.IPAddr
 	ctx := context.Background()
-	ipaddrs, err := resolver.LookupIPAddr(ctx, host)
+	ipaddrs, err = resolver.LookupIPAddr(ctx, host)
```
erikdubbelboer: This doesn't need to be changed? It can stay like `:=`.

enchantner:
@erikdubbelboer `LookupAddr` spits out a list of strings, which are each an IPv4 address in "four numbers with dots" representation. That's totally okay, but `IPAddr` type is more common, as you can get both IPv4 and IPv6 from it, if present. It looks like fasthttp doesn't support IPv6 yet, but in this way we may be able to introduce it without changing this Resolver interface again.

erikdubbelboer:
`LookupAddr` goes from IP to domain. `LookupIPAddr` goes from domain to IP.

Where in fasthttp do you think we use `LookupAddr`?

fasthttp does support IPv6 just fine.

enchantner:
@erikdubbelboer yeah, sorry, I just got everything mixed up looking at the example code you've provided above for the interface. Thanks for merging this!

erikdubbelboer:
Ah sorry I didn't realize I put `LookupAddr` in my example interface. My bad!

PR #713, opened 2019-12-31
Set StatusRequestTimeout when net timeout error occurred

SilverRainZ:
It is confusing that defaultErrorHandler returns "Error when parsing request" when request timeout.


erikdubbelboer:
Thanks!

PR #715, opened 2020-01-08
Optimized some logic when illegal request method.

wangfakang:
In `RequestHeader.parseFirstLin` when the parsed method can be terminated early for illegal length.


erikdubbelboer:
The HTTP spec says:
```
The set of common methods for HTTP/1.1 is defined below. Although this set can be expanded, additional methods cannot be assumed to share the same semantics for separately extended clients and servers.
```
So someone using method `XXXXXXXX` for their service is perfectly fine according to the spec.

I'm sorry but we can not do this optimization.

PR #720, opened 2020-01-10
Forward context in fasthttpadaptor

lovromazgon:
Right now the request in fasthttpadaptor has a background context. It would be beneficial to set it to `fasthttp.RequestCtx` or at least forward values that are already set in the context. This would allow us to retrieve any value that was set in middleware before the request reached the `http.Handler`.

### DIFF
--- a/fasthttpadaptor/adaptor_test.go
+++ b/fasthttpadaptor/adaptor_test.go
@@ -32,6 +32,8 @@ func TestNewFastHTTPHandler(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %s", err)
 	}
+	expectedContextKey := "contextKey"
+	expectedContextValue := "contextValue"
 
 	callsCount := 0
 	nethttpH := func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/valyala/fasthttp
+module github.com/lovromazgon/fasthttp
```
erikdubbelboer: We obviously can't merge this pull request with this change.

lovromazgon: Sorry, removed it!

fasthttpadaptor/adaptor_test.go
```diff
@@ -128,3 +134,10 @@ func TestNewFastHTTPHandler(t *testing.T) {
 		t.Fatalf("unexpected response body %q. Expecting %q", resp.Body(), expectedResponseBody)
 	}
 }
+
+func setContextValueMiddleware(next fasthttp.RequestHandler, key string, value interface{}) fasthttp.RequestHandler{
+	return func(ctx *fasthttp.RequestCtx) {
+		ctx.SetUserValue(key, value)
+		next(ctx)
+	}
+}
```
erikdubbelboer: Can you please `go fmt` this file.

erikdubbelboer:
Thanks!

PR #729, opened 2020-01-15
Make a little bit trivial code simplification

panjf2000:
As title.


erikdubbelboer:
Thanks for the pull request but I'm sorry but I don't think it makes it simpler. The 3 in one line assignment is harder to follow than 3 lines. I feel like this commit doesn't really add anything and just makes our commit history messier. So I'm afraid I won't merge this.

PR #733, opened 2020-01-21
Add third-party router / framework to README.md

Fenny:
Could you add [Fiber](https://gofiber.github.io/fiber) to the list?

Thanks!


erikdubbelboer:
I'm sorry but your framework doesn't seem to have any users and most commits are changes to the readme or the license. I'm afraid we can't merge this for now. When your framework is more mature and has more users we can reconsider.

Fenny:
I'm not sure if you have the correct library, because it has 80 stars and is well tested.
https://github.com/gofiber/fiber

PR #738, opened 2020-02-07
increase correctness of GetOpenConnectionsCount while Server is shutting down

forestgagnon:
Currently, it appears that GetOpenConnectionsCount underreports the
count by 1 after Shutdown has been called. This leads to a confusing
value when servers are gracefully terminating. For instance, if a server
is stuck in graceful termination, and there is one open connection
keeping it alive, GetOpenConnectionsCount would report zero.

This fix removes the decrement while the server is shutting down. It is
not perfect due to the use of two sequential atomic loads, but in the
common case it should return a more correct value overall.


erikdubbelboer:
Thanks!

PR #741, opened 2020-02-09
Add prefork utility

savsgio:


prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+var Child bool
+
+// Prefork implements fasthttp server prefork
+//
+// Preforks master process (with all cores) between several child processes
+// increases performance significantly, because Go doesn't have to share
+// and manage memory between cores
+//
+// WARNING: Does not recommended for servers with in-memory cache,
+// because the cache will be duplicated in each process
```
erikdubbelboer: Please change this warning to something like:
```
WARNING: using prefork prevents the use of any global state! Things like in-memory caches won't work.
```
It's not just about in-memory caches. Everything that uses global state won't work.

prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if !p.Reuseport {
+		if err := p.setTCPListenerFiles(addr); err != nil {
+			return err
+		}
+
+		defer p.ln.Close()
+	}
+
+	for i := 0; i < runtime.NumCPU(); i++ {
```
erikdubbelboer: Shouldn't this be `GOMAXPROCS`? So the user can control how many cores will be used just like with normal Go processes.

prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+	"runtime"
+
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/reuseport"
+)
+
+const preforkChildFlag = "prefork-child"
+const defaultNetwork = "tcp4"
+
+var Child bool
```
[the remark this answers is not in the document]
savsgio: I did it  public for know if the current thread/worker is a child or not, because sometimes could be useful to change some things if it's a master or a worker.
You could check the `preforkChildFlag` but you need to know the name of it and parse the input, so what do you think about this solution:

```go
var child bool

func isChild() bool {
	return child
}
```

prefork/prefork.go
savsgio: Done!

prefork/prefork.go
savsgio: Done!

savsgio:
> Thanks for the pull request. Some minor changes needed.
> Can you also add a README that explains how this works and maybe some benchmarks which show the speed improvements. Thanks

Ok, i will do it.

savsgio:
README added!

prefork/prefork.go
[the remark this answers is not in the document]
savsgio: What do you think about it?

prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+	ServeFunc         func(ln net.Listener) error
+	ServeTLSFunc      func(ln net.Listener, certFile, keyFile string) error
+	ServeTLSEmbedFunc func(ln net.Listener, certData, keyData []byte) error
+
+	ln    net.Listener
+	files []*os.File
+}
+
+func init() { // nolint:gochecknoinits
+	flag.BoolVar(&child, preforkChildFlag, false, "is child proc")
```
erikdubbelboer: This is never going to work be cause [`flag.Parse()`](https://golang.org/pkg/flag/#Parse) is never called.

prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+			chErr <- cmd.Run()
+		}()
+	}
+
+	return <-chErr
+}
+
+// ListenAndServe serves HTTP requests from the given TCP addr
+func (p *Prefork) ListenAndServe(addr string) error {
+	if child {
```
erikdubbelboer: Replace this with:
```go
if isChild() {
```
And
```
func isChild() bool {
	for _, arg := range os.Args {
		if arg == preforkChildFlag {
			return true
		}
	}
	return false
}
```
And just make `preforkChildFlag = "-prefork-child"` etc...

prefork/prefork.go
```diff
@@ -0,0 +1,179 @@
...
+// Prefork implements fasthttp server prefork
+//
+// Preforks master process (with all cores) between several child processes
+// increases performance significantly, because Go doesn't have to share
+// and manage memory between cores
+//
+// WARNING: using prefork prevents the use of any global state!
+// Things like in-memory caches won't work.
+type Prefork struct {
+	Addr string
```
erikdubbelboer: This member isn't needed anywhere and doesn't even work if you set it.

prefork/prefork.go
savsgio: Yes, i didn't understand it well before.
Fixed!

prefork/prefork.go
savsgio: Done!

prefork/prefork.go
savsgio: I changed it to `flag.Bool(preforkChildFlag[1:], false, "is a child process")`, because if the user define some extra flags in their program and pass `-prefork-child`, the program crash.

savsgio: The definition of the flag is mandatory

prefork/prefork.go
```diff
@@ -0,0 +1,182 @@
...
+	ServeFunc         func(ln net.Listener) error
+	ServeTLSFunc      func(ln net.Listener, certFile, keyFile string) error
+	ServeTLSEmbedFunc func(ln net.Listener, certData, keyData []byte) error
+
+	ln    net.Listener
+	files []*os.File
+}
+
+func init() { //nolint:gochecknoinits
+	flag.Bool(preforkChildFlag[1:], false, "is a child process")
```
erikdubbelboer: Why is this still here? It doesn't do anything?

prefork/prefork.go
```diff
@@ -0,0 +1,182 @@
...
+	if !p.Reuseport {
+		if err := p.setTCPListenerFiles(addr); err != nil {
+			return err
+		}
+
+		defer p.ln.Close()
+	}
+
+	for i := 0; i < runtime.GOMAXPROCS(0); i++ {
+		cmd := exec.Command(strCmd, append(os.Args[1:], preforkChildFlag)...)
```
erikdubbelboer:
```suggestion
		cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
```

prefork/prefork_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"os"
+	"reflect"
+	"runtime"
+	"testing"
+
+	"github.com/valyala/fasthttp"
+)
+
+func init() { //nolint:gochecknoinits
+	setUp()
```
erikdubbelboer: Please move this into the tests. If I'm looking at `Test_ListenAndServe` for example, it is unclear to me that this is supposed to be a child process and not the master process. It's always a bad idea to do things in `init` in tests.

prefork/prefork.go
```diff
@@ -0,0 +1,186 @@
...
+type Prefork struct {
+	// The network must be "tcp", "tcp4" or "tcp6".
+	//
+	// By default is "tcp4"
+	Network string
+
+	// Flag to use a listener with reuseport, if not a File Listener will be used
+	//
+	// It's disabled by default
+	Reuseport bool
```
erikdubbelboer: Maybe add some documentation on why `Reuseport` is better than just sharing a single socket.
This link has a good explanation on why: https://www.nginx.com/blog/socket-sharding-nginx-release-1-9-1/

prefork/prefork.go
savsgio: Example:

```go
package main

import (
	"flag"
	"log"
	"os"

	"github.com/valyala/fasthttp"
	"github.com/valyala/fasthttp/prefork"
)

var (
	addr = flag.String("addr", ":8080", "TCP address to listen to")
)

func requestHandler(ctx *fasthttp.RequestCtx) {
	ctx.WriteString("Hello World")
}

func main() {
	flag.Parse()

	s := &fasthttp.Server{Handler: requestHandler}
	ps := prefork.New(s)

	log.Printf("Running on %s (pid: %d)", *addr, os.Getpid())
	if err := ps.ListenAndServe(*addr); err != nil {
		log.Fatalf("Error in ListenAndServe: %s", err)
	}
}
```

If i not define the flag `prefork-child`, i will take the following error:
```bash
2020/02/12 12:35:21 Running on :8080 (pid: 12547)
flag provided but not defined: -prefork-child
Usage of /var/folders/kl/zdgwjx956cv0yw1pt89zzxt80000gp/T/go-build831309057/b001/exe/helloworldserver:
  -addr string
        TCP address to listen to (default ":8080")
```

savsgio: Only in cases that the user create their own flags

erikdubbelboer: Ok then it makes sense. But please add a comment explaining why cause right now someone reading the code will not understand why that is there.

prefork/prefork_test.go
savsgio: Done!

prefork/prefork.go
savsgio: Done!

prefork/prefork.go
savsgio: Done

erikdubbelboer:
Thanks!

savsgio:
Thanks for merge it 😃

tsingson:
cool. thanks a lot. 

i try!  maybe i can use this to replace taskset  + reuseport + multiple instance fasthttp serv.

PR #755, opened 2020-03-03
Make the prefork mode more robust

panjf2000:
The main process will exit if one of the prefork child processes doesn't complete successfully under the current prefork mode, so it ought to make sure that all child processes run independently and the main process will only exit after all child processes are finished.

### DIFF
--- a/prefork/prefork.go
+++ b/prefork/prefork.go
@@ -12,8 +12,11 @@ import (
 	"github.com/valyala/fasthttp/reuseport"
 )
 
-const preforkChildFlag = "-prefork-child"
-const defaultNetwork = "tcp4"
+const (
+	preforkChildFlag        = "-prefork-child"
+	defaultNetwork          = "tcp4"
+	defaultRecoverThreshold = 10
+)
 
 // Prefork implements fasthttp server prefork
 //
@@ -34,6 +37,10 @@ type Prefork struct {
 	//
 	// It's disabled by default
 	Reuseport bool
+
+	// Child prefork processes may exit with failure and will be started over until the times reach
+	// the value of RecoverThreshold, then it will return and terminate the server.
+	RecoverThreshold int
 
 	ServeFunc         func(ln net.Listener) error
 	ServeTLSFunc      func(ln net.Listener, certFile, keyFile string) error
@@ -64,6 +71,7 @@ func IsChild() bool {
 func New(s *fasthttp.Server) *Prefork {
 	return &Prefork{
 		Network:           defaultNetwork,
+		RecoverThreshold:  defaultRecoverThreshold,
 		ServeFunc:         s.Serve,
 		ServeTLSFunc:      s.ServeTLS,
 		ServeTLSEmbedFunc: s.ServeTLSEmbed,
@@ -156,14 +164,21 @@ func (p *Prefork) prefork(addr string) (err error) {
 		}()
 	}
 
-	var completeProcs int
+	var brokenProcs, completeProcs int
 	for sig := range sigCh {
 		if sig.err != nil {
 			delete(childProcs, sig.pid)
 
 			log.Printf("one of the child prefork processes failed to complete, "+
 				"error: %v", sig.err)
-			
+
+			if brokenProcs++; brokenProcs > p.RecoverThreshold {
+				log.Printf("child prefork processes exit too many times, "+
+					"which exceeds the value of RecoverThreshold(%d), "+
+					"exiting the master process.\n", brokenProcs)
+				break
+			}
+
 			/* #nosec G204 */
 			cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
 			cmd.Stdout = os.Stdout
@@ -174,8 +189,7 @@ func (p *Prefork) prefork(addr string) (err error) {
 				sigCh <- procSig{cmd.Process.Pid, cmd.Run()}
 			}()
 		} else {
-			completeProcs++
-			if completeProcs == goMaxProcs {
+			if completeProcs++; completeProcs == goMaxProcs {
 				break
 			}
 		}
--- a/prefork/prefork.go
+++ b/prefork/prefork.go
@@ -124,8 +124,21 @@ func (p *Prefork) prefork(addr string) (err error) {
 		}()
 	}
 
+	type procSig struct {
+		pid int
+		err error
+	}
+
 	goMaxProcs := runtime.GOMAXPROCS(0)
-	errCh := make(chan error, goMaxProcs)
+	sigCh := make(chan procSig, goMaxProcs)
+	childProcs := make(map[int]*exec.Cmd)
+
+	defer func() {
+		for _, cmd := range childProcs {
+			_ = cmd.Process.Kill()
+		}
+	}()
+
 	for i := 0; i < goMaxProcs; i++ {
 		/* #nosec G204 */
 		cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
@@ -133,27 +146,34 @@ func (p *Prefork) prefork(addr string) (err error) {
 		cmd.Stderr = os.Stderr
 		cmd.ExtraFiles = p.files
 		if err = cmd.Start(); err != nil {
+			for _, proc := range childProcs {
+				_ = proc.Process.Kill()
+			}
 			log.Printf("failed to start a child prefork process, error: %v\n", err)
 			return
 		}
 
+		childProcs[cmd.Process.Pid] = cmd
 		go func() {
-			errCh <- cmd.Wait()
+			sigCh <- procSig{cmd.Process.Pid, cmd.Wait()}
 		}()
 	}
 
 	var completeProcs int
-	for procErr := range errCh {
-		if procErr != nil {
+	for sig := range sigCh {
+		if sig.err != nil {
+			delete(childProcs, sig.pid)
+
 			log.Printf("one of the child prefork processes failed to complete, "+
-				"error: %v", procErr)
+				"error: %v", sig.err)
 			/* #nosec G204 */
 			cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
 			cmd.Stdout = os.Stdout
 			cmd.Stderr = os.Stderr
 			cmd.ExtraFiles = p.files
+			childProcs[cmd.Process.Pid] = cmd
 			go func() {
-				errCh <- cmd.Run()
+				sigCh <- procSig{cmd.Process.Pid, cmd.Run()}
 			}()
 		} else {
 			completeProcs++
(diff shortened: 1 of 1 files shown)


savsgio:
HI @panjf2000,

It's not necessary to wait to the all child procs will finish, because if the master process dies, the children should die too.
Also, you are collected all error and just return the first one.

```go
return p.prefork(addr)[0]
```

Think that, if you wait to all child procs dies/end, you will lose performance because maybe you couldn't know it, but if master dies when any of child procs dies, you could know quickly that the server is dead and not lose performance due to the some cores doesn't work.

But we wait for @erikdubbelboer, to know their opinion.

panjf2000:
> It's not necessary to wait to the all child procs will finish, because if the master process dies, the children should die too.

I know that, but this PR doesn't intend to make sure all child processes to exit after the main process exit, it try to avoid that any child process to break the main process.

> Also, you are collected all error and just return the first one.

Yes, you are right but the root cause of it is that the definition of `ListenAndServe` only return one single `error` so I have to pick only one error to return:
https://github.com/valyala/fasthttp/blob/695f713fcf594b33fd8c126cfa8dc9f119d4391b/server.go#L88
also, it printed those else errors to stdout, unfortunately it is only way for now.

> Think that, if you wait to all child procs dies/end, you will lose performance because maybe you couldn't know it, but if master dies when any of child procs dies, you could know quickly that the server is dead and not lose performance due to the some cores doesn't work.

You got a point but in my opinion, it makes more sense to choose robustness over performance in this case because fasthttp is fast enough, it can still provide high-performance services even though it does lose a little bit of performance because of some child prefork processes breaking down, so it shouldn't shut the entire server down just because one or two prefork child processes break down.

@savsgio

savsgio:
What do you think @erikdubbelboer ?

erikdubbelboer:
I think you're both wrong 😄 . When a parent process dies doesn't mean the children also automatically dies: https://unix.stackexchange.com/questions/158727/is-there-any-unix-variant-on-which-a-child-process-dies-with-its-parent

I think if one child dies it should be restarted. Up to a certain (configuratble) limit at which point `ListenAndServe` should return with an error. But before that, it should clean up/kill all the children. That way developers using this can handle failure in the way that they would like.

panjf2000:
> When a parent process dies doesn't mean the children also automatically dies: https://unix.stackexchange.com/questions/158727/is-there-any-unix-variant-on-which-a-child-process-dies-with-its-parent

It is right in general, but we were talking about goroutines, you see, in the previous code the `cmd.Run()` was wrapped into a goroutine, so when the goroutine that starts other goroutines dies, other goroutines die, then the child processes started in goroutines become orphan processes and adopted by init process, which I think it is also a bad thing.

> I think if one child dies it should be restarted. Up to a certain (configuratble) limit at which point ListenAndServe should return with an error. But before that, it should clean up/kill all the children. That way developers using this can handle failure in the way that they would like.

So you are suggesting that I should add code that restarts child processes automatically when it finds any child process dies?

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+				err = errors.New("exceeding the value of RecoverThreshold")
+				break
+			}
+
+			/* #nosec G204 */
+			cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.ExtraFiles = p.files
+			childProcs[cmd.Process.Pid] = cmd
```
erikdubbelboer: Isn't `cmd.Process.Pid` here still the old value as `cmd.Run()` hasn't started yet?

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+	}()
+
+	for i := 0; i < goMaxProcs; i++ {
 		/* #nosec G204 */
 		cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 		cmd.ExtraFiles = p.files
+		if err = cmd.Start(); err != nil {
+			log.Printf("failed to start a child prefork process, error: %v\n", err)
```
erikdubbelboer: I would prefer to not use the `log` package. This was users don't have much control over the output of their program. We don't do this anywhere else in fasthttp either. Better is doing something similar to https://godoc.org/github.com/valyala/fasthttp#Server.Logger

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+
+	defer func() {
+		for _, proc := range childProcs {
+			_ = proc.Process.Kill()
+		}
+	}()
+
+	for i := 0; i < goMaxProcs; i++ {
 		/* #nosec G204 */
 		cmd := exec.Command(os.Args[0], append(os.Args[1:], preforkChildFlag)...)
```
erikdubbelboer: I would prefer if you could merge this `exec.Command` setup with the one below. Maybe put it in a separate function?

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
-		if err := p.setTCPListenerFiles(addr); err != nil {
-			return err
+		if err = p.setTCPListenerFiles(addr); err != nil {
+			return
 		}
 
-		defer p.ln.Close()
+		defer func() {
+			if err == nil {
+				err = p.ln.Close()
```
erikdubbelboer: Can you document why the listener only being closed if `!p.Reuseport`.

erikdubbelboer: And even if `err != nil` the listener still needs to be closed properly.

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+			delete(childProcs, sig.pid)
+
+			log.Printf("one of the child prefork processes failed to complete, "+
+				"error: %v", sig.err)
+
+			if brokenProcs++; brokenProcs > p.RecoverThreshold {
+				log.Printf("child prefork processes exit too many times, "+
+					"which exceeds the value of RecoverThreshold(%d), "+
+					"exiting the master process.\n", brokenProcs)
+				err = errors.New("exceeding the value of RecoverThreshold")
```
erikdubbelboer: This should be an exported package constant so users can compare against it.

prefork/prefork.go
```diff
@@ -110,30 +120,84 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+			cmd.Stdout = os.Stdout
+			cmd.Stderr = os.Stderr
+			cmd.ExtraFiles = p.files
+			childProcs[cmd.Process.Pid] = cmd
+			go func() {
+				sigCh <- procSig{cmd.Process.Pid, cmd.Run()}
+			}()
+		} else {
+			if completeProcs++; completeProcs == goMaxProcs {
+				break
```
erikdubbelboer: I'm not following this logic. Is this for when the process exists without an error? Why wouldn't that use the same counter as `brokenProcs`?

prefork/prefork.go
panjf2000: My oversight, fixed it.

prefork/prefork.go
panjf2000: Got it.

prefork/prefork.go
panjf2000: Good point.

prefork/prefork.go
panjf2000: Done.

prefork/prefork.go
panjf2000: > Can you document why the listener only being closed if !p.Reuseport.

You mean to add some comments on it?

erikdubbelboer: Yes please. I know the close was already there, but might as well add a comment if we're making changes here 😄

prefork/prefork.go
panjf2000: > Why wouldn't that use the same counter as brokenProcs?

Sorry, I didn't get your point, two counter are for different purposes, or you were saying that we shouldn't exit the master process even if all child processes exit successfully?

prefork/prefork.go
panjf2000: Done.

prefork/prefork.go
panjf2000: You prefer to start over any exited child process despite of it is successful or unsuccessful?

erikdubbelboer: Yes. Does `os/exec` even return an error if the process exists with an error? Doesn't it only throw an error if there was some other cause like not being able to start the other process?

prefork/prefork.go
```diff
@@ -110,30 +144,91 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
-		if err := p.setTCPListenerFiles(addr); err != nil {
-			return err
+		if err = p.setTCPListenerFiles(addr); err != nil {
+			return
 		}
 
-		defer p.ln.Close()
+		// defer for closing the net.Listener opened by setTCPListenerFiles.
+		defer func() {
+			e := p.ln.Close()
```
erikdubbelboer: Should the listener from reuseport not be closed?

erikdubbelboer:
Isn't it better if RecoverThreshold has been reached for the master process to kill all the remaining children before returning?

server.go
```diff
@@ -876,16 +876,21 @@ func (ctx *RequestCtx) FormFile(key string) (*multipart.FileHeader, error) {
 var ErrMissingFile = errors.New("there is no uploaded file associated with the given key")
 
 // SaveMultipartFile saves multipart file fh under the given filename path.
-func SaveMultipartFile(fh *multipart.FileHeader, path string) error {
-	f, err := fh.Open()
+func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
```
erikdubbelboer: I would prefer you make a separate pull request for these changes. I can them merge those into master and you can update this branch on top of master.

panjf2000: OK

panjf2000:
> Isn't it better if RecoverThreshold has been reached for the master process to kill all the remaining children before returning?

What's the problem of killing them in `defer`?

prefork/prefork.go
panjf2000: `p.ln` doesn't even get assigned an instance of `net.Listener` under reuseport mode.

panjf2000: Is the comment not good enough?

prefork/prefork.go
panjf2000: Done.

panjf2000:
Any progress？
@erikdubbelboer

erikdubbelboer:
Thanks!

savsgio:
Thanks!

PR #758, opened 2020-03-09
Add support for `NoDefaultDate` option that excludes the sending of the `Date` header

phuslu:
same as https://github.com/valyala/fasthttp/pull/629  but with tests.

### DIFF
--- a/server.go
+++ b/server.go
@@ -322,6 +322,13 @@ type Server struct {
 	// value is explicitly provided during a request.
 	NoDefaultServerHeader bool
 
+	// NoDefaultDate, when set to true, causes the default Date
+	// header to be excluded from the Response.
+	//
+	// The default Date header value is the current date value. When
+	// set to true, the Date will not be present.
+	NoDefaultDate bool
+
 	// NoDefaultContentType, when set to true, causes the default Content-Type
 	// header to be excluded from the Response.
 	//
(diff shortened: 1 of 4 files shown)


phuslu:
For changes in https://github.com/valyala/fasthttp/pull/758/files#diff-34c6b408d72845d076d47126c29948d1R919-R928 , just make GoSec happy.

server.go
```diff
@@ -907,9 +914,18 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) error {
...
 	_, err = copyZeroAlloc(ff, f)
-	return err
+	if err != nil {
+		return err
+	}
+
+	err = ff.Close()
+	if err != nil {
+		return err
+	}
```
erikdubbelboer: Now `ff.Close()` isn't called if `copyZeroAlloc` fails which could cause files to stay open (multipart files are backed by os files). I'm not sure why this wasn't an issue before. Doesn't changing it to `defer ff.Close() //nolint:errcheck` fix the issue?

phuslu: Done

phuslu:
Hi @erikdubbelboer, I amended the commit. please have a look.

erikdubbelboer:
Thanks!

phuslu:
With this PR, we could minimize fasthttp response by following codes
```golang
server := &fasthttp.Server{
	NoDefaultServerHeader: true,
	NoDefaultContentType:  true,
	NoDefaultDate:         true,
	DisableKeepalive:      false,
}

// ... ...

ctx.SetStatusCode(fasthttp.StatusNoContent)
```
will give output:
```
HTTP/1.1 204 No Content

```

PR #764, opened 2020-03-13
feat: make client to wait when no free connections

chanjarster:
Add a new field `MaxConnWaitTimeout` on Client. 
* If `MaxConnWaitTimeout` == 0, when all connections are busy, will return `ErrNoFreeConns` immediately (just like what it already does). 
* If `MaxConnWaitTimeout` > 0, when all connections are busy, Client will wait at most `MaxConnWaitTimeout` for a chance to get a connection, if Client failed, will return `ErrNoFreeConns`


client.go
```diff
@@ -1355,7 +1367,34 @@ func (c *HostClient) acquireConn() (*clientConn, error) {
...
-		return nil, ErrNoFreeConns
+		if c.MaxConnWaitTimeout <= 0 {
+			return nil, ErrNoFreeConns
+		}
+
+		// wait for a free connection
+		ctx, cancelFunc := context.WithTimeout(context.Background(), c.MaxConnWaitTimeout)
+		w := &wantConn{
+			ready: make(chan struct{}, 1),
+		}
```
erikdubbelboer: Instead of using a context (which generates a lot of garbage) please use AcquireTimer. For example see:
https://github.com/valyala/fasthttp/blob/695f713fcf594b33fd8c126cfa8dc9f119d4391b/client.go#L782-L794

chanjarster:
> First thing I notice is a lot of potential heap allocations which can slow things down. Did you try doing any benchmarks or checking the number of allocations with a test like this?:
> https://github.com/valyala/fasthttp/blob/38aa88ab52c4b6f61722e2e7809c17b6f7df2b2a/allocation_test.go#L40-L72

I've noticed that the current benchmark and allocation test are based on fast server. And, by default (which `MaxConnWaitTimeout=0`) this implementation doesn't queue the request and return ErrNoFreeConns immediately.

I'll provide a allocation test based on slow server and enable `MaxConnWaitTimeout` later.

client.go
chanjarster: Resolved, please see commit 014e836

chanjarster:
Hi @erikdubbelboer , I add `ClientGetEndToEndWaitConn` benchmarks to show heap allocations (see commit cb96d97), below is my test result.

```txt
$ GOMAXPROCS=4 go test -bench='kClientGetEndToEndWaitConn' -benchmem -benchtime=10s
goos: darwin
goarch: amd64
BenchmarkClientGetEndToEndWaitConn1Inmemory-4      	     237	  50495156 ns/op	     917 B/op	       4 allocs/op
BenchmarkClientGetEndToEndWaitConn10Inmemory-4     	     397	  25392398 ns/op	     397 B/op	       3 allocs/op
BenchmarkClientGetEndToEndWaitConn100Inmemory-4    	    4700	   2565375 ns/op	     193 B/op	       2 allocs/op
BenchmarkClientGetEndToEndWaitConn1000Inmemory-4   	   41949	    263421 ns/op	     183 B/op	       2 allocs/op
BenchmarkClientGetEndToEndWaitConn10kInmemory-4    	  414907	     29407 ns/op	     183 B/op	       2 allocs/op
PASS
ok  	github.com/valyala/fasthttp	87.533s
```

chanjarster:
Hi @erikdubbelboer , I add some benchmarks to show heap allocations (see commit cb96d97 and b3a83d1) when in situation of waiting for a free connection, below is my test result:

fasthttp:

```txt
$ GOMAXPROCS=4 go test -bench='kClientGetEndToEndWaitConn' -benchmem -benchtime=10s
goos: darwin
goarch: amd64
BenchmarkClientGetEndToEndWaitConn1Inmemory-4      	     237	  50495156 ns/op	     917 B/op	       4 allocs/op
BenchmarkClientGetEndToEndWaitConn10Inmemory-4     	     397	  25392398 ns/op	     397 B/op	       3 allocs/op
BenchmarkClientGetEndToEndWaitConn100Inmemory-4    	    4700	   2565375 ns/op	     193 B/op	       2 allocs/op
BenchmarkClientGetEndToEndWaitConn1000Inmemory-4   	   41949	    263421 ns/op	     183 B/op	       2 allocs/op
BenchmarkClientGetEndToEndWaitConn10kInmemory-4    	  414907	     29407 ns/op	     183 B/op	       2 allocs/op
PASS
ok  	github.com/valyala/fasthttp	87.533s
```

net/http:

```txt
$ GOMAXPROCS=4 go test -bench='HTTPClientGetEndToEndWaitConn' -benchmem -benchtime=10s
goos: darwin
goarch: amd64
BenchmarkNetHTTPClientGetEndToEndWaitConn1Inmemory-4      	     951	  12644110 ns/op	    9435 B/op	      76 allocs/op
BenchmarkNetHTTPClientGetEndToEndWaitConn10Inmemory-4     	    9501	   1262682 ns/op	   10328 B/op	      79 allocs/op
BenchmarkNetHTTPClientGetEndToEndWaitConn100Inmemory-4    	   91471	    133340 ns/op	   12020 B/op	      84 allocs/op
BenchmarkNetHTTPClientGetEndToEndWaitConn1000Inmemory-4   	  277612	     39097 ns/op	   13705 B/op	      89 allocs/op
PASS
ok  	github.com/valyala/fasthttp	74.091s
```

client_timing_test.go
```diff
@@ -640,3 +640,155 @@ func benchmarkPipelineClient(b *testing.B, parallelism int) {
...
+			err.Error(), "use of closed network connection") {
+			b.Errorf("error when serving requests: %s", err)
+		}
+		close(ch)
+	}()
+
+	c := &http.Client{
+		Transport: &http.Transport{
+			Dial:            func(_, _ string) (net.Conn, error) { return ln.Dial() },
+			MaxConnsPerHost: 1,
```
chanjarster: @erikdubbelboer travis got an error of `./client_timing_test.go:757:19: unknown field 'MaxConnsPerHost' in struct literal of type http.Transport` on this line in go version 1.9.x and 1.10.x. What should i do?

chanjarster: resloved, using `// +build go1.11`

client.go
```diff
@@ -1461,9 +1561,29 @@ var clientConnPool sync.Pool
...
+	// try to deliver an idle connection to a *wantConn
 	c.connsLock.Lock()
-	c.conns = append(c.conns, cc)
-	c.connsLock.Unlock()
+	defer c.connsLock.Unlock()
+	delivered := false
+	if q := c.connsWait; q != nil && q.len() > 0 {
+		for q.len() > 0 {
+			w := q.popFront()
+			if w.waiting() {
```
erikdubbelboer: `popFront` can return `nil` in which case this will panic.

client.go
```diff
@@ -1355,7 +1367,47 @@ func (c *HostClient) acquireConn() (*clientConn, error) {
...
+				timeoutOverridden = true
+		}
+
+		// wait for a free connection
+		tc := AcquireTimer(timeout)
+		defer ReleaseTimer(tc)
+
+		w := &wantConn{
+			ready: make(chan struct{}, 1),
+		}
```
erikdubbelboer: Would it be possible to reuse `wantConn` or the channel? Or is that going to make the code so complicated that its not worth it?

client.go
chanjarster: Since `q.len() > 0`, so `popFront` is impossible return `nil`

client.go
chanjarster: For the first question, if we want to reuse the `wantConn`, then we must reuse the channel, but AFAIK there's no way to reopen a closed channel (maybe).
In terms of the heap allocation, yes, this will cause heap allocation and slow things down, but if we want to waiting for a free connection, the waiting dominates the slow, not gc.

For the second question, many http clients provide connection pooling and waiting mechanism (by blocking). This waiting mechanism is very useful, it can simplify the code on the caller side, otherwise the caller have two options:

1. busy waiting, try again immediately at some max try count, which consumes cpu time.
2. sleep waiting, sleep for a while and try again, which can't get connection in time.

Both are not perfect. So I think the waiting mechanism is worth it.

For the complication of this PR, em ..., actually this PR is based on [net/http/transport.go][1], maybe there is a more simple way, we can have a discussion about that :)

[1]: https://github.com/golang/go/blob/master/src/net/http/transport.go#L95

erikdubbelboer: I have thought about it and I agree. Let's merge this!

PR #765, opened 2020-03-15
client: Implement DoRedirects

tsenart:
This commit adds a `DoRedirects` method to both `HostClient` and
`Client` as well as top level convenience function of the same name that
is called with the package level `defaultClient`.

Re-implementing this redirect logic in user code is harder than
necessary.

For context, this is the first of possibly many PRs that I'm contributing as part of my effort to integrate Vegeta with this library. See https://github.com/tsenart/vegeta/pull/500


erikdubbelboer:
Thanks!

PR #772, opened 2020-04-06
Fixed case when no hijacker is added but no-response is true

dgrr:
Hello,

It can be the case when the hijackerHandler is not set but hijackNoResponse is. If that case happens before this commit the server won't send any response to the client and then the connection will hang (client will be waiting, the server will be waiting too).

Thanks.


erikdubbelboer:
Nice find!

PR #778, opened 2020-04-11
Add configuration to not pre-read multipart form data

Markmerc:
Some servers want to treat the multipart form data as a binary blob, or at least not read/parse it until other decisions are made about the request.

### DIFF
--- a/http.go
+++ b/http.go
@@ -937,10 +937,10 @@ func (req *Request) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
 		return err
 	}
 
-	return req.readLimitBody(r, maxBodySize, false)
+	return req.readLimitBody(r, maxBodySize, false, true)
 }
 
-func (req *Request) readLimitBody(r *bufio.Reader, maxBodySize int, getOnly bool) error {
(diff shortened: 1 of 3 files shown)


http.go
```diff
@@ -979,24 +979,26 @@ func (req *Request) MayContinue() bool {
 //
 // If maxBodySize > 0 and the body size exceeds maxBodySize,
 // then ErrBodyTooLarge is returned.
-func (req *Request) ContinueReadBody(r *bufio.Reader, maxBodySize int) error {
+func (req *Request) ContinueReadBody(r *bufio.Reader, maxBodySize int, preReadMultipartForm bool) error {
```
erikdubbelboer, line 982: The only problem I see with this pull request is that this changes the API. Anyone using this function that upgrades fasthttp will have it break their build.

I guess you could make it a `...bool` so it becomes optional and backwards compatible.

Markmerc: switched to ...bool (though this seems kind of ugly)

Markmerc:
> I think the term `pre-read` is a bit wrong (also in the original code). Can you change it to `pre-parse`?

Fixed

erikdubbelboer:
Thanks!

Issue #783, opened 2020-04-17
Expect 100 Continue example/support?


mmacdermaid:
Does fasthttp actually support expect 100 continue workflow, or does it just automatically return the continue response and read the incoming body?

Not sure if this is the right section.
https://github.com/valyala/fasthttp/blob/master/server.go#L2036-L2067

There's a chance to further optimize my use case by reading the headers inside of my handler, and it would be nice to not process the bodies coming my direction.

If it does support the handling of the 100 continues, it would be great to see an example.

erikdubbelboer:
fasthttp automatically handles Expect 100-continue and reads the incoming body. It currently doesn't have a callback or something that you could use to reject the body. This shouldn't be too hard to implement in the lines you linked. A pull request is always welcome 😄

mmacdermaid:
@erikdubbelboer :+1: I will see if I can come up with something if I get a second.  I appreciate the response.

PR #787, opened 2020-04-21
allow the expect 100 continue workflow to deny requests

mmacdermaid:
Fixes https://github.com/valyala/fasthttp/issues/783

An okay solution right now which allows servers to deny reading the request body after reading the request headers via a server callback configuration.

A more long term solution later on might be to give full control to the user during their handler callback, but that would definitely require a lot more work and be more error prone for the user.

One unfortunate issue with this is, if you do any major work in processing headers before accepting or denying a request, you likely will have to re-process the headers on the other side.  It might be worth sticking a

````
func() interface[}
````
callback out of the deny which can be passed to a new handler, so people can pass state between the two.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1946,8 +1946,8 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 		connectionClose bool
 		isHTTP11        bool
 
-		reqReset      bool
-		deniedRequest bool
+		reqReset               bool
+		continueReadingRequest bool = true
 	)
 	for {
 		connRequestNum++
@@ -2059,7 +2059,7 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 
 			// Allow the ability to deny reading the incoming request body
 			if s.ContinueHandler != nil {
-				if deniedRequest = s.ContinueHandler(&ctx.Request.Header); deniedRequest {
+				if continueReadingRequest = s.ContinueHandler(&ctx.Request.Header); !continueReadingRequest {
 					if br != nil {
 						br.Reset(ctx.c)
 					}
@@ -2068,7 +2068,7 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 				}
 			}
 
-			if !deniedRequest {
+			if continueReadingRequest {
 				if bw == nil {
 					bw = acquireWriter(ctx)
 				}
@@ -2115,7 +2115,7 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 		ctx.time = time.Now()
 
 		// If a client denies a request the handler should not be called
-		if !deniedRequest {
+		if continueReadingRequest {
 			s.Handler(ctx)
 		}
 
--- a/server.go
+++ b/server.go
@@ -172,6 +172,17 @@ type Server struct {
 	// non zero RequestConfig field values will overwrite the default configs
 	HeaderReceived func(header *RequestHeader) RequestConfig
 
+	// DenyRequest is called after receiving the Expect 100 Continue Header
+	//
+	//https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: server.go [hunk 1]
```diff
@@ -2041,37 +2053,53 @@ func (s *Server) serveConn(c net.Conn) (err error) {
...
-			}
-			err = ctx.Request.ContinueReadBody(br, maxRequestBodySize, !s.DisablePreParseMultipartForm)
-			if (s.ReduceMemoryUsage && br.Buffered() == 0) || err != nil {
-				releaseReader(s, br)
-				br = nil
+			//Allow the ability to deny reading the incoming request body
+			if s.DenyRequest != nil {
+				if deniedRequest = s.DenyRequest(&ctx.Request.Header); deniedRequest {
+					if br != nil {
+						br.Reset(ctx.c)
```
mmacdermaid, line 2063: @erikdubbelboer After debugging the strange relationship between the headers and request bodies, this ended up being the "correct" way to reproduce whatever is currently happening inside of serveConn.  

Someone more familiar with the code definitely needs to weigh in on this piece here.

server.go
```diff
@@ -2041,37 +2053,53 @@ func (s *Server) serveConn(c net.Conn) (err error) {
...
+					}
+
+					ctx.SetStatusCode(StatusExpectationFailed)
```
mmacdermaid, line 2066: This seemed to be the most reasonable and clean way to allow setting the response code without calling flush etc. and letting the connection continue as it would had someone's handler been called and called the same thing.

server.go
```diff
@@ -172,6 +172,17 @@ type Server struct {
 	// non zero RequestConfig field values will overwrite the default configs
 	HeaderReceived func(header *RequestHeader) RequestConfig
 
+	// DenyRequest is called after receiving the Expect 100 Continue Header
+	//
+	//https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8.2.3
+	//https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.1.1
+	//Using DenyRequest a server can make decisioning on whether or not
+	//to read a potentially large request body based on the headers
```
erikdubbelboer, line 180: There are spaces missing after the `//`, in many other places as well.

server.go
```diff
@@ -172,6 +172,17 @@ type Server struct {
 	// non zero RequestConfig field values will overwrite the default configs
 	HeaderReceived func(header *RequestHeader) RequestConfig
 
+	// DenyRequest is called after receiving the Expect 100 Continue Header
```
erikdubbelboer, line 175: I think `DenyRequest` is not a clear name for this. Maybe just call it `ContinueHandler`?

server.go
erikdubbelboer: Is this really needed? Shouldn't the reader be empty at this point?

mmacdermaid: From what I have tested it is 100% needed.  If not the loop will come back around, attempt a call on ````if err = ctx.Request.Header.Read(br); err == nil {```` and fail with ````server_test.go:1705: Unexpected error from serveConn: error when reading request headers: EOF.````

I believe this is because br is not nil so
````
br = acquireReader(ctx)
````
Does not get called, which means it's already in a place somewhere but hasn't been fully read or reset.  The function that would normally change it's position or use it later on ````ctx.Request.ContinueReadBody```` hasn't been called in this case because the request was abandoned.

At least that's what I think?

server.go
mmacdermaid: :+1:

erikdubbelboer:
Thanks!

mmacdermaid:
:+1: Happy to help.

vivere-dally:
I want to use a mechanism to deny requests as well, but as far as I can understand this PR, this adds a ContinueHandler for the entire server, and it cannot be controlled per route? Is there any way to have some sort of control of denying 100-continue requests in a handler per route?

Issue #795, opened 2020-04-29
data race when setting an empty UserAgent


junxiang-viper:
Hi,

A data race condition was found with the following code when setting an empty UserAgent for a request header.

## code

```go
package main

import (
	"sync"
	"time"

	"github.com/valyala/fasthttp"
)

func main() {

	go fasthttp.ListenAndServe("0.0.0.0:8080",
		fasthttp.RequestHandler(func(ctx *fasthttp.RequestCtx) {}))

	cl := fasthttp.Client{}

	get := func(flip *bool) {

		req := fasthttp.AcquireRequest()
		resp := fasthttp.AcquireResponse()

		defer fasthttp.ReleaseRequest(req)
		defer fasthttp.ReleaseResponse(resp)

		req.URI().Update("http://localhost:8080")

		if *flip {
			req.Header.SetUserAgent("hello")
		} else {
			req.Header.SetUserAgent("")
		}

		*flip = !*flip

		cl.Do(req, resp)

		return
	}

	var wg sync.WaitGroup
	for i := 0; i < 2; i++ {
		wg.Add(1)
		go func() {
			defer wg.Done()
			flip := false
			for j := 0; j < 1000; j++ {
				get(&flip)
			}
			time.Sleep(time.Microsecond)
		}()
	}

	wg.Wait()
}

```
## run

```
go1.11.5 build -race -gcflags='all=-N -l' race.go
env GORACE="halt_on_error=1" ./race
```

## output
```
==================
WARNING: DATA RACE
Write at 0x000001964610 by goroutine 8:
  runtime.slicestringcopy()
      /usr/local/go/src/runtime/slice.go:256 +0x0
  github.com/valyala/fasthttp.(*RequestHeader).SetUserAgent()
      /users/jzhou/go/src/github.com/valyala/fasthttp/header.go:448 +0x278
  main.main.func2()
      /Users/jzhou/go/src/test/fasthttp-race/main.go:28 +0x1bd
  main.main.func3()
      /Users/jzhou/go/src/test/fasthttp-race/main.go:47 +0xc9

Previous read at 0x000001964610 by goroutine 7:
  runtime.slicecopy()
      /usr/local/go/src/runtime/slice.go:221 +0x0
  github.com/valyala/fasthttp.appendHeaderLine()
      /users/jzhou/go/src/github.com/valyala/fasthttp/header.go:1648 +0x726
  github.com/valyala/fasthttp.(*RequestHeader).AppendBytes()
[...]

PR #796, opened 2020-04-29
Copy to the req.Header.userAgent from the defaultUserAgent

junxiang-viper:
This avoids the req.Header.userAgent sharing an underlying array with the
global defaultUserAgent

This should fix the #795


erikdubbelboer:
Good find! Thanks!

PR #802, opened 2020-05-05
Make FS return a redirect for directories without trailing slash

### LINKED ISSUE #792: FS: redirect directories to have a trailing slash
Currently, I can access a file `/test/index.html` either through `/test/` or through `/test`. In the latter case, relative paths in the document will refer to completely different files, which makes this unappropriate in my opinion - `/test` should yield a 301 redirect to `/test/`. This will be a breaking change if someone uses something like `ctx.Request.SetRequestURI(path.Clean(p))` before calling the FSHandler though, so I guess making it an option would make sense?.

moqmar:
Fixes #792

### DIFF
--- a/fs_test.go
+++ b/fs_test.go
@@ -765,6 +765,6 @@ func TestServeFileDirectoryRedirect(t *testing.T) {
 	ctx.Response.Reset()
 	ServeFile(&ctx, "fs.go")
 	if ctx.Response.StatusCode() != StatusOK {
-		t.Fatalf("Unexpected status code %d for file '/.git'. Expecting %d.", ctx.Response.StatusCode(), StatusOK)
+		t.Fatalf("Unexpected status code %d for file '/fs.go'. Expecting %d.", ctx.Response.StatusCode(), StatusOK)
 	}
 }
--- a/fs_test.go
+++ b/fs_test.go
@@ -738,3 +738,33 @@ func TestServeFileContentType(t *testing.T) {
 		t.Fatalf("Unexpected Content-Type, expected: %q got %q", expected, resp.Header.ContentType())
 	}
 }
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks. Can you add a test case please?

moqmar:
I've added a test case and actually found an issue due to that :D

I'm not completely sure though if there are other functions making use of `fs` though that way (overriding the request URI before letting it serve a file or directory) - I haven't found any yet at least.

fs_test.go
```diff
@@ -738,3 +738,33 @@ func TestServeFileContentType(t *testing.T) {
...
+	ServeFile(&ctx, ".git/")
+	if ctx.Response.StatusCode() != StatusOK {
+		t.Fatalf("Unexpected status code %d for directory '/.git/' with trailing slash. Expecting %d.", ctx.Response.StatusCode(), StatusOK)
+	}
+
+	ctx.Request.Reset()
+	ctx.Response.Reset()
+	ServeFile(&ctx, "fs.go")
+	if ctx.Response.StatusCode() != StatusOK {
+		t.Fatalf("Unexpected status code %d for file '/.git'. Expecting %d.", ctx.Response.StatusCode(), StatusOK)
```
erikdubbelboer, line 768:
```suggestion
		t.Fatalf("Unexpected status code %d for file '/fs.go'. Expecting %d.", ctx.Response.StatusCode(), StatusOK)
```

kimkit:
@moqmar is it better to use origin path for redirect ?

https://github.com/valyala/fasthttp/blob/master/fs.go#L738

```go
ctx.RedirectBytes(append(ctx.Path(), '/'), StatusFound)
```

PR #817, opened 2020-05-27
Add custom retry decider for fail http requests

vural:
Hi,

The client will be able to use a custom retry decider with this pull request so the client will now be able to retry for non-idempotent requests.


erikdubbelboer:
Thanks!

PR #821, opened 2020-06-01
Update TechEmpower benchmark from 18 to 19 round

panjf2000:
Round 19 of the TechEmpower Framework Benchmarks project is now available and this is the detailed blog: https://www.techempower.com/blog/2020/05/28/framework-benchmarks-round-19/


erikdubbelboer:
Thanks!

PR #822, opened 2020-06-02
Support Windows SO_REUSEADDR

panjf2000:
Tested on Windows successfully, details shown below:
image


go.mod
```diff
@@ -8,4 +8,5 @@ require (
 	github.com/valyala/bytebufferpool v1.0.0
 	github.com/valyala/tcplisten v0.0.0-20161114210144-ceec8f93295a
 	golang.org/x/net v0.0.0-20200324143707-d3edc9973b7e
+	golang.org/x/sys v0.0.0-20200323222414-85ca7c5b95cd
```
erikdubbelboer, line 11: Why is this older version added?

panjf2000: Have no idea about it, I use `go mod tidy` and it added automatically, fixing it.

panjf2000: I know the reason, it is because the `golang.org/x/sys` already existed in go.sum before so when I use `go mod tidy`, it kept the existing one, now it has been upgraded to the latest version.

panjf2000:
@erikdubbelboer PTAL

erikdubbelboer:
Thanks!

PR #823, opened 2020-06-02
Added Gearbox

muhammednagy:
Gearbox is a web framework written in Go with a focus on high performance and memory optimization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,6 +281,7 @@ with fasthttp support:
   * [lu](https://github.com/vincentLiuxiang/lu)
   * [atreugo](https://github.com/savsgio/atreugo)
   * [Fiber](https://github.com/gofiber/fiber)
+  * [Gearbox](https://github.com/gogearbox/gearbox)
 
   Net/http code with simple ServeMux is trivially converted to fasthttp code:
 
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
To be honest I'm wondering if we should just add everything that uses fasthttp to the readme. This framework has only 22 commits, and doesn't seem to have any users except for the creators.

muhammednagy:
I understand.
You do have a valid point. however, we are working hard on improving it every day and we promise something good ( and have good progress so far ).
If you would like you could keep this issue open till we release our first stable release ( in a month or less )

proyb6:
@muhammednagy Good luck! From SG.

muhammednagy:
Hey @erikdubbelboer 
We have released a stable version now.
I know 30 commits is not much but please give it a chance. we are offering some pretty cool features like LRU caching for routes and the TST routing which already makes it faster than almost all current Go frameworks.

muhammednagy:
Hey @erikdubbelboer 
So we have checked and apparently the map is faster.
So we switched to map now and we are waiting for a newer proper benchmark.
However, even when we were using TST we were the fifth fastest framework.
I think you should give it a chance :)

erikdubbelboer:
Ok here you go!

muhammednagy:
Thank you! @erikdubbelboer

PR #827, opened 2020-06-04
Make the ErrNothingRead to be exposed.

wangfakang:
hello @erikdubbelboer We used the `ReadLimitBody` API directly to read client data, and to avoid this [issue](https://github.com/valyala/fasthttp/commit/598a52272abafde3c5bebd7cc1972d3bead7a1f7?spm=a2o8d.corp_prod_req_detail.0.0.18df5f1ahGBAbU) we need to check the error type, as below

```
……
                err := request.ReadLimitBody(conn.br, defaultMaxRequestBodySize)
                if err != nil {
                        _, errNothingRead := err.(fasthttp.ErrNothingRead)
                        if err != io.EOF && !errNothingRead {
                                // write error response
                                conn.conn.Write(buffer.NewIoBufferBytes(strErrorResponse))

                                // close connection with flush
                                conn.conn.Close(api.FlushWrite, api.LocalClose)
                        }
                   ……
                }
……
```

So need to exposed the `ErrNothingRead` erroy type, thx.


header.go
```diff
@@ -2247,7 +2247,11 @@ var (
 	errSmallBuffer = errors.New("small read buffer. Increase ReadBufferSize")
 )
 
-type errNothingRead struct {
+// ErrNothingRead is returned that means just a keep-alive connection
+// closing down either because the remote closed it or because
+// or a read timeout on our side.
+// Either way just close the connection and don't return any error response.
```
erikdubbelboer, line 2253:
```suggestion
// ErrNothingRead is returned when a keep-alive connection is closed,
// either because the remote closed it or because of a read timeout.
```

erikdubbelboer:
Thanks!

PR #828, opened 2020-06-05
Try TravisCI Windows

erikdubbelboer:


fs_test.go
```diff
@@ -740,6 +741,10 @@ func TestServeFileContentType(t *testing.T) {
 }
 
 func TestServeFileDirectoryRedirect(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.SkipNow()
+	}
```
erikdubbelboer, line 746: This test fails because of `/` vs `\`.

prefork/README.md
```diff
@@ -6,6 +6,8 @@ Preforks master process between several child processes increases performance, b
 
 **WARNING: using prefork prevents the use of any global state!. Things like in-memory caches won't work.**
 
+Note: Prefork is not supported on Windows.
```
erikdubbelboer, line 9: See: https://golang.org/pkg/os/exec/#Cmd.ExtraFiles

.travis.yml
```diff
@@ -9,14 +9,13 @@ dist: bionic
...
   - osx
+  - windows
 go:
   - tip
   - 1.14.x
   - 1.13.x
   - 1.12.x
   - 1.11.x
-  - 1.10.x
-  - 1.9.x
```
erikdubbelboer, line 19: These versions doesn't support everything on windows. They are also multiple years old so I think we can drop support.

prefork/README.md
savsgio: I think so with reuseport (reuseaddr) is supported

Issue #832, opened 2020-06-10
Race condition on read timeout causes deadlock in PipelineClient.


lithdew:
`(*PipelineClient).Do` deadlocks after the first request when a read timeout is set. 

This likely happens because `(*PipelineClient).worker` is not restarted after the first read timeout passes, which then causes the subsequent request to be buffered in `c.chW` on line 2286, and thus deadlock on `<-w.done` on line 2304.

The line numbers refer to `client.go`.

```go
func TestClientTest2(t *testing.T) {
	ln, err := net.Listen("tcp", ":0")
	require.NoError(t, err)

	req := fasthttp.AcquireRequest()
	defer fasthttp.ReleaseRequest(req)
	req.SetHost(ln.Addr().String())

	res := fasthttp.AcquireResponse()
	defer fasthttp.ReleaseResponse(res)

	var client fasthttp.PipelineClient
	client.Addr = ln.Addr().String()
	client.DialDualStack = true
	client.ReadTimeout = 1 * time.Second

	defer func() {
		require.NoError(t, ln.Close())
	}()

	attempts := 32
	go func() {
		for i := 0; i < attempts; i++ {
			_, err := ln.Accept()
			require.NoError(t, err)
		}
	}()

	for i := 0; i < attempts; i++ {
		fmt.Println(client.Do(req, res))
	}
}
```

lithdew:
The only viable solution I can think of is running a garbage collection worker in a goroutine. The worker will, every 10 seconds, for example, check the number of inflight requests.

If there are inflight requests that exist and the worker is not running, kill the inflight requests. This will unblock `(*PipelineClient).Do`.

erikdubbelboer:
@lithdew what do you think of this? https://github.com/valyala/fasthttp/pull/834

PR #834, opened 2020-06-12
Restart PipelineClient worker on error

erikdubbelboer:
Fixes https://github.com/valyala/fasthttp/issues/832


lithdew:
I will check more thoroughly later today, apologies for the delay - on a quick skim though, should chW be reset when there are pending messages available?

erikdubbelboer:
`chW` is only reset when there are no message pending.

lithdew:
> `chW` is only reset when there are no message pending.

Tested it on my end and it works; LGTM?

erikdubbelboer:
Thanks

PR #836, opened 2020-06-14
fix: adaptor content type

dink10:
In case of needs to create router for swagger adaptor from repository doesn't work because it always returns
`Content-Type: text/plain`

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -82,6 +82,7 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 		r.URL = rURL
 
 		var w netHTTPResponseWriter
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
I'm confused, why is this required? Is it because `net/http` handlers might depends on `net/http` setting a `Content-Type` header if the handler doesn't set it? In that case shouldn't we check if the handler set a `Content-Type` header and otherwise use [`http.DetectContentType`](https://golang.org/pkg/net/http/#DetectContentType) to set it like `net/http` does?

dink10:
Yes, possibly we need firstly check Content-Type and otherwise use http.DetectContentType. I'll check that

erikdubbelboer:
This is fixed in https://github.com/valyala/fasthttp/commit/ef51a7e590d2c076a95b0d1a59832833758ef251

PR #859, opened 2020-08-05
Add a api `DisableNoDefaultContentType` to disable add default content type.

wangfakang:
@valyala Add api DisableNoDefaultContentType to disable default conten type if no `Content-Type` header.


header.go
```diff
@@ -601,6 +601,11 @@ func (h *ResponseHeader) DisableNormalizing() {
 	h.disableNormalizing = true
 }
 
+// Disable add default contentype if no Content-Type header.
+func (h *ResponseHeader) DisableNoDefaultContentType() {
```
erikdubbelboer, line 605: This method needs to be named `DisableDefaultContentType`, the current method name implies exactly what it doesn't do.
Even better would be a method `SetNoDefaultContentType(boolean)` so you can change it instead of only setting it to true.

wangfakang: @erikdubbelboer yeah, thanks your review.

header.go
```diff
@@ -601,10 +601,17 @@ func (h *ResponseHeader) DisableNormalizing() {
 	h.disableNormalizing = true
 }
 
+// SetNoDefaultContentType sets header's 'noDefaultContentType' flag to the given value.
+//
+// Disable add 'defaultContentType' if no Content-Type header when 'noDefaultContentType' is true.
```
erikdubbelboer, line 606:
```suggestion
// SetNoDefaultContentType allows you to control if a default Content-Type header will be set (false) or not (true).
```

erikdubbelboer:
Thanks!

PR #865, opened 2020-08-17
Check lbClient is under penalty

koykov:
We have experienced with a trouble with LBClient. We have a lot of destination hosts and some of them is under deploy/maintenance sometimes and are inaccessible. We noticed that inaccessible servers receives most of the requests due to LBClient thinks they are so fast in comparison nodes (in fact they answers with error code quickly).
The problem is here https://github.com/valyala/fasthttp/blob/master/lbclient.go#L138
, since pending request is zero and penalty counter can be maximum 300, therefore LBClient.get() method decides that inaccessible node is less busy.
This PR just ignores nodes that still are under penalty.

### DIFF
--- a/lbclient.go
+++ b/lbclient.go
@@ -125,6 +125,9 @@ func (cc *LBClient) get() *lbClient {
 		minN := minC.PendingRequests()
 		minT := atomic.LoadUint64(&minC.total)
 		for _, c := range cs[off:] {
+			if c.underPenalty() {
+				continue
+			}
 			n := c.PendingRequests()
 			t := atomic.LoadUint64(&c.total)
 			if n < minN || (n == minN && t < minT) {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Wouldn't another way to fix this is to remove `maxPenalty` since that doesn't make sense to me to have anyways?

koykov:
From my view the better way is to ban the node during it is under penalty, but remove maxPenalty is also good idea. I will test this way as well.

erikdubbelboer:
But should a node really be removed for 3 seconds each time it has 1 failure? Maybe there should be a threshold.

koykov:
As I understand, the decision is node alive or not is a responsibility of HealthCheck() method. Of course, it should be safe of false-positive results.
My version of that method counters how many success/fail requests occurred during last second and then checks ratio of success/fail response codes. If it is less that a threshold, then method returns false and I expect that LBClient will ban that node. But it keeps node in rotation.

PR #889, opened 2020-10-26
httpproxy: improved auth barrier storage and full https proxy support

mkorolyov:
* Improved encoded auth barrier storage for http and https proxy urls
* Moved new proxy to the v2.
* Added full https proxy support.

### DIFF
--- a/fasthttpproxy/proxy_env.go
+++ b/fasthttpproxy/proxy_env.go
@@ -14,23 +14,23 @@ import (
 	"github.com/valyala/fasthttp"
 )
 
-// FasthttpHTTPDialer returns a fasthttp.DialFunc that dials using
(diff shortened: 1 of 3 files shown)


mkorolyov:
Hi @erikdubbelboer . Added v2 as you proposed. 

Also fixed https proxy support. Replaced mutex with atomics storage as it fits better with the current use-case.

erikdubbelboer (changes requested):
I didn't mean a v2 for this subpackage, I suggest we should change all methods at the same time if we ever release a v2 of fasthttp that isn't backwards incompatible.
With this your import path wout be `github.com/valyala/fasthttp/fasthttpproxy/v2` which doesn't make sense.

If you can remove the v2 and just keep the auth barrier changes we can merge those.

erikdubbelboer:
Thanks!

PR #890, opened 2020-10-27
remove unnecessary type conversion

mdaliyan:
I know doing this causes some kind of code duplication but I was writing `[]byte` directly into the response and converting`[]byte` to `string` and again `string` to `[]byte` caused a little process and delay for my large data. And I think reusing a code just adds inconvenience in this case.

### DIFF
--- a/http.go
+++ b/http.go
@@ -477,7 +477,8 @@ func (resp *Response) BodyWriteTo(w io.Writer) error {
 //
 // It is safe re-using p after the function returns.
 func (resp *Response) AppendBody(p []byte) {
-	resp.AppendBodyString(b2s(p))
+	resp.closeBodyStream()     //nolint:errcheck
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
`b2s` and `s2b` don't copy any data, they basically just change between string and slice pointers. Did you actually benchmark if this change had any effect?

mdaliyan:
Changing this had no effect on profiling (memory/cpu) results but influenced response time a very small amount in benchmarks with the load tests I had with [gatling](gatling.io) and I had fewer spears that cannot be proved it was because of this.

erikdubbelboer:
I guess it's slightly faster as you save some function calls and stack allocations. Also the functions you are manually inlining here are right below it so it is easy to follow and make sure both stay the same.

Thanks for the pull request!

PR #897, opened 2020-10-30
fixed default schema for for req url

mkorolyov:
There is a bug with req url schema settings in the proxy dialer adapter.


erikdubbelboer:
Thanks!

PR #905, opened 2020-11-06
add nil check for LocalAddr

shohi:
Add nil check for `TCPDialer`'s LocalAddr. The PR will improve the error message for dial failure when LocalAddr is not set.

```
# error message before
err: dial tcp4 <nil>->127.0.0.1:8080: connect: connection refused

# error message after
err: dial tcp4 127.0.0.1:8080: connect: connection refused
```

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -327,7 +327,11 @@ func (d *TCPDialer) tryDial(network string, addr *net.TCPAddr, deadline time.Tim
 		defer func() { <-concurrencyCh }()
 	}
 
-	dialer := net.Dialer{LocalAddr: d.LocalAddr}
+	dialer := net.Dialer{}
+	if d.LocalAddr != nil {
+		dialer = net.Dialer{LocalAddr: d.LocalAddr}
+	}
+
 	ctx, cancel_ctx := context.WithDeadline(context.Background(), deadline)
 	defer cancel_ctx()
 	conn, err := dialer.DialContext(ctx, network, addr.String())


tcpdialer.go
```diff
@@ -327,7 +327,11 @@ func (d *TCPDialer) tryDial(network string, addr *net.TCPAddr, deadline time.Tim
 		defer func() { <-concurrencyCh }()
 	}
 
-	dialer := net.Dialer{LocalAddr: d.LocalAddr}
+	dialer := net.Dialer{}
+	if d.LocalAddr != nil {
+		dialer = net.Dialer{LocalAddr: d.LocalAddr}
+	}
```
erikdubbelboer, line 333: I'm sorry but this doesn't change anything? Fields are `nil` by default so with `net.Dialer{}` `LocalAddr` will be `nil` but when `d.LocalAddr` is `nil` and you do `net.Dialer{LocalAddr: d.LocalAddr}` it will be exactly the same as doing `net.Dialer{}`.

erikdubbelboer (changes requested):

tcpdialer.go
shohi: Actually, they are not the name. This is because `net.Dialer.LocalAddr` is an interface type while the `d.LocalAddr` is a concrete type.  The statement `net.Dialer{LocalAddr: d.LocalAddr}` with  nil`d.LocalAddr`  will set the underlying type in `Dialer.LocalAddr` to `*net.TCPAddr` and the data to nil,  which lead nil check for `Dialer.LocalAddr` to fail.

There is a test sample on golang playground - https://play.golang.org/p/E1tlCOb32P

tcpdialer.go
```diff
@@ -327,7 +327,11 @@ func (d *TCPDialer) tryDial(network string, addr *net.TCPAddr, deadline time.Tim
 		defer func() { <-concurrencyCh }()
 	}
 
-	dialer := net.Dialer{LocalAddr: d.LocalAddr}
+	dialer := net.Dialer{}
+	if d.LocalAddr != nil {
+		dialer = net.Dialer{LocalAddr: d.LocalAddr}
```
erikdubbelboer, line 332: You are right, my mistake. But please simplify the code a bit to:
```suggestion
		dialer.LocalAddr = d.LocalAddr
```

erikdubbelboer (changes requested):

tcpdialer.go
shohi: Sure, I'll close this PR and file another one, it's my bad to delete the forked repo accidentally :-)

shohi:
will create another one

PR #911, opened 2020-11-08
Add request body streaming. Fixes #622

kirillDanshin:

PR #915, opened 2020-11-19
Add test cases for StreamRequestBody

kiyonlin:
Co-authored-by: Fenny fenny@gofiber.io

Related to #911

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -3357,25 +3357,29 @@ func TestStreamRequestBody(t *testing.T) {
 	part2 := strings.Repeat("2", 1<<20-1<<10)
 	contentLength := len(part1) + len(part2)
 	next := make(chan struct{})
-	done := make(chan struct{})
 
 	s := &Server{
 		Handler: func(ctx *RequestCtx) {
 			checkReader(t, ctx.RequestBodyStream(), part1)
 			close(next)
-			<-done
 			checkReader(t, ctx.RequestBodyStream(), part2)
 		},
-		ReadTimeout:       time.Second * 5,
+		DisableKeepalive:  true,
 		StreamRequestBody: true,
 	}
 
-	rw := &readWriter{}
-	rw.r.WriteString(fmt.Sprintf("POST /foo2 HTTP/1.1\r\nHost: aaa.com\r\nContent-Length: %d\r\nContent-Type: aa\r\n\r\n%s", contentLength, part1))
+	pipe := fasthttputil.NewPipeConns()
+	cc, sc := pipe.Conn1(), pipe.Conn2()
+	go func() {
+		//write headers and part1 body
+		if _, err := cc.Write([]byte(fmt.Sprintf("POST /foo2 HTTP/1.1\r\nHost: aaa.com\r\nContent-Length: %d\r\nContent-Type: aa\r\n\r\n%s", contentLength, part1))); err != nil {
+			t.Error(err)
+		}
+	}()
 
 	ch := make(chan error)
 	go func() {
-		ch <- s.ServeConn(rw)
+		ch <- s.ServeConn(sc)
 	}()
 
 	select {
@@ -3384,8 +3388,11 @@ func TestStreamRequestBody(t *testing.T) {
 		t.Fatal("part1 timeout")
 	}
 
-	rw.r.WriteString(part2)
-	close(done)
+	go func() {
+		if _, err := cc.Write([]byte(part2)); err != nil {
+			t.Error(err)
+		}
+	}()
 
 	select {
 	case err := <-ch:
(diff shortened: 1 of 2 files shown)


Fenny (approved):

server_test.go
```diff
@@ -3350,6 +3350,81 @@ func TestMaxBodySizePerRequest(t *testing.T) {
 	}
 }
 
+func TestStreamRequestBody(t *testing.T) {
```
erikdubbelboer, line 3353: This tests checks if the result is correct. But it doesn't test if the body is actually being streamed or still completely read in one go.

It would be a better test to send the header then send part of the body, then wait for the `Handler` to be called. And only then send the rest of the body.

erikdubbelboer (changes requested):

server_test.go
erikdubbelboer: Maybe even send it and read it in chunks with some delay in between to really test the streaming.

kiyonlin: Thx for the review! Will improve them soon.

Fenny:
@erikdubbelboer, do you mind reviewing the changes?

server_test.go
```diff
@@ -3350,6 +3350,110 @@ func TestMaxBodySizePerRequest(t *testing.T) {
...
+	go func() {
+		ch <- s.ServeConn(rw)
+	}()
+
+	select {
+	case err := <-ch:
+		if err != nil {
+			t.Error(err)
+		}
+	case <-time.After(100 * time.Millisecond):
```
erikdubbelboer, line 3427: This timeout should be higher as it is constantly failing on travis-ci.

server_test.go
```diff
@@ -3350,6 +3350,110 @@ func TestMaxBodySizePerRequest(t *testing.T) {
...
+		Handler: func(ctx *RequestCtx) {
+			checkReader(t, ctx.RequestBodyStream(), reqBody)
+		},
+		ReadTimeout:       time.Second * 5,
+		StreamRequestBody: true,
+	}
+
+	rw := &readWriter{}
+	rw.r.WriteString(fmt.Sprintf("POST /foo2 HTTP/1.1\r\nHost: aaa.com\r\nContent-Length: %d\r\nContent-Type: aa\r\n\r\n%s", contentLength, part1))
+
```
erikdubbelboer, line 3372: What I would like to see if for you to do a read on the request body here and see if it matches `part1`. Then `rw.r.WriteString(part2)` and then read the rest of the request body and see if it matches `part2`. That way you test if you can already read the partial body before the full body it send. Right now you're not testing that.

erikdubbelboer (changes requested):

server_test.go
```diff
@@ -3350,6 +3350,117 @@ func TestMaxBodySizePerRequest(t *testing.T) {
...
+	part1 := strings.Repeat("1", 1<<10)
+	part2 := strings.Repeat("2", 1<<20-1<<10)
+
+	contentLength := len(part1) + len(part2)
+
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			checkReader(t, ctx.RequestBodyStream(), part1)
+			checkReader(t, ctx.RequestBodyStream(), part2)
```
erikdubbelboer, line 3364: You're still just writing and reading both parts after each other. What I would like to see is like this: https://play.golang.org/p/r1s8_h52e3q
So you really only send part2 when you know part1 is already read.

erikdubbelboer (changes requested):

kiyonlin:
When I don’t use done chan, the test occasionally fails. I suspect it is related to reading and writing rw.r at the same time.

server_test.go
```diff
@@ -3350,6 +3350,133 @@ func TestMaxBodySizePerRequest(t *testing.T) {
...
+	part2 := strings.Repeat("2", 1<<20-1<<10)
+	contentLength := len(part1) + len(part2)
+	next := make(chan struct{})
+	done := make(chan struct{})
+
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			checkReader(t, ctx.RequestBodyStream(), part1)
+			close(next)
+			<-done
```
erikdubbelboer, line 3366: Instead of adding the `done` channel I think it's probably a better idea to add a `sync.Mutex` to `readWriter` to make its operations safe just like reading and writing on a normal socket would be.

erikdubbelboer (changes requested):

server_test.go
kiyonlin: I think we can't use `sync.Mutex`. If handler get lock first, then second part can't be sent, and the test case will timeout.
So we should use a more real `Conn`. Let me try it.

kiyonlin: Use `fasthttputil.NewPipeConns()` now.

server_test.go
```diff
@@ -3350,6 +3350,125 @@ func TestMaxBodySizePerRequest(t *testing.T) {
...
+	}
+
+	pipe := fasthttputil.NewPipeConns()
+	cc, sc := pipe.Conn1(), pipe.Conn2()
+	go func() {
+		//write headers and part1 body
+		if _, err := cc.Write([]byte(fmt.Sprintf("POST /foo2 HTTP/1.1\r\nHost: aaa.com\r\nContent-Length: %d\r\nContent-Type: aa\r\n\r\n%s", contentLength, part1))); err != nil {
+			t.Error(err)
+		}
+	}()
```
erikdubbelboer, line 3378: There is no reason to do this in a goroutine. If you move the `ch <- s.ServeConn(sc)` gorourine up to before this Write then you don't have to do this Write or the one below it in goroutines.

erikdubbelboer (changes requested):

server_test.go
kiyonlin: You're right, thanks!

erikdubbelboer (approved):

PR #918, opened 2020-11-22
Fixing deletion of headers/queryargs having multiple values.

anshul-jain-aws:
In a HTTP Request Headers/QueryArgs can have multiple values for the same key. Currently [delAllArgs](https://github.com/valyala/fasthttp/blob/master/args.go#L387-L399) doesn't delete all the headers/queryargs with the same key. This change fixes it by setting the index `i` to be checked next for deletion, correctly.

I have extracted relevant parts of code in [playground](https://play.golang.org/p/JmE2-Y8vz2k) to show the before and after changes.

### DIFF
--- a/args.go
+++ b/args.go
@@ -391,6 +391,7 @@ func delAllArgs(args []argsKV, key string) []argsKV {
 			tmp := *kv
 			copy(args[i:], args[i+1:])
 			n--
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #922, opened 2020-12-02
initial gnet non-blocking io support

calbot:
I created an initial implementation of fasthttp using gnet to handle the non-blocking IO. Seems to work well. Still haven't benchmarked it though. The changes don't touch any of the existing workings of fasthttp since a new GnetConn instance is used to adapt the serveConn(net.Conn) method to gnet.  So it shouldn't affect any existing servers which don't want to use Gnet. Separate ListenAndServeGnet function was created.

I don't know whether or not it handles closing connections properly or timeouts. I did see connections seem to eventually close.

Interested in what people think. 😄


server.go
```diff
@@ -1488,6 +1491,156 @@ func (s *Server) ListenAndServe(addr string) error {
...
+}
+
+// React fires when a connection sends the server data.
+// Call c.Read() or c.ReadN(n) within the parameter:c to read incoming data from client.
+// Parameter:out is the return value which is going to be sent back to the client.
+func (es *GnetHTTP) React(frame []byte, c gnet.Conn) (out []byte, action gnet.Action) {
+	log.Println("REACTING!!!!!")
+	gconn := &GnetConn{
+		gnetConn:    c,
+		readBuffer:  bytes.NewBuffer(frame),
```
erikdubbelboer, line 1610: This is not going to work if the http request is spread over multiple packets in which case `React` might be called with a partial request in `frame` and `serveConn` will error.

server.go
```diff
@@ -1583,44 +1581,73 @@ type GnetHTTP struct {
...
+	log.Printf("HTTP server is listening on %s (multi-cores: %t, loops: %d)\n",
+		srv.Addr.String(), srv.Multicore, srv.NumEventLoop)
+	return
+}
 
-	return gconn.writeBuffer.Bytes(), action
+//React -
+func (es *GnetHTTP) React(frame []byte, c gnet.Conn) (out []byte, action gnet.Action) {
+	// handle the request
+	out = frame
```
erikdubbelboer, line 1645: This doesn't make sense to me? The documentation says: `Parameter:out is the return value which is going to be sent back to the client.`

calbot: The frame is the result of the Decode method. All of the logic was moved to Decode since the gnet.Conn ResetBuffer can be used to allow support for partial messages. In a typical (I assume) case the serveConn should work the same and only be called once. But if there is only a partial message and an error is returned from serveConn it now waits for more data and retries Decode and serveConn. A bit redundant in that case since the internal state of serveConn can't be saved without changes.

yuriy-yarosh:
@calbot  well, I'm also considering using fasthttp with a custom [F-Stack](https://github.com/F-Stack/f-stack) port for [nff-go](https://github.com/intel-go/nff-go), alongside gnet...  I think it's better to prep a modular API because there more than one stack that''s better than go's net, especially when we're talking about [DPDK](https://www.dpdk.org/) related stuff.

Feel free to contact me via [telegram](https://t.me/yuriyyarosh) for further discussion.

PR #925, opened 2020-12-07
(header) do case insensitive lookup of cookie header value

### LINKED ISSUE #924: collectCookies should lookup cookie header field in a case insensitive way
We have a setup were we are using fasthttp behind a nginx load balancer talking HTTP/2 to browsers. 

Since HTTP/2 [specifies](https://tools.ietf.org/html/rfc7540#section-8.1.2) HTTP Header Fields to be lower case, the Header Fields are exchanged lower-case between browser and load balancer.

Collecting cookies fails in this case because of the current implementation (cookie Header is sent lower-case but fasthttp uses a canonicalized version for the lookup).

Proposal: Do a case insensitive lookup of the cookie header field

dkuerner:
Fixes #924

### DIFF
--- a/header.go
+++ b/header.go
@@ -1958,7 +1958,7 @@ func (h *RequestHeader) collectCookies() {
 
 	for i, n := 0, len(h.h); i < n; i++ {
 		kv := &h.h[i]
-		if bytes.Equal(kv.key, strCookie) {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks!

Issue #919, opened 2020-11-23
Inconsistency in results of RequestURI() when using URI().SetQueryString(...) and URI().QueryArgs() in combination


anshul-jain-aws:
We have two ways to manipulate QueryString of a HTTP Request present in FastHTTP via [SetQueryString()](https://github.com/valyala/fasthttp/blob/master/uri.go#L145-L148) and via [QueryArgs.Add/Set/Del](https://github.com/valyala/fasthttp/blob/master/args.go#L171)

When using `SetQueryString` we set a boolean [parsedQueryArgs](https://github.com/valyala/fasthttp/blob/master/uri.go#L147) as false so that next call can construct it again. But we don't honor it RequestURI() method leading to querysting being constructed from old queryArgs resulting in stale Querystring being returned.

```golang
func main() {
	req := fasthttp.AcquireRequest()
	
        req.URI().SetQueryString("q1=foo&q2=bar")
	fmt.Println(string(req.URI().RequestURI())) 

        // Set q3=bar with QueryArgs
	req.URI().QueryArgs().Add("q3", "baz")
	fmt.Println(string(req.URI().RequestURI()))

        // Now overwrite everything with SetQueryString
	req.URI().SetQueryString("q1=foo&q2=bar&q4=quux")
	fmt.Println(string(req.URI().RequestURI()))
}
```
```
Output
-------------
/?q1=foo&q2=bar
/?q1=foo&q2=bar&q3=baz
/?q1=foo&q2=bar&q3=baz        <<< Should be q1=foo&q2=bar&q4=quux

```

The fix is simple by adding ```u.parsedQueryArgs``` check in [uri.go](https://github.com/valyala/fasthttp/blob/master/uri.go#L409). I can send in a PR.

erikdubbelboer:
Sorry for the slow reply, yes please make a pull request, thanks!

anshul-jain-aws:
Merged.

PR #937, opened 2021-01-02
Use QueryString while constructing RequestURI instead of QueryArgs if parsedQueryArgs is set to false

anshul-jain-aws:
https://github.com/valyala/fasthttp/issues/919


erikdubbelboer:
Thanks!

PR #939, opened 2021-01-05
Worker pool should exit immidiatly while sleeping

makasim:
I have some tests with [goleak](https://github.com/uber-go/goleak) tool. From time to time It complains on workerpool goroutine. This PR fixes the issue.


erikdubbelboer:
I'm curious, why did you close the pull request? The change seems to make sense to me.

makasim:
1) It cause allocation 
2) There are other places that leak goroutines and it's not that easy to fix them

PR #943, opened 2021-01-12
feat: keep req/resp Body syntax correctly

kaka19ace:
Hi, i think error handles in Request::Body() and Response::Body() is not a good idea.  
they should be returns with another error type , the user could check error before handle the body.

we could provides BodyE() and Body():  
1. BodyE() returns ([]byte, error) , the caller can validate error by him/her self.
2. Body() // wrap BodyE(),  if error occurs, only returns nil slice

usage scene:
```
body, err := request.BodyE()
if err != nil {
    // early  return
    // log error
    return err
}
err = json.Unmarshal(body, &model)
// ...
```

the current usage: with error in body
usage scene:
```
body, err := request.Body()
// err in body, it's not json format
err = json.Unmarshal(body, &model) // not unexpected, and it's a waste operation.
// ...
```

### DIFF
--- a/http.go
+++ b/http.go
@@ -315,20 +315,29 @@ func (resp *Response) LocalAddr() net.Addr {
 	return resp.laddr
 }
 
-// Body returns response body.
-//
-// The returned body is valid until the response modification.
-func (resp *Response) Body() []byte {
+// BodyOrErr returns response (body, nil) if success.
+// when error occurs, it returns nil, err.
+func (resp *Response) BodyOrErr() ([]byte, error) {
 	if resp.bodyStream != nil {
 		bodyBuf := resp.bodyBuffer()
 		bodyBuf.Reset()
 		_, err := copyZeroAlloc(bodyBuf, resp.bodyStream)
 		resp.closeBodyStream() //nolint:errcheck
 		if err != nil {
-			bodyBuf.SetString(err.Error())
+			return nil, err
 		}
 	}
-	return resp.bodyBytes()
+	return resp.bodyBytes(), nil
+}
+
+// Body returns response body.
+// if error occurs, the error message would be put into body.
+func (resp *Response) Body() []byte {
+	body, err := resp.BodyOrErr()
+	if err != nil {
+		return []byte(err.Error())
+	}
+	return body
 }
 
 func (resp *Response) bodyBytes() []byte {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Good idea, but I don't think `BodyE` is a good name. Maybe a better name would be `BodyOrErr`? I feel like `BodyErr` sounds like getting a body error and not just the body or the error.

kaka19ace:
> Good idea, but I don't think `BodyE` is a good name. Maybe a better name would be `BodyOrErr`? I feel like `BodyErr` sounds like getting a body error and not just the body or the error.

the style is borrowed from https://github.com/spf13/cast/blob/master/caste.go#L74:6. ToBoolE(v interface{}) (bool, error).  
Because is current signature Body() is used,  we need another to define func() ([]byte, error) . 

BodyOrError sounds good, i re-commit with new name.

kaka19ace:
> Good idea, but I don't think `BodyE` is a good name. Maybe a better name would be `BodyOrErr`? I feel like `BodyErr` sounds like getting a body error and not just the body or the error.

hi, i've added BodyOrErr method to Request/Response, could you please review the pull request again? :) @erikdubbelboer

http.go
```diff
@@ -315,20 +315,26 @@ func (resp *Response) LocalAddr() net.Addr {
...
 		}
 	}
-	return resp.bodyBytes()
+	return resp.bodyBytes(), nil
+}
+
+// Body returns response body.
+// if get body failed, returns nil.
+func (resp *Response) Body() []byte {
+	body, _ := resp.BodyOrErr()
```
erikdubbelboer, line 336: Now this function isn't backwards compatible anymore as it doesn't do the `bodyBuf.SetString(err.Error())` anymore. Can you change it back to how it was? I don't think it's easy to use `BodyOrErr` for that so I would just leave the function as it was.

erikdubbelboer (changes requested):

http.go
kaka19ace: i recommit again,  the Body() function returns body with err msg when error happens.

http.go
```diff
@@ -315,20 +315,29 @@ func (resp *Response) LocalAddr() net.Addr {
...
-	return resp.bodyBytes()
+	return resp.bodyBytes(), nil
+}
+
+// Body returns response body.
+// if error occurs, the error message would be put into body.
+func (resp *Response) Body() []byte {
+	body, err := resp.BodyOrErr()
+	if err != nil {
+		return []byte(err.Error())
```
erikdubbelboer, line 338: I'm afraid that this is not the same as `bodyBuf.SetString(err.Error())`. `bodyBuf.SetString` also modifies the `Response` body for if it gets used afterwards. I think the implementation of `Body()` should just stay the same as before.

erikdubbelboer (changes requested):

ibraheemdev:
`TryBody`?

PR #946, opened 2021-01-19
NetHTTPResponseWriter support hijackerConn for fasthttp adaptor

yoyofx:
if the request connection upgrade the hijacker connection, that handler needed direct return after  ServeHTTP handler process .

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -81,8 +81,11 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 		}
 		r.URL = rURL
 
-		var w netHTTPResponseWriter
+		w := NetHTTPResponseWriter{Ctx: ctx}
 		h.ServeHTTP(&w, r.WithContext(ctx))
+		if w.IsHijackerConn {
+			return
+		}
 
 		ctx.SetStatusCode(w.StatusCode())
 		haveContentType := false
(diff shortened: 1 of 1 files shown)


yoyofx:
> What is `hijackerConn`? Where is `IsHijackerConn` set to true? Why put `Ctx` in there if you're not going to use it?
 
1. because  the netHTTPResponseWriter is not implement Hijacker interface , that is call name  'hijackerConn' .
2. I  need call ctx.Hijack(func(netConn net.Conn) func in  http.Handler , upgrade it to websocket.
3. IsHijackerConn call adpter handler, that's 'Hijack' not needed anything. and then hold the net.conn in  http.Handler.

I used to this example： https://github.com/yoyofx/yoyogo/blob/dev/web/websocketupgrader.go#L49

yoyofx:
readme:
http.Hijacker interface, the Hijack() (net.Conn, *bufio.ReadWriter, error)  is sync func.

ctx.Hijack(func(netConn net.Conn) is callback func ,that is async.
In the 'adpter' .if don't direct return on after call ctx.Hijjack func ,  fasthttp is async process fasthttp.RequestHandler.
and then the net.Conn will not wait to be used.

erikdubbelboer:
I don't understand, with your pull `NetHTTPResponseWriter` still doesn't implement `http.Hijacker`. If that if your desire then why not implement it? Is there code missing in your pull maybe? Right now your just adding fields to the struct and not using them? Can you maybe add a test case that fails before your pull and succeeds after?

oSethoum:
@erikdubbelboer Hi, am having the same issue when using the fasthttpadaptor 
image
how can I fix it?

erikdubbelboer:
I tried implementing this, but it's impossible with the way hijacking works in net/http and fasthttp. They are just incompatible.

oSethoum:
@erikdubbelboer Thanks for trying.

PR #950, opened 2021-01-22
Do not start connsCleaner on SetConnectionClosed requests.

moredure:


client.go
```diff
@@ -1459,7 +1459,7 @@ func (c *HostClient) acquireConn(reqTimeout time.Duration) (cc *clientConn, err
 		if c.connsCount < maxConns {
 			c.connsCount++
 			createConn = true
-			if !c.connsCleanerRun {
+			if !(c.connsCleanerRun || connectionClose) {
```
erikdubbelboer, line 1462:
```suggestion
			if !c.connsCleanerRun && !connectionClose {
```

erikdubbelboer (changes requested):
Good optimization, just one small change request to make the condition a bit cleaner.

client.go
moredure: Just de morgana rule) I think readability depends on reader in this case) But ok, it's not a problem.

erikdubbelboer:
Thanks!

moredure:
Thank you too!

PR #954, opened 2021-01-24
Improve TLS handshake timeout handling.

moredure:


moredure:
Maybe we need some more granular use of SetReadDeadline and SetWriteDeadline to wrap single call of Handshake with some values and then update SetRead/WriteTimeout with values from HostClient and only in case of TLS connections...

moredure:
Smth like that?
```go
func tlsClientHandshake(rawConn net.Conn, tlsConfig *tls.Config, handshakeTimeout time.Duration) (net.Conn, error) {
	conn := tls.Client(rawConn, tlsConfig)
	if err := conn.SetDeadline(time.Now().Add(handshakeTimeout)); err != nil {
		rawConn.Close()
		return nil, err
	}
	if err := conn.Handshake(); err != nil {
		rawConn.Close()
		return nil, err
	}
	return conn, nil
}
```

moredure:
At least looks promising, because of no new goroutines for tlsHandshakeTimeout.)

moredure:
Or smth like that:
```go
tlsConn := tls.Client(conn, tlsConfig)
		if handshakeTimeout > 0 {
			if err := tlsConn.SetDeadline(time.Now().Add(handshakeTimeout)); err != nil {
				conn.Close()
				return nil, err
			}
			if err := tlsConn.Handshake(); err != nil {
				conn.Close()
				return nil, err
			}
			if err := tlsConn.SetDeadline(time.Time{}); err != nil {
				conn.Close()
				return nil, err
			}
		}
		return tlsConn, nil
```

client.go
```diff
@@ -687,6 +693,11 @@ type HostClient struct {
 	//
 	// By default request write timeout is unlimited.
 	WriteTimeout time.Duration
+	
+	// Maximum duration for full handshake reading and writing.
+	//
+	// By default request handshake timeout is unlimited.
+	HandshakeTimeout time.Duration
```
erikdubbelboer, line 700: I'm not sure if this is a good name. For http it now is the dial timeout, for https it is the dial timeout + handshake timeout.

erikdubbelboer: Also it should be backwards compatible which means it should default to `WriteTimeout` which was used before.

erikdubbelboer (changes requested):

client.go
moredure: Yes, I believe so, I just used such a naming in my fork, but what do you think we should name it? Or how to distinguish handshake read+write timeout , from write timeout? Or we should treat fasthttp writetimeout and readtimeout not only to request read/write but to handshake read/write operations as well?

erikdubbelboer: I think we should just keep using `WriteTimeout` for this as doing the handshake is kinda part of writing the request.

moredure: But then subtract time spent on handshake from writeTimeout ?

erikdubbelboer: That's maybe also not the right solution. Maybe we should call it DialTimeout instead cause that's more what it is. It is also the timeout passed to the dialer for the whole dial operation. And I think when `DialTimeout` is 0 it should default to `WriteTimeout` so current users don't suddenly loose their timeout when the update fasthttp.

PR #962, opened 2021-02-07
Release objects correctly in clientGetURLDeadline

kiyonlin:
Now `chv` and `req` won't be released if timeout case is hit. I move similar logic from [clientDoDeadline](https://github.com/valyala/fasthttp/blob/ed1cedd51e07e6ddedfe18d860bb9e837f133c26/client.go#L1121-L1203) to make sure they will be putted back to the pool correctly.

### DIFF
--- a/client.go
+++ b/client.go
@@ -856,8 +856,6 @@ func clientGetURLDeadline(dst []byte, url string, deadline time.Time, c clientDo
 	}
 	ch = chv.(chan clientURLResponse)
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!. There was still one race condition that needed to be fixed so I made a commit based on your pull request: https://github.com/valyala/fasthttp/commit/1494fdc53b7e30cf72ffe01e8f882a255997471d

PR #637, opened 2019-08-18
A faster s2b function

ZYunH:
A faster s2b function

The new function just use the return stack space to store the final value, without the allocation of a temporary struct. s2bFast is 100% faster if the Go compiler doesn't use deeper optimization in some code, if there is deeper optimization (inline for example), s2bFast is 5%~15% faster. The s2bFast is always same as s2b in any situations, both functions identically from the perspective of caller. You can see this in ASM, the new one has a smaller stack space and without locals.

Environment:  go1.12.7 darwin/amd64
(The go code)

> func s2b(s string) []byte {
	sh := (*StringHeader)(unsafe.Pointer(&s))
	bh := SliceHeader{
		Data: sh.Data,
		Len:  sh.Len,
		Cap:  sh.Len,
	}
	return *(*[]byte)(unsafe.Pointer(&bh))
}

> func s2bFast(s string) (b []byte) {
	bh := (*SliceHeader)(unsafe.Pointer(&b))
	sh := *(*StringHeader)(unsafe.Pointer(&s))
	bh.Data = sh.Data
	bh.Len = sh.Len
	bh.Cap = sh.Len
	return b
}

(In ASM)
> "".s2b STEXT nosplit size=88 args=0x28 locals=0x20
	0x0000 00000 (main.go:33)	TEXT	"".s2b(SB), NOSPLIT|ABIInternal, $32-40
	0x0000 00000 (main.go:33)	SUBQ	$32, SP
	0x0004 00004 (main.go:33)	MOVQ	BP, 24(SP)
	0x0009 00009 (main.go:33)	LEAQ	24(SP), BP
	0x000e 00014 (main.go:33)	FUNCDATA	$0, gclocals·9fad110d66c97cf0b58d28cccea80b12(SB)
	0x000e 00014 (main.go:33)	FUNCDATA	$1, gclocals·7d2d5fca80364273fb07d5820a76fef4(SB)
	0x000e 00014 (main.go:33)	FUNCDATA	$3, gclocals·ebb0e8ce1793da18f0378b883cb3e122(SB)
	0x000e 00014 (main.go:33)	FUNCDATA	$4, "".s2b.stkobj(SB)
	0x000e 00014 (main.go:35)	PCDATA	$2, $0
	0x000e 00014 (main.go:35)	PCDATA	$0, $0
	0x000e 00014 (main.go:35)	XORPS	X0, X0
	0x0011 00017 (main.go:35)	MOVUPS	X0, "".bh(SP)
[...]

PR #967, opened 2021-02-12
fix s2b go vet warning

zhangyunhao116:
The `go vet` will complain about `sh := *(*reflect.StringHeader)(unsafe.Pointer(&s))` (possible misuse of reflect.StringHeader). The extra `*` is added because it generate less code compared to `sh := (*reflect.StringHeader)(unsafe.Pointer(&s))`, see details in https://godbolt.org/z/xG3hbn   

They are actually do the same thing for now(almost same speed, previous implementation 1 CPU tick faster maybe), but if `go vet` complain about this, i think we can remove the extra `*` to prevent later Go release has inconsistent behavior(escape analysis etc).

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -343,7 +343,7 @@ func s2b(s string) (b []byte) {
 	/* #nosec G103 */
 	bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
 	/* #nosec G103 */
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks! Had to look up why we did that in the first place but it seems we talked about this last time you modified it: https://github.com/valyala/fasthttp/pull/637 I agree that the 1 CPU tick faster probably isn't worth it.

PR #969, opened 2021-02-12
Added Protocol() as a replacement of hardcoded strHTTP11

dgrr:


erikdubbelboer (changes requested):
`proto` needs to be reset in `RequestHeader.resetSkipNormalize`.
It needs to be copied in `RequestHeader.CopyTo`.
`RequestHeader.AppendBytes` still does `dst = append(dst, strHTTP11...)`.
`RequestHeader.IsHTTP11()` is now incorrect.
`RequestHeader.noHTTP11` should maybe be removed and changed to a comparison of `proto` and `strHTTP11`.
`proto` need to be updated in `RequestHeader.parseFirstLine`.

dgrr:
Hello @erikdubbelboer 

That's the thing... The library (fasthttp) only supports HTTP/1.1, but for example my HTTP/2 implementation uses the Request and Response from RequestCtx to support fasthttp.

In that topic, functions like `RequestHeader.parseFirstLine` will never be reached because the RequestCtx will be executed inside my library, and fasthttp will not handle that thanks to [NextProto](https://pkg.go.dev/github.com/valyala/fasthttp#Server.NextProto) which is executed like [this](https://github.com/valyala/fasthttp/blob/master/server.go#L1986)

dgrr:
* [x]  `proto` needs to be reset in `RequestHeader.resetSkipNormalize`.
* [x]  `proto` needs to be copied in `RequestHeader.CopyTo`.
* [x]  `RequestHeader.AppendBytes` still does `dst = append(dst, strHTTP11...)`.
* [x]  `RequestHeader.IsHTTP11()` is now incorrect.
* [x]  `RequestHeader.noHTTP11` should maybe be removed and changed to a comparison of `proto` and `strHTTP11`. **On every setProtocol the comparison changes**
* [ ]  `proto` need to be updated in `RequestHeader.parseFirstLine`.

dgrr:
Also, the behaviour of noHTTP11 in fasthttp serveConn is to use HTTP/1.0, instead of HTTP/1.1, so I don't know if we should modify that var.

erikdubbelboer:
I think we should modify it in `RequestHeader.parseFirstLine`. Right now calling `Protocol()` on an incomming HTTP 1/0 request would return a wrong value. What would be the harm in setting it correctly there?

dgrr:
What do you think about the last commit?

erikdubbelboer:
LGTM! Thanks!

PR #970, opened 2021-02-13
Streaming fixes

erikdubbelboer:
- Allow DisablePreParseMultipartForm in combination with
StreamRequestBody.
- Support streaming into MultipartForm instead of reading the whole body
  first.
- Support calling ctx.PostBody() when streaming is enabled.

### DIFF
--- a/http.go
+++ b/http.go
@@ -3,6 +3,7 @@ package fasthttp
 import (
 	"bufio"
 	"bytes"
+	"compress/gzip"
 	"encoding/base64"
 	"errors"
 	"fmt"
@@ -345,6 +346,15 @@ func (req *Request) bodyBytes() []byte {
 	if req.bodyRaw != nil {
 		return req.bodyRaw
 	}
+	if req.bodyStream != nil {
+		bodyBuf := req.bodyBuffer()
+		bodyBuf.Reset()
+		_, err := copyZeroAlloc(bodyBuf, req.bodyStream)
+		req.closeBodyStream() //nolint:errcheck
+		if err != nil {
+			bodyBuf.SetString(err.Error())
+		}
+	}
 	if req.body == nil {
 		return nil
 	}
@@ -630,14 +640,6 @@ func (req *Request) SwapBody(body []byte) []byte {
 func (req *Request) Body() []byte {
 	if req.bodyRaw != nil {
 		return req.bodyRaw
-	} else if req.bodyStream != nil {
-		bodyBuf := req.bodyBuffer()
-		bodyBuf.Reset()
-		_, err := copyZeroAlloc(bodyBuf, req.bodyStream)
-		req.closeBodyStream() //nolint:errcheck
-		if err != nil {
-			bodyBuf.SetString(err.Error())
-		}
 	} else if req.onlyMultipartForm() {
 		body, err := marshalMultipartForm(req.multipartForm, req.multipartFormBoundary)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


http.go
```diff
@@ -345,6 +346,15 @@ func (req *Request) bodyBytes() []byte {
...
 	}
+	if req.bodyStream != nil {
+		bodyBuf := req.bodyBuffer()
+		bodyBuf.Reset()
+		_, err := copyZeroAlloc(bodyBuf, req.bodyStream)
+		req.closeBodyStream() //nolint:errcheck
+		if err != nil {
+			bodyBuf.SetString(err.Error())
+		}
+	}
```
erikdubbelboer, line 357: Without this various functions such as `ctx.PostBody()` wouldn't work when streaming was enabled.

http.go
```diff
@@ -814,24 +816,43 @@ func (req *Request) MultipartForm() (*multipart.Form, error) {
...
 		}
-	} else if len(ce) > 0 {
-		return nil, fmt.Errorf("unsupported Content-Encoding: %q", ce)
-	}
 
-	f, err := readMultipartForm(bytes.NewReader(body), req.multipartFormBoundary, len(body), len(body))
-	if err != nil {
-		return nil, err
+		mr := multipart.NewReader(bodyStream, req.multipartFormBoundary)
+		req.multipartForm, err = mr.ReadForm(8 * 1024)
```
erikdubbelboer, line 834: Don't use `AppendGunzipBytes` and `readMultipartForm` which requires the whole body to be in memory. Using `gzip.Reader` and `multipart.NewReader` we can stream the body and allow `multipart` to keep big files on disk and reduce memory usage.

http.go
```diff
@@ -1033,39 +1057,7 @@ func (req *Request) readBodyStream(r *bufio.Reader, maxBodySize int, getOnly boo
...
-		return nil
-	}
-
-	bodyBuf := req.bodyBuffer()
-	bodyBuf.Reset()
-
-	req.bodyStream = acquireRequestStream(bodyBuf, r, contentLength)
-
-	return nil
+	return req.ContinueReadBodyStream(r, maxBodySize, preParseMultipartForm)
```
erikdubbelboer, line 1060: Change this to be similar to `readBody` which also just calls `ContinueReadBody` instead of having the logic inlined.

http.go
```diff
@@ -1170,21 +1162,15 @@ func (req *Request) ContinueReadBodyStream(r *bufio.Reader, maxBodySize int, pre
...
 	bodyBuf.B, err = readBodyWithStreaming(r, contentLength, maxBodySize, bodyBuf.B)
-	bodyBufLen := maxBodySize
-	if contentLength < maxBodySize {
-		bodyBufLen = cap(bodyBuf.B)
-	}
 	if err != nil {
 		if err == ErrBodyTooLarge {
 			req.Header.SetContentLength(contentLength)
 			req.body = bodyBuf
-			req.bodyRaw = bodyBuf.B[:bodyBufLen]
```
erikdubbelboer, line 1181: No need to store the unread part in `bodyRaw` as it wasn't used and was causing bugs in functions that were checking if `bodyRaw != nil` and assume the whole body is in there.

http.go
```diff
@@ -1936,24 +1921,27 @@ func readBody(r *bufio.Reader, contentLength int, maxBodySize int, dst []byte) (
...
-	switch {
-	case contentLength >= 0 && maxBodySize >= contentLength:
-		readN := maxBodySize
-		if contentLength > 8*1024 {
-			readN = 8 * 1024
-		}
+
+	readN := maxBodySize
+	if readN > contentLength {
+		readN = contentLength
```
erikdubbelboer, line 1933: Without this we would wait forever on bodies smaller than 8kb as we would block on reading from a connection that never had any more data.

server_test.go
```diff
@@ -3413,8 +3425,8 @@ func TestMaxBodySizePerRequest(t *testing.T) {
 func TestStreamRequestBody(t *testing.T) {
 	t.Parallel()
 
-	part1 := strings.Repeat("1", 1<<10)
-	part2 := strings.Repeat("2", 1<<20-1<<10)
+	part1 := strings.Repeat("1", 1<<15)
```
erikdubbelboer, line 3428: Had to increase this size. If the body is big enough we always block on chunks of 8kb. So when `part1` is smaller than 8kb we get a read timeout.

streaming.go
```diff
@@ -45,14 +45,23 @@ func (rs *requestStream) Read(p []byte) (int, error) {
 	}
 	var n int
 	var err error
-	if int(rs.prefetchedBytes.Size()) > rs.totalBytesRead {
+	prefetchedSize := int(rs.prefetchedBytes.Size())
+	if prefetchedSize > rs.totalBytesRead {
+		left := prefetchedSize - rs.totalBytesRead
+		if len(p) > left {
+			p = p[:left]
```
erikdubbelboer, line 52: Never try to read more into `p` than we have left. Otherwise a big `p` would block on reading data that would never come.

streaming_test.go
```diff
@@ -6,10 +6,100 @@ import (
 	"io/ioutil"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/valyala/fasthttp/fasthttputil"
 )
 
+func TestStreamingPipeline(t *testing.T) {
```
erikdubbelboer, line 14: Streaming wasn't compatible with pipelining as the stream would already read data from the next request which would then fail the next request.

kirillDanshin (approved):

PR #975, opened 2021-02-22
Fix Client ms cleaner

kiyonlin:
Currently we only start a gorouting for cleaning Client.m, but no one for Client.ms.


erikdubbelboer:
Thanks! I have no idea how this went unnoticed for so long.

moredure:
@kiyonlin Great job!

moredure:
But it's not a fix) there is an another issue) 

if len(c.m) == 0 {
			mustStop = true				mustStop = true
		}			}
		etc (

moredure:
replace all c.m usages

moredure:
delete as well

moredure:
Opened https://github.com/valyala/fasthttp/pull/979/files

PR #978, opened 2021-02-22
byteSliceReader compatibility with flate.Reader interface

moredure:
Add compatibility with flate.Reader to reduce allocations of bufio.NewReader structs backed by default size slices of 4096 bytes for WriteGunzip, WriteDeflate.

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -342,6 +342,15 @@ func (r *byteSliceReader) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+func (r *byteSliceReader) ReadByte() (byte, error) {
+	if len(r.b) == 0 {
+		return 0, io.EOF
+	}
+	n := r.b[0]
+	r.b = r.b[1:]
+	return n, nil
+}
+
 func acquireStacklessDeflateWriter(w io.Writer, level int) stackless.Writer {
 	nLevel := normalizeCompressLevel(level)
 	p := stacklessDeflateWriterPoolMap[nLevel]


dgrr:
This is going to increase allocation. It'd be better to use: `r.b = append(r.b[:0], r.b[1:]...)`. Right?

erikdubbelboer:
I don't think this is going to cause an allocation. The same backing array will be used and the data pointer in slice will just be overwritten.

I had no idea `flate.Reader` did this, good catch!

dgrr:
Right! Though the content was appended at the end of the slice.

moredure:
I'd also want to contribute change with sharing sync.Pools of bufio.Readers, across Client and HostClient, because of my code accessing tons of different hostnames concurrently (my fork already doing this, but not sure it will fit well here, because of pointers to sync.Pools breaks compatibility, maybe need to set check whether fasthttp.Client provides it in hostclient field, etc...).

moredure:
Also looks like ```go

type byteSliceReader []byte

func (z *byteSliceReader) Read(p []byte) (int, error) {
	if len(*z) == 0 {
		return 0, io.EOF
	}
	n := copy(p, *z)
	*z = (*z)[n:]
	return n, nil
}

func (z *byteSliceReader) ReadByte() (byte, error) {
	if len(*z) == 0 {
		return 0, io.EOF
	}
	n := (*z)[0]
	*z = (*z)[1:]
	return n, nil
}
```

Should perform a few nanoseconds better and one allocation less than a struct I guess

**Comment by dgrr** [comment]
@moredure I think few nanos doesn't make that much of a difference, but I think cache-wise should be better right? I don't know exactly how Golang manages memory, but having a pointer to a struct looks like it makes the computer perform 2 lookups (?). One lookup for the pointer where the structure is and then another lookup for the pointer to the slice.
If that assertion is correct, I imagine having just the slice makes more sense.

**Comment by moredure** [comment]
Anyway there will be a need to get address (&) of slicer instance to allow passing to interface function parameters (io.Reader, flate.Reader, etc)

**Comment by moredure** [comment]
If flate readers using ReadByte frequently maybe bytes.NewReader is better) than custom solution)

**Comment by moredure** [comment]
In the end
```go

type byteSliceReader2 struct {
	b []byte
	i int64
}

func (r *byteSliceReader2) Read(b []byte) (n int, err error) {
	if r.i >= int64(len(r.b)) {
		return 0, io.EOF
	}
	n = copy(b, r.b[r.i:])
	r.i += int64(n)
	return n, nil
}

func (r *byteSliceReader2) ReadByte() (byte, error) {
	if r.i >= int64(len(r.b)) {
		return 0, io.EOF
	}
	b := r.b[r.i]
	r.i++
	return b, nil
}
```
performs best)

erikdubbelboer:
I think the current change is good enough and is in line with how the rest works. If we want to try and speed up byteSliceReader as a whole that should be a different pull request.

PR #979, opened 2021-02-24
Fix mCleaner not cleaning Ms fix.

moredure:
Fixing fix https://github.com/valyala/fasthttp/pull/975

### DIFF
--- a/client.go
+++ b/client.go
@@ -544,10 +544,10 @@ func (c *Client) mCleaner(m map[string]*HostClient) {
 			v.connsLock.Unlock()
 
 			if shouldRemove {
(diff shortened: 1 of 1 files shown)


kiyonlin:
Well done! I can't believe I missed other 2 `c.m` 😢

moredure:
Yeap, I also missed them at first) But understanding comes later))

moredure:
I hope that we don't need fix to fix the fix)

erikdubbelboer:
Good catch, not sure how I missed those either, I remember Ctrl+f'ing `c.m`.

Issue #976, opened 2021-02-22
Export info about HostClient.connsCount


kiyonlin:
I'm [implementing Fiber http client](https://github.com/gofiber/fiber/pull/1177) on top of fasthttp HostClient, and I want to reuse HostClient too. But I can't clean them because I have no access to [connsCount](https://github.com/valyala/fasthttp/blob/master/client.go#L740). So can we export this information?

PS I can make the PR.

kiyonlin:
@erikdubbelboer Could you please take a look 😃 ?

erikdubbelboer:
Can you please make a pull request to add a function that returns this number? Thanks.

kirillDanshin:
maybe it's better to provide a Reset() method to allow reusing HostClient? 🤔

kiyonlin:
> maybe it's better to provide a Reset() method to allow reusing HostClient? 🤔 

But when to call Reset() if there are current connections?

PR #992, opened 2021-03-09
Provide host client pool and Reset method to reuse HostClient instance

kiyonlin:
Implement [this](https://github.com/valyala/fasthttp/issues/976#issuecomment-785440286)

> maybe it's better to provide a Reset() method to allow reusing HostClient? 🤔

@erikdubbelboer @kirillDanshin Please take a look.


client.go
```diff
@@ -1881,6 +1881,67 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
...
+	hc.RetryIf = nil
+	hc.clientName = atomic.Value{}
+	hc.lastUseTime = 0
+	hc.connsCount = 0
+	hc.conns = hc.conns[:0]
+	if hc.connsWait != nil {
+		hc.connsWait.head = hc.connsWait.head[:0]
+		hc.connsWait.headPos = 0
+		hc.connsWait.tail = hc.connsWait.tail[:0]
+	}
```
erikdubbelboer, line 1935: I'm wondering if we should do something here when `hc.connsCount` isn't `0`. It means there are still open connections and just resetting all fields means they can't be cleaned up and closed anymore.

We could close all connections here. And we would also have to make sure no goroutine is waiting on `connsWait` anymore.
The alternative is just panicing as the caller needs to make sure the HostClient isn't in use anymore. If they don't it's a serious flaw in their program and just closing connections might just make that flaw harder to see.

erikdubbelboer (changes requested):

client.go
kiyonlin: My intention is that the user decides whether to release the host, because we have exposed connsCount. Maybe I should make it clearer in the document.

erikdubbelboer: I think it's better to also just panic here in case the user didn't wait until `connsCount == 0`. Cause thats a bug you won't easily notice otherwise.

kiyonlin: You're right. I'll add panic and warning in docs.

kiyonlin: I rethought `hostClient.ConnsCount()` again. If user use hostClient concurrently, we can not ensure we get the correct connsCount. Any ideas?

erikdubbelboer: Why not?

erikdubbelboer:
Can you fix the conflicts in client.go so this can be merged?

kiyonlin:
> Can you fix the conflicts in client.go so this can be merged?

Yeah sure. I'll let you know when I finish.

kiyonlin:
@erikdubbelboer Done. And I kept all #1000 implementation code.

erikdubbelboer (changes requested):
Seems like you have a race condition. `TestClientHTTPSConcurrent`, `TestClientGetTimeoutErrorConcurrent` and `TestClientGetTimeoutError` fail. You can test by running `go test -race`.

kiyonlin:
> Seems like you have a race condition. `TestClientHTTPSConcurrent`, `TestClientGetTimeoutErrorConcurrent` and `TestClientGetTimeoutError` fail. You can test by running `go test -race`.


When mCleaner is started, the HostClient has not yet started to connect, so the cleanup operation cannot be done immediately. We can postpone the cleanup.

```go
func (c *Client) mCleaner(m map[string]*HostClient) {
	mustStop := false

	for range time.NewTicker(time.Second * 10).C {
		// ...
		//time.Sleep(10 * time.Second)
	}
}
```

erikdubbelboer:
I'm not sure if that's the issue, it still exists: https://travis-ci.org/github/valyala/fasthttp/jobs/770199422

kiyonlin:
> I'm not sure if that's the issue, it still exists: https://travis-ci.org/github/valyala/fasthttp/jobs/770199422

I see. I tested locally and got all test cases passed. I'll take a look tomorrow.

kiyonlin:
This reset idea may has problems. Like I said before:
> I rethought hostClient.ConnsCount() again. If user use hostClient concurrently, we can not ensure we get the correct connsCount. Any ideas?

`connsCount == 0` isn't the right time to release `HostClient` because we use `sync.Mutex`. There may be some goroutines waiting for starting new connections once they fetch the `Lock`.

```go
func (c *Client) mCleaner(m map[string]*HostClient) {
	for {
		// ...
		for k, v := range m {
			v.connsLock.Lock()
			shouldRemove := v.connsCount == 0
			v.connsLock.Unlock()
			// Some other goroutines may start connections here
			if shouldRemove {
				ReleaseHostClient(v)
				delete(m, k)
			}
		}
		// ...
	}
}
```

Maybe it's time to give up this reuse feature.

PR #995, opened 2021-03-15
Immediate headers flush fixes

Bobochka:
Make `ImmediateHeaderFlush` to work correctly when no actual body will be sent.


erikdubbelboer:
Thanks!

PR #997, opened 2021-03-17
Fix unexpected panic when calling Do of a PipelineClient

### LINKED ISSUE #996: Unexpected panic when using nil response in PipelineClient
When calling `Do()` of a `PipelineClient` with nil as the second param, we always get a panic.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1255f99]
...
```
It's easy to reproduce using following code.

```golang
package main

import (
	"log"

	"github.com/valyala/fasthttp"
)

func main() {
	cli := fasthttp.PipelineClient{
		Addr:  "github.com",
		IsTLS: true,
	}

	request := fasthttp.AcquireRequest()
	request.URI().Update("https://github.com/")
	request.Header.SetMethod("GET")
	defer fasthttp.ReleaseRequest(request)

	err := cli.Do(request, nil)
	if err != nil {
		log.Fatal(err)
	}
}

```

blanet:
This PR fixes the unexpected panic issue when calling `Do()` of a `PipelineClient` with nil `Response` parameter.
And a new test case is added to avoid we run into similar issues in the future.

closes #996

### DIFF
--- a/client.go
+++ b/client.go
@@ -1443,8 +1443,7 @@ var (
 		"Make sure the server returns 'Connection: close' response header before closing the connection")
 )
 
-type timeoutError struct {
-}
+type timeoutError struct{}
 
 func (e *timeoutError) Error() string {
 	return "timeout"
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #999, opened 2021-03-19
Use bytes.IndexByte instead of bytes.Index for single byte lookup

moredure:


erikdubbelboer:
Thanks. I would expect the result to be the same as `bytes.Index` just calls `bytes.IndexByte` and the input is a constant. But it seems that as the compiler isn't smart enough to see that `strAt` never gets modified it isn't able to inline this. Even writing it as `bytes.Index(host, []byte("@"))` it isn't smart enough.

PR #1000, opened 2021-03-22
Improve Client/HostClient pooling when many HostClient structs created and removed during program execution

moredure:
@erikdubbelboer In case if #992 will be implemented, my fix almost (but may improve reusability of pools anyway, less number of objects inuse) won't needed, because of hostclients reused both with internal to HostClient pools, looking forward to one of our patches.

### DIFF
--- a/client.go
+++ b/client.go
@@ -293,6 +293,8 @@ type Client struct {
 	//
 	// By default will use isIdempotent function
 	RetryIf RetryIfFunc
+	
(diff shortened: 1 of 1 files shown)


moredure:
In case if https://github.com/valyala/fasthttp/pull/992 this ticket will be implemented, my fix will be not needed because of hostclients reused both with pools

erikdubbelboer:
Thanks, good improvement!

PR #1001, opened 2021-03-24
Adding support for securing error logs

dfirsht:
For some applications handling sensitive data, it can violate certain certifications, such as HIPAA, to log parts of requests and responses like the header and body.

This PR adds the option `SecureErrorLogMessage` which will replace the fully detailed error messages with more generic ones so that they can safely be logged.


erikdubbelboer:
Thanks!

PR #1009, opened 2021-04-21
Add option for middleware to set custom remote address

### LINKED ISSUE #1008: Add option to set value of RemoteAddr/RemoteIP by middleware
Maybe pointer to  `net.Addr` could be added to `RequestCtx` set that it could be set by middleware that if set would be used by `RequestCtx.RemoteAddr()` instead of one from `c`.

This would allow implementing more robust and generic middlewares for handling `X-Real-IP` and/or `X-Forwarded-For` etc

lafriks:
Fixes #1008

### DIFF
--- a/server.go
+++ b/server.go
@@ -2536,6 +2536,7 @@ func (s *Server) acquireCtx(c net.Conn) (ctx *RequestCtx) {
 // See https://github.com/valyala/httpteleport for details.
 func (ctx *RequestCtx) Init2(conn net.Conn, logger Logger, reduceMemoryUsage bool) {
 	ctx.c = conn
+	ctx.remoteAddr = nil
 	ctx.logger.logger = logger
 	ctx.connID = nextConnID()
 	ctx.s = fakeServer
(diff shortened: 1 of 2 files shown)


lafriks:
lint failure does not seem to be related to my PR :)

erikdubbelboer (changes requested):
It's probably best to also set `ctx.remoteAddr = nil` in `RequestCtx.Init2`.

lafriks:
> It's probably best to also set `ctx.remoteAddr = nil` in `RequestCtx.Init2`.

ok, will add. Should I remove that from `releaseCtx`?

lafriks:
@erikdubbelboer fixed

erikdubbelboer:
Thanks!

lafriks:
Thanks for prompt review

lafriks:
would be nice to tag version so that it would be easier to use it :)

erikdubbelboer:
Done

PR #1015, opened 2021-04-28
Fix chunked streaming

roman-khimov:
`io.Reader` interface provided by `requestStream` is broken for chunks exceeding `len(p)` which can lead to various panics in code using `Read()`.

### DIFF
--- a/http.go
+++ b/http.go
@@ -2072,25 +2072,35 @@ func parseChunkSize(r *bufio.Reader) (int, error) {
 		if c == ' ' {
 			continue
 		}
-		if c != '\r' {
+		if err := r.UnreadByte(); err != nil {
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Can you add a test case that fails before your change and succeeds after?

roman-khimov:
`TestRequestStream` with `128 * 1024` bytes body does exactly that --- fails without the fix and works with it. I can make this change (`3` -> `128*1024`) a separate commit and place it before the actual fix if you want to. At the moment I'm trying more to keep build/test passing at every commit from this set.

roman-khimov:
Fixed linter warning.

erikdubbelboer:
Thanks!

PR #1019, opened 2021-05-07
Fix method should not set content-length

morefreeze:
In https://tools.ietf.org/html/rfc7230#section-3.3.2, there is a sentence like this:

> A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

I'm sure GET method should set content-length if there is no body, so I submit this PR

### DIFF
--- a/http.go
+++ b/http.go
@@ -41,7 +41,7 @@ type Request struct {
 
 	multipartForm         *multipart.Form
 	multipartFormBoundary string
-	secureErrorLogMessage        bool
+	secureErrorLogMessage bool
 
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
I don't think we should change this. The spec says a `Content-Length` header shouldn't be send if it is a GET or HEAD request **AND** there is no body. Now you're always removing the header, even if a body is set.

There are API's such as this one from ElasticSearch which require a body on a GET request: https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-body.html

PR #1021, opened 2021-05-11
Fix adaptor doesn't convert header values correctly

juan-chan:
Fix(adaptor): Fixed an issue where the adapter did not convert all values of the header successfully when converting the header.

Change `Header.Set()` to `Header.Add()` to support header that has multiple values such as `Accept`

Relative reference: https://github.com/gofiber/fiber/issues/1329

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -68,7 +68,7 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 			case "Transfer-Encoding":
 				r.TransferEncoding = append(r.TransferEncoding, sv)
 			default:
-				hdr.Set(sk, sv)
+				hdr.Add(sk, sv)
 			}
 		})
 		r.Header = hdr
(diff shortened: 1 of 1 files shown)


ReneWerner87:
@erikdubbelboer with this change a bug has crept in

by doing the Add instead of Set, the ContentType is no longer set and the default ContentType is used, although it was already set by the Add.

https://github.com/gofiber/fiber/pull/1351#issuecomment-851015621
https://github.com/gofiber/fiber/pull/1351#issuecomment-851016558
https://github.com/gofiber/fiber/pull/1351#issuecomment-851018546

We just use
https://github.com/gofiber/fiber/blob/master/middleware/pprof/pprof.go#L13

and our test is broken 
https://github.com/gofiber/fiber/blob/master/middleware/pprof/pprof_test.go#L43

although the content type "text/html; charset=utf-8" was set inside
https://golang.org/src/net/http/pprof/pprof.go?s=11637:11687#L377

ReneWerner87:
@juan-chan can you look at this again and make a new correction to the initial problem with an eye on this resulting error?

erikdubbelboer:
@ReneWerner87 a fix is in progress here https://github.com/valyala/fasthttp/pull/1036 and will be merged next week of it doesn't break for anyone.

ReneWerner87:
Thanks for the quick response and the link

PR #1022, opened 2021-05-11
Fix ignoreBody still set content length

morefreeze:
Come from #1019  @erikdubbelboer I added test to prove my changes work. I'm not always removing `Content-Length` header, I remove it when `contentLength == -2` which means no body. 

This is not a perfect solution, because I found If any `Request.SetContentLength` is called then `contentLength` must be printed when `RequestHeader.Write`. I suppose we may define -3 or sth don't set `contentLengthBytes` any more.

Shall we close MR only after we have the same conclusion?

### DIFF
--- a/http.go
+++ b/http.go
@@ -41,7 +41,7 @@ type Request struct {
 
 	multipartForm         *multipart.Form
 	multipartFormBoundary string
-	secureErrorLogMessage        bool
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks. I think this is a good fix for now.

PR #1023, opened 2021-05-11
Use proper content-type when it is not present

morefreeze:
refer to https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5 

> A sender that generates a message containing a payload body SHOULD
   generate a Content-Type header field in that message unless the
   intended media type of the enclosed representation is unknown to the
   sender.  If a Content-Type header field is not present, the recipient
   MAY either assume a media type of "application/octet-stream"
   ([RFC2046], Section 4.5.1) or examine the data to determine its type.

default content type should be "application/octet-stream" instead of "application/x-www-form-urlencoded" which will mess body up.

I found this because when I sent curl command like `curl my/proxy/with/fasthttp -d abcd -XPOST` I noticed content-type had been modified by fasthttp to `application/x-www-form-urlencoded`, and body changed to `abcd=`.

### DIFF
--- a/header.go
+++ b/header.go
@@ -32,10 +32,10 @@ type ResponseHeader struct {
 	noDefaultContentType bool
 	noDefaultDate        bool
 
-	statusCode         int
-	contentLength      int
-	contentLengthBytes []byte
-	secureErrorLogMessage     bool
+	statusCode            int
+	contentLength         int
+	contentLengthBytes    []byte
+	secureErrorLogMessage bool
 
 	contentType []byte
 	server      []byte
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
I'm a little bit afraid that this might potentially break for some people who use `application/x-www-form-urlencoded` bodies and don't set the Content-Type header.

But I agree it's a better default so I'll merge it on Monday and I'll tag a new release later that week.

morefreeze:
> I'm a little bit afraid that this might potentially break for some people who use `application/x-www-form-urlencoded` bodies and don't set the Content-Type header.
> 
> But I agree it's a better default so I'll merge it on Monday and I'll tag a new release later that week.

Thanks for reply. I agree your concern. rfc https://datatracker.ietf.org/doc/html/rfc7231#section-3.1.1.5 also said
> Implementers are encouraged to provide a
   means of disabling such "content sniffing" when it is used.`
Maybe we can add `contentSniffing bool` and turn it on default, in this way that'll make sense if set `application/x-www-form-urlencoded` when body is present.

PR #1024, opened 2021-05-11
feature: add ConvertRequest func

nicolascb:
To easy method to convert ctx.RequestCtx to http.Request

### DIFF
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -1,0 +1,51 @@
+package fasthttpadaptor
+
+import (
+	"bytes"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/valyala/fasthttp"
+)
+
+// ConvertRequest convert a fasthttp.Request to http.Request
+// setURI is optionally because your use in client mode
+func ConvertRequest(ctx *fasthttp.RequestCtx, setURI bool) (*http.Request, error) {
+	var r http.Request
+
+	r.Method = string(ctx.Method())
+	r.Proto = "HTTP/1.1"
+	r.ProtoMajor = 1
+	r.ProtoMinor = 1
+	r.ContentLength = int64(len(ctx.PostBody()))
+	r.RemoteAddr = ctx.RemoteAddr().String()
+	r.Host = string(ctx.Host())
+
+	if setURI {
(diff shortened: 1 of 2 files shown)


fasthttpadaptor/request.go
```diff
@@ -0,0 +1,51 @@
...
+	"io/ioutil"
+	"net/http"
+	"net/url"
+
+	"github.com/valyala/fasthttp"
+)
+
+// ConvertRequest convert a fasthttp.Request to http.Request
+// setURI is optionally because your use in client mode
+func ConvertRequest(ctx *fasthttp.RequestCtx, setURI bool) (*http.Request, error) {
```
erikdubbelboer, line 14:
```suggestion
// ConvertRequest convert a fasthttp.Request to an http.Request
// forServer should be set to true when the http.Request is going to passed to a http.Handler.
func ConvertRequest(ctx *fasthttp.RequestCtx, forServer bool) (*http.Request, error) {
```

erikdubbelboer (changes requested):

fasthttpadaptor/request.go
```diff
@@ -0,0 +1,51 @@
...
+	var r http.Request
```
savsgio, line 15: Maybe, the `http.Request` could be part of the `Convert Request()` parameters to can use a `sync.Pool` if it's needed.

savsgio (changes requested):

fasthttpadaptor/request.go
savsgio: Somethig like this:

```go
func ConvertRequest(ctx *fasthttp.RequestCtx, setURI bool, req *http.Request) error
```

nicolascb:
@savsgio @erikdubbelboer Thanks for review!

Can you revise again?

fasthttpadaptor/request.go
erikdubbelboer: I'm not sure I agree with this. `http.Request` contains unexported fields. It's not trivial to reuse an `http.Request` with `sync.Pool` and the standard library gives no guarantee that this might break in the future.

But I guess it's up to the user to make this work properly.

erikdubbelboer (approved):

savsgio:
LGTM!

savsgio (approved):

PR #1027, opened 2021-05-16
Make sure to reset the userValues always and at the exact time

savsgio:
Reference: https://github.com/elastic/apm-agent-go/pull/957#discussion_r622694247


server.go
```diff
@@ -2237,8 +2237,6 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 		hijackNoResponse = ctx.hijackNoResponse && hijackHandler != nil
 		ctx.hijackNoResponse = false
 
-		ctx.userValues.Reset()
```
erikdubbelboer, line 2240: Now all user values will persist between requests on the same connection. This breaks what the documentation for [`SetUserValue`](https://pkg.go.dev/github.com/valyala/fasthttp?utm_source=godoc#RequestCtx.SetUserValue) states:
```
All the values are removed from ctx after returning from the top RequestHandler.
```

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

server.go
sunhailin-Leo: Why not use a config variable to control this reset operation? In some cases, users want to persist values on all connections.

PR #1034, opened 2021-05-24
[fasthttpadaptor] Set Request.TLS and try to avoid a new alloc if Request.Header is alr…

savsgio:
…eady allocated


erikdubbelboer:
Thanks!

PR #1036, opened 2021-05-26
Fix header .Add functions

### LINKED ISSUE #1033: 1.25.0: fasthttpadaptor duplicate content-type with go stdlib embed
So in 1.25.0 we noticed the `content-type` header with things using the fasthttpadaptor not only had two headers, but the first one was always text/plain when used in conjunction with go stdlib embed.

This is specifically (for an unknown reason) caused by this change:

https://github.com/valyala/fasthttp/commit/b2f111bd987535e673df3172e34ec06fed46ae0f#diff-e6ec50a76313ea3e95c7fa583dd3909aae602d31e74d55631becdab50949a519R95

I have put together an example application that will showcase the issue which I've attached [here](https://github.com/valyala/fasthttp/files/6527456/example.zip).

Edit: Also as a note, part of the above commit was reverted theoretically unintentionally here:

https://github.com/valyala/fasthttp/commit/04cde74c419c81185c47e28683da50c8a152d01d#diff-e6ec50a76313ea3e95c7fa583dd3909aae602d31e74d55631becdab50949a519L71

https://github.com/valyala/fasthttp/commit/04cde74c419c81185c47e28683da50c8a152d01d#diff-f9e55e6c9441df72de54af693adaa40e418460e97707b7b86b20c78b6947ca47R40

erikdubbelboer:
These functions should take the headers that are handled differently into account.

Fixes #1033

After this change it is not possible to have multiple 
Content-Type, Content-Length, Connection, Server, Set-Cookie, Transfer-Encoding and Date headers. Having other headers multiple times is still fine. For these headers it should never happen anyways and I don't think we have any users who do this. But if they do it would break for then.

### DIFF
--- a/fasthttpadaptor/adaptor_test.go
+++ b/fasthttpadaptor/adaptor_test.go
@@ -20,7 +20,6 @@ func TestNewFastHTTPHandler(t *testing.T) {
 	expectedRequestURI := "/foo/bar?baz=123"
 	expectedBody := "body 123 foo bar baz"
 	expectedContentLength := len(expectedBody)
-	expectedTransferEncoding := "encoding"
 	expectedHost := "foobar.com"
 	expectedRemoteAddr := "1.2.3.4:6789"
 	expectedHeader := map[string]string{
@@ -56,8 +55,8 @@ func TestNewFastHTTPHandler(t *testing.T) {
 		if r.ContentLength != int64(expectedContentLength) {
 			t.Fatalf("unexpected contentLength %d. Expecting %d", r.ContentLength, expectedContentLength)
 		}
-		if len(r.TransferEncoding) != 1 || r.TransferEncoding[0] != expectedTransferEncoding {
-			t.Fatalf("unexpected transferEncoding %q. Expecting %q", r.TransferEncoding, expectedTransferEncoding)
+		if len(r.TransferEncoding) != 0 {
+			t.Fatalf("unexpected transferEncoding %q. Expecting []", r.TransferEncoding)
 		}
 		if r.Host != expectedHost {
 			t.Fatalf("unexpected host %q. Expecting %q", r.Host, expectedHost)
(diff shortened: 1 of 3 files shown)


savsgio:
LGTM!

savsgio (approved):

james-d-elliott (approved):

Fenny:
Nice catch! Lgtm

ReneWerner87:
old code:
```go
	case HeaderTransferEncoding:
		// Transfer-Encoding is managed automatically.
	case HeaderDate:
		// Date is managed automatically.
	default:
		h.h = setArgBytes(h.h, key, value, argsHasValue)
```

new code:
```go
	case 't':
		if caseInsensitiveCompare(strTransferEncoding, key) {
			// Transfer-Encoding is managed automatically.
			return true
		}
		.....
	case 'd':
		if caseInsensitiveCompare(strDate, key) {
			// Date is managed automatically.
			return true
		}
	}
	
	....
```


@erikdubbelboer 
i think there is a mistake here, the return value should be false, outside of that you break if you have the value true and don't set a header

https://github.com/valyala/fasthttp/blob/7bd43f6c54ed4ae951c07fccaa7c6a7e25c0a27a/header.go#L1252
https://github.com/valyala/fasthttp/blob/7bd43f6c54ed4ae951c07fccaa7c6a7e25c0a27a/header.go#L1296

i saw this, because of our tests
https://github.com/gofiber/adaptor/pull/66/checks?check_run_id=2734201815

we expect
https://github.com/gofiber/adaptor/blob/master/main_test.go#L29
https://github.com/gofiber/adaptor/blob/master/main_test.go#L66
https://github.com/gofiber/adaptor/blob/master/main_test.go#L110

ReneWerner87:
@erikdubbelboer

erikdubbelboer:
To be honest I think your test is wrong. `Transfer-Encoding` can't just be set to something random like that. `Transfer-Encoding` control how the body is transferred between server and client which is something fasthttp controls itself. Otherwise you would be able to set `Transfer-Encoding: chunked` while fasthttp does an identity encoding and the whole thing fails.

ReneWerner87:
ok, thanks 

then I will adjust our tests times, just wanted to be sure

you had that also in your tests and I thought it should be settable

thx

PR #1042, opened 2021-06-05
added AcquireTestRequestCtx 

joicemjoseph:
added AcquireTestRequestCtx function to share a fakeServer for testing handler with server context

### DIFF
--- a/client.go
+++ b/client.go
@@ -294,9 +294,9 @@ type Client struct {
 	// By default will use isIdempotent function
 	RetryIf RetryIfFunc
 
-	mLock sync.Mutex
-	m     map[string]*HostClient
-	ms    map[string]*HostClient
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
What would be the use of this?

joicemjoseph:
I was testing a handler. `r.RequestCtx` was being used as value `context.Context`. while using `AcquireRequest`, `s` was nil and hence my test panicked.

erikdubbelboer:
You can use `RequestCtx.Init` or `RequestCtx.Init2` which sets the `fakeServer` just like you want: https://github.com/valyala/fasthttp/blob/b8b065b0d71acaa5c77f06e88db3fa4c3e86feb5/server.go#L2541

Issue #972, opened 2021-02-15
Support Transport-like mechanism.


dgrr:
Hello,

For future extensions of the HTTP protocol (HTTP/2 or HTTP/3), it would be nice if fasthttp can support a Transport-like mechanism like the one in net/http.
I've been thinking about adding the following to the [fasthttp.Client](https://pkg.go.dev/github.com/valyala/fasthttp#Client) and to the [fasthttp.HostClient](https://pkg.go.dev/github.com/valyala/fasthttp#HostClient) too.
```go
type Transport interface {
  Do(io.Reader, io.Writer, *Request, *Response) error
}

type Client struct {
  ...
  Transport Transport
  ...
}
```
And probably somewhere around [here](https://github.com/valyala/fasthttp/blob/a88030b8fbcabe7010ff0a8f037e46c64475e56e/client.go#L1373) do:
```go
if c.Transport != nil {
  c.Transport(br, bw, req, res)
}
```

I think the transport should be an interface. Because some Transport implementation may want to keep the state of the connection.
I have questions about the implementation of the interface. First of all:
- The `Do` function should receive an `io.Reader` and `io.Writer`? Why not receiving the net.Conn? The Reader/Writer can cause problems on the Writer side, what if the Writer is buffering (like bufio.Writer) and the Transport is never able to send the full payload?
- I prefer fasthttp to be the one handling the connections, that way we lift some weight on the transport to be the one handling the TCP connections. That way the Transport is only that, a Transport. And not a whole implementation of another library that needs to fully manage TCP connections + the protocol + any other extra costs like a object pooling, goroutine scheduling, etc...

I'd like to know your thoughts about it. Thanks

dgrr:
https://github.com/dgrr/fasthttp/commit/512bc2fc2fb0f39c4483674c8d6ddaceb930bb56

erikdubbelboer:
The way your commit implements it now the transport layer has to do all the connection managing as well. To see if this is really possible you could try and turn the current functions into a http1 transport library.

dgrr:
> The way your commit implements it now the transport layer has to do all the connection managing as well. To see if this is really possible you could try and turn the current functions into a http1 transport library.

Yes, while implementing the HTTP/2 transport I realised that the HTTP/1.1 `connection` is equivalent to `stream` in HTTP/2. In that way, the transport should manage the connections. In the HTTP/2 case, the HTTP/2 library manages one single connection, but every requests goes through a different stream.

[...]

PR #1045, opened 2021-06-13
Defined Transport for the client

dgrr:
Solves https://github.com/valyala/fasthttp/issues/972

### DIFF
--- a/client.go
+++ b/client.go
@@ -294,9 +294,9 @@ type Client struct {
 	// By default will use isIdempotent function
 	RetryIf RetryIfFunc
 
-	mLock sync.Mutex
-	m     map[string]*HostClient
-	ms    map[string]*HostClient
+	mLock      sync.Mutex
+	m          map[string]*HostClient
+	ms         map[string]*HostClient
 	readerPool sync.Pool
 	writerPool sync.Pool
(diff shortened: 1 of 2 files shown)


dgrr:
Notice that this implementation only works with the `HostClient`

erikdubbelboer:
What is the reason for not adding it to `Client` as well?

dgrr:
> What is the reason for not adding it to `Client` as well?

I'll in the future, but using the HostClient.Transport, and also I need more time for that. Doing so now in the Client is a little bit problematic. The ideal scenario would be to have something like the Server's NextProtos in the Client. So, on every new connection the HostClient creates, we'll need to check the TLS ALPN fields and if there's any extension for that has been added to fasthttp through a 3rd party library, then put it as TransportFunc.

The main problem is that the Client doesn't own the connections but the HostClient does. And the HTTP/2 lib manages it's own connections, therefore we assume all the `Transport` implementations are going to manage their own connections. Because at the end, the Transport call should only be used to send requests and get a response back.

As said, I'll try to support Transport in Client, but I'll need more time.

dgrr:
@erikdubbelboer

PR #1046, opened 2021-06-14
TCPDialer :: DNSCacheDuration option

eeertekin:
TCPDialer{} extended to support DNSCacheDuration

image

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -153,7 +153,7 @@ type TCPDialer struct {
 	// }
 	Resolver Resolver
 
-	// DNSCacheDuration may be used to override the default DNS cache duration (DefaultDNSCacheDuration = time.Minute)
+	// DNSCacheDuration may be used to override the default DNS cache duration (DefaultDNSCacheDuration)
 	DNSCacheDuration time.Duration
 
 	tcpAddrsLock sync.Mutex
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -358,7 +358,7 @@ type tcpAddrEntry struct {
 
 // DefaultDNSCacheDuration is the duration for caching resolved TCP addresses
 // by Dial* functions.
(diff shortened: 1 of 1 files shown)


erikdubbelboer (changes requested):
I would rather you add `DNSCacheDuration` as option to `TCPDialer`.

tcpdialer.go
```diff
@@ -153,6 +153,8 @@ type TCPDialer struct {
 	// }
 	Resolver Resolver
 
+	DNSCacheDuration time.Duration
```
erikdubbelboer, line 156: Can you add some documentation here that also mentions that it will default to `DefaultDNSCacheDuration`.

erikdubbelboer (changes requested):

tcpdialer.go
eeertekin: Hey @erikdubbelboer,

Comment is added,

Thanks!

tcpdialer.go
```diff
@@ -153,6 +153,9 @@ type TCPDialer struct {
 	// }
 	Resolver Resolver
 
+	// DNSCacheDuration may be used to override the default DNS cache duration (DefaultDNSCacheDuration = time.Minute)
```
erikdubbelboer, line 156:
```suggestion
	// DNSCacheDuration may be used to set the DNS cache duration (default is DefaultDNSCacheDuration)
```

erikdubbelboer: I prefer not to mention the minute here as that might change in the future and will be forgotten to change here as well.

erikdubbelboer (changes requested):

tcpdialer.go
eeertekin: @erikdubbelboer fixed.

erikdubbelboer:
Thanks!

PR #1049, opened 2021-06-18
fix: set content-length properly when StreanRequestBody was enabled

### LINKED ISSUE #1048: Bug: incorrect content-length while enable StreamRequestBody.
The content length of a request is wrong when below conditions are fulfilled:

* Enable StreamRequestBody
* MaxRequestBodySize > real RequestBodySize
* DisablePreParseMultipartForm

## How to reproduce
Step1. Start a fasthttp server using this code
```golang
package main

import (
	"fmt"

	"github.com/valyala/fasthttp"
)

func main() {
	server := &fasthttp.Server{
			fmt.Println("req.content-legnth:", ctx.Request.Header.ContentLength())
		},
		DisablePreParseMultipartForm: true,
		MaxRequestBodySize:           10 * 1024 * 1024,
		StreamRequestBody:            true,
	}

	server.ListenAndServe(":8081")
}
```
Step2. Use curl to send request and ensure the request body size is less than 10M which is server's MaxRequestBodySize.
 ```Bash
# dd if=/dev/zero of=5m.img bs=5m count=1
curl -F 'file=@5m.img' http://127.0.0.1:8081/
```
Then we see the server's output:
```
req.content-legnth: 8192
```
The `content-length` is always 8192 no matter how many requests are sent.

## Possible cause
```golang
func (req *Request) ContinueReadBodyStream(r *bufio.Reader, maxBodySize int, preParseMultipartForm ...bool) error {
    // ignore unrelated code.

    bodyBuf := req.bodyBuffer()
    bodyBuf.Reset()
    bodyBuf.B, err = readBodyWithStreaming(r, contentLength, maxBodySize, bodyBuf.B)
    if err != nil {
        if err == ErrBodyTooLarge {
            req.Header.SetContentLength(contentLength)
            req.body = bodyBuf
            req.bodyStream = acquireRequestStream(bodyBuf, r, contentLength)
            return nil
        }
        if err == errChunkedStream {
            req.body = bodyBuf
            req.bodyStream = acquireRequestStream(bodyBuf, r, -1)
            return nil
        }
        req.Reset()
        return err
    }

    req.body = bodyBuf
    req.bodyStream = acquireRequestStream(bodyBuf, r, contentLength)
    req.Header.SetContentLength(len(bodyBuf.B)) // should be req.Header.SetContentLength(contentLength)
    return nil
}
```

lemonlinger:
Fix #1048 

Please take a look. @valyala @kirillDanshin 
Thanks.

### DIFF
--- a/http.go
+++ b/http.go
@@ -1185,7 +1185,7 @@ func (req *Request) ContinueReadBodyStream(r *bufio.Reader, maxBodySize int, pre
 
 	req.body = bodyBuf
 	req.bodyStream = acquireRequestStream(bodyBuf, r, contentLength)
-	req.Header.SetContentLength(len(bodyBuf.B))
+	req.Header.SetContentLength(contentLength)
 	return nil
 }
 
--- a/server_test.go
+++ b/server_test.go
@@ -1108,7 +1108,6 @@ Host: asbd
 Connection: close
 
 `
-
 		ln := fasthttputil.NewInmemoryListener()
 
 		s := &Server{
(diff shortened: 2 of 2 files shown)


erikdubbelboer (changes requested):
Can you please add a test case that fails before the fix and passes after.

lemonlinger:
> Can you please add a test case that fails before the fix and passes after.

Done

erikdubbelboer:
Thanks!

PR #1056, opened 2021-06-28
Add ResetUserValues() and related test

sujit-baniya:
`ResetUserValues()` function will allow users to reset the user values from RequestCtx.


erikdubbelboer:
I'm curious what is the use case for this? Is it for testing?

sujit-baniya:
@erikdubbelboer
In a handler, the user value might have been already set. Based on some condition, I need to clear eexisting user value and pass new values.

PR #1064, opened 2021-08-04
Handle perIPConn in RequestCtx.IsTLS() specially

### LINKED ISSUE #1062: `RequestCtx.IsTLS()` doesn't recognize`perIPConn` when TLS is used
Steps To Reproduce: See https://github.com/gofiber/fiber/issues/1456

Description: 

When `Server.MaxConnsPerIP` is set to a non-zero value, the server wraps every incoming connection to a `perIPConn` in `acquirePerIPConn`. `perIPConn` has the struct definition: 

```go
type perIPConn struct {
	net.Conn

	ip               uint32
	perIPConnCounter *perIPConnCounter
}
```
When a `tls.Conn` comes in, `acquirePerIPConn` assigns the `tls.Conn` to the field `Conn` in the return value, which restricts the `tls.Conn` to `net.Conn`.

https://github.com/valyala/fasthttp/blob/9466cd7657c9b3e3ec7e2601114ffaaa739aeed4/peripconn.go#L48-L59

Therefore, the `IsTLS()` check on `perIPConn` will fail, as it doesn't implement the `connTLSer` any more. 

https://github.com/valyala/fasthttp/blob/9466cd7657c9b3e3ec7e2601114ffaaa739aeed4/server.go#L709-L720

Furthermore, adding the following failing test in `server_test.go` confirmed the discovery:
```go

func TestRequestCtxIsTLS(t *testing.T) {
	t.Parallel()

	var ctx RequestCtx

        // ... omitted

	ctx.c = &struct {
		net.Conn
		fooBar bool
	}{Conn: &tls.Conn{}}
	if !ctx.IsTLS() {
		t.Fatal("IsTLS must return true") // fails
	}
}
```

tysg:
Fixes #1062

### DIFF
--- a/server.go
+++ b/server.go
@@ -715,6 +715,13 @@ func (ctx *RequestCtx) IsTLS() bool {
 	//
 	//     // other custom fields here
 	// }
+
+	// perIPConn wraps the net.Conn in the Conn field
+	if pic, ok := ctx.c.(*perIPConn); ok {
+		_, ok := pic.Conn.(connTLSer)
+		return ok
+	}
+
 	_, ok := ctx.c.(connTLSer)
 	return ok
 }
--- a/server_test.go
+++ b/server_test.go
@@ -529,6 +529,11 @@ func TestRequestCtxIsTLS(t *testing.T) {
 	if !ctx.IsTLS() {
 		t.Fatal("IsTLS must return true")
 	}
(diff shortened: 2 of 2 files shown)


erikdubbelboer:
Thanks!

PR #1077, opened 2021-08-21
Optimize size of Server by moving bool fields

lancoLiu:


moredure:
the server is not an object instantiated thousands/millions of times I guess so it's not worth to optimise the alignment

erikdubbelboer:
It can't hurt either 😄

PR #1095, opened 2021-09-06
remove unnecessary  ` clear `

tylitianrui:
since  Cookie has been  cleared by  method `Reset` . i think it  is  unnecessary to  clear it ,  twice.


erikdubbelboer:
Thanks!

PR #1096, opened 2021-09-09
Add Timeouts methods to Client

erikdubbelboer:
The current Timeout/Deadline methods starting a new goroutine is
something most users don't seem to understand. They also don't get the
performance implications of using this vs setting a
ReadTimeout/WriteTimeout.

Setting A ReadTimeout/WriteTimeout is always better. Maybe we shoud
deprecate the old methods.


erikdubbelboer:
@kirillDanshin @valyala what do you think of this?

PR #1098, opened 2021-09-11
Client connection cleaner

chhquan:
Client and HostClient register itself to cleaner to clean resource, instead of creating a goroutine. new interface `resourceClean` will help a client to clean resource. Client implements `resourceClean` can register to the global cleaner.


moredure:
I love this idea

moredure:
Maybe change this ```go
func (c *Client) hasResource() bool {
	c.mLock.Lock()
	defer c.mLock.Unlock()
	if len(c.m) > 0 || len(c.ms) > 0 {
		return true
	}
	return false
}
```
to this
```go
func (c *Client) hasResource() bool {
	c.mLock.Lock()
	defer c.mLock.Unlock()
	return len(c.m) > 0 || len(c.ms) > 0
}
```

moredure:
And in case !c.connsCleanerRun && !connectionClose new connection is connectionClosed we do not need to count it as a connection?

moredure:
I mean replace this
```go
if c.connsCount == 1 {
    startCleaner = true
}
```
maybe with this not sure regarding connectionCounting
```go
if c.connsCount > 0 && !connectionClose {
    startCleaner = true
}
```

client.go
```diff
@@ -2907,3 +2902,75 @@ func releasePipelineWork(pool *sync.Pool, w *pipelineWork) {
...
+
+// Register Client to cleaner.
+func (c *clientCleaner) register(client resourceClean) {
+	if client == nil {
+		return
+	}
+	// Check if init
+	c.init()
+
+	// Store client
```
erikdubbelboer, line 2948:
```suggestion
```
The statement below is already super clear, it doesn't need a comment like this.

client.go
```diff
@@ -2907,3 +2902,75 @@ func releasePipelineWork(pool *sync.Pool, w *pipelineWork) {
...
+		time.Sleep(10 * time.Second)
+	}
+}
+
+// Register Client to cleaner.
+func (c *clientCleaner) register(client resourceClean) {
+	if client == nil {
+		return
+	}
+	// Check if init
```
erikdubbelboer, line 2945:
```suggestion

```
Not really a check and not really a needed comment.

client.go
```diff
@@ -2907,3 +2902,75 @@ func releasePipelineWork(pool *sync.Pool, w *pipelineWork) {
...
+	c.init()
+
+	// Store client
+	c.clients.Store(client, struct{}{})
+}
+
+// Clean Client when the client has no resource.
+func (c *clientCleaner) cleanClient() {
+	// Find all clients
+	allClients := make([]resourceClean, 0)
```
erikdubbelboer, line 2955: What is this slice for? Merge the loops below so this isn't needed anymore.

client.go
```diff
@@ -2907,3 +2902,75 @@ func releasePipelineWork(pool *sync.Pool, w *pipelineWork) {
 	w.err = nil
 	pool.Put(w)
 }
+
+// Resource Clean interface.
+// resourceClean can register into clientCleaner for cleaning resources.
+type resourceClean interface {
```
erikdubbelboer, line 2908: Maybe call it `cleanableClient`? Either you use `resource` everywhere and make it all generic. Or use `client` and make it specific for clients.

erikdubbelboer (changes requested):

client.go
```diff
@@ -2907,3 +2902,75 @@ func releasePipelineWork(pool *sync.Pool, w *pipelineWork) {
...
+	// clients to clean
+	clients        sync.Map
+}
+
+// Global cleaner.
+var cCleaner = &clientCleaner{}
+
+// Initialize cleaner.
+func (c *clientCleaner) init() {
+	c.initOnce.Do(func() {
```
tylitianrui, line 2927: I think  `if atomic.CompareAndSwapUint32(&o.done, 0, 1) )` is better

clientCleaner  
```
type clientCleaner struct {
	done       int32
	// clients to clean
	clients        sync.Map
}
```
method init:
```
func (c *clientCleaner) init() {
	if atomic.CompareAndSwapUint32(&o.done, 0, 1) {
               go c.cleaner()
}
```

tylitianrui (commented):
it  can  avoid  mutex .  ` sync.Once ` is  implemented   by  `Mutex `

client.go
erikdubbelboer: Good catch, `sync.Once` is only really needed if what you are doing inside it is really required for the rest of the code to continue. This is not the case here, it's perfectly fine if the goroutine is started after calls to `register` have completed already.

chhquan:
@moredure @erikdubbelboer @tylitianrui  ok! i will change the code for your suggestion.

client.go
```diff
@@ -539,37 +540,34 @@ func (c *Client) CloseIdleConnections() {
...
 
-		if mustStop {
-			break
-		}
-		time.Sleep(sleep)
+// Whether a Client has any HostClient.
+func (c *Client) hasResource() bool {
+	c.mLock.Lock()
+	defer c.mLock.Unlock()
+	if len(c.m) > 0 || len(c.ms) > 0 {
```
moredure, line 567: simplify `return len(c.m) > 0 || len(c.ms) > 0`

PR #1102, opened 2021-09-15
statusMessageMin: instead of 0

tylitianrui:
use  `statusMessageMin`  instead of 0

### DIFF
--- a/status.go
+++ b/status.go
@@ -173,7 +173,7 @@ func init() {
 }
 
 func statusLine(statusCode int) []byte {
-	if statusCode < 0 || statusCode > statusMessageMax {
+	if statusCode < statusMessageMin || statusCode > statusMessageMax {
 		return invalidStatusLine(statusCode)
 	}
 


erikdubbelboer:
The loop that fills `statusLines` starts at `0` so this if statement is correct as it is:
https://github.com/valyala/fasthttp/blob/5d73da31aed12047d2625e86bf405a0cd1f77f2b/status.go#L170-L172

tylitianrui:
thanks  for reviewing. 

i am sorry, i left some code.
i  originally thought to commit the code below,  and now  i  think it is not good.
``` 
func init() {
	// Fill all valid status lines
	for i := statusMessageMin; i < len(statusLines); i++ {
		statusLines[i] = []byte(fmt.Sprintf("HTTP/1.1 %d %s\r\n", i, StatusMessage(i)))
	}
}

func statusLine(statusCode int) []byte {
	if statusCode < statusMessageMin || statusCode > statusMessageMax {
		return invalidStatusLine(statusCode)
	}

	return statusLines[statusCode]
}

```

PR #1106, opened 2021-09-17
feat: improve  TCPDialer  by `sync.map` instead of  `map+mutex`

tylitianrui:


erikdubbelboer:
The downside of `sync.Map` is that it requires two allocations instead of one:
https://play.golang.org/p/RVUVvngcHFi
vs
https://play.golang.org/p/OW7hGm0J1JN

Also this mutex isn't very much in contention. So I don't think we should do this.

tylitianrui:
> The downside of `sync.Map` is that it requires two allocations instead of one:
> https://play.golang.org/p/RVUVvngcHFi
> vs
> https://play.golang.org/p/OW7hGm0J1JN
> 
> Also this mutex isn't very much in contention. So I don't think we should do this.

yes,  `sync.map`  requires more memory allocation than `map` , because of   its downside implement .   `tcpAddrEntry.tcpAddrsMap` is  a  cache ( read more , write less ) of `net.TCPAddr ` . so  i  think we should focus on  `cache-read` more than  `cache-write` 

-  read 

syncMap read：
```
var syncMap sync.Map
func BenchmarkSyncMap(b *testing.B) {
	syncMap.Store("example.com:443", &tcpAddrEntry{})
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			syncMap.Load("example.com:443")
		}
	})
}
```
benchmark
```
goos: darwin
goarch: amd64
BenchmarkSyncMap
BenchmarkSyncMap-8   	173760667	         7.35 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/valyala/fasthttp	2.218s
```

Map read：
```
var (
	mu sync.Mutex
	m  = make(map[string]*tcpAddrEntry)
)

func BenchmarkMap(b *testing.B) {
	m["example.com:443"] = &tcpAddrEntry{}
	b.RunParallel(func(pb *testing.PB) {
		for pb.Next() {
			mu.Lock()
			_ = m["example.com:443"]
			mu.Unlock()
		}
	})
}
```
benchmark
```
goos: darwin
goarch: amd64
BenchmarkMap
BenchmarkMap-8   	17126959	        69.7 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/valyala/fasthttp	1.525s
```

when  `fasthttp.Dial ` is call  in concurrent gorountine, as it faces me in   real work  ,  `Map + mutex`  is not a better choice . when  not in concurrency ， `Map + mutex`  is better.

erikdubbelboer:
Since the extra allocation isn't in a hot path and it makes the cache slightly faster I'll merge it.

erikdubbelboer:
Thanks!

PR #1108, opened 2021-09-19
add support for Windows filepath in fs handlers

### LINKED ISSUE #1101: FS handlers don't cooperate well with handleRequest on Windows
'Root' field within fasthttp.FS requires existing filepath provided in Windows way ("C:\\Users\\...") however additonal data is added in unix way. Resolving in not finding any html files besides "/" path.
For example: My endpoint is set to /v1/login. I have my template files in dir "C:\\Users\\myfiles" (so i set fasthttp.FS{Root: C:\\Users\\myfiles}. When I make request to endpoint /v1/login -> additonal request path is added to filepath where my html templates should been. Resulting in final filepath: "C:\\Users\\myfiles/v1/login".  Resolving in not finding any template files.
Check implementation on fs.go line 796 version 1.29.0

Landryna:
Not quite sure about any other way around it. Open to discuss
Fixes #1101


erikdubbelboer (changes requested):
I don't think this is all that needs to be fixed. I also see places like https://github.com/valyala/fasthttp/blob/4b3460a1a305545545ddfef040dfdc391d730ed8/fs.go#L983
And I see `/` being checked in a lot of places.

You also don't need the `runtime.GOOS` check, `filepath.FromSlash` takes care of that.

fs.go
```diff
@@ -794,6 +795,10 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 	if !ok {
 		pathStr := string(path)
 		filePath := h.root + pathStr
+		switch runtime.GOOS {
+		case "windows":
+			filePath = filepath.FromSlash(filePath)
```
erikdubbelboer, line 800: You don't need to check `runtime.GOOS` if you use `filepath.FromSlash`. It will take care of that.

erikdubbelboer (changes requested):
Like I said before, there are many other places where `/` is being used or checked as well. These will all need to be changed for this to work.

I think we also need some tests to check if everything works on Windows. At the moment [all FS tests are disabled on windows](https://github.com/valyala/fasthttp/blob/f0b0cfe9b2dcae6fa59ee5b18dc509cce2037efe/fs_test.go#L1).

PR #1115, opened 2021-10-01
Update http.go

moredure:
Is it make sense to append instead of allocate new? Or it's suboptimal?


erikdubbelboer:
That's just exactly the same thing written in a single, harder to read, line.

PR #1116, opened 2021-10-06
feat:no  need  to  store nil

tylitianrui:
I  dont think it is necessary  to  store  key & value into  `userdata ` ,when value  is  `nil` . 
-  method `Get` will  return  nil , if  key does not  exist.    same result with not    storing  key & value into  `userdata `.
-  there is not  a flag  indicates whether value was found in the   `userdata ` . for example ` func (d *userData) Get(key string) (value interface{} , exist bool )`

### DIFF
--- a/userdata.go
+++ b/userdata.go
@@ -21,6 +21,9 @@ func (d *userData) Set(key string, value interface{}) {
 			return
 		}
 	}
+	if value == nil {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks! Nice optimization.

PR #1117, opened 2021-10-08
feat: userData new api "delete"

tylitianrui:
[issue ](https://github.com/valyala/fasthttp/commit/f307299246e62cf484214d83c9fc41c2a106511b#commitcomment-57611238)   
a new API  for  deleting a key instead of setting  nil. 
- a key  is  removed logically .  adding a new attribute `status ` to   `userDataKV`  which  report  whether the `userDataKV`  is deleted or not.  
-  the  memory of  the  deleted one will  be  reuse,

### DIFF
--- a/userdata.go
+++ b/userdata.go
@@ -2,11 +2,19 @@ package fasthttp
 
 import (
 	"io"
+	"sync/atomic"
+)
+
+const (
+	statusAlive = iota
+	statusLocked
+	statusDeleted
 )
 
 type userDataKV struct {
-	key   []byte
-	value interface{}
+	key    []byte
+	value  interface{}
+	status uint32
 }
 
 type userData []userDataKV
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
I'm not sure what you mean with memory migration but rearranging the slice isn't a bad thing you can just copy the last entry of the slice to the one that gets deleted and decrease the len by 1. Just make sure to set the `value` of that last entry to `nil` before you decrease the length so the GC doen't keep it alive.

userdata.go
```diff
@@ -72,3 +108,22 @@ func (d *userData) Reset() {
...
 	*d = (*d)[:0]
 }
+
+func (d *userData) Remove(key string) {
+	args := *d
+	n := len(args)
+	for i := 0; i < n; i++ {
+		kv := &args[i]
+		if string(kv.key) == key {
+			kv.status = statusDeleted
```
erikdubbelboer, line 118: So here something like:
```go
n--
args[i] = args[n]
args[n].value = nil // Make sure not to keep the value alive for the GC.
args = args[:n]
*d = args
```

PR #1125, opened 2021-10-11
feat: use skipmap in tcpAddrsMap

zhangyunhao116:
[skipmap](https://github.com/zhangyunhao116/skipmap) is a high-performance, scalable concurrent map, and the `Load` and `Range` are wait-free(skipmap don't need to lock the entire map compared to sync.Map).

Because we call `tcpAddrsMap.Range` per second, which can boost a lot in this situation. We can see `1Range9Delete90Store900Load` in the [benchmark](https://github.com/zhangyunhao116/skipmap#benchmark).

skipmap has been running for a few months in the production environment at my work, hope this can help :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/klauspost/compress v1.13.4
 	github.com/valyala/bytebufferpool v1.0.0
 	github.com/valyala/tcplisten v1.0.0
+	github.com/zhangyunhao116/skipmap v0.7.0
 	golang.org/x/crypto v0.0.0-20210513164829-c07d793c2f9a
 	golang.org/x/net v0.0.0-20210510120150-4163338589ed
 	golang.org/x/sys v0.0.0-20210514084401-e8d321eab015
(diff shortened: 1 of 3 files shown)


zhangyunhao116:
The CI failed, but it looks like not related to this pull request,  maybe we could re-run these jobs?

erikdubbelboer:
Adding a bunch of new dependencies requires a lot of convincing. This is what I'm seeing in your benchmark:
```
String/1Range9Delete90Store900Load/skipmap-16   41.0ns ±21%
String/1Range9Delete90Store900Load/sync.Map-16  1.17µs ± 8%
```
I'm thinking this code in fasthttp is called so little that 1µs vs 40ns probably isn't going to make any difference at all. This really only makes a difference if you're doing millions of map operations per second which I don't see happening with fasthttp. So I'm afraid unless you can show me a profile where this would have mattered I'm going to say no to this for now.

zhangyunhao116:
Agree. Just a minor improvement for fasthttp but requires more dependencies.

PR #1126, opened 2021-10-13
feat: ability to edit status messages

valchulen:
I'm in need to change the message associated with the status of a request. In my context: when millions of RPS are answered with 204, every byte of the response counts towards reducing AWS transfer out costs, so "NC" instead of "No Content" is a small change with big impact.

To do so, I've created `SetStatusMessage`. Also I've extracted the set of the `statusLines` to reuse it in said function.

I await for you input! Thanks!

### DIFF
--- a/header.go
+++ b/header.go
@@ -137,6 +137,14 @@ func (h *ResponseHeader) SetStatusCode(statusCode int) {
 	h.statusCode = statusCode
 }
 
+// StatusLine returns response status line.
+func (h *ResponseHeader) StatusLine() []byte {
+	if len(h.statusLine) > 0 {
+		return h.statusLine
+	}
+	return statusLine(h.StatusCode())
+}
+
 // SetStatusLine sets response status line bytes.
 func (h *ResponseHeader) SetStatusLine(statusLine []byte) {
 	h.statusLine = append(h.statusLine[:0], statusLine...)
--- a/header_test.go
+++ b/header_test.go
@@ -52,8 +52,8 @@ func TestResponseHeaderMultiLineValue(t *testing.T) {
 		t.Fatalf("parse response using net/http failed, %s", err)
 	}
 
-	if !bytes.Equal(header.statusLine, []byte("SuperOK")) {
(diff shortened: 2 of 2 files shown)


header.go
```diff
@@ -136,6 +137,11 @@ func (h *ResponseHeader) SetStatusCode(statusCode int) {
 	h.statusCode = statusCode
 }
 
+// SetStatusLine sets response status line bytes.
+func (h *ResponseHeader) SetStatusLine(statusLine []byte) {
+	h.statusLine = statusLine
```
erikdubbelboer, line 142: This makes it impossible to reuse the bytes you provide as parameter to this function. It's better to use a pattern like the other functions: https://github.com/valyala/fasthttp/blob/0ceee57875aaf69f7e98bccb845e04fd4b1de559/header.go#L298

header.go
```diff
@@ -1639,7 +1645,12 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 	if statusCode < 0 {
 		statusCode = StatusOK
 	}
-	dst = append(dst, statusLine(statusCode)...)
+
+	if h.statusLine != nil {
+		dst = append(dst, h.statusLine...)
```
tylitianrui, line 1650: i think it  is  better 
```
if  len(h.statusLine) == 0  {
      dst = append(dst, statusLine(statusCode)...)
} else {
      dst = append(dst, h.statusLine...)
}
```

header.go
```diff
@@ -1859,6 +1872,9 @@ func (h *ResponseHeader) parseFirstLine(buf []byte) (int, error) {
 		}
 		return 0, fmt.Errorf("unexpected char at the end of status code. Response %q", buf)
 	}
+	if len(b) > n+1 && !bytes.Equal(b[n+1:], statusLine(h.statusCode)) {
+		h.SetStatusLine(b[n+1:])
+	}
```
erikdubbelboer, line 1877: If you're going to parse it here then I suggest adding a getter method as well `func (h *ResponseHeader) StatusLine() []byte {`.

Issue #1132, opened 2021-10-24
Status Line does not have CRLF appended properly to it


ShivanshVij:
I think I've found a bug where a response status line is not properly formatted then the `ResponseHeader.AppendBytes` function is called. 

A screenshot of the error is attached:
<img width="1316" alt="Screen Shot 2021-10-24 at 4 23 23 AM" src="https://user-images.githubusercontent.com/27162109/138586317-0725a109-6f8b-495d-9a3b-61e66cd0d45b.png">

You can see the error in the bottom part of the above screenshot 

What seems to be happening is pretty clear, the response is parsed correctly (I have confirmed this), but when the `AppendBytes` function is being called, a newline (CRLF) is not being appended in the following lines:

<img width="1171" alt="Screen Shot 2021-10-24 at 4 26 56 AM" src="https://user-images.githubusercontent.com/27162109/138586375-89032e79-63d7-4caa-98b4-bae0d223b65b.png">

PR #1135, opened 2021-10-24
fix: Status Line parsing and writing

ShivanshVij:
This is a reference to https://github.com/valyala/fasthttp/issues/1132 and it corrects a number of bugs (while keeping zero-allocation)

First, it corrects https://github.com/loopholelabs/fasthttp/commit/36ba831347092bfa941c4c8afa6ffd8bb3477ef8#diff-445b7d49a79bb217c576f137077e18262a0a2d58e424239fcc01a437397188f7R1883 where instead of comparing the status message to the formatted status message (which is what is returned by the `statusLine` function), it instead compares  the status message to the `statusLines` map entry. 

Next, it modifies the `SetStatusLine` function (https://github.com/loopholelabs/fasthttp/commit/36ba831347092bfa941c4c8afa6ffd8bb3477ef8#diff-445b7d49a79bb217c576f137077e18262a0a2d58e424239fcc01a437397188f7R149) to require a status code as well as a status message, and this is used to generate a properly formatted status line using the new `formatStatusLine` function (https://github.com/loopholelabs/fasthttp/commit/36ba831347092bfa941c4c8afa6ffd8bb3477ef8#diff-99087af835ba003ecaa8e8e178095e102304dd3b2aef74476ab06ea1e615601cR176). 

Finally, in order to convert the `statusCode` integer to a byte slice with zero-allocations, a new `i2b` function has been created (https://github.com/loopholelabs/fasthttp/commit/36ba831347092bfa941c4c8afa6ffd8bb3477ef8#diff-24779539a7f9c30594f395bab66c981b5069e92b84157cea9fa77ca1a6dbc9f1R369), which is just a slightly modified version of `strconv.formatBits`


header.go
```diff
@@ -1880,8 +1894,8 @@ func (h *ResponseHeader) parseFirstLine(buf []byte) (int, error) {
 		}
 		return 0, fmt.Errorf("unexpected char at the end of status code. Response %q", buf)
 	}
-	if len(b) > n+1 && !bytes.Equal(b[n+1:], statusLine(h.statusCode)) {
-		h.SetStatusLine(b[n+1:])
+	if len(b) > n+1 {
```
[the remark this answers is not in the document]
ShivanshVij: So it looks like the `Reason-Phrase` from the HTTP RFC is entirely optional (https://stackoverflow.com/questions/17517086/can-an-http-response-omit-the-reason-phrase), 

So, if we force the h.SetStatusLine to always get called then either:

1) we have to throw an error if the phrase is omitted (which causes the `testResponseHeaderReadSuccess` in `header_test.go` to fail) or
2) fasthttp panics if we just always call `h.SetStatusLine(b[n+1:])` and the `Reason-Phrase` is omitted

Because of this, I don't think we can throw an error if `len(b) < n+1`, and we can't just assume that `SetStatusMessage` is always called when this function returns without an error.

To get around this, I've modified the `formatStatusLine` function so that if the given `statusText` is nil, it grabs a default status message for the given `statusCode` using the the `StatusMessage` function:

```go
func formatStatusLine(dst []byte, protocol []byte, statusCode int, statusText []byte) []byte {
	dst = append(dst, protocol...)
	dst = append(dst, ' ')
	dst = strconv.AppendInt(dst, int64(statusCode), 10)
	dst = append(dst, ' ')
	if statusText == nil {
		dst = append(dst, s2b(StatusMessage(statusCode))...)
	} else {
		dst = append(dst, statusText...)
	}
	return append(dst, strCRLF...)
}
```

header.go
```diff
@@ -746,7 +758,8 @@ func (h *ResponseHeader) CopyTo(dst *ResponseHeader) {
 	dst.noDefaultDate = h.noDefaultDate
 
 	dst.statusCode = h.statusCode
-	dst.statusLine = append(dst.statusLine, h.statusLine...)
+	dst.statusMessage = append(dst.statusMessage[:0], h.statusMessage...)
```
erikdubbelboer, line 761:
```suggestion
	dst.statusMessage = append(dst.statusMessage, h.statusMessage...)
```
Sorry to nitpick but `dst.Reset()` is already called above which makes `dst.statusMessage` length 0. So the `[:0]` isn't needed.

header.go
```diff
@@ -746,7 +758,8 @@ func (h *ResponseHeader) CopyTo(dst *ResponseHeader) {
...
+	dst.protocol = append(dst.protocol[:0], h.protocol...)
```
erikdubbelboer, line 762:
```suggestion
	dst.protocol = append(dst.protocol, h.protocol...)
```

header.go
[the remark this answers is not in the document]
erikdubbelboer: Sounds good, the only thing is that you need to check if `len(statusText) == 0` because `[]byte{} == nil` is `false` and the slice can be empty if the header is being reused.

erikdubbelboer (changes requested):

ShivanshVij:
Alright, I think that should be everything. Sorry for the delay in those last few changes!

header_test.go
```diff
@@ -52,8 +52,29 @@ func TestResponseHeaderMultiLineValue(t *testing.T) {
...
+	}
+
+	if !bytes.Equal(header.appendStatusLine(nil), []byte("HTTP/3.3 200 SuperOK\r\n")) {
+		t.Errorf("parse status line with non-default value failed, got: '%s' want: 'HTTP/3.3 200 SuperOK'", header.Protocol())
+	}
+
+	header.SetStatusMessage(nil)
+
+	if !bytes.Equal(header.appendStatusLine(nil), []byte("HTTP/3.3 200 \r\n")) {
+		t.Errorf("parse status line with default protocol value failed, got: '%s' want: 'HTTP/3.3 200 '", header.appendStatusLine(nil))
```
erikdubbelboer, line 71:
```suggestion
	if !bytes.Equal(header.appendStatusLine(nil), []byte("HTTP/3.3 200 OK\r\n")) {
		t.Errorf("parse status line with default protocol value failed, got: '%s' want: 'HTTP/3.3 200 OK'", header.appendStatusLine(nil))
```

erikdubbelboer (changes requested):
Just one last test that is failing, after that I'll merge it.

header_test.go
ShivanshVij: Tests seem to be running properly now

erikdubbelboer:
Thanks!

PR #1145, opened 2021-11-03
use sync.map is better

liu-song:
The same as the pull request of https://github.com/valyala/fasthttp/pull/1106,in the fileslockmap read is more than write ,so sync.map will be better

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1370,18 +1370,15 @@ func fsModTime(t time.Time) time.Time {
 	return t.In(time.UTC).Truncate(time.Second)
 }
 
-var (
-	filesLockMap     = make(map[string]*sync.Mutex)
-	filesLockMapLock sync.Mutex
-)
+var filesLockMap sync.Map
 
 func getFileLock(absPath string) *sync.Mutex {
-	filesLockMapLock.Lock()
-	flock := filesLockMap[absPath]
-	if flock == nil {
-		flock = &sync.Mutex{}
-		filesLockMap[absPath] = flock
+	v, exist := filesLockMap.Load(absPath)
+	if exist == false {
+		flock := &sync.Mutex{}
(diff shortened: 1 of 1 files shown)


erikdubbelboer (changes requested):
These are not the same I'm afraid. Something that went wrong in the previous pull request as well and was fixed here: https://github.com/valyala/fasthttp/commit/6d4db9bb7ebb00084c15ad164a204452cacd1bd4

With your new code it is possible for two goroutines to reach the `exists == false` state at the same time.

On way to fix this is to use `LoadOrStore` instead of just `Load`.

liu-song:
Thanks for review ！
    LoadOrStore will be better, I hava changed .here is my points about race condition,I'm not sure it's right，in this scenario when two goroutines  reach the `exists == false` state at the same time, both of goroutines  will store the value of *sync.Mutex, because of the implementation of the sync.map.store,there will be  only one goroutine store success  at the same time .
    Another scenario , when two goroutine both return sync.Mutex to 
https://github.com/valyala/fasthttp/blob/master/fs.go#L1124 , because of the implementation of the sync.mutex,  there will only one goroutine get the mutex ,so it is safe for file  when the value is mutex ? Is it different when the value of sync.map is *tcpAddrEntry?

erikdubbelboer:
If both goroutines reach `exists == false` then only one of the mutexes will be stored in the map. But both goroutines will continue to lock their own mutex and think they have exclusive access to the file while they don't.

The `tcpAddrEntry` case is slightly different because it also checks some properties on it (`.pending`). That's why there I fixed it with an atomic check. Because of the use of atomic there only one goroutine will end up storing it's entry in the map.

erikdubbelboer:
Thanks!

PR #1148, opened 2021-11-07
http.go: Request.SetURI() (Fix #1141)

### LINKED ISSUE #1141: client: URI is always parsed
To perform an HTTP requests each time the `SetRequestURIBytes()` must be called. But internally the method just saves an URL and it is then parsed inside of `parseURI()`:
```go
func (req *Request) parseURI() error {
	if req.parsedURI {
		return nil
	}
	req.parsedURI = true
	return req.uri.parse(req.Header.Host(), req.Header.RequestURI(), req.isTLS)
}
```
My application has a bunch of fixed URLs like `http://example.com/api/` and I can call the ` req.uri.parse()` only once on initialization. And then I would like to just set this `fasthttp.URI` objects before sending the request with a new method `SetRequestURI(preparsedUri)`.

Is any reason why this wasn't implemented? Will you implement this or accept a PR with such functionality?

stokito:
Currently, the only way to set URI for a request is to call SetRequestURI(string).
Then when a request performed the string will be parsed into a fasthttp.URI struct.
If there are many requests with the same URI then we'll waste CPU for a parsing of the same URI string.
With the new SetURI(*URI) method we can once parse a URI string into a fasthttp.URI struct and then reuse it for many requests.
Unfortunately the URI will be copied because may be modified inside the request.
But anyway this will be more lightweight than parsing.

### DIFF
--- a/http.go
+++ b/http.go
@@ -774,6 +774,20 @@ func (req *Request) URI() *URI {
 	return &req.uri
 }
 
+// SetURI initializes request URI
+// Use this method if a single URI may be reused across multiple requests.
+// Otherwise, you can just use SetRequestURI() and it will be parsed as new URI.
+// The URI is copied and can be safely modified later.
+func (req *Request) SetURI(newUri *URI) {
+	if newUri != nil {
+		newUri.CopyTo(&req.uri)
+		req.parsedURI = true
+		return
+	}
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #1150, opened 2021-11-08
uri: isHttps() and isHttp()

stokito:
Use them instead of manual schema comparison.
This makes code cleaner but also opens a door for next optimizations

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,7 +4,6 @@ package fasthttp
 
 import (
 	"bufio"
-	"bytes"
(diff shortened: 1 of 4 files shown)


erikdubbelboer:
Seems like `TestClientDoWithCustomHeaders` is failing. I'm not sure why, maybe you can investigate.

stokito:
Sorry, did the change before going to sleep :)

erikdubbelboer:
Thanks!

PR #1151, opened 2021-11-09
Add string and bytes buffer convert trick in README

### LINKED ISSUE #1031: Consider using copy-less []byte/string conversion
Many of the optimizations in fasthttp are based on using byte arrays instead of strings. That is wise!

However, there is an assumption that translating these into strings (or back) requires a copy and should thus be discouraged.

With some caveats, this is not true. There are in fact reliable ways to avoid copying. Please see the code I use, [here](https://github.com/tliron/kutil/blob/main/util/string.go). I urge you to consider these or similar functions to handle conversion, and perhaps to benchmark the memory use and total throughput accordingly.

(The major caveat is the use of "unsafe". Some projects decide, as a matter of principle, to not use it. However, for the same reason fasthttp doesn't try to copy Go's http library, I believe that "unsafe" would be acceptable here.)

For completion, here are some references to code and discussion on this topic:

* https://go101.org/article/unsafe.html
* https://github.com/golang/go/issues/25484
* https://github.com/golang/go/issues/19367
* https://golang.org/src/strings/builder.go#L45

ichxxx:
close #1031

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,6 +485,28 @@ statusCode, body, err := fasthttp.Get(nil, "http://google.com/")
 uintBuf := fasthttp.AppendUint(nil, 1234)
 ```
 
+* String and `[]byte` buffer may converts wihtout memory allocation
+```go
+func b2s(b []byte) string {
+    return *(*string)(unsafe.Pointer(&b))
+}
+
+func s2b(s string) (b []byte) {
+    bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+    sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
+    bh.Data = sh.Data
+    bh.Cap = sh.Len
+    bh.Len = sh.Len
+    return b
+}
+```
+The underlying structure of `[]byte` buffer has only one `Len` field more than String.
+So, we can construct one directly from the other.
+
+Note:
+This is an unsafe way, 
+please make sure not to modify the bytes in the `[]byte` buffer if the result string still survives.
+
 ## Related projects
 
   * [fasthttp](https://github.com/fasthttp) - various useful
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -485,6 +485,28 @@ statusCode, body, err := fasthttp.Get(nil, "http://google.com/")
 uintBuf := fasthttp.AppendUint(nil, 1234)
 ```
 
+* String and `[]byte` buffer may converts wihtout memory allocation
```
erikdubbelboer, line 488:
```suggestion
* String and `[]byte` buffers may converted without memory allocations
```

README.md
```diff
@@ -485,6 +485,28 @@ statusCode, body, err := fasthttp.Get(nil, "http://google.com/")
...
+    bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+    sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
+    bh.Data = sh.Data
+    bh.Cap = sh.Len
+    bh.Len = sh.Len
+    return b
+}
+```
+The underlying structure of `[]byte` buffer has only one `Len` field more than String.
+So, we can construct one directly from the other.
```
erikdubbelboer, line 504: I'm not sure if it's useful to mention here.

README.md
```diff
@@ -485,6 +485,28 @@ statusCode, body, err := fasthttp.Get(nil, "http://google.com/")
...
+
+Note:
+This is an unsafe way, 
+please make sure not to modify the bytes in the `[]byte` buffer if the result string still survives.
```
erikdubbelboer, line 508: I would like to see this warning bigger and bold as it's really important! It's not just a note.

erikdubbelboer (changes requested):

README.md
ichxxx: I think some people are wondering how it works, but if people want to know more details, they'll Google it. It's ok to delete.

README.md
ichxxx: Updated

erikdubbelboer:
Thanks!

PR #1155, opened 2021-11-12
feat: close idle connections when server shutdown

### LINKED ISSUE #1103: Closing idle connections (and improving shutdown)
I've been trying to get proper shutdown on fasthttp, and given the number of bugs with people confused about this, maybe something better could be done? 

The current situation (please correct me if I'm wrong) seems to be that: Unless keepalive connections are completely disabled or `readTimeout` is set to something, there's no way to gracefully shutdown fasthttp. Also, setting those two things right before calling `Shutdown()` seems to be ignored. Finally, I am not sure what `CloseOnShutdown` was supposed to do, but whatever it is, it doesn't seem to be doing it. :)

This makes `Shutdown()` very unfriendly to use.

My code is something like:
```go
serverCtx, serverStopCtx := context.WithCancel(context.Background())

quitCtx, quitCancel := signal.NotifyContext(serverCtx,
    syscall.SIGHUP, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)

go func() {
    <-quitCtx.Done()
    quitCancel()

    shutdownCtx, cancel := context.WithTimeout(serverCtx, 60*time.Second)
    defer cancel()

    go func() {
      server.Shutdown()
      serverStopCtx()
    }()

    <-shutdownCtx.Done()
    if shutdownCtx.Err() == context.DeadlineExceeded {
      log.Printf("graceful shutdown timed out... forcing exit.")
    }
    serverStopCtx()
}()

if err = server.ListenAndServe(":8080"); err != nil {
  log.Panic(err)
}

<-serverCtx.Done()

// shutdown code, defer, etc, etc...
```

This is a bit more convoluted than the usual examples, because it includes a hard timeout on Shutdown.

If I start this code and `ctrl+c` before any requests, if works perfectly. Same if I `curl` and then `ctrl+c`. But if I open a request on a browser, then keepalive kicks in and then this will only shutdown on the hard timeout (60s) OR if I set a `ReadTimeout` for fasthttp (in which case it will wait for the read timeout to pass and then quit).

I'm thinking about some options on how to address this:

1. Would it be possible to implement something similar to: [closeIdleConns](https://github.com/golang/go/blob/release-branch.go1.15/src/net/http/server.go#L2743) from net/http? Would that solve the issue? Or would "pending" requests still hang until their timeout kicks in?

2. Would it be possible for `Shutdown()` to "update" all connections to not be keepalive anymore? I.e., to finish as soon as the current request is gone?

ichxxx:
fixes #1103

### DIFF
--- a/server.go
+++ b/server.go
@@ -416,13 +416,14 @@ type Server struct {
 
 	// We need to know our listeners and idle connections so we can close them in Shutdown().
 	ln []net.Listener
-	idleConns map[net.Conn]struct{}
 
-	connsMu sync.Mutex
-	mu      sync.Mutex
-	open    int32
-	stop    int32
-	done    chan struct{}
+	idleConns   map[net.Conn]struct{}
+	idleConnsMu sync.Mutex
+
+	mu   sync.Mutex
+	open int32
+	stop int32
+	done chan struct{}
 }
 
 // TimeoutHandler creates RequestHandler, which returns StatusRequestTimeout
@@ -2429,15 +2430,6 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 }
 
 func (s *Server) setState(nc net.Conn, state ConnState) {
-	s.connsMu.Lock()
-	s.trackConn(nc, state)
-	s.connsMu.Unlock()
-	if hook := s.ConnState; hook != nil {
-		hook(nc, state)
-	}
-}
-
-func (s *Server) setStateLocked(nc net.Conn, state ConnState) {
 	s.trackConn(nc, state)
 	if hook := s.ConnState; hook != nil {
 		hook(nc, state)
@@ -2809,6 +2801,7 @@ func (s *Server) writeErrorResponse(bw *bufio.Writer, ctx *RequestCtx, serverNam
 }
 
 func (s *Server) trackConn(c net.Conn, state ConnState) {
+	s.idleConnsMu.Lock()
 	switch state {
 	case StateIdle:
 		if s.idleConns == nil {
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -2794,6 +2808,28 @@ func (s *Server) writeErrorResponse(bw *bufio.Writer, ctx *RequestCtx, serverNam
...
+	default:
+		delete(s.idleConns, c)
+	}
+}
+
+func (s *Server) closeIdleConns() {
+	s.connsMu.Lock()
+	for c := range s.idleConns {
+		c.Close()
+		s.setStateLocked(c, StateClosed)
```
erikdubbelboer, line 2828: This is going to call `s.ConnState` which shouldn't happen here as it's already happening here: https://github.com/valyala/fasthttp/blob/2a6f7db5bbc4d7c11f1ccc0cb827e145b9b7d7ea/workerpool.go#L237

erikdubbelboer (changes requested):

server.go
ichxxx: Oh, I missed it. 
Updated.

server_test.go
```diff
@@ -3347,6 +3347,49 @@ func TestShutdownErr(t *testing.T) {
...
+
+	shutdownErr := make(chan error)
+	go func() {
+		shutdownErr <- s.Shutdown()
+	}()
+
+	timer := time.NewTimer(time.Second)
+	select {
+	case <-timer.C:
+		t.Fatalf("idel connections don't close when shutdown")
```
erikdubbelboer, line 3382:
```suggestion
		t.Fatal("idle connections not closed on shutdown")
```

server_test.go
```diff
@@ -3347,6 +3347,49 @@ func TestShutdownErr(t *testing.T) {
...
+	}()
+	conn, err := ln.Dial()
+	if err != nil {
+		t.Fatalf("unexepcted error: %s", err)
+	}
+
+	if _, err = conn.Write([]byte("GET / HTTP/1.1\r\nHost: google.com\r\n\r\n")); err != nil {
+		t.Errorf("unexpected error: %s", err)
+	}
+	time.Sleep(100 * time.Millisecond)
```
erikdubbelboer, line 3372: Instead of sleeping do
```go
	br := bufio.NewReader(conn)
	verifyResponse(t, br, StatusOK, "aaa/bbb", "real response")
```
here so you know the connection is idle after that.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

edwvee:
Great thanks! Now my app can safely stop.

PR #1162, opened 2021-11-17
fix: reset request after reset user values on keep-alive connections

savsgio:
This fix the use cases that reference a `RequestCtx` into user values that implements `io.Closer`

Example:
```go
func handler(ctx *fasthttp.RequestCtx) {
	rd := acquireRequestDuration()
	rd.ctx = ctx
	rd.start = time.Now()

	ctx.SetUserValue("__http::requestDuration__", rd)
}

func (rd *requestDuration) Close() error {
	duration := time.Since(rd.start)
	statusCode := strconv.Itoa(rd.ctx.Response.StatusCode())
	method := string(rd.ctx.Request.Header.Method())  // ISSUE: INVALID VALUE
	path := string(rd.ctx.Request.URI().Path())  // ISSUE: INVALID VALUE

	requestsDurationHistogramVec.WithLabelValues(
		statusCode, method, path,
	).Observe(duration.Seconds())

	releaseRequestDuration(rd)

	return nil
}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -2118,6 +2118,7 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 			br, err = acquireByteReader(&ctx)
 		}
 
+		reqReset = false
 		ctx.Request.isTLS = isTLS
 		ctx.Response.Header.noDefaultContentType = s.NoDefaultContentType
 		ctx.Response.Header.noDefaultDate = s.NoDefaultDate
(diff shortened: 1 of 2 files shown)


savsgio:
The test in go 1.15 fails because timeout. Could you run it again, please!

erikdubbelboer:
I'm not sure I understand what this is for. Can you maybe add a test case that fails before your change and passes afterwards?

erikdubbelboer:
Thanks, good fix!

PR #1164, opened 2021-11-23
fix: cannot use custom host in header

sikasjc:
Hi, I need fasthttp client to support custom Host in header, so I used the code change from #676 and I just modified a test case.

I think if the host in the request header exists, the host in the url should not be used.
I'm not sure, is this change ok?

Thanks

### DIFF
--- a/http.go
+++ b/http.go
@@ -1356,11 +1356,13 @@ func (req *Request) onlyMultipartForm() bool {
 func (req *Request) Write(w *bufio.Writer) error {
 	if len(req.Header.Host()) == 0 || req.parsedURI {
 		uri := req.URI()
-		host := uri.Host()
-		if len(host) == 0 {
-			return errRequestHostRequired
+		if len(req.Header.Host()) == 0 {
+			host := uri.Host()
+			if len(host) == 0 {
+				return errRequestHostRequired
+			}
+			req.Header.SetHostBytes(host)
 		}
-		req.Header.SetHostBytes(host)
 		req.Header.SetRequestURIBytes(uri.RequestURI())
 
 		if len(uri.username) > 0 {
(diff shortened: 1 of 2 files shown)


http.go
```diff
@@ -1356,11 +1356,13 @@ func (req *Request) onlyMultipartForm() bool {
...
-		host := uri.Host()
-		if len(host) == 0 {
-			return errRequestHostRequired
+		if len(req.Header.Host()) == 0 {
+			host := uri.Host()
+			if len(host) == 0 {
+				return errRequestHostRequired
+			}
+			req.Header.SetHostBytes(host)
 		}
```
erikdubbelboer, line 1365:
```suggestion
		host := uri.Host()
		if len(host) > 0 {
			req.Header.SetHostBytes(host)
		} else if len(req.Header.Host()) == 0 {
			return errRequestHostRequired
		}
```
The change you made isn't backwards compatible. There might be people who set both for some reason and expect the host from the URI to be used. What we can do is use my suggestion here where we don't return an error if the Host header is already set while the URI host is empty.

erikdubbelboer (changes requested):

http.go
sikasjc: Your suggestion should not meet my needs, which requires that there should be no host in the url so that the host in the header can be used.

I understand that backwards compatibility is important, but as far as I know, the normal behavior is supposed to be to use the host in the url only if there is no host in the header.

I think people explicitly set the host in the Header just to not use the host in the url, and I've seen many people mention this before, is it possible to break this backwards compatibility？

erikdubbelboer: Sorry I agree that the current usage isn't optimal, I would also have designed the API differently, but this is what we have and we can't break it 😞

Issue #1078, opened 2021-08-21
Chunked response with trailer result in "cannot find crlf at the end of chunk" error


pckhoi:
Hello,

While using fasthttp.Client to talk to a server that produce chunked response with trailer headers I see "cannot find crlf at the end of chunk" error. I suspects that this happens because fasthttp's implementation does not expect to see trailer headers after the last chunk, which is not according to [chunked transfer coding spec](https://www.rfc-editor.org/rfc/rfc2616.html#section-3.6.1).

This can ideally be fixed by adding https://github.com/valyala/fasthttp/issues/210. Otherwise `readBodyChunked` need to be mindful of trailers.

Let me know if I should produce an example.

Regards.

Issue #210, opened 2016-12-08
Support for HTTP Trailers


shashwathi:
Hi,

We were trying to implement a reverse proxy using fasthttp library and we noticed that the client does not support reading HTTP trailers from the response. 

Sample backend application 
```
package main

import (
       	"fmt"
       	"io"
       	"net/http"
       	"os"
)

func main() {
       	http.HandleFunc("/", hello)
       	fmt.Println("listening...")
       	err := http.ListenAndServe(":"+os.Getenv("PORT"), nil)
       	if err != nil {
       		panic(err)
       	}
}

func hello(w http.ResponseWriter, req *http.Request) {
       	w.Header().Set("Trailer", "AtEnd1, AtEnd2")

       	w.Header().Set("Content-Type", "text/plain; charset=utf-8") // normal header
       	w.WriteHeader(http.StatusOK)

       	w.Header().Set("AtEnd1", "value 1")
       	io.WriteString(w, "This HTTP response has both headers before this text and trailers at the end.\r\n")
       	w.Header().Set("AtEnd2", "value 2")
}
```

fasthttp client
```
package main

import (
       	"fmt"
       	"github.com/valyala/fasthttp"
)

func main() {
       	req := fasthttp.AcquireRequest()
       	resp := fasthttp.AcquireResponse()
       	defer fasthttp.ReleaseRequest(req)
       	defer fasthttp.ReleaseResponse(resp)
       	req.SetHost("localhost:9090")
       	req.Header.SetMethod("GET")
       	err := fasthttp.Do(req, resp)
       	if err != nil {
       		fmt.Printf("Error making HTTP request: %s\n", err.Error())
       	}
       	fmt.Printf("Response %#v", resp)
}
```

Is this feature anywhere in the roadmap for the project? 

Regards
Shash && Edwin (@flawedmatrix)

shalako:
Hello @valyala 
Are you open to a PR to add support for HTTP trailers?

erikdubbelboer:
Yes we would take PR to add support for trailers.

rof20004:
@shashwathi @erikdubbelboer 

I saw somewhere that is not possible to write to response more than once.

You wrote the response status, than you can not write to response anymore.

I will search where I saw this.

rof20004:
I founded: https://stackoverflow.com/questions/39427544/golang-http-response-headers-being-removed

Spec: https://golang.org/pkg/net/http/#ResponseWriter

rof20004:
[...]

PR #1165, opened 2021-11-24
Add trailer support

ichxxx:
This is in reference to #1078 , #210 .
This will allow FastHTTP to support trailer for chunked transfers.

### Write trailer
For example:
```go
resp := &Response{}
resp.Header.Set("Trailer", "Foo")
// write body
resp.Header.Set("Foo", "bar")

req := &Request{}
req.Header.SetTrailer("Foo")
// write body
req.Header.Set("Foo", "bar")
```
Header `Foo` will send after the chunked body.

### Read trailer
For example:
```go
func(ctx *RequestCtx) {
    ioutil.ReadAll(ctx.RequestBodyStream())
    ctx.Request.Header.VisitAllTrailer(func(key []byte) {
        trailer := ctx.Request.Header.PeekBytes(key)
        // do something
    })
}
```
Trailer can be found by inspecting Header after reading all chunked body.


header.go
```diff
@@ -382,6 +385,48 @@ func (h *RequestHeader) SetMultipartFormBoundaryBytes(boundary []byte) {
...
+		for len(key) > 0 && key[0] == ' ' {
+			key = key[1:]
+		}
+		for len(key) > 0 && key[len(key)-1] == ' ' {
+			key = key[:len(key)-1]
+		}
+		h.bufKV.key = append(h.bufKV.key[:0], key...)
+		normalizeHeaderKey(h.bufKV.key, h.disableNormalizing)
+		h.trailer = setArgBytes(h.trailer, h.bufKV.key, nil, argsNoValue)
+	}
```
erikdubbelboer, line 406:
```suggestion
	h.SetTrailerBytes(s2b(trailer))
```

header.go
```diff
@@ -530,6 +575,48 @@ func (h *RequestHeader) SetRequestURIBytes(requestURI []byte) {
...
```
erikdubbelboer, line 596:
```suggestion
	h.SetTrailerBytes(s2b(trailer))
```

header.go
```diff
@@ -382,6 +385,48 @@ func (h *RequestHeader) SetMultipartFormBoundaryBytes(boundary []byte) {
...
+		key := trailer[:i]
+		for len(key) > 0 && key[0] == ' ' {
+			key = key[1:]
+		}
+		for len(key) > 0 && key[len(key)-1] == ' ' {
+			key = key[:len(key)-1]
+		}
+		h.bufKV.key = append(h.bufKV.key[:0], key...)
+		normalizeHeaderKey(h.bufKV.key, h.disableNormalizing)
+		h.trailer = setArgBytes(h.trailer, h.bufKV.key, nil, argsNoValue)
```
erikdubbelboer, line 426: Shouldn't this be `appendArgBytes`? And then `h.trailer` should be reset before the `for` loop. And should there also be a `AddTrailerBytes`?

header.go
```diff
@@ -811,12 +902,31 @@ func (h *ResponseHeader) VisitAll(f func(key, value []byte)) {
 			f(strSetCookie, v)
 		})
 	}
+	if len(h.trailer) > 0 {
+		visitArgs(h.trailer, func(k, v []byte) {
+			f(strTrailer, k)
+		})
```
erikdubbelboer, line 908: Shouldn't this just call `f(strTailer, appendArgsKeyBytes(nil, h.trailer, strCommaSpace))`?

header.go
```diff
@@ -992,6 +1111,10 @@ func (h *ResponseHeader) setSpecialHeader(key, value []byte) bool {
 			// Transfer-Encoding is managed automatically.
 			return true
 		}
+		if caseInsensitiveCompare(strTrailer, key) {
```
erikdubbelboer, line 1114:
```suggestion
		} else if caseInsensitiveCompare(strTrailer, key) {
```

header.go
```diff
@@ -1037,6 +1160,10 @@ func (h *RequestHeader) setSpecialHeader(key, value []byte) bool {
...
```
erikdubbelboer, line 1163:
```suggestion
		} else if caseInsensitiveCompare(strTrailer, key) {
```

header.go
```diff
@@ -1060,6 +1187,8 @@ func (h *RequestHeader) setSpecialHeader(key, value []byte) bool {
 // the Content-Type, Content-Length, Connection, Server, Set-Cookie,
 // Transfer-Encoding and Date headers can only be set once and will
 // overwrite the previous value.
+//
+// If the header is in Trailer, it will send after the chunked response body.
```
erikdubbelboer, line 1191:
```suggestion
// If the header is set as a Trailer, it will be send after the chunked response body.
```
Same for the other comments below.

header.go
```diff
@@ -1656,6 +1933,26 @@ func (h *ResponseHeader) Header() []byte {
...
+// The returned value is valid until the request is released,
+// either though ReleaseRequest or your request handler returning.
+// Do not store references to returned value. Make copies instead.
+func (h *ResponseHeader) TrailerHeader() []byte {
+	h.bufKV.value = h.bufKV.value[:0]
+	for _, t := range h.trailer {
+		value := h.peek(t.key)
+		h.bufKV.value = appendHeaderLine(h.bufKV.value, t.key, value)
+	}
+	return append(h.bufKV.value, strCRLF...)
```
erikdubbelboer, line 1953:
```suggestion
	}
	h.bufKV.value = append(h.bufKV.value, strCRLF...)
	return h.bufKV.value
```
In case this allocates a new slice we want to save that in `bufKV.value` for future reuse.

header.go
```diff
@@ -1837,6 +2185,37 @@ func (h *ResponseHeader) parse(buf []byte) (int, error) {
 	return m + n, nil
 }
 
+func (h *ResponseHeader) parseTrailer(buf []byte) (int, error) {
+	if buf[0] == '0' {
```
erikdubbelboer, line 2189: Shouldn't this check `if len(buf) > 0`?

header.go
```diff
@@ -1859,6 +2238,77 @@ func (h *RequestHeader) parse(buf []byte) (int, error) {
 	return m + n, nil
 }
 
+func (h *RequestHeader) parseTrailer(buf []byte) (int, error) {
+	if buf[0] == '0' {
```
erikdubbelboer, line 2242: Shouldn't this check if `len(buf) > 0`?

header.go
```diff
@@ -382,6 +385,48 @@ func (h *RequestHeader) SetMultipartFormBoundaryBytes(boundary []byte) {
...
+		if i < 0 {
+			i = len(trailer)
+		}
+		key := trailer[:i]
+		for len(key) > 0 && key[0] == ' ' {
+			key = key[1:]
+		}
+		for len(key) > 0 && key[len(key)-1] == ' ' {
+			key = key[:len(key)-1]
+		}
```
erikdubbelboer, line 402: Should there be a `isBadTrailer` check here?

header.go
```diff
@@ -2028,6 +2478,10 @@ func (h *ResponseHeader) parseHeaders(buf []byte) (int, error) {
 					}
 					continue
 				}
+				if caseInsensitiveCompare(s.key, strTrailer) {
```
erikdubbelboer, line 2481:
```suggestion
				} else if caseInsensitiveCompare(s.key, strTrailer) {
```

header.go
```diff
@@ -2116,6 +2570,10 @@ func (h *RequestHeader) parseHeaders(buf []byte) (int, error) {
...
```
erikdubbelboer, line 2573:
```suggestion
				} else if caseInsensitiveCompare(s.key, strTrailer) {
```

erikdubbelboer (changes requested):
Lint is failing: https://github.com/valyala/fasthttp/runs/4310069291?check_suite_focus=true

Can you add some clear documentation that trailers are only supported with chunked bodies?

header.go
ichxxx: This was checked by `mustPeekBuffered()` above.
```go
func mustPeekBuffered(r *bufio.Reader) []byte {
	buf, err := r.Peek(r.Buffered())
	if len(buf) == 0 || err != nil {
		panic(fmt.Sprintf("bufio.Reader.Peek() returned unexpected data (%q, %v)", buf, err))
	}
	return buf
}
```

header.go
ichxxx: Sure.
Having a `AddTrailerBytes` will be better.

header.go
ichxxx: The previous judgement has a `return`, would it be clearer not to write `else` ?

header.go
ichxxx: The previous judgement has a `continue`, would it be clearer not to write `else` ?

ichxxx:
> Lint is failing: https://github.com/valyala/fasthttp/runs/4310069291?check_suite_focus=true
> 
> Can you add some clear documentation that trailers are only supported with chunked bodies?

Sure.

header.go
```diff
@@ -382,6 +384,102 @@ func (h *RequestHeader) SetMultipartFormBoundaryBytes(boundary []byte) {
...
+		key := trailer[:i]
+		for len(key) > 0 && key[0] == ' ' {
+			key = key[1:]
+		}
+		for len(key) > 0 && key[len(key)-1] == ' ' {
+			key = key[:len(key)-1]
+		}
+		if isBadTrailer(key) {
+			continue
+		}
```
ichxxx, line 476: Should here collect the bad trailers and return an error to user?

ichxxx, line 476: Should here collect the bad trailers and return an error to user?

header.go
erikdubbelboer: Behavior wise it doesn't matter, but it's better to follow the same style as the code around it which does use `else`.

header.go
erikdubbelboer: Yes that would be better to do.

header.go
erikdubbelboer: I still think it's better if this calls `f` once with all the trailer values separated by a comma. This is also how it will go over the wire.

header.go
erikdubbelboer: Since these are new functions I think it would be ok if the return an `error`.

header.go
ichxxx: Sorry, I forgot to remove `visitArgs`

ichxxx:
Seems the tests are a bit flaky.

erikdubbelboer:
> Seems the tests are a bit flaky.

Yeah they are. You can ignore that.

header.go
```diff
@@ -471,13 +480,21 @@ func (h *ResponseHeader) AddTrailerBytes(trailer []byte) {
...
+			badTrailers = append(badTrailers, strCommaSpace...)
 			continue
 		}
 		h.bufKV.key = append(h.bufKV.key[:0], key...)
 		normalizeHeaderKey(h.bufKV.key, h.disableNormalizing)
 		h.trailer = appendArgBytes(h.trailer, h.bufKV.key, nil, argsNoValue)
 	}
+
+	if len(badTrailers) > 0 {
+		return fmt.Errorf("forbidden trailers: %s", badTrailers[:len(badTrailers)-len(strCommaSpace)])
```
erikdubbelboer, line 495: I think it's better to define a ʼErrBadTrailerʼ and return that so we don't have any allocations here either. Users can figure out themselves which trailer was bad.

header.go
```diff
@@ -2593,7 +2638,7 @@ func (h *ResponseHeader) parseHeaders(buf []byte) (int, error) {
 					continue
 				}
 				if caseInsensitiveCompare(s.key, strTrailer) {
-					h.SetTrailerBytes(s.value)
+					_ = h.SetTrailerBytes(s.value)
```
erikdubbelboer, line 2641: These functions return an error so we could return it here. It also returns an error on a bad content-length for example.

erikdubbelboer (changes requested):

erikdubbelboer:
Sorry for asking for so many changes. I'm super happy that you wrote this code. But it's also quite a lot and important to get right I think.

ichxxx:
This is not a problem at all. It's always better to be strict with code. I have also learned a lot from this pr.

header.go
```diff
@@ -2159,13 +2770,15 @@ func (h *RequestHeader) collectCookies() {
...
 func parseContentLength(b []byte) (int, error) {
 	v, n, err := parseUintBuf(b)
 	if err != nil {
-		return -1, err
+		return -1, fmt.Errorf("cannot parse Content-Length: %s", err)
 	}
 	if n != len(b) {
-		return -1, fmt.Errorf("non-numeric chars at the end of Content-Length")
+		return -1, fmt.Errorf("cannot parse Content-Length: %s", errNonNumericChars)
 	}
```
ichxxx, line 2782: I think the previous bad Content-Length err was a little unclear, so I made this change.

header.go
```diff
@@ -2159,13 +2770,15 @@ func (h *RequestHeader) collectCookies() {
 	h.cookiesCollected = true
 }
 
+var errNonNumericChars = errors.New("non-numeric chars found")
+
 func parseContentLength(b []byte) (int, error) {
 	v, n, err := parseUintBuf(b)
 	if err != nil {
-		return -1, err
+		return -1, fmt.Errorf("cannot parse Content-Length: %s", err)
```
erikdubbelboer, line 2778:
```suggestion
		return -1, fmt.Errorf("cannot parse Content-Length: %w", err)
```

header.go
```diff
@@ -2159,13 +2770,15 @@ func (h *RequestHeader) collectCookies() {
...
 	}
 	if n != len(b) {
-		return -1, fmt.Errorf("non-numeric chars at the end of Content-Length")
+		return -1, fmt.Errorf("cannot parse Content-Length: %s", errNonNumericChars)
```
erikdubbelboer, line 2781:
```suggestion
		return -1, fmt.Errorf("cannot parse Content-Length: %w", errNonNumericChars)
```

server_test.go
```diff
@@ -3784,6 +3784,67 @@ func TestIncompleteBodyReturnsUnexpectedEOF(t *testing.T) {
...
+		"AtEnd3": "3333",
+	}
+
+	h := func(ctx *RequestCtx) {
+		ctx.Response.Header.DisableNormalizing()
+		ctx.Response.Header.Set("Transfer-Encoding", "chunked")
+		for k := range trailer {
+			err := ctx.Response.Header.AddTrailer(k)
+			if err != nil {
+				t.Fatalf("unexpected error: %s", err)
```
erikdubbelboer, line 3802:
```suggestion
				t.Errorf("unexpected error: %s", err)
```
Calling `t.Fatal` is not allowed in a gorountine. See: https://pkg.go.dev/testing#B.FailNow (which Fatal calls)

server_test.go
```diff
@@ -3784,6 +3784,67 @@ func TestIncompleteBodyReturnsUnexpectedEOF(t *testing.T) {
...
+			}
+		}
+		ctx.Response.SetBodyStreamWriter(func(w *bufio.Writer) {
+			for i := 0; i < 3; i++ {
+				fmt.Fprintf(w, "message %d", i)
+				if err := w.Flush(); err != nil {
+					return
```
erikdubbelboer, line 3809: Why not `t.Error(err)`?

streaming_test.go
```diff
@@ -142,6 +143,70 @@ func getChunkedTestEnv(t testing.TB) (*fasthttputil.InmemoryListener, []byte) {
...
+%s
+`, chunkedBody)
+
+	ln := fasthttputil.NewInmemoryListener()
+	s := &Server{
+		StreamRequestBody: true,
+		Handler: func(ctx *RequestCtx) {
+			all, err := ioutil.ReadAll(ctx.RequestBodyStream())
+			if err != nil {
+				t.Fatalf("unexpected error: %s", err)
```
erikdubbelboer, line 169:
```suggestion
				t.Errorf("unexpected error: %s", err)
```

streaming_test.go
```diff
@@ -142,6 +143,70 @@ func getChunkedTestEnv(t testing.TB) (*fasthttputil.InmemoryListener, []byte) {
...
+			}
+			if !bytes.Equal(all, body) {
+				t.Fatalf("unexpected body %q. Expecting %q", all, body)
```
erikdubbelboer, line 172:
```suggestion
				t.Errorf("unexpected body %q. Expecting %q", all, body)
```

streaming_test.go
```diff
@@ -142,6 +143,70 @@ func getChunkedTestEnv(t testing.TB) (*fasthttputil.InmemoryListener, []byte) {
...
+			}
+
+			for k, v := range expectedTrailer {
+				r := ctx.Request.Header.Peek(k)
+				if string(r) != v {
+					t.Fatalf("unexpected trailer %s. Expecting %s. Got %q", k, v, r)
```
erikdubbelboer, line 178:
```suggestion
					t.Errorf("unexpected trailer %s. Expecting %s. Got %q", k, v, r)
```

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1169, opened 2021-12-04
Allow to set Host header for Client

Tolyar:
Allow to change Host header when using Client. Currently only HostClient allows to change Host header.

### DIFF
--- a/http.go
+++ b/http.go
@@ -57,8 +57,8 @@ type Request struct {
 	// if <= 0, means not set
 	timeout time.Duration
 
-	// Allow to change Host header with parsedURI == true (E.g. Client for type).
-	AllowToChangeHostHeader bool
+	// Use Host header (request.Header.SetHost) instead of the host from SetRequestURI, SetHost, or URI().SetHost
+	UseHostHeader bool
 }
 
 // Response represents HTTP response.
@@ -1363,7 +1363,7 @@ func (req *Request) Write(w *bufio.Writer) error {
 		if len(host) == 0 {
 			return errRequestHostRequired
 		}
-		if len(req.Header.Host()) == 0 || !req.AllowToChangeHostHeader {
+		if len(req.Header.Host()) == 0 || !req.UseHostHeader {
 			req.Header.SetHostBytes(host)
 		}
 		req.Header.SetRequestURIBytes(uri.RequestURI())
--- a/http.go
+++ b/http.go
@@ -1360,10 +1360,13 @@ func (req *Request) Write(w *bufio.Writer) error {
 	if len(req.Header.Host()) == 0 || req.parsedURI {
 		uri := req.URI()
 		host := uri.Host()
-		if len(host) == 0 {
-			return errRequestHostRequired
-		}
-		if len(req.Header.Host()) == 0 || !req.UseHostHeader {
+		if len(req.Header.Host()) == 0 {
+			if len(host) == 0 {
+				return errRequestHostRequired
+			} else {
+				req.Header.SetHostBytes(host)
+			}
+		} else if !req.UseHostHeader {
 			req.Header.SetHostBytes(host)
 		}
 		req.Header.SetRequestURIBytes(uri.RequestURI())
--- a/http.go
+++ b/http.go
@@ -56,6 +56,9 @@ type Request struct {
 	// Request timeout. Usually set by DoDeadline or DoTimeout
 	// if <= 0, means not set
 	timeout time.Duration
+
+	// Allow to change Host header with parsedURI == true (E.g. Client for type).
+	AllowToChangeHostHeader bool
 }
 
 // Response represents HTTP response.
@@ -1360,7 +1363,7 @@ func (req *Request) Write(w *bufio.Writer) error {
 		if len(host) == 0 {
 			return errRequestHostRequired
 		}
-		if len(req.Header.Host()) == 0 {
+		if len(req.Header.Host()) == 0 || !req.AllowToChangeHostHeader {
 			req.Header.SetHostBytes(host)
 		}
 		req.Header.SetRequestURIBytes(uri.RequestURI())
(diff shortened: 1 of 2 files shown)


Tolyar:
Comments for failed tests. Test TestRequestUpdateURI do r.Header.SetHost("aaa.bbb") but checking header Host: foobar.com This correct for the old behaviour but not for the new. I do not want touch tests before PR will be checked.

erikdubbelboer:
There was already a discussion about this here: https://github.com/valyala/fasthttp/pull/1164
I think the change I posted there is the only acceptable solution, the change you made now would break for all users who somehow have the Host header set already but depend on it being change to the host of the url that they specify.

Tolyar:
I've checked your version. It does not works for me with Client. I've get error 'lookup : no such host'. req.URI().SetHost("") allow to change Host header, but breaks connection address. May be possible to add new attribute to Request type for enable to change Host header? This should not break compatibility with old versions with correct defaults.

Tolyar:
I've pushed new version without tests violation.

http.go
```diff
@@ -56,6 +56,9 @@ type Request struct {
 	// Request timeout. Usually set by DoDeadline or DoTimeout
 	// if <= 0, means not set
 	timeout time.Duration
+
+	// Allow to change Host header with parsedURI == true (E.g. Client for type).
+	AllowToChangeHostHeader bool
```
erikdubbelboer, line 61: I think something like this is cleaner?
```suggestion
	// Use Host header (request.Header.SetHost) instead of the host from SetRequestURI, SetHost, or URI().SetHost
	UseHostHeader bool
```

http.go
```diff
@@ -1360,7 +1363,9 @@ func (req *Request) Write(w *bufio.Writer) error {
 		if len(host) == 0 {
 			return errRequestHostRequired
 		}
-		req.Header.SetHostBytes(host)
+		if len(req.Header.Host()) == 0 || !req.AllowToChangeHostHeader {
+			req.Header.SetHostBytes(host)
```
erikdubbelboer, line 1367: Shouldn't the `if len(host) == 0 { return errRequestHostRequired` check then also be inside the `if req.AllowToChangeHostHeader`?

erikdubbelboer (changes requested):

http.go
Tolyar: Good comment. I've rename attribute and add test for it.

http.go
Tolyar: I don't think so. The uri.Host() used as connection address at least for Client type. If host will be empty but Host header is not, connection will be broken. I've checked such code in http.go

		if len(req.Header.Host()) == 0 || !req.UseHostHeader {
			req.Header.SetHostBytes(host)
		} else if len(host) == 0 {
			return errRequestHostRequired
		}

And such usage for this code

		req.UseHostHeader = true
		req.SetHost("")
		req.Header.SetHost(host)

There is result - lookup : no such host

We can use headers for setting uri.Host() but I am not sure that it is correct.

Tolyar: I've check code again. It is not important for Client, because error raised inside dialAddr. But I am still not sure that it is correct allow uri.Host() to be empty if Host header presents.

erikdubbelboer: But shouldn't `len(host) == 0` be allowed when `req.UseHostHeader === true`? It makes more sense to me to put that inside the if?

_Sorry I'm replying bit slow, I'm traveling._

Tolyar: I've checked logic again and pushed a new version. errRequestHostRequired should be returned only if both uri.Host() and req.Header.Host() are empty. And it does not depend from UseHostHeader. If Header.Host() is empty we always need to set header from uri.Host(). And only when both are not empty we need to check UseHostHeader.

erikdubbelboer:
Thanks! this makes sense.

sikasjc:
Hi, it looks like `UseHostHeader` is not works. `Client.Do` uses reqCopy, but `Request.copyToSkipBody` does not copy the `UseHostHeader`

Tolyar:
> Hi, it looks like `UseHostHeader` is not works. `Client.Do` uses reqCopy, but `Request.copyToSkipBody` does not copy the `UseHostHeader`

Could you give me an example where is it does not works?  Looks like that it may not works for pipelined client but I am not sure yet. I've using it in Client.Do and it works for me.

sikasjc:
> Could you give me an example where is it does not works? Looks like that it may not works for pipelined client but I am not sure yet. I've using it in Client.Do and it works for me.

I made a mistake, I used `Client.DoTimeout()` not `Client.Do()`.
 `Client.Do()` is correct, but `Client.DoTimeout()` does not work. Here is an example:

```Go
testServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	fmt.Println(r.Host)
	// output:
	// 127.0.0.1:port from Client.DoTimeout 
	// SomeHost       from Client.Do
}))
defer testServer.Close()

client := &fasthttp.Client{}
defer fasthttp.ReleaseRequest(req)

req.SetRequestURI(testServer.URL)
req.UseHostHeader = true
req.Header.SetHost("SomeHost")
if err := client.DoTimeout(req, resp, 10*time.Second); err != nil {
	log.Print(err)
}
if err := client.Do(req, resp); err != nil {
	log.Print(err)
}
```

Tolyar:
> Here is an example:

Thank you. I will check and fix it soon.

Tolyar:
> Hi, it looks like `UseHostHeader` is not works. `Client.Do` uses reqCopy, but `Request.copyToSkipBody` does not copy the `UseHostHeader`

I've created #1184 that fixes the problem.

sikasjc:
> I've created #1184 that fixes the problem.

Thanks
But the UseHostHeader was not reset, which could affect subsequent requests.

```Go
req.UseHostHeader = true
fasthttp.ReleaseRequest(req)

newReq := fasthttp.AcquireRequest()
defer fasthttp.ReleaseRequest(req)
fmt.Println(newReq.UseHostHeader)
// output:
//     true, should be false
```

Tolyar:
> But the UseHostHeader was not reset, which may affect subsequent requests.

You are right again. Thanks a lot. There is fix #1185

PR #1176, opened 2021-12-14
fix: reset response after reset user values on keep-alive connections

savsgio:
Same as the issue https://github.com/valyala/fasthttp/pull/1162 but with the response!

### DIFF
--- a/server.go
+++ b/server.go
@@ -2079,7 +2079,7 @@ func (s *Server) serveConn(c net.Conn) (err error) {
 		connectionClose bool
 		isHTTP11        bool
 
-		reqReset               bool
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

ShivanshVij:
@erikdubbelboer Can we make a new release from the master branch with this change?

erikdubbelboer:
I was not planning on tagging a new release of fasthttp until next year, just to be sure we don't break anything for anyone around this busy time where not everyone works.

ShivanshVij:
That makes a lot of sense, thank you for thinking about that!

PR #1181, opened 2021-12-20
HostClient: addMissingPort() refactoring

stokito:
Please see commit messages for details
Now instead of
```
&HostClient{
		Addr: "localhost:80",
```
We can use simply:

```
&HostClient{
		Addr: "localhost",
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -299,7 +299,6 @@ type Client struct {
 
 	mLock      sync.Mutex
 	m          map[string]*HostClient
-	ms         map[string]*HostClient
 	readerPool sync.Pool
 	writerPool sync.Pool
 }
@@ -460,7 +459,7 @@ func (c *Client) Do(req *Request, resp *Response) error {
 		return ErrorInvalidURI
 	}
 
-	host := uri.Host()
+	host := uri.HostWithPort()
 
 	isTLS := false
 	if uri.isHttps() {
@@ -472,22 +471,13 @@ func (c *Client) Do(req *Request, resp *Response) error {
 	startCleaner := false
 
 	c.mLock.Lock()
-	m := c.m
-	if isTLS {
-		m = c.ms
+	if c.m == nil {
+		c.m = make(map[string]*HostClient)
 	}
-	if m == nil {
-		m = make(map[string]*HostClient)
-		if isTLS {
-			c.ms = m
-		} else {
-			c.m = m
-		}
-	}
-	hc := m[string(host)]
+	hc := c.m[string(host)]
 	if hc == nil {
 		hc = &HostClient{
-			Addr:                          addMissingPort(string(host), isTLS),
+			Addr:                          string(host),
 			Name:                          c.Name,
 			NoDefaultUserAgentHeader:      c.NoDefaultUserAgentHeader,
 			Dial:                          c.Dial,
@@ -510,8 +500,8 @@ func (c *Client) Do(req *Request, resp *Response) error {
 			clientReaderPool:              &c.readerPool,
 			clientWriterPool:              &c.writerPool,
 		}
-		m[string(host)] = hc
-		if len(m) == 1 {
+		c.m[string(host)] = hc
+		if len(c.m) == 1 {
 			startCleaner = true
 		}
 	}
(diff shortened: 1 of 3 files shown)


stokito:
Damn there is failed tests that is not so easy to fix. 
TestAllocationClient: now failed because there is an allocation of a new addr string with port.
TestHostClientGet: uses a Unix socket and the added 80 port breaks this.

Please tell me if the PR is valuable and it's make sense to try to fix the tests

erikdubbelboer (changes requested):
I'm sorry but these changes don't make any sense. We can't use the same HostClient for both http and https connections as a HostClient uses persistent connections and sending a http request to a https port (and the other way around) will return in an error.

stokito:
Yes, you are right and here is a bug. I rethought the PR and here is my proposal:
1. The hidden allocation inside `dialAddr()` when `dial == nil` the `addMissingPort()` is called must be removed.
2. We may store both http and tls host clients into the same map but before we must add a port to host. I.e. `example.com` must be extended to `example.com:80` or `example.com:443`. So the key will be unique and we won't have a clash. So the rule is to always add a port even if the port is default for the protocol.
3. To avoid a memory allocation on each `client.Do()` call because of an additional  `addMissingPort()` call we may add a new method `HostAndPort()` to `URI`. We should add the `:80` or `:443` port part inside of `parseHost()` method.

erikdubbelboer:
Sounds reasonable. But keep in mind that `HostClient` can also be used without `Client`. So you'll also need to fix things like [`nextAddr`](https://github.com/valyala/fasthttp/blob/f7c354c760cf6974f0a954eb2e024dceb842f002/client.go#L1897)

stokito:
Hi @erikdubbelboer I made the changes as we discussed.
First of all it looks like it's better to keep the addMissingPort() for dialer because in the HostClient.Addr field clearly said that users may not add a port. So this is out of scope of the initial task.

I made the changes fully backward compatible. Please see commit comments for details

erikdubbelboer (changes requested):
I would prefer it if instead of all the complicated logic and the extra `portPos` field you would just add an internal `func (u *URI) hostWithPort() string` function that `Client` can then use. This function can then contain the logic to add the default ports and just return a string. Don't you think that is simpler?

stokito:
Yes, that's what the first commit does.
Still adding a port is an allocation. And previously there wasn't alloc in this place.
Here I see that code complication is not so big as an optimization.
But it's up to you to decide.

erikdubbelboer:
I would prefer to see something like the first commit but then keeping `HostWithPort` internal and returning a string so there are no additional allocations.

stokito:
Hi @erikdubbelboer could you please elaborate more? I didn't get your idea

erikdubbelboer:
So like https://github.com/valyala/fasthttp/pull/1181/commits/bec03ca3ea1f101d60dc8e2dcbb305a40487a4cf but then change `HostWithPort` to `hostWithPort` and make it return a string.

stokito:
I made it return a byte slice because converting to a string will result into an allocation. The `host` is then used for a lookup so ideally it must be fast

erikdubbelboer:
But the place where it's used it's cast to a string anyways: https://github.com/valyala/fasthttp/blob/bec03ca3ea1f101d60dc8e2dcbb305a40487a4cf/client.go#L477-L480
So instead of casting it to a `[]byte` in `hostWithPort` and then back to a `string` twice there it's better to just not cast it at all.

stokito:
Yes, you right but please note that the `m[string(host)]` is a special case that compiler optimizes to avoid an allocation:

https://github.com/golang/go/issues/3512

So while it looks ugly it's better to keep it like that

erikdubbelboer:
But it's originally a string. So it doesn't make sense to convert it to a byte slice if it is only used as a string.

stokito:
The `host` internally is a byte slice and it's difficult to change it. Maybe I missed you point, if you find the change useful please do the changes yourself as you think it would be better.
Thank you anyway

erikdubbelboer:
Not in this commit: https://github.com/valyala/fasthttp/commit/bec03ca3ea1f101d60dc8e2dcbb305a40487a4cf which I suggested you use. Here you use `addMissingPort` which returns a string and you then convert it to a `[]byte` just to convert it back to a string again.

stokito:
The `addMissingPort` has an allocation. That's why in the next commit I changed it to just return a field. I'll close the PR since it can't be applied.

PR #1197, opened 2022-01-17
Fixed statusCode = 200 if request failed / did`t have response code

### LINKED ISSUE #1196: (h *ResponseHeader) StatusCode() int Leads to misunderstanding
header.go
```
// StatusCode returns response status code.
func (h *ResponseHeader) StatusCode() int {
	if h.statusCode == 0 {
		return StatusOK
	}
	return h.statusCode
}
```
leads to misunderstanding, because while I debug my code, I expect StatusCode() will return the the value with same sense as h.statusCode but it returns HTTP response code. I propose to rename this method to HTTPStatusCode or assign to h.statusCode another name.

waclawthedev:
Closes #1196

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -920,6 +920,7 @@ func testPipelineClientDoConcurrent(t *testing.T, concurrency int, maxBatchDelay
 
 	s := &Server{
 		Handler: func(ctx *RequestCtx) {
+			ctx.SetStatusCode(200)
 			ctx.WriteString("OK") //nolint:errcheck
 		},
 	}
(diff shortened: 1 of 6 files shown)


client_test.go
```diff
@@ -920,6 +920,7 @@ func testPipelineClientDoConcurrent(t *testing.T, concurrency int, maxBatchDelay
 
 	s := &Server{
 		Handler: func(ctx *RequestCtx) {
+			ctx.SetStatusCode(200)
```
erikdubbelboer, line 923: Sorry, this is not what I had in mind. If this is required for the change to work then it will break for all users that aren't doing this at the moment. That in unacceptable. I think 200 should be set as the default status code if nothing has been set once the handler returns.

erikdubbelboer (changes requested):

client_test.go
waclawthedev: Ok, i proposed second solution (PR #1198). Maybe it will be softer :)

PR #1198, opened 2022-01-17
Moved setting of the statusCode from (h *ResponseHeader) StatusCode()

### LINKED ISSUE #1196: (h *ResponseHeader) StatusCode() int Leads to misunderstanding

waclawthedev:
Closes  #1196

### DIFF
--- a/header.go
+++ b/header.go
@@ -129,9 +129,6 @@ func (h *RequestHeader) SetByteRange(startPos, endPos int) {
 
 // StatusCode returns response status code.
 func (h *ResponseHeader) StatusCode() int {
-	if h.statusCode == 0 {
-		return StatusOK
-	}
 	return h.statusCode
 }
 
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -2503,6 +2500,9 @@ func (h *ResponseHeader) parseFirstLine(buf []byte) (int, error) {
 		}
 		return 0, fmt.Errorf("cannot parse response status code: %w. Response %q", err, buf)
 	}
+	if h.statusCode == 0 {
```
erikdubbelboer, line 2503: Why is parseUintBuf returning 0 if it's a valid response? That needs more investigating. This feels like a quick patch for a bug somewhere else.

erikdubbelboer (changes requested):

header.go
waclawthedev: Because tests (for example) ignore the setting of response status - then it will be 0.

erikdubbelboer: But setting the response status shouldn't be needed, it should default to 200.
I would like to see in `server.go` after `s.Handler(ctx)` a check if the status code was set, and if it wasn't set it to 200.
Otherwise all handlers that don't explicitly set a statuscode will return 0 as status code and you're change will break for a LOT of users. We don't want that.

PR #1199, opened 2022-01-18
fix(hijack): reset userValues after hijack handler execution

savsgio:


erikdubbelboer:
Makes sense I guess. The requestCtx isn't reused but you want `Close` to be called on the values.

savsgio:
> Makes sense I guess. The requestCtx isn't reused but you want `Close` to be called on the values.

That's it, because something values that implement the `io.Closer` will never close.

Maybe, the `RequestCtx` can be put back into the pool afterwards to, what do you thing?

savsgio:
Can this be the reason for not put back the RequestCtx to the pool?
https://github.com/valyala/fasthttp/blob/master/server.go#L2383

erikdubbelboer:
The reason that comment is there is that we can't return `ctx` to the pool while the hijack handler is still running. That is what not setting `ctx = nil` there would do. But releasing it to the pool after the hijack handler is done should be ok I think. But please test it well 😄

savsgio:
Ok, I'll make the PR to test it 😉

PR #1203, opened 2022-01-28
Fix scheme check for not yet parsed requests

ArminBTVS:
At this point the request might not be parsed yet and set. In that case uri is empty and isHttps() returns always false. I don't expect this is intended?

Otherwise URL() must be called before actually passing the request to client.Do()


erikdubbelboer:
That makes sense. Would it be possible for you to add a test case that fails before and passes after your fix?

ArminBTVS:
@erikdubbelboer sure. Added a test which would fail without the fix

erikdubbelboer:
Sorry one last thing, can you fix the lining error? Thanks!

ArminBTVS:
> Sorry one last thing, can you fix the lining error? Thanks!

Done

erikdubbelboer:
Thanks!

Issue #1205, opened 2022-02-03
Panics at on context close, v1.33.0


Gaudeamus:
Hello, unfortunately there are no more details except this message I could find.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x200 pc=0x901667]

goroutine 19473 [running]:
github.com/valyala/fasthttp.(*RequestCtx).Done(0x8c4639)
        /go/src/dsp/vendor/github.com/valyala/fasthttp/server.go:2691 +0x7
context.propagateCancel.func1()
        /usr/local/go/src/context/context.go:280 +0x50
created by context.propagateCancel
        /usr/local/go/src/context/context.go:278 +0x1d0
```

Gaudeamus:
I believe it's due to latest `func (ctx *RequestCtx) reset()` where `ctx.s = nil` is now set

moredure:
https://github.com/valyala/fasthttp/commit/436977654aa1d51bf45507353e5ff34a4f54ca1a

moredure:
https://github.com/valyala/fasthttp/pull/1201

moredure:
Are you calling Done() function inside request handler or in some goroutine/task scheulder (ants, etc) where requestctx may be already released?

Gaudeamus:
on the init of handler it goes like:
```
	handlerData.requestCtx = fasthttpCtx
	handlerData.ctx, handlerCtx.cancel = context.WithTimeout(fasthttpCtx, time.Millisecond*500)
```
on the defer of the handler there is release context like this:
```
	handlerData.requestCtx = nil //ensure not use context anymore
	if handlerData.cancel != nil {
		handlerData.cancel()
	}
	if handlerData.ctx != nil {
		select {
		case <-handlerData.ctx.Done():
		default:
			logs.Warning("context is not released")
		}
	}
	handlerData.cancel = nil
	handlerData.ctx = nil
```

moredure:
as I understand it logging "context is not released", then panics after 500ms

moredure:
I'll make pull request to golang to fix it)

moredure:
https://github.com/golang/go/pull/50989/files

moredure:
But in some cases parent.Err() will panic anyway, since server may close s.done channel, and in parallel child cancel then sometimes it might call parent.Err() when s already nil

moredure:
[...]

PR #1206, opened 2022-02-03
Use context instead of done channel to avoid RequestCtx race condition

moredure:
Fixes https://github.com/valyala/fasthttp/issues/1205 (propagateCancel will return ok and no goroutine causing race condition won't be created). Truth be told, I don't like the idea of embedding context into the structure: bad practices, etc. But it's seems the only way to fix it, while maintaining requestCtx compatibility with context.Context interface

### DIFF
--- a/server.go
+++ b/server.go
@@ -424,10 +424,11 @@ type Server struct {
 	idleConns   map[net.Conn]struct{}
 	idleConnsMu sync.Mutex
 
-	mu   sync.Mutex
-	open int32
-	stop int32
-	done chan struct{}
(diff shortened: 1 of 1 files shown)


moredure:
Not fixes because of Context will become nil on requestCtx reset anyway...

moredure:
@erikdubbelboer please check.

moredure:
As @erikdubbelboer found a simpler solution, probably this patch can be used as a solution to not create goroutine per request in cases WithCancel or WithDeadline used on requestCtx.

PR #1207, opened 2022-02-03
Use atomic s field

moredure:
Fixes https://github.com/valyala/fasthttp/issues/1205


moredure:
atomics won't help here since race condition

PR #1209, opened 2022-02-07
uri.go fast scheme comparison

stokito:
The methods isHttps() and isHttp() checks that all bytes are equals bytes.Equal().
If the URI is reused for many requests then the bytes.Equal() call is performed on each request.
Another downside is that on SetScheme() memory is once allocated and copied.
FastHttp supports only http and https so instead we may store a strHTTP(S) constant and then just compare by reference.
But in Go byte slices doesn't have doesn't have an identity so we must use a pointer to a byte slice instead.

Also for a custom schemes (ftp, dav) the memory will be allocated instead of pointing to a constant. If really needed this may be changed in the future.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -224,21 +224,7 @@ func (u *URI) SetScheme(scheme string) {
 
 // SetSchemeBytes sets URI scheme, i.e. http, https, ftp, etc.
 func (u *URI) SetSchemeBytes(scheme []byte) {
-	if len(scheme) == 0 {
-		u.scheme = nil
-		return
-	}
-	if bytes.Equal(scheme, strHTTP) {
-		u.scheme = &strHTTP
-		return
-	}
-	if bytes.Equal(scheme, strHTTPS) {
-		u.scheme = &strHTTPS
-		return
-	}
-	newScheme := append([]byte(nil), scheme...) // copy scheme
-	lowercaseBytes(newScheme)
-	u.scheme = &newScheme
+	u.SetScheme(b2s(scheme))
 }
 
 func (u *URI) isHttps() bool {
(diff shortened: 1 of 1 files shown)


uri.go
```diff
@@ -197,37 +197,62 @@ func (u *URI) PathOriginal() []byte {
...
+		u.scheme = &strHTTP
+		return
+	}
+	if bytes.Equal(scheme, strHTTPS) {
+		u.scheme = &strHTTPS
+		return
+	}
+	newScheme := append([]byte(nil), scheme...) // copy scheme
+	lowercaseBytes(newScheme)
+	u.scheme = &newScheme
```
erikdubbelboer, line 241:
```suggestion
	u.SetScheme(b2s(scheme))
```

erikdubbelboer (changes requested):

erikdubbelboer:
Do you really think this is worth complicating the code over? It's basically optimizing `bytes.Equal` with 4 bytes to a pointer compare. Have you done any benchmark to see what the actual difference is? I wouldn't be surprised if it's in nanoseconds. And in that case the work being done by the GC for having the extra pointer probably takes up more CPU time.

stokito:
Yes, you are right: difference in nanoseconds. Interesting that the `SetScheme` works much faster than `SetSchemeBytes`.
```
func BenchmarkSetScheme(b *testing.B) {
	var u URI
	u.Parse(nil, []byte("http://google.com/")) //nolint:errcheck
	scheme := []byte("http")
	scheme2 := []byte("https")
	u.SetSchemeBytes(scheme2)
	for i := 0; i < b.N; i++ {
		u.SetSchemeBytes(scheme)
		u.isHttp()
		u.isHttps()
		u.SetSchemeBytes(scheme2)
		u.isHttp()
		u.isHttps()
	}
}
```

The initial idea was to have something similar to enum but the resulting code is too complicated.

PR #1212, opened 2022-02-08
uri_test.go use example.com for clearness

stokito:
To simplify understanding of test use more meaningful names instead of asdf


erikdubbelboer:
Thanks! Makes sense.

PR #1214, opened 2022-02-10
support configure HostClient

farss:
ConfigureClient configures the fasthttp.HostClient. 
So I can use https://github.com/dgrr/http2 in fasthttp.Client

### DIFF
--- a/client.go
+++ b/client.go
@@ -297,6 +297,9 @@ type Client struct {
 	// By default will use isIdempotent function
 	RetryIf RetryIfFunc
 
+	// ConfigureClient configures the fasthttp.HostClient.
+	ConfigureClient func(hc *HostClient) error
+
 	mLock      sync.Mutex
 	m          map[string]*HostClient
 	ms         map[string]*HostClient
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
I'm curious, what do you need to configure that you can't configure on the `fasthttp.Client` itself?

erikdubbelboer:
In that case I would prefer it if you added `Client.Transport` that gets copied to the `HostClient` then, just like all the other options.

farss:
Client 

> In that case I would prefer it if you added `Client.Transport` that gets copied to the `HostClient` then, just like all the other options.

Client don't know Addr of HostClient, So need callback function to config HostClient

erikdubbelboer:
I'm not sure I understand. `HostClient.Transport` is a `func(*Request, *Response) error`. Why can't that be configured on `Client` and passed to `HostClient`? In the Transport function you get the `Request` so you can make decisions based on that?

farss:
HostClient.Transport works on all requests. Sometimes want to config HostClient only after created

erikdubbelboer:
Ok you convinced me 😄

farss:
thank you!

zxpdmw:
如何将fasthttp.Client配置成支持http2协议的, 用新增的ConfigureClient可以吗, 可以提供个demo

PR #1215, opened 2022-02-14
[workerpool] Improve performance

OmidHekayati:
[workerpool] Improve performance >>
- Don't use a global lock to prevent stop-the-world problems same as in the GC mechanism
- Due to don't use the global lock, clean mechanism don't stop whole the app
- Move idle workers pool to idleWorkers(sync.Pool) field instead of ready([]*workerChan) field of workerPool struct
- Instead compare of last usage time for each worker, just compare idle workers numbers in two clean period
- Some minor fixes

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -1,0 +1,219 @@
+package pool
+
+import (
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+// LIFO pool, i.e. the most recently Put() item will return in call Get()
+// Such a scheme keeps CPU caches hot (in theory).
+// 
+// Due to LIFO behaviors and MaxItems logic can't implement with sync/atomic like sync.Pool
+type LIFO struct {
+	MaxItems    int64
+	MinItems    int64
+	MaxIdles    int64
+	IdleTimeout time.Duration
+	// New optionally specifies a function to generate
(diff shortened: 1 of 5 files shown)


erikdubbelboer:
Sorry I haven't had the time yet to read though the changes. But can you fix the linting errors first?

OmidHekayati:
Due to my misunderstanding of the behaviors `sync.Pool` that can drop all unused items in the two periods of GC and some other mechanisms explained here https://github.com/golang/go/issues/22950, I need more time to add a new commit to fix my idea to reach correctly to limit the number of concurrent workers alongside improve performance.

erikdubbelboer:
That's a lot of complicated changes. Do you have some benchmarks to show the improvements?

OmidHekayati:
I will try to complete the idea (improve pool mechanism performance) in [our organization Go library](https://github.com/GeniusesGroup/libgo/blob/master/pool/lifo.go) and add enough tests and benchmarks before reopening a PR to discuss more.
Meanwhile, you can see my first [benchmark here](https://github.com/GeniusesGroup/libgo/blob/master/pool/bench_realloc-vs-empty_test.go) that show allocating new idle items slices in a pool with many idle items and many clean in short periods will be a better solution. But we need more benchmarks to discover more aspects.

PR #1216, opened 2022-02-15
feature: Keep the memory usage of the service at a stable level

Rennbon:
add request and response body size limit, it prevents the large body from slowly stretching the memory of the entire service

### DIFF
--- a/http.go
+++ b/http.go
@@ -22,8 +22,8 @@ var (
 	responseBodyPoolSizeLimit = -1
 )
 
-// SetBodySizePoolLimit set the max body limit in request and response
-// if the body size larger than the limit,it will be released
+// SetBodySizePoolLimit set the max body size for bodies to be returned to the pool.
+// If the body size is larger it will be released instead of put back into the pool for reuse.
 func SetBodySizePoolLimit(reqBodyLimit, respBodyLimit int) {
 	requestBodyPoolSizeLimit = reqBodyLimit
 	responseBodyPoolSizeLimit = respBodyLimit
--- a/http.go
+++ b/http.go
@@ -18,13 +18,13 @@ import (
 )
 
 var (
-	requestBodyMaxLimit  = 0
-	responseBodyMaxLimit = 0
+	requestBodyMaxLimit  = -1
+	responseBodyMaxLimit = -1
 )
 
-// SetBodyLimit set the max body limit in request and response
+// SetBodySizePoolLimit set the max body limit in request and response
 // if the body size larger than the limit,it will be released
-func SetBodyLimit(reqBodyMaxLimit, respBodyMaxLimit int) {
+func SetBodySizePoolLimit(reqBodyMaxLimit, respBodyMaxLimit int) {
 	requestBodyMaxLimit = reqBodyMaxLimit
 	responseBodyMaxLimit = respBodyMaxLimit
 }
@@ -969,7 +969,7 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 
 // Reset clears request contents.
 func (req *Request) Reset() {
-	if requestBodyMaxLimit > 0 && req.body != nil {
+	if requestBodyMaxLimit >= 0 && req.body != nil {
 		req.ReleaseBody(requestBodyMaxLimit)
 	}
 	req.Header.Reset()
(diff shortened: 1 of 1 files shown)


Rennbon:
#1225

http.go
```diff
@@ -17,6 +17,18 @@ import (
 	"github.com/valyala/bytebufferpool"
 )
 
+var (
+	requestBodyMaxLimit  = 0
+	responseBodyMaxLimit = 0
```
erikdubbelboer, line 22: I would use `-1` as a default so people can set it to `0` to prevent any bodies from being kept.

http.go
```diff
@@ -957,6 +969,9 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 
 // Reset clears request contents.
 func (req *Request) Reset() {
+	if requestBodyMaxLimit > 0 && req.body != nil {
```
erikdubbelboer, line 972: This then becomes `requestBodyMaxLimit >= 0`.

http.go
```diff
@@ -17,6 +17,18 @@ import (
...
 )
 
+var (
+	requestBodyMaxLimit  = 0
+	responseBodyMaxLimit = 0
+)
+
+// SetBodyLimit set the max body limit in request and response
+// if the body size larger than the limit,it will be released
+func SetBodyLimit(reqBodyMaxLimit, respBodyMaxLimit int) {
```
erikdubbelboer, line 27: This is very confusing. The name of the function and the documentation makes me think this limits the body sizes I can send or receive.

I would call the method something like `SetBodySizePoolLimit` or `SetMaxBodySizeForPool`.

erikdubbelboer (changes requested):
I'm also wondering, why can't you call `Request.ReleaseBody()` or `Response.ReleaseBody()` yourself?

Rennbon:
> 
Because fasthttp is well developed and there are many other supporting respositories available. For a business project, we want to configure the limit of request and response uniformly, so that the upper limit of the mem of the request response can be handled from a macro perspective, instead of repeating the coding in each place for recycling the critical value.

http.go
Rennbon: Yeah, the naming here is misleading, I've modified it

Rennbon:
> I'm also wondering, why can't you call `Request.ReleaseBody()` or `Response.ReleaseBody()` yourself?
In addition, some data collection handlers depend on the server, such as prometheus, etc. The body of some scenarios will also be very large.

http.go
```diff
@@ -957,6 +969,9 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 
 // Reset clears request contents.
 func (req *Request) Reset() {
+	if requestBodyMaxLimit >= 0 && req.body != nil {
+		req.ReleaseBody(requestBodyMaxLimit)
```
erikdubbelboer, line 973:
```suggestion
	if requestBodyPoolSizeLimit >= 0 && req.body != nil {
		req.ReleaseBody(requestBodyPoolSizeLimit)
```

http.go
```diff
@@ -986,6 +1001,9 @@ func (req *Request) RemoveMultipartFormFiles() {
 
 // Reset clears response contents.
 func (resp *Response) Reset() {
+	if responseBodyMaxLimit >= 0 && resp.body != nil {
+		resp.ReleaseBody(responseBodyMaxLimit)
```
erikdubbelboer, line 1005:
```suggestion
	if responseBodyPoolSizeLimit >= 0 && resp.body != nil {
		resp.ReleaseBody(responseBodyPoolSizeLimit)
```

http.go
```diff
@@ -17,6 +17,18 @@ import (
 	"github.com/valyala/bytebufferpool"
 )
 
+var (
+	requestBodyMaxLimit  = -1
+	responseBodyMaxLimit = -1
+)
+
+// SetBodySizePoolLimit set the max body limit in request and response
+// if the body size larger than the limit,it will be released
```
erikdubbelboer, line 26:
```suggestion
// SetBodySizePoolLimit set the max body size for bodies to be returned to the pool.
// If the body size is larger it will be released instead of put back into the pool for reuse.
```

erikdubbelboer:
Thanks!

Rennbon:
> 

My pleasure

Issue #873, opened 2020-09-05
[Bug/Question] Hostclient sends content type header on DELETE request even if it is not set


shkreios:
Hey,
I just started to learn **golang** as well as **fashttp**. I tried to build a **fasthttp reverse proxy** using this [libary](https://github.com/yeqown/fasthttp-reverse-proxy) and later using the code from this [issue](https://github.com/valyala/fasthttp/issues/64#issuecomment-194880575).

All requests work without any problems, only on delete requests the **hostclient** adds the header `content-type `with the value `application/x-www-form-urlencoded`. 

The server to which the request is proxyed responds with `406`, is there a way to disable this behavior?

Thank you in advance for any help.

**Code example:**
In the code example I don't use a proxy just the `fashttp.Client` to make a request with the request method set to `DELETE`. I have tested the behavior with a **fashttp** and **express** server, the code is shown below.

`sender.go`
```golang
package main

import (
	"fmt"

	"github.com/valyala/fasthttp"
)

func main() {
	req := fasthttp.AcquireRequest()

	resp := fasthttp.AcquireResponse()
	req.SetRequestURI("http://localhost:3000/")

	req.Header.SetMethod("DELETE")

	req.Header.Del("content-type")

	client := &fasthttp.Client{}
	fmt.Printf("Content-Type: %s\n", string(req.Header.ContentType()))
	client.Do(req, resp)
	fmt.Printf("Content-Type: %s\n", string(req.Header.ContentType()))

	fmt.Println(resp.StatusCode(), string(resp.Body()))
	fmt.Println(string(resp.Body()))

}
```
------------------------------------------------------------
**Output:**
```
Content-Type: 
Content-Type: 
200 OK
OK
```
`receiver.go`

```golang
package main

import (
	"fmt"

	"github.com/valyala/fasthttp"
)

func fastHTTPHandler(ctx *fasthttp.RequestCtx) {

	fmt.Println(string(ctx.Request.Header.Header()))

	fmt.Fprintf(ctx, "Hi there! RequestURI is %q", ctx.RequestURI())

}

func main() {
	fmt.Println("server started...")

	server := &fasthttp.Server{
		Handler: fastHTTPHandler,
	}

	err := server.ListenAndServe(":3000")
	if err != nil {
		panic(err)
	}

}
```
------------------------------------------------------------
**Output:**
```
DELETE / HTTP/1.1
User-Agent: fasthttp
Host: localhost:3000
[...]

PR #1218, opened 2022-02-16
RequestHeader support set no default ContentType

j40951:
#873

In the definition of restful style api, the deletion of resources is generally defined as follows: 

```go
delete /students/{student_id}
```

there is no need to pass in a body, so add a func SetNoDefaultContentType to RequestHeader.


erikdubbelboer:
Thanks!

PR #1221, opened 2022-02-18
Fix http proxy behavior

Aoang:
Add Host header fix behavior

https://datatracker.ietf.org/doc/html/rfc7230#section-5.4


Aoang:
HTTP Proxy needs to follow the RFC7230 standard.

https://datatracker.ietf.org/doc/html/rfc7230

But long before RFC7230 was formulated in 2014, the CONNECT tunnel principle was widely implemented.

erikdubbelboer:
Thanks!

PR #1224, opened 2022-02-20
fix: (useless check), skip Response body if http method HEAD

sten4eg:
The header can't be both a HEAD and GET so checking both doesn't make any sense.


erikdubbelboer:
Thanks!

PR #1233, opened 2022-03-03
Read response when client closes connection #1232

### LINKED ISSUE #1232: Response not returned when server forcefully closes connection
When using fasthttp.HostClient to send a request, a response is ignored if the remote side forcefully closes the connection (via RST flag) right after sending a response.

Instead of returning the error message: "An existing connection was forcibly closed by the remote host", I would expect the response to be read and a nil error to be returned from `client.Do(req, resp)`.

According to the [HTTP spec 8.2.2](https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html) the client SHOULD monitor the network connection for an error status while it is transmitting the request. In case of fasthttp the response is not read after writing failed due to a closed connection.

ArminBTVS:
Closes #1232

### DIFF
--- a/client.go
+++ b/client.go
@@ -1439,7 +1439,8 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		err = bw.Flush()
 	}
 	c.releaseWriter(bw)
-	if err != nil && !errors.Is(err, syscall.ECONNRESET) {
+	isECONNRESET := errors.Is(err, syscall.ECONNRESET)
+	if err != nil && !isECONNRESET {
 		c.closeConn(cc)
 		return true, err
 	}
@@ -1471,7 +1472,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		return retry, err
 	}
 
-	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() {
+	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() || isECONNRESET {
 		c.closeConn(cc)
 	} else {
 		c.releaseConn(cc)
--- a/client.go
+++ b/client.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 )
 
@@ -1437,12 +1438,11 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 	if err == nil {
 		err = bw.Flush()
 	}
-	if err != nil {
-		c.releaseWriter(bw)
+	c.releaseWriter(bw)
+	if err != nil && !errors.Is(err, syscall.ECONNRESET) {
 		c.closeConn(cc)
 		return true, err
 	}
-	c.releaseWriter(bw)
 
 	if c.ReadTimeout > 0 {
 		// Set Deadline every time, since golang has fixed the performance issue
@@ -1462,14 +1462,14 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 	}
 
 	br := c.acquireReader(conn)
-	if err = resp.ReadLimitBody(br, c.MaxResponseBodySize); err != nil {
-		c.releaseReader(br)
+	err = resp.ReadLimitBody(br, c.MaxResponseBodySize)
+	c.releaseReader(br)
+	if err != nil && !errors.Is(err, syscall.ECONNRESET) {
 		c.closeConn(cc)
 		// Don't retry in case of ErrBodyTooLarge since we will just get the same again.
 		retry := err != ErrBodyTooLarge
 		return retry, err
 	}
-	c.releaseReader(br)
 
 	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() {
 		c.closeConn(cc)
@@ -1477,7 +1477,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		c.releaseConn(cc)
 	}
 
-	return false, err
+	return false, nil
 }
 
 var (
--- a/client.go
+++ b/client.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-	"syscall"
 	"time"
 )
 
@@ -1439,8 +1438,8 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		err = bw.Flush()
 	}
 	c.releaseWriter(bw)
-	isECONNRESET := errors.Is(err, syscall.ECONNRESET)
-	if err != nil && !isECONNRESET {
+	isConnRST := isConnectionReset(err)
+	if err != nil && !isConnRST {
 		c.closeConn(cc)
 		return true, err
 	}
@@ -1472,7 +1471,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 		return retry, err
 	}
 
-	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() || isECONNRESET {
+	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() || isConnRST {
 		c.closeConn(cc)
 	} else {
 		c.releaseConn(cc)
--- a/client.go
+++ b/client.go
@@ -1464,7 +1464,7 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
 	br := c.acquireReader(conn)
 	err = resp.ReadLimitBody(br, c.MaxResponseBodySize)
 	c.releaseReader(br)
-	if err != nil && !errors.Is(err, syscall.ECONNRESET) {
+	if err != nil {
 		c.closeConn(cc)
 		// Don't retry in case of ErrBodyTooLarge since we will just get the same again.
 		retry := err != ErrBodyTooLarge
--- a/client_test.go
+++ b/client_test.go
@@ -2839,7 +2839,7 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
 
 	_, err := client.doNonNilReqResp(req, &Response{})
 	if err != nil {
-		t.Fatalf("https requests with IsTLS client must succeed")
+		t.Fatal("https requests with IsTLS client must succeed")
 	}
 }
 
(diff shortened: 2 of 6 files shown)


erikdubbelboer (changes requested):
Can you rebase your pull request on master.

client.go
```diff
@@ -1462,22 +1462,22 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
...
 		// Don't retry in case of ErrBodyTooLarge since we will just get the same again.
 		retry := err != ErrBodyTooLarge
 		return retry, err
 	}
-	c.releaseReader(br)
 
 	if resetConnection || req.ConnectionClose() || resp.ConnectionClose() {
 		c.closeConn(cc)
 	} else {
 		c.releaseConn(cc)
```
erikdubbelboer, line 1477: Now when a connection is closed with `ECONNRESET` it will still be released to the pool here (as there was no error before) and will try to be used again.

erikdubbelboer (changes requested):

client.go
moredure: ```if resetConnection || req.ConnectionClose() || resp.ConnectionClose() || isECONNRESET```

http.go
```diff
@@ -1290,13 +1291,19 @@ func (resp *Response) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
 
 	if !resp.mustSkipBody() {
 		err = resp.ReadBody(r, maxBodySize)
+		if errors.Is(err, syscall.ECONNRESET) {
```
moredure, line 1294: Can be embedded into next if err != nil block

http.go
```diff
@@ -1290,13 +1291,19 @@ func (resp *Response) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
...
+			return nil
+		}
 		if err != nil {
 			return err
 		}
 	}
 
 	if resp.Header.ContentLength() == -1 {
 		err = resp.Header.ReadTrailer(r)
+		if errors.Is(err, syscall.ECONNRESET) {
```
moredure, line 1304: Can be embedded into next if err != nil block

client.go
ArminBTVS: @moredure true. Did not see this case. I did make the change as @moredure suggested. Thanks!

http.go
ArminBTVS: Did change this

http.go
ArminBTVS: Did change this

moredure:
Cool, but still needed to be rebase.

http.go
```diff
@@ -1291,20 +1291,14 @@ func (resp *Response) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
 
 	if !resp.mustSkipBody() {
 		err = resp.ReadBody(r, maxBodySize)
-		if errors.Is(err, syscall.ECONNRESET) {
-			return nil
-		}
-		if err != nil {
+		if err != nil && !errors.Is(err, syscall.ECONNRESET) {
```
moredure, line 1294: It's a little bit not what I meant, 
```go
if err != nil {
    if errors.Is(err, syscall.ECONNRESET) {
	    return nil
    }
    return err
}
```

http.go
```diff
@@ -1291,20 +1291,14 @@ func (resp *Response) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
...
 		}
 	}
 
 	if resp.Header.ContentLength() == -1 {
 		err = resp.Header.ReadTrailer(r)
-		if errors.Is(err, syscall.ECONNRESET) {
-			return nil
-		}
-		if err != nil && err != io.EOF {
+		if err != nil && err != io.EOF && !errors.Is(err, syscall.ECONNRESET) {
```
moredure, line 1301: same as above

client.go
moredure: Ack

http.go
moredure: @ArminBTVS please check latest comments

http.go
ArminBTVS: @moredure OK. Changed to your recomendation.

ArminBTVS:
@erikdubbelboer @moredure It seems like the bahaviour on windows is different than on unix. My test never fails on linux even when I run it for multiple times, but never passes on windows. I will further investigate the issue using an windows OS.

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+
+	srv, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	go func() {
+		conn, err := srv.Accept()
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2859:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Read at least one byte of the header
+		// Otherwise we would have an unsolicited response
+		_, err = ioutil.ReadAll(io.LimitReader(conn, 1))
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2866:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Respond
+		_, err = conn.Write([]byte("HTTP/1.1 418 Teapot\r\n\r\n"))
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2872:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Forcefully close connection
+		err = conn.(*net.TCPConn).SetLinger(0)
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2878:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+
+	srv, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err.Error())
+	}
+
+	go func() {
+		conn, err := srv.Accept()
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2917:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Read at least one byte of the header
+		// Otherwise we would have an unsolicited response
+		_, err = ioutil.ReadAll(io.LimitReader(conn, 1))
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2924:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Respond with incomplete header
+		_, err = conn.Write([]byte("Http"))
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2930:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+		}
+
+		// Forcefully close connection
+		err = conn.(*net.TCPConn).SetLinger(0)
+		if err != nil {
+			t.Errorf(err.Error())
```
erikdubbelboer, line 2936:
```suggestion
			t.Error(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
 }
+
+// See issue #1232
+func TestRstConnResponseWhileSending(t *testing.T) {
+	const expectedStatus = http.StatusTeapot
+	const payload = "payload"
+
+	srv, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err.Error())
```
erikdubbelboer, line 2853:
```suggestion
		t.Fatal(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+	resp := AcquireResponse()
+	defer ReleaseResponse(resp)
+
+	req.Header.SetMethod("POST")
+	req.SetBodyStream(strings.NewReader(payload), len(payload))
+	req.SetRequestURI(svrUrl)
+
+	err = client.Do(req, resp)
+	if err != nil {
+		t.Fatal(err.Error())
```
erikdubbelboer, line 2898:
```suggestion
		t.Fatal(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+	}
+}
+
+// See issue #1232
+func TestRstConnClosedWithoutResponse(t *testing.T) {
+	const payload = "payload"
+
+	srv, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatal(err.Error())
```
erikdubbelboer, line 2911:
```suggestion
		t.Fatal(err)
```

client_test.go
```diff
@@ -2840,3 +2842,118 @@ func TestHttpsRequestWithoutParsedURL(t *testing.T) {
...
+	defer ReleaseResponse(resp)
+
+	req.Header.SetMethod("POST")
+	req.SetBodyStream(strings.NewReader(payload), len(payload))
+	req.SetRequestURI(svrUrl)
+
+	err = client.Do(req, resp)
+
+	if !isConnectionReset(err) {
+		t.Fatalf("Expected connection reset error")
```
erikdubbelboer, line 2957:
```suggestion
		t.Fatal("Expected connection reset error")
```

erikdubbelboer (changes requested):

client_test.go
ArminBTVS: Thanks! Did change all formatted strings to not formatted in tests as suggested.

erikdubbelboer:
Looks like the tests are still failing on windows:
```
--- FAIL: TestRstConnResponseWhileSending (0.00s)
    client_test.go:2898: the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection
```

ArminBTVS:
> Looks like the tests are still failing on windows:
> 
> ```
> --- FAIL: TestRstConnResponseWhileSending (0.00s)
>     client_test.go:2898: the server closed connection before returning the first response byte. Make sure the server returns 'Connection: close' response header before closing the connection
> ```

It seems timing dependent. Bad!

Was able to reproduce this behavior on windows only when running the tests for 100 times in a loop. I will further look into it.

ArminBTVS:
@erikdubbelboer so I further investigated on the failing test on Windows. When I repeat the test for >100 times it will fail. But only on Windows. It does not happen on Linux. The issue goes all the way down to the actual syscalls.

So what happens on Windows is that even after I called `_, err = conn.Write([]byte("HTTP/1.1 418 Teapot\r\n\r\n"))` (which used [WSASend](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsasend)) in the server go routine, the read of the connection will immediately fail with an connection reset error, even if I closed the connection after conn.Write succeeded without any error.

What I wanted to simulate in the test is the following simplified traffic:

```
client -> server REQ
client <- server RESP
client <- server RST
```

But it seems that I am not really able to do so on Windows. I could run the test on Linux only? Or add a sleep between write and forcefully closing the connection? Or do you have any thoughts/ideas on this?

erikdubbelboer:
Could also have to do with using sockets on the same machine instead of between different machines. I have seen TCP behave differently that case other times before (even on Linux). Just ignore the test on windows, we don't have great windows support either way 😅

erikdubbelboer:
Thanks!

PR #1238, opened 2022-03-05
Add connection pool queuing strategies in HostClient.

### LINKED ISSUE #1236: Configurable LIFO, FIFO client connection pooling
Hello, 

we use fasthttp in our gateway and in front of the gateway is a nginx. While monitoring one of our more load intensive applications we could see, that the load balance is distributed equal after nginx, but after our gateway the load balance was some how screwed up.
After digging into the code we found out, that the cause of this is, that fasthttp is using a LIFO strategy for the connection pooling.

It would be great if the connection pooling strategy could be made configurable so that the user can decide which pooling strategy should be used.

As you can see in the pictures below after the nginx the load balance is equal over the different connections
Load_balance_after_nginx

But after the Gateway with fasthttp the load balance looks like this:
Load_balance_after_gateway_with_fasthttp
.

u5surf:
Fixes #1236 
* At first, we implement LIFO(conventional) and FIFO which indicate in #1236

### DIFF
--- a/client.go
+++ b/client.go
@@ -618,6 +618,14 @@ type RetryIfFunc func(request *Request) bool
 // TransportFunc wraps every request/response.
 type TransportFunc func(*Request, *Response) error
 
+// ConnPoolStrategyType define strategy of connection pool enqueue/dequeue
+type ConnPoolStrategyType int
+
+const (
+	FIFO ConnPoolStrategyType = iota
+	LIFO
+)
+
 // HostClient balances http requests among hosts listed in Addr.
 //
 // HostClient may be used for balancing load among multiple upstream hosts.
@@ -771,6 +779,9 @@ type HostClient struct {
 
 	// Transport defines a transport-like mechanism that wraps every request/response.
 	Transport TransportFunc
+
+	// Connection pool strategy. Can be either LIFO or FIFO (default).
+	ConnPoolStrategy ConnPoolStrategyType
 
 	clientName  atomic.Value
 	lastUseTime uint32
@@ -1497,6 +1508,10 @@ var (
 	// to broken server.
 	ErrConnectionClosed = errors.New("the server closed connection before returning the first response byte. " +
 		"Make sure the server returns 'Connection: close' response header before closing the connection")
+
+	// ErrConnPoolStrategyNotImpl is returned when HostClient.ConnPoolStrategy is not implement yet.
+	// If you see this error, then you need to check your HostClient configuration.
+	ErrConnPoolStrategyNotImpl = errors.New("connection pool strategy is not implement")
 )
 
 type timeoutError struct{}
@@ -1544,10 +1559,20 @@ func (c *HostClient) acquireConn(reqTimeout time.Duration, connectionClose bool)
 			}
 		}
 	} else {
-		n--
-		cc = c.conns[n]
-		c.conns[n] = nil
-		c.conns = c.conns[:n]
+		switch c.ConnPoolStrategy {
+		case LIFO:
+			n--
+			cc = c.conns[n]
+			c.conns[n] = nil
+			c.conns = c.conns[:n]
+		case FIFO:
+			cc = c.conns[0]
+			copy(c.conns, c.conns[1:])
+			c.conns[n-1] = nil
+			c.conns = c.conns[:n-1]
+		default:
+			return nil, ErrConnPoolStrategyNotImpl
+		}
 	}
 	c.connsLock.Unlock()
 


erikdubbelboer:
I'm curios, have you tested if FIFO results in a better distribution in your case?

u5surf:
Hi, @erikdubbelboer 
I'm not sure FIFO can be better for fairness of using their connections.
@bssd7 Can you test for this patch on your environment to resolve your problem?

bssd7:
i will check it

bssd7:
How can i add this fork into my go.mod file? 
When i try to pull it i get following error:

go get github.com/u5surf/fasthttp@issue-1236                                                                                                                                                        
go get: github.com/u5surf/fasthttp@v1.33.1-0.20220307014346-46010ca954f2: parsing go.mod:
        module declares its path as: github.com/valyala/fasthttp
                but was required as: github.com/u5surf/fasthttp

erikdubbelboer:
It should work if you add this to your `go.mod` file:
```
replace (
    github.com/valyala/fasthttp => github.com/u5surf/fasthttp v1.33.1-0.20220307014346-46010ca954f2
)
```

bssd7:
Hello,

the load balancing is much better now and is working fine see pictures below.

load balancing after nginx:
image

loadbalancing after fasthttp:
image

As you can see the load balancing is much smoother now. The differences in the load balancing which are left are properly an issue from the difference in the connection pool minimum size. Nginx and other http clients are working with a pre allocated pool minimum size and fasthttp does not support a pool minimum size. 

Thank you, great work.

erikdubbelboer:
That is great. The last thing I'm wondering now is if we should maybe make this the default. It seems like this is an issue most people won't be able to recognize and change the configurations. And I don't really see how it could break someone's current use case. So I think it's best if we make this the default, can you change that?

u5surf:
@erikdubbelboer @bssd7 Of course, if maintainers agree with making this default, I'll change such that. Thanks all!

client.go
```diff
@@ -1544,10 +1559,19 @@ func (c *HostClient) acquireConn(reqTimeout time.Duration, connectionClose bool)
...
+		switch c.ConnPoolStrategy {
+		case LIFO:
+			n--
+			cc = c.conns[n]
+			c.conns[n] = nil
+			c.conns = c.conns[:n]
+		case FIFO:
+			cc = c.conns[0]
+			c.conns[0] = nil
+			c.conns = c.conns[1:]
```
erikdubbelboer, line 1571:
```suggestion
			cc = c.conns[0]
			copy(c.conns, c.conns[1:])
			c.conns[n] = nil
			c.conns = c.conns[:n]
```
Otherwise `c.conns = c.conns[1:]` causes an allocation of a new backing slice.

client.go
```diff
@@ -772,6 +780,9 @@ type HostClient struct {
 	// Transport defines a transport-like mechanism that wraps every request/response.
 	Transport TransportFunc
 
+	// Connection pool strategy determines whether LIFO or FIFO storategy for connection pool
+	ConnPoolStrategy ConnPoolStrategyType
```
erikdubbelboer, line 784:
```suggestion
	// Connection pool strategy. Can be either LIFO or FIFO (default).
	ConnPoolStrategy ConnPoolStrategyType
```

client.go
```diff
@@ -1497,6 +1508,10 @@ var (
 	// to broken server.
 	ErrConnectionClosed = errors.New("the server closed connection before returning the first response byte. " +
 		"Make sure the server returns 'Connection: close' response header before closing the connection")
+
+	// ErrConnPoolStrategyNotImpl is returned when HostClient.ConnPoolStrategy is not implement yet.
+	// If you see this error, then you need to check HostClient configuration once.
```
erikdubbelboer, line 1513:
```suggestion
	// If you see this error, then you need to check your HostClient configuration.
```

erikdubbelboer (changes requested):

client.go
erikdubbelboer: See: https://go.dev/play/p/bcrFeRc1I-n

u5surf: @erikdubbelboer 
Thanks reviewing! Your suggestion which use copy without a new allocation is completely right indeed, But your suggested code has failed following test on my environment.

```
--- PASS: TestHostClientMaxConnWaitTimeoutError (0.01s)
=== CONT  TestHostClientConcurrent
panic: runtime error: index out of range [1] with length 1

goroutine 594 [running]:
github.com/valyala/fasthttp.(*HostClient).acquireConn(0xc0005891e0, 0x0, 0x0)
        /home/u5surf/fasthttp/client.go:1571 +0xd45
github.com/valyala/fasthttp.(*HostClient).doNonNilReqResp(0xc0005891e0, 0xc00012e380, 0xc000023800)
        /home/u5surf/fasthttp/client.go:1417 +0x5b2
github.com/valyala/fasthttp.(*HostClient).do(0x1f, 0x20, 0xc000023800)
        /home/u5surf/fasthttp/client.go:1368 +0xc9
github.com/valyala/fasthttp.(*HostClient).Do(0xc0005891e0, 0xc00012e380, 0xc0003064a0)
        /home/u5surf/fasthttp/client.go:1315 +0x114
github.com/valyala/fasthttp.doRequestFollowRedirects(0xc00012e380, 0xc000023800, {0xc0003064a0, 0x1f}, 0x10, {0xc06480, 0xc0005891e0})
        /home/u5surf/fasthttp/client.go:1034 +0x1da
github.com/valyala/fasthttp.doRequestFollowRedirectsBuffer(0xf85100, {0xc000306460, 0x1f, 0x20}, {0xc0003064a0, 0x1f}, {0xc06480, 0xc0005891e0})
        /home/u5surf/fasthttp/client.go:1015 +0x2a5
github.com/valyala/fasthttp.clientGetURL({0xc000306460, 0x1f, 0x20}, {0xc0003064a0, 0x1f}, {0xc06480, 0xc0005891e0})
        /home/u5surf/fasthttp/client.go:886 +0x133
github.com/valyala/fasthttp.(*HostClient).Get(0xb2fb99, {0xc000306460, 0x1f, 0x20}, {0xc0003064a0, 0x1f})
        /home/u5surf/fasthttp/client.go:838 +0x88
github.com/valyala/fasthttp.testClientGet(0xc000224680, {0xc064a0, 0xc0005891e0}, {0xb30193, 0x11}, 0x1e)
        /home/u5surf/fasthttp/client_test.go:2326 +0x175
github.com/valyala/fasthttp.testHostClientGet(...)
        /home/u5surf/fasthttp/client_test.go:2410
github.com/valyala/fasthttp.TestHostClientConcurrent.func1()
        /home/u5surf/fasthttp/client_test.go:2315 +0xd3
created by github.com/valyala/fasthttp.TestHostClientConcurrent
        /home/u5surf/fasthttp/client_test.go:2313 +0x145
FAIL    github.com/valyala/fasthttp     1.046s
```

my go version
```
 go version
go version go1.17.8 linux/amd64
```

Thus, I just modified following using copy, then passed all tests.
```
@@ -1568,7 +1568,8 @@ func (c *HostClient) acquireConn(reqTimeout time.Duration, connectionClose bool)
                case FIFO:
                        cc = c.conns[0]
                        c.conns[0] = nil
-                       c.conns = c.conns[1:]
+                       copy(c.conns, c.conns[1:])
+                       c.conns = c.conns[:len(c.conns)-1]
                default:
                        return nil, ErrConnPoolStrategyNotImpl
                }
```

As it relates to this patch, I found a curious article about performance of FIFO implementation with copy.
https://www.reddit.com/r/golang/comments/24dfn8/someones_right_building_a_queue_list_or_slices/ch62raz/?context=8&depth=9

erikdubbelboer: My bad, it should have been n-1. You already have the length, no need to call len again.

erikdubbelboer:
Thanks for building this!

Issue #504, opened 2018-12-28
hao  to dynamic add client 


miniframework:
code example:
```go
c := &fasthttp.HostClient{Addr: addr}
lbc.Clients = append(lbc.Clients, c)
lbc.Do()  
c1 := &fasthttp.HostClient{Addr: addr}
lbc.Clients = append(lbc.Clients, c1)
lbc.Do()  
```
i see code ：
cc.once.Do(cc.init)      run once init
so c1 not add in cc.cs

hao to dynamic add client  to lbClient 
service discovery need

erikdubbelboer:
I'm afraid this is not possible at the moment. It is fairly easy to add a `Add` method to `LBClient` to add new clients after first use. I don't have time to write the code at the moment. Pull requests are always welcome!

miniframework:
easy add a client to LBClient.

func (cc *LBClient) AddClient(client BalancingClient) {
	cc.cs = append(cc.cs, &lbClient{
		c:           client,
		healthCheck: cc.HealthCheck,
	})
}

Hao to del client by client.Addr      
question,  if i implement other client interface , I have to change this function， 
I don't want to do that. hao?

func (cc *LBClient) DelClientByAddr(string addr) {
   k,c := range cc.cs {
       switch  c.(type) {
        case HostClient:
                  if c.(HostClient).Addr == addr {
                      cc.cs = append(cc.cs[:k], cc.cs[k+1:]...)   
                   }
        case PipelineClient:
                .....
      }
  }
}
@erikdubbelboer

erikdubbelboer:
Is deleting by address really useful? I would think passing a `BalancingClient` and removing that from the list makes more sense? If someone then wants to delete by address they can just loop though the `Clients` and find the one with the address that they want to have removed.

Also right now you have a race condition if you do this from multiple goroutines. You will need to add a `sync.Mutex` to `LBClient` to protect against this. And since `Clients` is public I guess this mutex should also be public even though that's kinda ugly.

miniframework:
type BalancingClient interface {
	DoDeadline(req *Request, resp *Response, deadline time.Time) error
	PendingRequests() int
       Addr() string     ? 
}
HostClient add Addr()
PipelineClient add Addr()

miniframework:
@erikdubbelboer  
i used  service discovery.   after-end register addr to etcd， proxy add addr to lbClient ,  after-end hung, proxy del addr eq client from lbClient.

type BalancingClient interface {
[...]

PR #1243, opened 2022-03-10
support adding/removing clients from LBClient

treethought:
addresses https://github.com/valyala/fasthttp/issues/504 and is a followup to https://github.com/valyala/fasthttp/pull/556, addressing the outstanding feedback.

Adds support for adding/removing clients from LBClient utilizing a mutex as well as a user-defined callback for determining which HostClients to be removed

### DIFF
--- a/lbclient.go
+++ b/lbclient.go
@@ -50,7 +50,7 @@ type LBClient struct {
 	cs []*lbClient
 
 	once sync.Once
-	mu   sync.Mutex
+	mu   sync.RWMutex
 }
 
 // DefaultLBClientTimeout is the default request timeout used by LBClient
@@ -121,6 +121,7 @@ func (cc *LBClient) RemoveClients(rc func(BalancingClient) bool) int {
 func (cc *LBClient) get() *lbClient {
 	cc.once.Do(cc.init)
 
+	cc.mu.RLock()
 	cs := cc.cs
 
 	minC := cs[0]
@@ -135,6 +136,7 @@ func (cc *LBClient) get() *lbClient {
 			minT = t
 		}
 	}
+	cc.mu.RUnlock()
 	return minC
 }
 
--- a/lbclient.go
+++ b/lbclient.go
@@ -50,6 +50,7 @@ type LBClient struct {
 	cs []*lbClient
 
 	once sync.Once
+	mu   sync.Mutex
 }
 
 // DefaultLBClientTimeout is the default request timeout used by LBClient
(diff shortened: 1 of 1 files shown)


erikdubbelboer (changes requested):
You also have add Lock/Unlock to `.get`. And since it will be more than 99% reads using the Mutex I would a [`sync.RWMutex`](https://pkg.go.dev/sync#RWMutex)

lbclient.go
```diff
@@ -80,6 +81,7 @@ func (cc *LBClient) Do(req *Request, resp *Response) error {
 }
 
 func (cc *LBClient) init() {
+	cc.mu.Lock()
```
erikdubbelboer, line 84: Since this function can panic I think it's better to `defer cc.mu.Unlock()`.

erikdubbelboer (changes requested):

lbclient.go
```diff
@@ -91,9 +94,36 @@ func (cc *LBClient) init() {
...
+}
+
+// RemoveClients removes clients using the provided callback
+// if rc returns true, the passed client will be removed
+// returns the new total number of clients
+func (cc *LBClient) RemoveClients(rc func(BalancingClient) bool) int {
+	cc.mu.Lock()
+	for idx, cs := range cc.cs {
+		if rc(cs.c) {
+			cc.cs = append(cc.cs[:idx], cc.cs[idx+1])
```
erikdubbelboer, line 116: This is not a correct way to filter a slice. See: https://go.dev/play/p/9y9DyCJ_xDD

erikdubbelboer (changes requested):

lbclient.go
treethought: ah, my bad. addressed

erikdubbelboer (changes requested):
That is still not a correct way to filter a slice, see: https://go.dev/play/p/JM8EdsOgVMK
For a correct way see: https://github.com/golang/go/wiki/SliceTricks#filter-in-place
But make sure you set the unused entries at the end to `nil` before you resize the slice so there are no references to those clients anymore.

treethought:
thanks for the tip, updated the filtering. https://go.dev/play/p/UhjinrfN1qB

lbclient.go
```diff
@@ -91,9 +94,42 @@ func (cc *LBClient) init() {
...
+	cc.mu.Lock()
+	n := 0
+	for _, cs := range cc.cs {
+		if rc(cs.c) {
+			cs = nil
+			continue
+		}
+		cc.cs[n] = cs
+		n++
+	}
```
erikdubbelboer, line 122:
```suggestion
			continue
		}
		cc.cs[n] = cs
		n++
	}
	for i := n; i < len(cc.cs); i++ {
		cc.cs[i] = nil
	}
```

erikdubbelboer: I mean like this so that the `cc.cs` entries that you remove from the slice really aren't referenced anymore.

erikdubbelboer (changes requested):

PR #1249, opened 2022-03-16
Imporve AppendHTMLEscape fast path

zhangyunhao116:
The `<` is occurs more frequently in the HTML string.

> Move `<`

```
name                 old time/op  new time/op  delta
AppendHTMLEscape-16  99.3ns ± 1%  93.8ns ± 1%  -5.50%  (p=0.000 n=10+10)
```

And I am wondering maybe we could disable the fast path? Looks like it is not used in most cases.

>  Disable fast path

```
name                 old time/op  new time/op  delta
AppendHTMLEscape-16  99.3ns ± 1%  82.3ns ± 3%  -17.10%  (p=0.000 n=10+10)
```

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -11,7 +11,6 @@ import (
 	"math"
 	"net"
 	"reflect"
-	"strings"
 	"sync"
 	"time"
 	"unsafe"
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
How about using [`strings.IndexAny`](https://pkg.go.dev/strings#IndexAny) to find the first special character. If it returns `-1` you can `return append(dst, s...)`. Otherwise you can `dst = append(dst, s[:i]...)` and continue the `for` loop from there. I'm not 100% sure if `strings.IndexAny` doesn't allocate though.

zhangyunhao116:
The `strings.IndexByte` actually use assembly directly(SSE AVX etc), but `strings.IndexAny` is more complicate (contains more Go codes), it will be slower in this case.

>  use `strings.IndexAny`

```
name                 old time/op  new time/op   delta
AppendHTMLEscape-16  99.3ns ± 1%  106.2ns ± 1%  +6.92%  (p=0.000 n=10+10)
```

erikdubbelboer:
In that case I'm in favor of removing the fast path as well.

But in that case isn't it also possible to use `strings.IndexByte` to find the next index character to encode instead of looping though the whole string one by one?

zhangyunhao116:
> But in that case isn't it also possible to use `strings.IndexByte` to find the next index character to encode instead of looping though the whole string one by one?

I haven't come up with a good plan yet, the strategy seems complicated, maybe we could do it in the future.
 
This PR removes the fast path now.
```
name                 old time/op  new time/op  delta
AppendHTMLEscape-16  99.3ns ± 1%  82.3ns ± 3%  -17.10%  (p=0.000 n=10+10)
```

erikdubbelboer:
Thanks. I like simplifying the code if possible so I think this is also a good solution.

PR #1250, opened 2022-03-16
bytesconv: add appropriate build tags for s390x

enr0n:
The bytesconv 32-bit tests fail on s390x, because it is a 64-bit
architecture. Add the appropriate build flags so that 32-bit tests do
not run on this architecture.


erikdubbelboer:
Thanks!

Issue #1241, opened 2022-03-10
Extra goroutine after server gracefull shutdown


denvolj:
Seems that updateServerDate() creating goroutine that can't be exited.
This is causing some test tools (like goleak) to report about extra goroutines even after server's gracefull shutdown.

denvolj:
Dirtyhack: you can ignore by this way
`
	goleak.VerifyNone(
		t,
		goleak.IgnoreTopFunction("time.Sleep"),
	)
`

erikdubbelboer:
I'm open to a pull request to somehow fix this. But I don't see any other way than an extra method to stop it. And you'll have to build your re-settable version of sync.Once I guess.

PR #1251, opened 2022-03-16
Fix goroutine leaks of updateServsrDate (#1241).

u5surf:
Fixes: #1241 

* updateServerDate's goroutine makes available to catch s.done closing

### DIFF
--- a/header.go
+++ b/header.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"sync"
+	"sync/atomic"
 	"time"
 )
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
I wish it was this simple but what now if someone calls `ResponseHeader.AppendBytes` without using `Server`? For example when they are using `Client` only. Or maybe they don't use `Server` or `Client` either and are just using `ResponseHeader` for something else. They will get invalid values.

erikdubbelboer:
But this breaks `ResponseHeader.AppendBytes` without `Server` so I can never merge it like this.

erikdubbelboer:
Sorry I missed that. I guess this could work for a subset of the problem. But I'm not sure if its worth making the code more complex just to fix a subset of this problem. I still don't see this one update routing running in the background as an issue. It's better to just ignore it with the problems that check for goroutine leaks like someone else suggested.

PR #1263, opened 2022-04-07
Update tlsClientHandshake

moredure:
I have such a code in a production crawlers for more than 1 year and it was just fine.
Also we can cleanup peercertificates+verifiedChains slice with nils since they are not used after handshake and in case of many concurrent connections to the same server may occupy some space.
```go
tlsConn.Handshake()
...
state := tlsConn.ConnectionState()
for i := range state.PeerCertificates {
    state.PeerCertificates[i] = nil
}
for i := range state.VerifiedChains { // they contains references to peerCertificates
    state.VerifiedChains[i] = nil
}
```
in some cases it can free up pretty high amount of in-use space

### DIFF
--- a/client.go
+++ b/client.go
@@ -1994,35 +1994,35 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
 // ErrTLSHandshakeTimeout indicates there is a timeout from tls handshake.
 var ErrTLSHandshakeTimeout = errors.New("tls handshake timed out")
 
-func tlsClientHandshake(rawConn net.Conn, tlsConfig *tls.Config, deadline time.Time) (net.Conn, error) {
-	conn := tls.Client(rawConn, tlsConfig)
-	err := conn.SetReadDeadline(deadline)
+func tlsClientHandshake(rawConn net.Conn, tlsConfig *tls.Config, deadline time.Time) (conn net.Conn, err error) {
+	defer func() {
+		if err != nil {
+			rawConn.Close()
+			return nil, err
+		}
+	}()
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1994,41 +1994,26 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
...
 	conn := tls.Client(rawConn, tlsConfig)
-
-	go func() {
-		ch <- conn.Handshake()
-	}()
-
-	select {
-	case <-tc.C:
+	_ = conn.SetReadDeadline(deadline)
+	_ = conn.SetWriteDeadline(deadline)
```
erikdubbelboer, line 2000: Might as well return the error if there is any since this function has an error return value.

erikdubbelboer (changes requested):

client.go
```diff
@@ -1994,41 +1994,41 @@ func (c *HostClient) cachedTLSConfig(addr string) *tls.Config {
...
+	}()
+	conn := tls.Client(rawConn, tlsConfig)
+	err := conn.SetReadDeadline(deadline)
+	if err != nil {
+		return nil, err
+	}
+	err = conn.SetWriteDeadline(deadline)
+	if err != nil {
+		return nil, err
+	}
```
erikdubbelboer, line 2011: Sorry bit pedantic but you can combine these with `conn.SetDeadline`. Same below. After that I can merge it.

erikdubbelboer (changes requested):

client.go
moredure: Indeed!

erikdubbelboer:
Thanks!

PR #1264, opened 2022-04-08
Added Windows support and removed some panics

### LINKED ISSUE #1101: FS handlers don't cooperate well with handleRequest on Windows

mxmauro:
This PR aims to fix #1101 and improve the changes proposed by #1108 

List of changes:

* Handling and proper conversion from/to backslashes.
* `rootFS.Root` is now empty and `ServeFile` has the control to convert a relative path to absolute.
* Using `filepath.IsAbs` to detect if a path is absolute or relative.

Extras:

* Some minor code linting.
* Removed `panic`s and now an error is returned. Crashing an app because the underlying OS calls does not work as expected, is not good. Also the function that may panic, already returns an errors for other situations.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -99,10 +99,13 @@ func ServeFile(ctx *RequestCtx, path string) {
 	rootFSOnce.Do(func() {
 		rootFSHandler = rootFS.NewRequestHandler()
 	})
-	if len(path) == 0 || path[0] != '/' {
+
+	if len(path) == 0 || !filepath.IsAbs(path) {
 		// extend relative path to absolute path
-		hasTrailingSlash := len(path) > 0 && path[len(path)-1] == '/'
+		hasTrailingSlash := len(path) > 0 && (path[len(path)-1] == '/' || path[len(path)-1] == '\\')
+
 		var err error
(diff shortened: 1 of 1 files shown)


mxmauro:
Hi, I seeing the Windows test some times throws timeout in different PRs. Is it possible to run the checks again? BTW: Tested also on my Windows machine.

erikdubbelboer:
Looks sane to me. Thanks!

mxmauro:
You are welcome and thanks also to you.

Issue #1265, opened 2022-04-08
nil pointer when fasthttp.(*Request).ReleaseBody(...)


zzzzwc:
```
panic: runtime error: invalid memory address or nil pointer dereference

[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xabd0ef]

goroutine 598556 [running]:
github.com/valyala/fasthttp.(*Request).ReleaseBody(...)
        /.../github.com/valyala/fasthttp@v1.31.0/http.go:585
...
 previous code in handler
```

may be there  need a nil check of req.body at https://github.com/valyala/fasthttp/blob/7a5afddf5b805a022f8e81281c772c11600da2f4/http.go#L585 and  https://github.com/valyala/fasthttp/blob/7a5afddf5b805a022f8e81281c772c11600da2f4/http.go#L600

DuckRui:
> ```
> panic: runtime error: invalid memory address or nil pointer dereference
> 
> [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xabd0ef]
> 
> goroutine 598556 [running]:
> github.com/valyala/fasthttp.(*Request).ReleaseBody(...)
>         /.../github.com/valyala/fasthttp@v1.31.0/http.go:585
> ...
>  previous code in handler
> ```
> 
> may be there need a nil check of req.body at
> 
> https://github.com/valyala/fasthttp/blob/7a5afddf5b805a022f8e81281c772c11600da2f4/http.go#L585
> 
> and
> https://github.com/valyala/fasthttp/blob/7a5afddf5b805a022f8e81281c772c11600da2f4/http.go#L600

Hello, I'm curious how you found this problem, as I did not find which function call the ReleaseBody func. Would you like tell me about process of finding this problem? Thank you very much!

PR #1266, opened 2022-04-08
add nil check of req.body and resp.body on ReleaseBody

zzzzwc:
resolve: https://github.com/valyala/fasthttp/issues/1265


erikdubbelboer:
Thanks!

PR #1269, opened 2022-04-16
Fix for neverending date updater goroutine

mxmauro:
This patch aims to fix issue #1241 

Instead of launching a never ending go routine, the same routine is started and destroyed manually in the main listener loop.

Because connections can be processed by an external listener using the `ServeConn`, a specific behavior is used for this scenario. The slow path (because of mutex usage) returns a buffer that is updated every second.

BTW: Some linting was applied to minimize warnings.

### DIFF
--- a/allocation_test.go
+++ b/allocation_test.go
@@ -47,7 +47,9 @@ func TestAllocationClient(t *testing.T) {
 		Handler: func(ctx *RequestCtx) {
 		},
 	}
-	go s.Serve(ln) //nolint:errcheck
+	go func() {
+		_ = s.Serve(ln) //nolint:errcheck
+	}()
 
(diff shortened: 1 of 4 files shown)


erikdubbelboer (commented):
I'm sorry but it feels like a bit overkill for something that isn't really a problem. As shown in https://github.com/valyala/fasthttp/issues/1241#issuecomment-1063845743 you can just ignore the goroutine in goleak. I just don't think it's worth the complexity and making the slow path a lot slower.

mxmauro:
Hi @erikdubbelboer, no problem. I'll close this.

erikdubbelboer:
Thanks for trying tho!

PR #1270, opened 2022-04-17
Update fileserver.go

Yariya:


erikdubbelboer:
Thanks for your pull request. But I'm sorry it's just a sample so it doesn't need a proper shutdown. A proper shutdown would also require the server to be shut down properly and that wouls just complicate the example too much for what it really tries to show.

erikdubbelboer:
Thanks for your pull request. But I'm sorry it's just a sample so it doesn't need a proper shutdown. A proper shutdown would also require the server to be shut down properly and that wouls just complicate the example too much for what it really tries to show.

PR #1272, opened 2022-04-19
optimize

tylitianrui:
optimize execution logic.  
Imaging  the  index   of char`%` in  slice `src` is   `len(src)-1`, the most  loops   is unnecessary

PR #1275, opened 2022-04-21
optimize

tylitianrui:
optimize  the logic of `fast path`. similar [pr #1272 ](https://github.com/valyala/fasthttp/pull/1272)

### DIFF
--- a/args.go
+++ b/args.go
@@ -544,13 +544,22 @@ func (s *argsScanner) next(kv *argsKV) bool {
 }
 
 func decodeArgAppend(dst, src []byte) []byte {
-	if bytes.IndexByte(src, '%') < 0 && bytes.IndexByte(src, '+') < 0 {
+	idxPercent := bytes.IndexByte(src, '%')
+	idxPlus := bytes.IndexByte(src, '+')
+	if idxPercent < 0 && idxPlus < 0 {
 		// fast path: src doesn't contain encoded chars
 		return append(dst, src...)
 	}
 
+	idx := min(idxPercent, idxPlus)
+	if idx > 0 {
+		dst := append(dst, src[:idx]...)
+	} else {
+		idx = 0
+	}
+
 	// slow path
-	for i := 0; i < len(src); i++ {
+	for i := idx; i < len(src); i++ {
 		c := src[i]
 		if c == '%' {
 			if i+2 >= len(src) {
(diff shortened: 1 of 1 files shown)


args.go
```diff
@@ -544,13 +544,22 @@ func (s *argsScanner) next(kv *argsKV) bool {
...
-	if bytes.IndexByte(src, '%') < 0 && bytes.IndexByte(src, '+') < 0 {
+	idxPercent := bytes.IndexByte(src, '%')
+	idxPlus := bytes.IndexByte(src, '+')
+	if idxPercent < 0 && idxPlus < 0 {
 		// fast path: src doesn't contain encoded chars
 		return append(dst, src...)
 	}
 
+	idx := min(idxPercent, idxPlus)
+	if idx > 0 {
```
erikdubbelboer, line 555: If one of them is -1 you can still start at the other. So I would change this block to be smarter without the min func.

erikdubbelboer (changes requested):

args.go
tylitianrui: thx,   it  would  be better

args.go
```diff
@@ -544,13 +544,31 @@ func (s *argsScanner) next(kv *argsKV) bool {
...
 func decodeArgAppend(dst, src []byte) []byte {
-	if bytes.IndexByte(src, '%') < 0 && bytes.IndexByte(src, '+') < 0 {
+	idxPercent := bytes.IndexByte(src, '%')
+	idxPlus := bytes.IndexByte(src, '+')
+	if idxPercent < 0 && idxPlus < 0 {
 		// fast path: src doesn't contain encoded chars
 		return append(dst, src...)
 	}
 
+	idx := 0
```
tylitianrui, line 554: i dont think   it  is  graceful. better solution is  welcome

erikdubbelboer: I would do this, which has less comparisons:
```go
idx := 0
if idxPercent == -1 {
    if idxPlus != -1 {
        idx = idxPlus
    }
} else if idxPlus == -1 {
    idx = idxPercent
} else if idxPercent > idxPlus {
    idx = idxPlus
} else {
    idx = idxPercent
}
```

erikdubbelboer (changes requested):

args.go
tylitianrui: thx，it  be  better. i  will  change it.
but the block
```
if idxPlus != -1 {   
     idx = idxPlus   
} 
```
is  unnecessary.    it is  not  possible the  `idxPercent `  and `idxPlus`  both  are  -1

erikdubbelboer: You could combine it with the check above to do less comparisons.

tylitianrui: > You could combine it with the check above to do less comparisons.

thanks @erikdubbelboer

erikdubbelboer:
Thanks!

PR #1298, opened 2022-05-11
fix userdata kv.key undefined behavior

yandd:
In https://github.com/valyala/fasthttp/blob/master/userdata.go#L33, appending kv.key[:0] is undefined behavior.

### DIFF
--- a/userdata.go
+++ b/userdata.go
@@ -16,7 +16,7 @@ func (d *userData) Set(key string, value interface{}) {
 	n := len(args)
 	for i := 0; i < n; i++ {
 		kv := &args[i]
-		if string(kv.key) == key {
+		if b2s(kv.key) == key {
 			kv.value = value
 			return
 		}
@@ -30,14 +30,14 @@ func (d *userData) Set(key string, value interface{}) {
 	if c > n {
 		args = args[:n+1]
 		kv := &args[n]
-		kv.key = append(kv.key[:0], key...)
+		kv.key = s2b(key)
 		kv.value = value
 		*d = args
 		return
 	}
 
 	kv := userDataKV{}
-	kv.key = append(kv.key[:0], key...)
+	kv.key = s2b(key)
 	kv.value = value
 	*d = append(args, kv)
 }
@@ -51,7 +51,7 @@ func (d *userData) Get(key string) interface{} {
 	n := len(args)
 	for i := 0; i < n; i++ {
 		kv := &args[i]
-		if string(kv.key) == key {
+		if b2s(kv.key) == key {
 			return kv.value
 		}
 	}
@@ -79,7 +79,7 @@ func (d *userData) Remove(key string) {
 	n := len(args)
 	for i := 0; i < n; i++ {
 		kv := &args[i]
-		if string(kv.key) == key {
+		if b2s(kv.key) == key {
 			n--
 			args[i] = args[n]
 			args[n].value = nil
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Why do you think it would be undefined behavior? `kv.key` is just a byte slice that we reuse. Yes it might contain a previous value but that doesn't matter cause at that point the `kv` is unused. The `b2s` you added instead of string casts don't change anything either as Go has an optimization to not allocate when it's used in a comparison like that.

yandd:
>`kv.key.Data` is a wild pointer and `kv.key.Cap` also can be a number greater than 0, `append(kv.key[:0], key...)` will direct write `key` to `kv.key.Data`, here can be a undefined behavior!

yandd:
> Why do you think it would be undefined behavior? `kv.key` is just a byte slice that we reuse. Yes it might contain a previous value but that doesn't matter cause at that point the `kv` is unused. The `b2s` you added instead of string casts don't change anything either as Go has an optimization to not allocate when it's used in a comparison like that.

Please check this test:
```
func TestUserDataKv(t *testing.T) {
	var (
		keys = map[string]struct{}{
			"method":                  {},
			"serverSpan":              {},
			"uber-trace-id":           {},
			"originClientRequestHost": {},
			"methodOrigin":            {},
			"baseRule":                {},
			"mappedPath":              {},
			"retryRequestFlag":        {},
			"requestExecCount":        {},
			"excludeInstanceIdMap":    {},
			"hostStrategy":            {},
			"instanceKey":             {},
			"responseError":           {},
		}
		ud   userData
		dump = func(d userData) {
			for _, kv := range d {
				t.Log((*reflect.SliceHeader)(unsafe.Pointer(&(kv.key))))
			}
		}
		test = func(d userData) {
			for _, kv := range d {
				if _, ok := keys[string(kv.key)]; !ok {
					dump(d)
					t.Fatalf("invaid key: %s", kv.key)
				}
			}
		}
	)

	for {
		test(ud)
		ud.Reset()
		test(ud)
		ud.Set("method", "")
		test(ud)
		ud.Set("serverSpan", "")
		test(ud)
		ud.Set("uber-trace-id", "")
		test(ud)
		ud.Set("originClientRequestHost", "")
		test(ud)
		ud.Set("methodOrigin", "")
		test(ud)
		ud.Set("baseRule", "")
		test(ud)
		ud.Set("mappedPath", "")
		test(ud)
		ud.Set("retryRequestFlag", "")
		test(ud)
		ud.Set("requestExecCount", "")
		test(ud)
		ud.Set("excludeInstanceIdMap", "")
		test(ud)
		ud.Set("hostStrategy", "")
		test(ud)
		ud.Set("instanceKey", "")
		test(ud)
		ud.Set("responseError", "")
		test(ud)
		ud.Remove("serverSpan")
		test(ud)
	}
}
```

yandd:
Or just need to fix `Remove`?
```
func (d *userData) Remove(key string) {
	args := *d
	n := len(args)
	for i := 0; i < n; i++ {
		kv := &args[i]
		if string(kv.key) == key {
			n--
			args[i] = args[n]
			args[n].key = nil               // <= add this line
			args[n].value = nil
			args = args[:n]
			*d = args
			return
		}
	}
}
```

erikdubbelboer:
Can you add a test case to `userdata_test.go` that fails before your change and passes after?

yandd:
> Can you add a test case to `userdata_test.go` that fails before your change and passes after?

Done.

erikdubbelboer:
Thanks! I get it now. But in that case I prefer this much easier fix: https://github.com/valyala/fasthttp/commit/99610791968a1b10dd28aa7dbf426ea3bad72aed This way the allocations for `key` can be reused.

PR #1305, opened 2022-05-26
add the functions to get host clients

xuxiao415:
After obtaining the client, we can monitor the client's indicators

### DIFF
--- a/client.go
+++ b/client.go
@@ -307,6 +307,16 @@ type Client struct {
 	writerPool sync.Pool
 }
 
+// GetHostClients return the map of Http HostClient
+func (c *Client) GetHostClients() map[string]*HostClient {
+	return c.m
+}
+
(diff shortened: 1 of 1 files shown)


erikdubbelboer (changes requested):
I'm curious, what indicators do you want to monitor?

Doing this is very unsafe as access to `c.m` and `c.ms` are protected by a `sync.Mutex` that isn't exposed. So using these functions will certainly lead to race conditions the way you implemented them now.

client.go
```diff
@@ -307,6 +307,16 @@ type Client struct {
 	writerPool sync.Pool
 }
 
+// GetHostClients return the map of Http HostClient
+func (c *Client) GetHostClients() map[string]*HostClient {
+	return c.m
```
tylitianrui, line 312: i dont think it is a good idea. it expose  what is protected

PR #1311, opened 2022-06-04
Response.ContentEncoding(): store as field and avoid using Header.SetCanonical()

stokito:
The CE is not so often used for plain APIs responses and even not so often used for static files and on the fly compression.
But still it should be checked each time so instead of peek headers map it will be faster to check a field.
Also having a dedicated field getter and setter simplifies code.

### DIFF
--- a/header.go
+++ b/header.go
@@ -104,7 +104,7 @@ func (h *ResponseHeader) SetContentRange(startPos, endPos, contentLength int) {
 	b = AppendUint(b, contentLength)
 	h.bufKV.value = b
 
-	h.SetCanonical(strContentRange, h.bufKV.value)
+	h.setNonSpecial(strContentRange, h.bufKV.value)
 }
 
 // SetByteRange sets 'Range: bytes=startPos-endPos' header.
@@ -126,7 +126,7 @@ func (h *RequestHeader) SetByteRange(startPos, endPos int) {
 	}
 	h.bufKV.value = b
 
-	h.SetCanonical(strRange, h.bufKV.value)
+	h.setNonSpecial(strRange, h.bufKV.value)
 }
 
 // StatusCode returns response status code.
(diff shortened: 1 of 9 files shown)


header.go
```diff
@@ -1262,6 +1289,11 @@ func (h *ResponseHeader) setSpecialHeader(key, value []byte) bool {
 	return false
 }
 
+// putHeader directly put into map i.e. not a basic header
+func (h *ResponseHeader) putHeader(key []byte, value []byte) {
```
erikdubbelboer, line 1293: I think the name is a bit unclear. I would just call it `setNonSpecial`.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

stokito:
Thank you, Eric. You may noticed that for the Request I didn't added the `ConentEncoding()`.
The reason why I not did it is because the [compressed requests aren't so often used](https://serverfault.com/questions/56700/is-it-possible-to-enable-http-compression-for-requests).
So adding a new field here may be questionable.
In the OpenRTB spec they are mentioned

> To enable compression on the bid request, it must first be agreed upon between the exchange and the
> bidder that this is supported. This is similar to when a custom data format is used since the exchange
> has to know both format and encoding before sending the bid request. If the bidder supports it, the
> exchange should indicate it is sending a gzip compressed bid request by setting the HTTP 1.1 ContentEncoding header. The encoding value used should be “gzip”.
> Content-Encoding: gzip
> If this header is not set then it is assumed that the request content isn’t encoded. In HTTP 1.1, the
> Content-Encoding header is usually only used for response content. However by using this header for
> the request content as well we are able to indicate a request is compressed regardless of the data
> format used. This is useful since even binary data formats can benefit from being compressed.
> 

And some providers supports this e.g. https://docs.openx.com/demand-partners/compression-traffic-gzip/ or https://protocol.bidswitch.com/rtb/data-format.html

Even while not so many such http clients but under a heavy load this may have a difference.
Adding a new field has a downside because makes `Peek()` or `VisitAll()` slower.
So I don't know if it worth to add a separate field.
Maybe just to keep Resp/Req method similar we can add the `ContentEncoding()` method that internally will just peek it from the headers map. Similarly to as `Referer()` works.

erikdubbelboer:
Lets leave it for now until someone complains about it.

Issue #1320, opened 2022-06-14
use tcpKeepaliveListener and reuseport together 


hawell:
I need the functionality of both tcpKeepaliveListener and reuseport.
something like this:
```
func ListenAndServe() error {
	s := &fasthttp.Server{
		Handler: handler,
		TCPKeepalive: true,
		TCPKeepalivePeriod: 30 * time.Second,
	}
	ln, err := reuseport.Listen("tcp4", ":8080")
	if err != nil {
		return err
	}
	if tcpln, ok := ln.(*net.TCPListener); ok {
		return s.Serve(tcpKeepaliveListener{
			TCPListener:     tcpln,
			keepalive:       s.TCPKeepalive,
			keepalivePeriod: s.TCPKeepalivePeriod,
		})
	}
	return s.Serve(ln)
}

```
is there a reason why tcpKeepaliveListener is private?
is there another way to have both functionalities that I'm missing here?
thanks a lot

stokito:
@erikdubbelboer could you please tell me if I should use the reuseport?
I have a heavy loaded OpenRtb service with huge QPS of short requests that are processed in subsecond.
I can't understand if the reuseport is great then why it's not used by default?
What are cons?

erikdubbelboer:
If you have to ask if you should use reuseport then you don't need to. This is only for really specific cases where you have multiple processes listening on the same port.

erikdubbelboer:
Fixed in #1324

PR #1324, opened 2022-06-20
Consolidate TCPKeepalive in server.Serve (#1320)

u5surf:
#1320 

consolidate Accept behavior in server.Serve(acceptConn).


erikdubbelboer:
Thanks!

Issue #1190, opened 2022-01-10
Shutdown() function freezes execution if at least one request was performed (v1.31.0)


waclawthedev:
How to reproduce:

**1. Execute the next code:** 
```
func main() {
	server := fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) {
		ctx.WriteString("OK")
		return
	}}
	go server.ListenAndServe("0.0.0.0:8000")
	println("Press ENTER to shutdown")
	bufio.NewReader(os.Stdin).ReadString('\n')
	println("Shutdown ...")
	startTime := time.Now()
	server.Shutdown()
	fmt.Printf("%v", time.Since(startTime))
}
```

**2. Try to press ENTER after performing requests on 127.0.0.1:8000. The output something like that:**

```
Press ENTER to shutdown

Shutdown ...
36.1746735s
```

_**If you will try to press ENTER just after server started (without requests on 127.0.0.1:8000) - everything is ok:**_
```
Press ENTER to shutdown

Shutdown ...
101.211041ms
```

erikdubbelboer:
This has been fixed by https://github.com/valyala/fasthttp/pull/1155 in the just released `v1.32.0`

PR #1326, opened 2022-06-24
fix shutdown fail

zc2638:
When the server accepts the request, calling shutdown will permanently fail to exit.

Here is my test code.
```golang
	wg, ctx := errgroup.WithContext(context.Background())
	wg.Go(func() error {
		ch := make(chan os.Signal, 1)
		signal.Notify(ch, syscall.SIGTERM, syscall.SIGQUIT, syscall.SIGINT)
		select {
		case <-ctx.Done():
			return nil
		case sig := <-ch:
			return fmt.Errorf("server shutdown: %v", sig)
		}
	})
	s := &fasthttp.Server{Handler: r.Handler}
	wg.Go(func() error {
		return s.ListenAndServe(":8080")
	})
	wg.Go(func() error {
		<-ctx.Done()
		return s.Shutdown()
	})
	log.Fatal(wg.Wait())
```

In the logic of Shutdown, after the server accepts the request, `s.open` will never be `0`.
So the server process will not be able to exit.
```golang
	for {
		if open := atomic.LoadInt32(&s.open); open == 0 {
			break
		}
		// This is not an optimal solution but using a sync.WaitGroup
		// here causes data races as it's hard to prevent Add() to be called
		// while Wait() is waiting.
		time.Sleep(time.Millisecond * 100)
	}
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -428,6 +428,8 @@ type Server struct {
 	open int32
 	stop int32
 	done chan struct{}
+
+	wp *workerPool
 }
 
 // TimeoutHandler creates RequestHandler, which returns StatusRequestTimeout
@@ -1756,7 +1758,7 @@ func (s *Server) Serve(ln net.Listener) error {
 	}
 	s.mu.Unlock()
 
-	wp := &workerPool{
+	s.wp = &workerPool{
 		WorkerFunc:            s.serveConn,
 		MaxWorkersCount:       maxWorkersCount,
 		LogAllErrors:          s.LogAllErrors,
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Why do you think `s.open` will never be `0`?

`s.open` is incremented when a listener is still listening, and decremented when it isn't. And it's incremented for each connection that is open.

Your solution would shut down the server while connections are still being processed.

zc2638:
Yes, you are right.
And I have the same problem as this issue #1190.
But I waited longer, and probably forever.

I found that I have a request stuck in this method, do you have any insight?
ref: https://github.com/valyala/fasthttp/blob/master/server.go#L2125

zc2638:
I created an issue for this, and I will close this PR.

PR #1331, opened 2022-07-01
Introduce FS.CompressRoot

mojatter:
I want to serve files on read-only Root with FS.Compress.

This may be one of solution, could you review it?

```go
fs := &fasthttp.FS {
    Root: "/read-only/html",
    Compress: true,
    CompressRoot: "/tmp/compressed",
}
```

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -797,7 +797,7 @@ func cleanCacheNolock(cache map[string]*fsFile, pendingFiles, filesToRelease []*
 }
 
 func (h *fsHandler) pathToFilePath(path string) string {
-	return h.root + filepath.FromSlash(path)
+	return filepath.FromSlash(h.root + path)
 }
 
 func (h *fsHandler) filePathToCompressed(filePath string) string {
@@ -807,7 +807,7 @@ func (h *fsHandler) filePathToCompressed(filePath string) string {
 	if !strings.HasPrefix(filePath, h.root) {
 		return filePath
 	}
-	return h.compressRoot + filePath[len(h.root):]
+	return filepath.FromSlash(h.compressRoot + filePath[len(h.root):])
 }
 
 func (h *fsHandler) handleRequest(ctx *RequestCtx) {
--- a/fs.go
+++ b/fs.go
@@ -796,6 +796,20 @@ func cleanCacheNolock(cache map[string]*fsFile, pendingFiles, filesToRelease []*
 	return pendingFiles, filesToRelease
 }
 
+func (h *fsHandler) pathToFilePath(path string) string {
+	return h.root + filepath.FromSlash(path)
+}
+
+func (h *fsHandler) filePathToCompressed(filePath string) string {
+	if h.root == h.compressRoot {
+		return filePath
+	}
+	if !strings.HasPrefix(filePath, h.root) {
+		return filePath
+	}
+	return h.compressRoot + filePath[len(h.root):]
+}
+
 func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 	var path []byte
 	if h.pathRewrite != nil {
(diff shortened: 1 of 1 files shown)


fs.go
```diff
@@ -1130,7 +1147,7 @@ const (
 	fsMaxCompressibleFileSize = 8 * 1024 * 1024
 )
 
-func (h *fsHandler) compressAndOpenFSFile(filePath string, fileEncoding string) (*fsFile, error) {
+func (h *fsHandler) compressAndOpenFSFile(path, filePath string, fileEncoding string) (*fsFile, error) {
```
erikdubbelboer, line 1150: Can you come up wit ha better name for `path` here?

fs.go
```diff
@@ -1232,7 +1259,7 @@ func (h *fsHandler) newCompressedFSFile(filePath string, fileEncoding string) (*
 	return h.newFSFile(f, fileInfo, true, fileEncoding)
 }
 
-func (h *fsHandler) openFSFile(filePath string, mustCompress bool, fileEncoding string) (*fsFile, error) {
+func (h *fsHandler) openFSFile(path string, filePath string, mustCompress bool, fileEncoding string) (*fsFile, error) {
```
erikdubbelboer, line 1262: Can you come up wit ha better name for `path` here?

erikdubbelboer (changes requested):

fs.go
mojatter: Which do you think better?
- pathStr https://github.com/valyala/fasthttp/blob/b23c5e90838fdcb504032d8932c905ca7bd3b7ae/fs.go#L833
- pathOrig
- uri

fs.go
```diff
@@ -1153,7 +1170,17 @@ func (h *fsHandler) compressAndOpenFSFile(filePath string, fileEncoding string)
 		return h.newFSFile(f, fileInfo, false, "")
 	}
 
-	compressedFilePath := filePath + h.compressedFileSuffixes[fileEncoding]
+	var compressedFilePath string
+	if h.root == h.compressRoot {
+		compressedFilePath = filePath
+	} else {
+		compressedFilePath = filepath.FromSlash(h.compressRoot + path)
```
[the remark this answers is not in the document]
mojatter: You are right. https://github.com/valyala/fasthttp/pull/1331/commits/5f432d94344509d6b15804f54db7df79d059d423

fs.go
```diff
@@ -1153,7 +1170,17 @@ func (h *fsHandler) compressAndOpenFSFile(filePath string, fileEncoding string)
 		return h.newFSFile(f, fileInfo, false, "")
 	}
 
-	compressedFilePath := filePath + h.compressedFileSuffixes[fileEncoding]
+	var compressedFilePath string
+	if h.root == h.compressRoot {
+		compressedFilePath = filePath
+	} else {
+		compressedFilePath = h.compressRoot + path
```
erikdubbelboer, line 1177: Isn't it possible here instead to do something like `if h.root != h.compressRoot { compressedFilePath = strings.Replace(filePath, h.root, h.compressedRoot, 1) }`? That way you don't have to pass the original path in each function.

erikdubbelboer (changes requested):

fs.go
mojatter: Introduce pathToFilePath and filePathToCompressed https://github.com/mojatter/fasthttp/commit/3dd25f3a3bc0b5693d01d2abcd99b22d88eeefe5

mojatter: I found unnecessary changes, reverted them. https://github.com/valyala/fasthttp/pull/1331/commits/fa6652bf9e2714060e580e92b4954f0673ba15a0

fs.go
```diff
@@ -780,6 +796,20 @@ func cleanCacheNolock(cache map[string]*fsFile, pendingFiles, filesToRelease []*
 	return pendingFiles, filesToRelease
 }
 
+func (h *fsHandler) pathToFilePath(path string) string {
+	return h.root + filepath.FromSlash(path)
```
erikdubbelboer, line 800: Shouldn't this be `filepath.FromSlash(h.root + path)` like it was in the code where you call this function now?

And then `strings.HasPrefix(filePath, h.root)` should also use `filepath.FromSlash(` I guess.

erikdubbelboer (changes requested):

fs.go
mojatter: OK. Join root and path, and then calls filepath.FromSlash https://github.com/valyala/fasthttp/pull/1331/commits/38b41a8133e7ae2c3cc0744011727482c50d033d

root and compressRoot are passed through filepath.FromSlash in normalizeRoot. (eg. `C:\path\to\root`)
The path is part of the URL, that is separated by slashes. (eg. `/path/to/file`)
The path needs filepath.FromSlash to join root. (eg. `C:\path\to\root` + `\path\to\file`)

I think the best way is to use "filepath.Join", but I used "+" and "filepath.FromSlash" to respect the original code.

```go
func (h *fsHandler) pathToFilePath(path string) string {
	return filepath.Join(h.root, path)
}

func (h *fsHandler) filePathToCompressed(filePath string) string {
	if h.root == h.compressRoot {
		return filePath
	}
	if !strings.HasPrefix(filePath, h.root) {
		return filePath
	}
	return filepath.Join(h.compressRoot, filePath[len(h.root):])
}
```

erikdubbelboer:
Thanks!

PR #1334, opened 2022-07-01
fix timeout lower than writetimeout or readtimeout

fare83:


erikdubbelboer:
The change makes sense but can you please have a look at all the failing checks.

PR #1342, opened 2022-07-09
Add BodyStream() method to Response

### LINKED ISSUE #1340: fasthttp.Response does not offer a method to access the body as a stream
When making a request using fasthttp, for example (only relevant lines included):

```go
err := client.Do(req, resp)
```

There's no way to access in `resp` (of type `*fasthttp.Response`) the body as a stream.

ItalyPaleAle:
Fixes #1340

@erikdubbelboer Please let me know if this works. I don't believe it should have an impact on performance since all it does is providing another accessor to a property that is otherwise already there.

PS: I also fixed a panic in `RequestCtx.RequestBodyStream()` if the bodyStream was nil (e.g. if the data was set as non-stream)

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -240,6 +240,80 @@ func TestClientGetWithBody(t *testing.T) {
 	}
 }
 
+func TestClientGetWithBodyStream(t *testing.T) {
+	t.Parallel()
+	t.Run("body set as string", func(t *testing.T) {
+		t.Parallel()
+		testClientGetWithBodyStream(t, "test", func(req *Request) {
+			req.SetBodyString("test")
+		})
+	})
+	t.Run("body set as stream", func(t *testing.T) {
+		t.Parallel()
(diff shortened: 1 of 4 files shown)


erikdubbelboer (changes requested):
I'm sorry but this doesn't make any sense. This doesn't read the body in a streaming way from the socket. This still reads the full body in memory and then just does `bytes.NewReader(...)` on that in memory body. If you want that you can just do `bytes.NewReader(resp.bodyBytes())` yourself?

Actually reading the body in a streaming way, without reading it all into memory first, is much more complicated.

Please check https://github.com/valyala/fasthttp/blob/b23c5e90838fdcb504032d8932c905ca7bd3b7ae/http.go#L1145-L1160, it would need to be something like that but then for `Response`.

ItalyPaleAle:
Thanks for the review. It wasn't clear that `Response.bodyStream` was actually used as the request body and not the response body.

I'll see if I can try something else.

PR #1344, opened 2022-07-12
change req.timeout to public and add min deadline in pipeline

fare83:
change req.timeout to public for Do method support timeout; 
add timeout in clientGetURLDeadline;
add min deadline in pipeline

### DIFF
--- a/client.go
+++ b/client.go
@@ -2456,6 +2456,7 @@ func (c *pipelineConnClient) DoDeadline(req *Request, resp *Response, deadline t
 	w := acquirePipelineWork(&c.workPool, timeout)
 	w.respCopy.Header.disableNormalizing = c.DisableHeaderNamesNormalizing
 	w.req = &w.reqCopy
+	w.req.Timeout = timeout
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Can you explain why you need `Request.Timeout` to be public?

fare83:
when i use hostclient.Do ，i can't set timeout

erikdubbelboer:
But that internal timeout is only used as a timeout to acquire a connection if you have `MaxConns` set. What I would actually much rather do is make that timeout equal to `WriteTimeout` and completely get rid of `req.timeout`.

fare83:
I want to set timeout for current request, but writetimeout is for all host and request.

erikdubbelboer:
What do you think of this instead: https://github.com/valyala/fasthttp/pull/1346
`HostClient` already has `DoTimeout` and `DoDeadline`. What this pull request does it just make them use `req.timeout` instead of the separate goroutine that I have always hated.

fare83:
#1346 LGTM

PR #1355, opened 2022-08-14
Add Go 1.19 Support

Aoang:


Aoang:
TBR

doc.go
```diff
@@ -3,35 +3,53 @@ Package fasthttp provides fast HTTP server and client API.
...
+
+  - Per-request logger.
+
+  - Unique request id.
+
+  - Request start time.
+
+  - Connection start time.
+
+  - Request sequence number for the current connection.
```
erikdubbelboer, line 47: These need to be sub bullet points like the original.

doc.go
```diff
@@ -3,35 +3,53 @@ Package fasthttp provides fast HTTP server and client API.
...
+
+  - Maximum request header size.
+
+  - Maximum request body size.
+
+  - Maximum request execution time.
+
+  - Maximum keep-alive connection lifetime.
+
+  - Early filtering out non-GET requests.
```
erikdubbelboer, line 33: These need to be sub bullet points like the original.

erikdubbelboer (changes requested):
Thanks! Super nice pull request, just 2 minor fixes needed.

Aoang:
@erikdubbelboer Go doc comments do not support nested lists. https://tip.golang.org/doc/comment#mistakes

Maybe we can mix list markers like the example? Since bullet markers do not introduce list items in a numbered list.

erikdubbelboer:
Thanks!

.github/workflows/lint.yml
```diff
@@ -9,14 +9,40 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
-    - uses: actions/setup-go@v2
+    - uses: actions/setup-go@v3
       with:
-        go-version: 1.18.x
+        go-version: 1.19.x
+
+    - name: Get Go cache paths
```
codefromthecrypt, line 16: @Aoang curious why you ended up doing this? In wazero, we're noticing the cache doesn't seem to work for golangci-lint and wondering if there's some difference between this and hard-coding (considering it is only run with ubuntu). Any thoughts appreciated!

https://github.com/tetratelabs/wazero/blob/main/.github/workflows/commit.yaml#L46-L52

Aoang: continue https://github.com/tetratelabs/wazero/pull/804

Issue #1359, opened 2022-08-23
HeaderReceived will permanently overwrite the server's maxRequestBodySize values for that connection


GeraltXLi:
Hi, I have an issue with the `HeaderReceived`.

https://github.com/valyala/fasthttp/blob/af94725d117b19e761d0c6e9c1b157516c482a71/server.go#L2187-L2197

While `maxRequestBodySize` is modified by `HeaderReceived`，in the current connection, it keeps the current value until the next modification. This causes the configuration value of the server itself to no longer be in effect.

For example, the server's `maxRequestBodySize` is set to 10MB, and in `HeaderReceived`, I set the `maxRequestBodySize` to 1MB while the path is `/upload`. So that once an upload request occurs, the size of all other request bodies under this connection cannot be larger than 1MB, which is not expected.

If it is indeed a bug, may I submit a PR to fix it?

erikdubbelboer:
This is indeed a bug, a pull request would be welcome. I think it's as easy as
```go
if reqConf.MaxRequestBodySize > 0 { 
 	maxRequestBodySize = reqConf.MaxRequestBodySize 
 } else if s.MaxRequestBodySize > 0 {
 	maxRequestBodySize = s.MaxRequestBodySize
} else {
 	maxRequestBodySize = DefaultMaxRequestBodySize
}
```

gaby:
@erikdubbelboer This is fixed by #1360

PR #1360, opened 2022-08-23
Fix(server): reset maxRequestBodySize to the server's config

GeraltXLi:
Hi, this patch aims to fix issue https://github.com/valyala/fasthttp/issues/1359


erikdubbelboer (changes requested):
Can you do the same with `writeTimeout`?

GeraltXLi:
Hi, I have reset the `writeTimeout`

erikdubbelboer:
Thanks!

PR #1375, opened 2022-09-13
byte to string unsafe conversion in fasthttpadaptor ConvertRequest method

mstrYoda:
I changed the code to prevent string allocation by using unsafe conversion.

I wrote a benchmark test against changes.

Before:

image

After unsafe conversion: 

image

### DIFF
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"unsafe"
 
 	"github.com/valyala/fasthttp"
 )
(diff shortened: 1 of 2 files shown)


fasthttpadaptor/request.go
```diff
@@ -57,3 +58,7 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 
 	return nil
 }
+
+func b2s(b []byte) string {
```
Aoang, line 62: A comment needs to be added here for gosec check to pass https://github.com/valyala/fasthttp/blob/v1.40.0/bytesconv.go#L326

Can copy from https://github.com/valyala/fasthttp/blob/v1.40.0/bytesconv.go#L320-L328

mstrYoda: Thank you, added it 👍

fasthttpadaptor/request.go
```diff
@@ -2,31 +2,31 @@ package fasthttpadaptor
...
 	"io/ioutil"
 	"net/http"
 	"net/url"
-
-	"github.com/valyala/fasthttp"
+	"unsafe"
 )
 
 // ConvertRequest convert a fasthttp.Request to an http.Request
 // forServer should be set to true when the http.Request is going to passed to a http.Handler.
```
erikdubbelboer, line 13:
```suggestion
// ConvertRequest convert a fasthttp.Request to an http.Request
// forServer should be set to true when the http.Request is going to passed to a http.Handler.
//
// The http.Request must not be used after the fasthttp handler has returned!
// Memory in use by the http.Request will be reused after your handler has returned!
```

This was already unsafe before your change because of the use of `ctx.PostBody()`. But with your change it becomes even more dangerous so it's best if we clearly document this.

fasthttpadaptor/request.go
```diff
@@ -2,31 +2,31 @@ package fasthttpadaptor
 
 import (
 	"bytes"
+	"github.com/valyala/fasthttp"
 	"io/ioutil"
 	"net/http"
 	"net/url"
-
-	"github.com/valyala/fasthttp"
+	"unsafe"
```
erikdubbelboer, line 9: Moving the github import to a separate block is good. But can you move the `unsafe` import together with the block of other std imports?

erikdubbelboer (changes requested):

fasthttpadaptor/request.go
mstrYoda: Thank you for review, I made the changes 👍

erikdubbelboer (changes requested):
Can you fix lint failing.

erikdubbelboer:
Thanks!

PR #1376, opened 2022-09-14
Add "GetString" method for Args

xiaojun207:
Add "GetString" method for Args

### DIFF
--- a/args.go
+++ b/args.go
@@ -338,6 +338,11 @@ func (a *Args) GetUfloatOrZero(key string) float64 {
 	return f
 }
 
+// GetString returns string value for the given key.
+func (a *Args) GetString(key string) string {
+	return string(a.Peek(key))
+}
+
 // GetBool returns boolean value for the given key.
 //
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
What is the reason you want to add this method? Why can't users of fasthttp just do `string(a.Peek(key))` themselves?

xiaojun207:
> What is the reason you want to add this method? Why can't users of fasthttp just do `string(a.Peek(key))` themselves?

1. For most http frameworks, the preferred method for parameter acquisition is Get or GetString, which helps users use fasthttp. For example, go gin: (c * gin. Context) GetString (key string) (s string)
2. When using fasthttp to receive input parameters, there are many string types used, so GetString is used frequently.

I really hope to pass, thank you

erikdubbelboer:
But there are no other methods that return strings. Everything returns bytes to prevent allocations. If this is too difficult for someone they *really shouldn't be using fasthttp*.

gaby:
@erikdubbelboer This can probably be closed.

Issue #586, opened 2019-06-01
Any tarpit / purposefully dropping connection for malicious attacks?


hiqbn:
Any tarpit / purposefully dropping connection for malicious attacks?
I'm looking for fasthttp equivalent of tarpitting / dropping slowloris kind of attacks.

kirillDanshin:
I've tested slowloris kind of attacks for fasthttp couple of weeks ago and I wasn't able to make any damage or DDoS. If you were able, please see [SECURITY.md](https://github.com/valyala/fasthttp/blob/master/SECURITY.md) for instructions, we'll be glad to receive such report.

If we look into another level of attacks, such as application-level vectors, you can see [Gramework](github.com/gramework/gramework)'s Protection as an example implementation of Whitelists, Blacklists and more.

Injection attacks tho, kinda difficult to detect and filter while maintaining low false-positives and high correct detection rate. E.g, mod_security can be bypassed and also easy to misconfigure, which bumps false-positives, etc. This kind of attacks can be detected and dropped using third-party WAF provider, but it depends of your project.

hiqbn:
if u are referring to this
https://github.com/gramework/gramework/blob/dev/app_protection.go

i gave a quick glance and saw you didnt have an option to unban or untime the ban...
it's ok. it's well written anyhow.

back to my question
Any tarpit / purposefully dropping connection for malicious attacks? (how do i achieve this) what do we do after we blacklist the remote ip?

how to "drop" connection for the purpose of hiding the fact that we "dropped" the connection silently from the malicious attacker's point of view.

would like a suggestion on how to do tarpit too, with fasthttp

erikdubbelboer:
This is more in scope for a separate library or a framework. Fasthttp is really meant as a low level library for http, nothing more.

PR #1382, opened 2022-09-18
WIP: Add defense against slowloris attack

mxmauro:
Hi @valyala and @erikdubbelboer, I would like to know your opinion about this addition.

The goal is to implement a protection for the slowloris attack. It is a work-in-progress but need your approval and point of view to continue working on this.

I read [this discussion](https://github.com/valyala/fasthttp/issues/586) about the scope of fasthttp but I also think this modification is aligned with the timeouts the dev can set.

Also setting timeouts are not enough when a server can receive small and large bodies.

Kind regards,
Mauro.


kirillDanshin:
I believe that is very much outside the scope of fasthttp. I believe it should be done at lower level (e.g. tcp/udp conn wrapper). having this built into fasthttp will impact some of the major use cases, for example, rtb can't have that much allocations just  suddenly appear out of nowhere, and having this type of defense in place wouldn't work for mid/high profile HLS applications.

we can probably make an addon library for this, publish it in github.com/fasthttp, and put a link in the readme.

mxmauro:
> I believe that is very much outside the scope of fasthttp. I believe it should be done at lower level (e.g. tcp/udp conn wrapper). having this built into fasthttp will impact some of the major use cases, for example, rtb can't have that much allocations just suddenly appear out of nowhere, and having this type of defense in place wouldn't work for mid/high profile HLS applications.
> 
> we can probably make an addon library for this, publish it in github.com/fasthttp, and put a link in the readme.

Hi @kirillDanshin no idea on how to create an add-on (because it affects the connection at a low level) but my intention, for the final version, is to not wrap `net.Conn` if configured values are zero.

erikdubbelboer:
In my opinion this shouldn't really be an issue with the use cases fasthttp is for. **fasthttp is not mean to be used for public webservers**. It's intended use is for some high QPS environments such as OpenRTB or other RTB use cases. There you always have connections between known servers, none of which would try a slowloris attack as you are customers of each other. If you have to expose fasthttp to the open web I would always put something like nginx in front of it. If only for TLS termination. Doing TLS in Go completely goes against the performance gains you would get from fasthttp. fasthttp isn't really meant to be used with https as crypto is still slower in Go than in C.

mxmauro:
Hi @erikdubbelboer I understand.

Thanks to both for explaining the details and the goal of the library.

Kind regards.

PR #1383, opened 2022-09-20
feat: add ShutdownWithContext

li-jin-gou:
pr for https://github.com/valyala/fasthttp/issues/1389
- add `ShutdownWithContext` function to support shutdown with timeout and timeout happen  after all listener already close.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1816,6 +1816,17 @@ func (s *Server) Serve(ln net.Listener) error {
 //
 // Shutdown does not close keepalive connections so its recommended to set ReadTimeout and IdleTimeout to something else than 0.
 func (s *Server) Shutdown() error {
+	return s.ShutdownWithContext(context.Background())
+}
+
+// ShutdownWithContext gracefully shuts down the server without interrupting any active connections.
+// ShutdownWithContext works by first closing all open listeners and then waiting for all connections to return to idle or context timeout and then shut down.
+//
+// When ShutdownWithContext is called, Serve, ListenAndServe, and ListenAndServeTLS immediately return nil.
+// Make sure the program doesn't exit and waits instead for Shutdown to return.
+//
+// ShutdownWithContext does not close keepalive connections so its recommended to set ReadTimeout and IdleTimeout to something else than 0.
+func (s *Server) ShutdownWithContext(ctx context.Context) (err error) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1856,6 +1856,57 @@ func (s *Server) Shutdown() error {
...
+		if open := atomic.LoadInt32(&s.open); open == 0 {
+			break
+		}
+		// This is not an optimal solution but using a sync.WaitGroup
+		// here causes data races as it's hard to prevent Add() to be called
+		// while Wait() is waiting.
+		select {
+		case <-ctx.Done():
+			err = ctx.Err()
+			break END
```
erikdubbelboer, line 1899: I'm wondering, should this then force close all connections? What does `net/http` do in this case?

pjebs:
`func (s *Server) Shutdown() error {` should be equivalent to `s.ShutdownWithContext (context.Background())`

server.go
pjebs: https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/net/http/server.go;l=2731

li-jin-gou: > I'm wondering, should this then force close all connections? What does `net/http` do in this case?

The use of `shutdown` does not force all connections to close, but rather uses context to control the timeout after all `listener` have been closed.

server_test.go
```diff
@@ -3597,6 +3597,50 @@ func TestShutdownCloseIdleConns(t *testing.T) {
 	}
 }
 
+func TestShutdownWithContext(t *testing.T) {
+	t.Parallel()
+
+	ln := fasthttputil.NewInmemoryListener()
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			ctx.Success("aaa/bbb", []byte("real response"))
```
erikdubbelboer, line 3606: Can you add a `time.Sleep(time.Millisecond*100)` here and test that `ShutdownWithContext` returns correctly after the timeout? The test you have now just tests the code that is similar to what `TestShutdown` already tests. Thanks!

erikdubbelboer (changes requested):

server_test.go
li-jin-gou: thanks

erikdubbelboer:
Thanks!

pjebs:
When will it be tagged?

erikdubbelboer:
Next week I'll tag a release.

skydiver1980 (approved):

skydiver1980 (approved):

PR #1386, opened 2022-09-28
optimize code

tylitianrui:
- reuse []byte memory 
-  reduce  redundant. code

### DIFF
--- a/args.go
+++ b/args.go
@@ -105,7 +105,7 @@ func (a *Args) ParseBytes(b []byte) {
 
 // String returns string representation of query args.
 func (a *Args) String() string {
-	return string(a.QueryString())
+	return b2s(a.QueryString())
 }
 
 // QueryString returns query string for the args.
--- a/client.go
+++ b/client.go
@@ -492,10 +492,11 @@ func (c *Client) Do(req *Request, resp *Response) error {
 			c.m = m
 		}
 	}
-	hc := m[string(host)]
+	host_str := b2s(host)
+	hc := m[host_str]
 	if hc == nil {
 		hc = &HostClient{
-			Addr:                          addMissingPort(string(host), isTLS),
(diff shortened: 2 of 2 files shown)


client.go
```diff
@@ -492,10 +492,11 @@ func (c *Client) Do(req *Request, resp *Response) error {
 			c.m = m
 		}
 	}
-	hc := m[string(host)]
+	host_str := b2s(host)
```
erikdubbelboer, line 495: This doesn't work. The string will be used for a map key which means the map will keep a reference to it. But that's unsafe as the underlying bytes might be modified.
See: https://go.dev/play/p/q4yRaxJuq3c

erikdubbelboer: What might be an improvement is only doing the `string(host)` conversion once inside `if hc == nil {`. And then doing `hc := m[b2s(host)]` even though I'm not sure if Go allocates for this string cast to access a map, you could test this.

args.go
```diff
@@ -105,7 +105,7 @@ func (a *Args) ParseBytes(b []byte) {
 
 // String returns string representation of query args.
 func (a *Args) String() string {
-	return string(a.QueryString())
+	return b2s(a.QueryString())
```
erikdubbelboer, line 108: I think this is a dangerous change. There could in theory be users that currency use this string and assume it is always safe to use. This change would make this string get modified when the Args get reused. So this change would be backwards incompatible. Therefor I would prefer not to do this.

erikdubbelboer (changes requested):

client.go
oschwald: Go has not allocated for `m[string(b)]` since 1.3 or so.

args.go
tylitianrui: thx for your review

Issue #179, opened 2016-10-04
Peek Headers with the same name


nsagnett:
Hi ! 

It's possible to get multiple headers with the same name without parsing all headers in requestHeader struct (without using visitAll function) ?

SshPwd:
```go
ctx.Request.Header.Peek("X-Your-Some-Header")
```

nsagnett:
It returns the first header with "X-Your-Some-Header" name. I would like same method returns all headers named "X-Your-Some-Header"

kataras:
@nsagnett What do you mean, you have multiple headers with the same key?

nsagnett:
It's possible !

SshPwd:
Can you show example your a request with multiple headers?

nsagnett:
curl -v -H "X-Your-Some-Header: foo" -H "X-Your-Some-Header: bar", you can see two headers with the same key

SshPwd:
Could you show reference to the rfc part please?

nsagnett:
https://tools.ietf.org/html/rfc3864 (2.2.1)

SshPwd:
I found https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 tnx

kataras:
Yes, multiple values are allowed, multiple keys are not possible, I though you meant multiple keys at my first  reading, but I misunderstood, sorry. Practical you have right then, if fasthttp doesn't supports multiple headers values yet this is a problem but you can fix it with a PR, don't be afraid, we are here!

AlskiOnTheWeb:
@kataras  "Yes, multiple values are allowed, multiple keys are not possible"

Actually, the HTTP RFC does state that multiple (duplicated) keys MAY exist in an HTTP header. They very carefully explain that proxies are not allowed to change the order of duplicates as the value order may be significant:

Multiple message-header fields with the same field-name MAY be
   present in a message if and only if the entire field-value for that
   header field is defined as a comma-separated list [i.e., #(values)].
   It MUST be possible to combine the multiple header fields into one
   "field-name: field-value" pair, without changing the semantics of the
   message, by appending each subsequent field-value to the first, each
   separated by a comma. The order in which header fields with the same
   field-name are received is therefore significant to the
   interpretation of the combined field value, and thus a proxy MUST NOT
   change the order of these field values when a message is forwarded.

aiddroid:
[...]

PR #1394, opened 2022-10-12
feat: add header.PeekAll

li-jin-gou:
pr for https://github.com/valyala/fasthttp/issues/179

### DIFF
--- a/args.go
+++ b/args.go
@@ -623,3 +623,13 @@ func decodeArgAppendNoPlus(dst, src []byte) []byte {
 	}
 	return dst
 }
+
+func peekAllArgBytesToDst(dst [][]byte, h []argsKV, k []byte) [][]byte {
+	for i, n := 0, len(h); i < n; i++ {
+		kv := &h[i]
+		if bytes.Equal(kv.key, k) {
+			dst = append(dst, kv.value)
+		}
+	}
+	return dst
+}
--- a/header.go
+++ b/header.go
@@ -42,6 +42,7 @@ type ResponseHeader struct {
 	contentType     []byte
 	contentEncoding []byte
 	server          []byte
+	mulHeader       [][]byte
 
 	h       []argsKV
 	trailer []argsKV
@@ -79,6 +80,7 @@ type RequestHeader struct {
 	host        []byte
 	contentType []byte
 	userAgent   []byte
+	mulHeader   [][]byte
 
 	h       []argsKV
 	trailer []argsKV
@@ -974,6 +976,7 @@ func (h *ResponseHeader) resetSkipNormalize() {
 	h.h = h.h[:0]
 	h.cookies = h.cookies[:0]
 	h.trailer = h.trailer[:0]
+	h.mulHeader = h.mulHeader[:0]
 }
 
 // SetNoDefaultContentType allows you to control if a default Content-Type header will be set (false) or not (true).
@@ -1002,6 +1005,7 @@ func (h *RequestHeader) resetSkipNormalize() {
 	h.contentType = h.contentType[:0]
 	h.userAgent = h.userAgent[:0]
 	h.trailer = h.trailer[:0]
+	h.mulHeader = h.mulHeader[:0]
 
 	h.h = h.h[:0]
 	h.cookies = h.cookies[:0]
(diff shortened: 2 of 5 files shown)


li-jin-gou:
refer to https://github.com/valyala/fasthttp/issues/179

args.go
```diff
@@ -260,7 +260,7 @@ func (a *Args) PeekBytes(key []byte) []byte {
 func (a *Args) PeekMulti(key string) [][]byte {
 	var values [][]byte
 	a.VisitAll(func(k, v []byte) {
-		if string(k) == key {
+		if b2s(k) == key {
```
erikdubbelboer, line 263: This is not needed. `string(bv) == sv` is optimized in the compiler and doesn't allocate: https://go.dev/play/p/WYcbS08dv_a
I haven't tested it but I wouldn't be surprised if it's even faster than with `b2s`.

args.go
```diff
@@ -343,7 +343,7 @@ func (a *Args) GetUfloatOrZero(key string) float64 {
 // true is returned for "1", "t", "T", "true", "TRUE", "True", "y", "yes", "Y", "YES", "Yes",
 // otherwise false is returned.
 func (a *Args) GetBool(key string) bool {
-	switch string(a.Peek(key)) {
+	switch b2s(a.Peek(key)) {
```
erikdubbelboer, line 346: Same here, this is also already optimized by the compiler not to allocate: https://go.dev/play/p/fkD2_0estoT

header.go
```diff
@@ -1793,6 +1793,69 @@ func (h *RequestHeader) peek(key []byte) []byte {
 	}
 }
 
+// PeekAll returns header value for the given key.
+func (h *RequestHeader) PeekAll(key string) [][]byte {
+	k := getHeaderKeyBytes(&h.bufKV, key, h.disableNormalizing)
```
erikdubbelboer, line 1798: I wonder if we should do the same trick here as with `bufKV`. Where we put a `[][]byte` in `RequestHeader` and reuse that instead of allocating the return value on the heap.

erikdubbelboer (changes requested):

args.go
li-jin-gou: thanks

header.go
li-jin-gou: thanks,You're right

li-jin-gou: I modified the logic and added a `mulHeader` filed.

args.go
```diff
@@ -105,7 +105,7 @@ func (a *Args) ParseBytes(b []byte) {
 
 // String returns string representation of query args.
 func (a *Args) String() string {
-	return string(a.QueryString())
+	return b2s(a.QueryString())
```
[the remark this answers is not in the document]
li-jin-gou: done

args.go
li-jin-gou: done

header.go
erikdubbelboer: Can you add a comment like this to these functions?
```
// The returned value is not safe to use after the handler returns, the request is released or other calls to PeekAll.
```

erikdubbelboer (changes requested):

li-jin-gou:
Comments have been modified.

header.go
li-jin-gou: thanks

args.go
```diff
@@ -623,3 +623,12 @@ func decodeArgAppendNoPlus(dst, src []byte) []byte {
...
 }
+
+func peekAllArgBytesToDst(dst *[][]byte, h []argsKV, k []byte) {
+	for i, n := 0, len(h); i < n; i++ {
+		kv := &h[i]
+		if bytes.Equal(kv.key, k) {
+			*dst = append(*dst, kv.value)
+		}
+	}
+}
```
erikdubbelboer, line 634: Sorry one minor last change. I don't think it's needed to pass a pointer to the `[][]byte`. Just append to it locally and return it. That's the pattern that is used in a lot of other places as well. It shouldn't cause any allocations.

```suggestion
func peekAllArgBytesToDst(dst [][]byte, h []argsKV, k []byte) [][]byte {
	for i, n := 0, len(h); i < n; i++ {
		kv := &h[i]
		if bytes.Equal(kv.key, k) {
			dst = append(dst, kv.value)
		}
	}
	return dst
}
```

erikdubbelboer (changes requested):

args.go
li-jin-gou: Thanks and done

erikdubbelboer:
Thanks!

header_test.go
```diff
@@ -2861,3 +2861,65 @@ func verifyTrailer(t *testing.T, r *bufio.Reader, expectedTrailers map[string]st
...
+	h.Add(HeaderContentType, "aaa/bbb")
+	h.Add(HeaderContentEncoding, "gzip")
+	h.Add(HeaderConnection, "close")
+	h.Add(HeaderContentLength, "1234")
+	h.Add(HeaderServer, "aaaa")
+	h.Add(HeaderSetCookie, "cccc")
+	h.Add("aaa", "aaa")
+	h.Add("aaa", "bbb")
+
+	expectResponseHeaderAll(t, h, HeaderContentType, [][]byte{s2b("aaa/bbb")})
```
codefromthecrypt, line 2909: does it make sense to add a test for when content type is deleted, whether the expected result is the default type or empty? see https://github.com/valyala/fasthttp/issues/1402 as it is ambiguous.

li-jin-gou: Thanks for reminding

Issue #1312, opened 2022-06-05
does fasthhttp support multipart/byteranges?


byene0923:
i see in [fs.go](https://github.com/valyala/fasthttp/blob/master/fs.go#L912) that fasthttp may not handle[ multipart/byteranges](https://datatracker.ietf.org/doc/html/rfc7233#page-18) .
will fasthttp support it ? if so, i'm willing to do it

erikdubbelboer:
A pull is always welcome!

byene0923:
support [multipart/byteranges](https://datatracker.ietf.org/doc/html/rfc7233#page-18) is a bit complex

the content of multipart/byteranges  like
```
--3d6b6a416f9b5
Content-Type: text/html
Content-Range: bytes 0-50/1270

<!DOCTYPE html>
<html lang="en-US">
<head>
    <title>Example Do
--3d6b6a416f9b5
Content-Type: text/html
Content-Range: bytes 100-150/1270

eta http-equiv="Content-type" content="text/html; c
--3d6b6a416f9b5--
```
this is quiet different from [Single part ranges](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests#single_part_ranges), we need to add Content-Type, Content-Range and Boundary of each range
however, fasthttp use []byte to handle content and we may not read all the content in one time. so we need carefully handle content and add extra information.
also, we need to calculate the length which include the extra information

so i seperate the `fsReader` and `RangeFsReader`. `fsReader`  handle the complete file and `RangeFsReader` handle the range file
i will pr later

byene0923:
there exists bugs in my pull request, i reopen it

PR #1398, opened 2022-10-17
feat: support mulit/range

byene0923:
pr for https://github.com/valyala/fasthttp/issues/1312

- change `SetByteRange`, support multipart/byteranges
- change `bigFileReader` and `fsSmallFileReader` which only handle complete files
- add `smallRangeReader` and `bigRangeReader` which only handle range files
- change `byteRangeUpdater` interface to `byteRangeHandler` 
**test**
- change `TestFSByteRangeSingleThread`, add `bigFileReader` and `fsSmallFileReader` test of `read` and `WriteTo`
- add `TestFSMultiByteRangeConcurrent` and `TestFSMultiByteRangeSingleThread`
- change `TestParseByteMultiRangeSuccess` and `TestParseByteRangeError`

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -977,7 +977,7 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 	statusCode := StatusOK
 	contentLength := ff.contentLength
 	if h.acceptByteRange {
-		hdr.SetCanonical(strAcceptRanges, strBytes)
+		hdr.setNonSpecial(strAcceptRanges, strBytes)
 		if len(byteRange) > 0 {
 			staList, endList, err := ParseByteRanges(byteRange, contentLength)
 			if err != nil {
--- a/fs.go
+++ b/fs.go
@@ -2,6 +2,7 @@ package fasthttp
 
 import (
 	"bytes"
+	"crypto/rand"
 	"errors"
 	"fmt"
 	"html"
@@ -501,7 +502,8 @@ type fsHandler struct {
 	cacheGzip   map[string]*fsFile
 	cacheLock   sync.Mutex
 
-	smallFileReaderPool sync.Pool
+	smallFileReaderPool  sync.Pool
+	smallRangeReaderPool sync.Pool
 }
 
 type fsFile struct {
@@ -520,6 +522,9 @@ type fsFile struct {
 
 	bigFiles     []*bigFileReader
 	bigFilesLock sync.Mutex
+
+	rangeBigFiles     []*bigRangeReader
+	rangeBigFilesLock sync.Mutex
 }
 
 func (ff *fsFile) NewReader() (io.Reader, error) {
@@ -540,9 +545,8 @@ func (ff *fsFile) smallFileReader() (io.Reader, error) {
 	}
 	r := v.(*fsSmallFileReader)
 	r.ff = ff
-	r.endPos = ff.contentLength
-	if r.startPos > 0 {
-		return nil, errors.New("bug: fsSmallFileReader with non-nil startPos found in the pool")
+	if r.offset > 0 {
+		panic("BUG: fsSmallFileReader with non-nil offset found in the pool")
 	}
 	return r, nil
 }
@@ -580,7 +584,6 @@ func (ff *fsFile) bigFileReader() (io.Reader, error) {
 	return &bigFileReader{
 		f:  f,
 		ff: ff,
-		r:  f,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


fs.go
```diff
@@ -935,25 +977,29 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 	statusCode := StatusOK
 	contentLength := ff.contentLength
 	if h.acceptByteRange {
-		hdr.setNonSpecial(strAcceptRanges, strBytes)
+		hdr.SetCanonical(strAcceptRanges, strBytes)
```
Aoang, line 980: If `fsSmallFileReader` does not support `ByteRange`, cannot set `Accept-Ranges: bytes`.

header.go
```diff
@@ -113,18 +113,23 @@ func (h *ResponseHeader) SetContentRange(startPos, endPos, contentLength int) {
 //
 //   - If startPos is negative, then 'bytes=-startPos' value is set.
 //   - If endPos is negative, then 'bytes=startPos-' value is set.
-func (h *RequestHeader) SetByteRange(startPos, endPos int) {
+func (h *RequestHeader) SetByteRange(startPos, endPos []int) {
```
erikdubbelboer, line 116: This is a backwards compatible change we can not make.

header_test.go
```diff
@@ -2922,4 +2922,4 @@ func expectResponseHeaderAll(t *testing.T, h *ResponseHeader, key string, expect
 	if !reflect.DeepEqual(h.PeekAll(key), expectedValue) {
 		t.Fatalf("Unexpected value for key %q: %q. Expected %q", key, h.PeekAll(key), expectedValue)
 	}
-}
+}
```
erikdubbelboer, line 2925: Why are you removing all the newlines at the end of files? They should be there.

erikdubbelboer (changes requested):

header_test.go
byene0923: I use goland ide, the format is different from github lint, i will fix it soon

header.go
byene0923: yeah, add  method of  `SetByteRanges`?

erikdubbelboer (changes requested):

fs.go
```diff
@@ -607,143 +610,175 @@ func (ff *fsFile) decReadersCount() {
...
+	return ff.smallRangeReader(), nil
+}
+
+func (ff *fsFile) smallRangeReader() io.Reader {
+	v := ff.h.smallRangeReaderPool.Get()
+	if v == nil {
+		v = &smallRangeReader{
+			sta: make([]int, 64),
+			end: make([]int, 64),
+			buf: make([]byte, 4096),
```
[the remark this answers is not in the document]
byene0923: 4096 keep the same with [`copyBufPool`](https://github.com/valyala/fasthttp/blob/master/http.go#L2052), this use for `copyZeroAlloc`

fs.go
```diff
@@ -1451,3 +1498,499 @@ func getFileLock(absPath string) *sync.Mutex {
...
+		cPos += n
+		r.buf = r.buf[:0]
+	}
+	return cPos, io.EOF
+}
+
+func (r *bigRangeReader) WriteTo(w io.Writer) (int64, error) {
+	if rf, ok := w.(io.ReaderFrom); ok {
+		return rf.ReadFrom(r)
+	}
```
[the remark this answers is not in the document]
byene0923: this is for `bytebufferpool`, do as this 
```
bb := bytebufferpool.Get()
_, err := reader.WriteTo(bb)
```
and `smallRangeReader` has also do it, in fs.go  line 1690. `smallRangeReader` do the same as `smallFileReader`, handle directory first, then handle `readfrom`, last handle `write`

fs.go
```diff
@@ -1451,3 +1498,499 @@ func getFileLock(absPath string) *sync.Mutex {
...
+		}
+
+		if r.IsMultiRange() && !r.hasCurRangeBodyHeader {
+			r.hasCurRangeBodyHeader = true
+			multiRangeBodyHeader(&r.buf, r.sta[i], r.end[i], r.ff.contentLength, r.ff.contentType, r.boundary, r.hf)
+			r.hf = false
+			cLen = len(p[cPos:])
+			n = copy(p[cPos:], r.buf)
+			cPos += n
+			if len(r.buf) > cLen {
```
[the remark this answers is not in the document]
byene0923: this is a bit complex
the multibyte/range like
```
--3d6b6a416f9b5
Content-Type: text/html
Content-Range: bytes 0-15/1270

<!DOCTYPE html>
--3d6b6a416f9b5
Content-Type: text/html
Content-Range: bytes 0-10/1270

<!DOCTYPE 
--3d6b6a416f9b5--
```
for each range, we need to add `boundary`, `Content-Type`, `Content-Range` first, then add http content, at last range, we need to and `boundary--`

however, the length of `p` is limited, so we need to check whether `p` is full when we add `multiRangeBodyHeader` which contain `boundary`, `Content-Type`, `Content-Range` and http content, and `multiRangeBodyEnd` which contain `boundary--`

the `r.buf` is the rest content of `multiRangeBodyHeader` or `multiRangeBodyEnd` that has not copy into `p` at last time.
`cLen` is the rest lengh of `p`,  and we have done `n = copy(p[cPos:], r.buf)`, if `cLen` is less then `len(r.buf)`, we can copy all the content of `r.buf` to `p`

by the way,
method `read` like a streaming which the `p` is limited
first, we read the `multiRangeBodyHeader`, then we read http content, if we are in last range, we read the `multiRangeBodyEnd`

the code which handle range content begins  in line fs.go 1603

erikdubbelboer (changes requested):
It seems like `TestFSMultiByteRangeConcurrent` always gets a timeout, can you fix that any other tests that might fail?

byene0923:
> 
sure,  the  test number of concurrent and multirange may be  too large

erikdubbelboer:
Thanks!

erikdubbelboer:
@byene0923 I had to revert this change as it caused a critical bug: https://github.com/valyala/fasthttp/issues/1445

Can you have a look at adding a test case for this and fixing it?

byene0923:
ok, sorry for missing the full tests. i will look deep into it

MindMayhem:
@byene0923  maybe it will be useful to you, but when we request large static file, then in request headers an interesting heading appears - Range: 3896-3896
on others files there's no such header

also, it's worth noting that my file was 5mb in size

byene0923:
@MindMayhem thanks a lot, but i can't  reproduce this by your files in https://github.com/valyala/fasthttp/issues/1445. my system is  macOS 10.15.7 i will do it in different systems

byene0923:
@MindMayhem thanks a lot ,but i can not reproduce it  by your files in https://github.com/valyala/fasthttp/issues/1445 now. should any changes i do in your code ?

MindMayhem:
@byene0923 that's not my example, but i pretty sure it shows that error, just check your browser console
i hope @smalloff can provide more info to you

erikdubbelboer:
I was able to reproduce it on MacOS using the files in the zipfile in this comment: https://github.com/valyala/fasthttp/issues/1445#issuecomment-1328800669 Only thing I did was `go mod tidy` and `go run main.go`.

li-jin-gou:
> I was able to reproduce it on MacOS using the files in the zipfile in this comment: [#1445 (comment)](https://github.com/valyala/fasthttp/issues/1445#issuecomment-1328800669) Only thing I did was `go mod tidy` and `go run main.go`.

I'll try.😏

byene0923:
thanks a lot to @MindMayhem and @erikdubbelboer, i reproduce it, i will look into it soon

byene0923:
hello, @erikdubbelboer @MindMayhem i have found the bug in fasthttp@1.42.0 fs.go, line 749
```
func (r *fsSmallFileReader) WriteTo(w io.Writer) (int64, error) {
	if r.offset != 0 {
		panic("BUG: non-zero offset! Read() mustn't be called before WriteTo()")
	}
```
when file size  in (4096, 8192) will cause this bug.  because writeTo() will be called twice, and the second call `r.offset` is not zero, so this will cause panic, when file size > 8192, it use `bigFileReader`

many thanks to @MindMayhem  and @smalloff ! i will fix it soon and add unit tests

byene0923:
it did a strange bug, and pass all the unit tests !

byene0923:
i delete this check. and i'm doing the unit test, you could change the go.mod of the zipfile in this comment:  https://github.com/valyala/fasthttp/issues/1445#issuecomment-1328800669

just add this at the last line of go.mod

```
replace github.com/valyala/fasthttp => github.com/byene0923/fasthttp v1.23.1-0.20221202163206-7e40fb23488f
```
if @erikdubbelboer @MindMayhem have interest, can have a try

PR #1403, opened 2022-10-21
optimize: adjust the behavior of PeekAll based on VisitAll

li-jin-gou:
refer to  https://github.com/valyala/fasthttp/pull/1394#discussion_r1001335924

### DIFF
--- a/header.go
+++ b/header.go
@@ -1811,11 +1811,17 @@ func (h *RequestHeader) peekAll(key []byte) [][]byte {
 	h.mulHeader = h.mulHeader[:0]
 	switch string(key) {
 	case HeaderHost:
-		h.mulHeader = append(h.mulHeader, h.Host())
+		if host := h.Host(); len(host) > 0 {
+			h.mulHeader = append(h.mulHeader, host)
+		}
 	case HeaderContentType:
-		h.mulHeader = append(h.mulHeader, h.ContentType())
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -1827,7 +1833,6 @@ func (h *RequestHeader) peekAll(key []byte) [][]byte {
 	case HeaderCookie:
 		if h.cookiesCollected {
 			h.mulHeader = append(h.mulHeader, appendRequestCookieBytes(nil, h.cookies))
-			return [][]byte{appendRequestCookieBytes(nil, h.cookies)}
```
li-jin-gou, line 1830: I apologize for a problem here that I didn't find before.

header.go
```diff
@@ -1853,11 +1858,17 @@ func (h *ResponseHeader) peekAll(key []byte) [][]byte {
 	h.mulHeader = h.mulHeader[:0]
 	switch string(key) {
 	case HeaderContentType:
-		h.mulHeader = append(h.mulHeader, h.ContentType())
+		if contentType := h.ContentType(); len(contentType) > 0 {
```
li-jin-gou, line 1861: This is a modification from VisitAll, because I found that when using PeekAll, if the field is empty he returns `[][]byte{ '' }` instead of `[][]byte{}`.

erikdubbelboer:
Thanks!

PR #1406, opened 2022-10-24
fix: ignore body should not set content-length of streaming 

byene0923:
https://github.com/valyala/fasthttp/pull/1022 only handle ContinueReadBody(), this pr handle ContinueReadBodyStream()

### DIFF
--- a/http.go
+++ b/http.go
@@ -1287,6 +1287,8 @@ func (req *Request) ContinueReadBodyStream(r *bufio.Reader, maxBodySize int, pre
 		// the end of body is determined by connection close.
 		// So just ignore request body for requests without
 		// 'Content-Length' and 'Transfer-Encoding' headers.
+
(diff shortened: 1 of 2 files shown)


http.go
```diff
@@ -1287,7 +1287,9 @@ func (req *Request) ContinueReadBodyStream(r *bufio.Reader, maxBodySize int, pre
 		// the end of body is determined by connection close.
 		// So just ignore request body for requests without
 		// 'Content-Length' and 'Transfer-Encoding' headers.
-		req.Header.SetContentLength(0)
+		if !req.Header.ignoreBody() {
```
li-jin-gou, line 1290: add comment `// refer to https://tools.ietf.org/html/rfc7230#section-3.3.2`

erikdubbelboer:
Thanks!

PR #1407, opened 2022-10-26
Add ResponseBodyStream

wzshiming:


http.go
```diff
@@ -478,6 +478,10 @@ func (ctx *RequestCtx) RequestBodyStream() io.Reader {
 	return ctx.Request.bodyStream
 }
 
+func (ctx *RequestCtx) ResponseBodyStream() io.Reader {
```
li-jin-gou, line 481: You can describe the use of PR

wzshiming:
``` golang
fasthttp.Serve(conn, fasthttp.RequestHandler(func(ctx *fasthttp.RequestCtx) {
  ctx.Request.SetHost("other")
  requestBodyStream := ctx.RequestBodyStream()
  // wrap the requestBodyStream
  _ = fasthttp.Do(&ctx.Request, &ctx.Response)
  responseBodyStream := ctx.ResponseBodyStream()
  // wrap the responseBodyStream
  ctx.SetBodyStream(responseBodyStream)
}))
```
This is a simple example, I want to do some processing of the request, then send it to another server, and for the response of the other server, do some processing and return

erikdubbelboer (changes requested):
Can you add some tests to test the behavior you would expect from this function?

wzshiming:
Done

server_test.go
```diff
@@ -763,6 +763,112 @@ func TestServerResponseBodyStream(t *testing.T) {
...
+	go func() {
+		if err := Serve(lnProxy, h); err != nil {
+			t.Errorf("unexpected error: %v", err)
+		}
+		close(serverCh)
+	}()
+
+	serverProxyCh := make(chan struct{})
+	go func() {
+		if err := Serve(ln, hProxy); err != nil {
```
erikdubbelboer, line 808: Shouldn't your test set [`Server.StreamRequestBody = true`](https://pkg.go.dev/github.com/valyala/fasthttp#Server.StreamRequestBody)? That's what the test should be testing, if that works correctly. Otherwise it's isn't testing something new as it just tests the `bytes.NewReader(ctx.Response.bodyBytes())` line of your function which is trivial.

server_test.go
```diff
@@ -763,6 +763,112 @@ func TestServerResponseBodyStream(t *testing.T) {
...
+		err := cli.Do(&ctx.Request, &ctx.Response)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		responseBodyStream := ctx.ResponseBodyStream()
+		ctx.Response.Body()
+		body, _ := io.ReadAll(responseBodyStream)
+		body = append(body, []byte(" append")...)
+		ctx.SetBodyStream(bytes.NewBuffer(body), len(body))
```
erikdubbelboer, line 795: Is it not possible to do `ctx.SetBodyStream(responseBodyStream, -1)` here? I feel like that's what people are going to want to do when we introduce this new function. So that the full body doesn't need to be read into memory anymore like you're still doing here.

erikdubbelboer (changes requested):

gaby:
@wzshiming @erikdubbelboer What's the status of this feature?

erikdubbelboer:
@gaby someone needs to fix the test and see if this code really works. I currently don't have time for this.

gaby:
> @gaby someone needs to fix the test and see if this code really works. I currently don't have time for this.

Isnt this already covered by https://github.com/valyala/fasthttp/pull/1414 ? Or am I confusing features

erikdubbelboer:
You're right. That change already added `Response.BodyStream() io.Reader` which is kinda what this pull request wanted to add as well in a slightly different way. So I'm going to close this.

PR #1410, opened 2022-10-27
fix: the length of trailer key should > 0

byene0923:
1. the length of trailer key should > 0
2. fix http_test words typo


erikdubbelboer:
Thanks!

PR #1414, opened 2022-11-02
support response body stream

Anthony-Dong:
why:
When developing HTTP proxy tools, we encountered the need to forward large packets, and the second is to support streaming forwarding, so we need to modify the sdk.

For security reasons,`Response.BodyStream()` function return `io.ReadCloser` and lets the user manually release the resource . 

Finally, I want request to support stream and open it up to developers.


streaming.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 type requestStream struct {
-	header          *RequestHeader
+	header interface {
+		ContentLength() int
+		ReadTrailer(r *bufio.Reader) error
+	}
```
erikdubbelboer, line 16: Probably better to define an interface type for this as it's used twice in this file.

erikdubbelboer (changes requested):
Looks good at a first glance, I still have to take a deep look at it.

streaming.go
Anthony-Dong: Thank you. I have modified it.

byene0923:
i'm also look deep into it, i think the code can be more elegant, do you think use `Requeststream`   in response is a good way?

Anthony-Dong:
> 我也在深入研究它，我认为代码可以更优雅，你认为使用`Requeststream` 响应是一个好方法吗？



> i'm also look deep into it, i think the code can be more elegant, do you think use `Requeststream` in response is a good way?

Anthony-Dong:
> i'm also look deep into it, i think the code can be more elegant, do you think use `Requeststream` in response is a good way?



> 

I think it's just a naming issue, since request stream was already supported.

byene0923:
> > i'm also look deep into it, i think the code can be more elegant, do you think use `Requeststream` in response is a good way?
> 
> > 
> 
> I think it's just a naming issue, since request stream was already supported.

you could change the name  such as `bodyStream`

Anthony-Dong:
@erikdubbelboer do you have any questions ?

http.go
```diff
@@ -293,6 +297,40 @@ func (resp *Response) BodyWriter() io.Writer {
...
 
+// BodyStream returns io.Reader
+//
+// You must close it after you use it.
+func (resp *Response) BodyStream() io.ReadCloser {
+	if resp.bodyStream == nil {
+		resp.bodyStream = bytes.NewReader(resp.Body())
+	}
+	return newCloseReader(resp.bodyStream, resp.closeBodyStream)
+}
```
erikdubbelboer, line 308: Is this method useful? We could also not add it for now.

http.go
```diff
@@ -92,6 +92,10 @@ type Response struct {
 	// Relevant for bodyStream only.
 	ImmediateHeaderFlush bool
 
+	// StreamBody enables response body streaming.
+	// Response.BodyStream() get response body.
```
erikdubbelboer, line 96:
```suggestion
	// StreamBody enables response body streaming.
	// Use SetBodyStream to set the body stream.
```

http.go
```diff
@@ -293,6 +297,40 @@ func (resp *Response) BodyWriter() io.Writer {
...
+	return &closeReader{Reader: r, closeFunc: closeFunc}
+}
+
+func (c *closeReader) Close() error {
+	if c.closeFunc == nil {
+		return nil
+	}
+	c.closeOnce.Do(func() {
+		c.err = c.closeFunc()
+	})
```
erikdubbelboer, line 330: Why is this `sync.Once` needed here? If `Close` can be called multiple times, shouldn't `closeFunc` then also not be able to be called multiple times?

erikdubbelboer (changes requested):

http.go
Anthony-Dong: if SetBodyStream is true, the clientConn does not immediately close.  need to close  it manually.  close clientConn does not need to be executed multiple times.

http.go
Anthony-Dong: We hope to read the response body through stream to prevent excessive memory pressure

http.go
Anthony-Dong: For example, `sync.Pool` has the same object and `clientConn` is closed multiple times

Anthony-Dong:
> 



> > > i'm also look deep into it, i think the code can be more elegant, do you think use `Requeststream` in response is a good way?
> > 
> > 
> > > 
> > 
> > 
> > I think it's just a naming issue, since request stream was already supported.
> 
> you could change the name such as `bodyStream`

`type responseStream requestStream` .  might be better

http.go
erikdubbelboer: I'm not sure I understand. Can you show how `Close` would be called multiple times?

Anthony-Dong: eg: `stream.Close()` Execute multiple times

```go
resp := AcquireResponse()
resp.StreamBody = true
request := AcquireRequest()
request.SetRequestURI(server.URL + "?chunked=true")
if err := Do(request, resp); err != nil {
	t.Fatal(err)
}
stream := resp.BodyStream()
defer func() {
	// exec multiple times
	stream.Close()
	stream.Close()
	stream.Close()
}()
content, _ := io.ReadAll(stream)
if string(content) != "0123456789" {
	t.Fatalf("unexpected body content, got: %#v, want: %#v", string(content), "0123456789")
}
```

Anthony-Dong: I thought it wouldn't be good to return an io.ReadCloser directly, so I added the CloseBodyStream method @erikdubbelboer

erikdubbelboer: I did some investigating into this issue, https://github.com/valyala/fasthttp/issues/1504, and it seems it is cause by Close being called multiple times. So I understand the `sync.Once` now. But I'm wondering if we should fix the double close instead.

Anthony-Dong: `perIPConn#Close()` method is also wrapped in `sync.Once` ?  It also seems reasonable and easier to deal with

Anthony-Dong: I think it depends on whether there is concurrent execution of 'closeBodyStream'. If there is, the same problem will indeed occur, but this is an unreasonable use behavior and concurrency framework defect.  @erikdubbelboer

erikdubbelboer:
Finally had the time to look, looks good. Thanks!

PR #1415, opened 2022-11-02
Request.SetTimeout

brian-armstrong-discord:
This functionally works the same as e.g. Client.DoTimeout(), but we can also use it for Client.DoRedirect(). There is no way as far as I can tell to set a timeout on a DoRedirect call, so this makes it possible.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -1566,6 +1566,56 @@ func TestClientFollowRedirects(t *testing.T) {
 		ReleaseResponse(resp)
 	}
 
+	for i := 0; i < 10; i++ {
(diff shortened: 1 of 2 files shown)


li-jin-gou:
Can you add some tests to test the behavior you would expect from this function and some comment?

brian-armstrong-discord:
@li-jin-gou Thanks for the feedback! I've added some tests both for the feature in isolation (in `client.Do()`) and in the context of Redirects. It looks to be working as expected.

erikdubbelboer:
Thanks!

Issue #1408, opened 2022-10-26
fasthttp client errors with io.Pipe()


lfld287:
test code here:

  

  
	  defer func() {
		  if r := recover(); r != nil {
			  t.Fatal("fatal recover triggered", zap.Any("recover", r))
		  }
	  }()
  
	  srcFilePath := "./test/download/test_download.img"
	  f, err := os.Open(srcFilePath)
	  if err != nil {
		  t.Fatal(err)
	  }
  
	  defer func() {
		  _ = f.Close()
	  }()
  
	  pr, pw := io.Pipe()
	  mw := multipart.NewWriter(pw)
  
	  go func() {
		  defer func() {
			  _ = mw.Close()
			  _ = pw.Close()
			  if err != nil {
				  t.Error("upload file write field failed", zap.Error(err))
			  }
		  }()
  
		  err = mw.WriteField("user", "test")
		  if err != nil {
			  return
		  }
  
		  err = mw.WriteField("type", "test")
		  if err != nil {
			  return
		  }
  
		  err = mw.WriteField("sha1", "")
		  if err != nil {
			  return
		  }
  
		  var filePart io.Writer
		  filePart, err = mw.CreateFormFile("file", "test.dat")
		  if err != nil {
			  return
		  }
  
		  _, err = io.Copy(filePart, f)
	  }()
  
	  req := fasthttp.AcquireRequest()
	  defer fasthttp.ReleaseRequest(req)
  
	  req.Header.SetContentType(mw.FormDataContentType())
	  req.Header.SetMethod("POST")
	  req.SetBodyStream(pr, -1)
	  req.SetRequestURI(fmt.Sprintf("%s/file/upload", "http://127.0.0.1:1234"))
  
	  //ignore response
	  err = fasthttp.Do(req, nil)
	  if err != nil {
		  t.Fatal(err)
	  }
  

and result:
=== RUN   TestPipe
    client_test.go:46: fatal recover triggered {recover 15 0 BUG: io.Reader returned 0, nil <nil>}
    client_test.go:67: upload file write field failed {error 26 0  io: read/write on closed pipe}
--- FAIL: TestPipe (0.00s)

lfld287:
i found this : https://github.com/golang/go/commit/5fbb54eaca57e2626e24f1982d28a7de9f91ac49

erikdubbelboer:
[...]

PR #1417, opened 2022-11-03
fix form empty field error when used with pipe

nick9822:
Fix towards issue #1408 

- Set the loop to continue instead of panic making room for next read
- Added test case for the same `TestRequestMultipartFormPipeEmptyFormField`


erikdubbelboer:
Thanks!

PR #1426, opened 2022-11-09
replace s2b function with the recommended one

efectn:
ref https://github.com/gofiber/fiber/issues/2180

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,12 +492,12 @@ func b2s(b []byte) string {
 }
 
 func s2b(s string) (b []byte) {
-    bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
-    sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
-    bh.Data = sh.Data
-    bh.Cap = sh.Len
-    bh.Len = sh.Len
-    return b
+	if s == "" {
+		return nil // or []byte{}
+	}
+	return (*[0x7fff0000]byte)(unsafe.Pointer(
+		(*reflect.StringHeader)(unsafe.Pointer(&s)).Data),
+	)[:len(s):len(s)]
 }
 ```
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
I'm going to hold of on this one until we can use [`unsafe.String`](https://pkg.go.dev/unsafe@master#String) and [`unsafe.StringData`](https://pkg.go.dev/unsafe@master#StringData) which are the real future proof ways to do this.

efectn:
> I'm going to hold of on this one until we can use [`unsafe.String`](https://pkg.go.dev/unsafe@master#String) and [`unsafe.StringData`](https://pkg.go.dev/unsafe@master#StringData) which are the real future proof ways to do this.

Agree but it will be available on Go 1.20. I think we can't use it on Fasthttp.

erikdubbelboer:
We can. We just put the current `s2b` etc functions in a file that only gets used with go 1.19 or lower and use `unsafe.String` etc in a file that gets used with go 1.20 or higher.

stokito:
Could you please elaborate, so starting from Go1.20 the s2b() will not work?
I'm asking because for my project I also needed the func and I copied it to a module
https://github.com/stokito/go-strbytes
So I have to update it too.

It would be great if the FastHttp makes the func public or extracted to a some utils module so that users can easily update.

erikdubbelboer:
`s2b` will still work fine in Go 1.20.

But what we can do is have 1 file with something like:
```go
//go:build go1.20

func s2b(s string) []byte {
	return unsafe.Slice(unsafe.StringData(s), len(s))
}
```
and another with the old `s2b` implementation and this build tag:
```go
//go:build !go1.20
```

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/pull/1481

PR #1432, opened 2022-11-15
Assert with *net.TCPConn instead of *net.TCPListener in acceptConn() for TCP sockets

panjf2000:
Make the code more succinct.


panjf2000:
@erikdubbelboer

PR #1433, opened 2022-11-15
Fix some potential pool leaks

panjf2000:
These cases could result in leaks of the pool when they return early, and the `DoDeadline()` forgot to call `releasePipelineWork()` when it's timed out waiting for the response.

### DIFF
--- a/client.go
+++ b/client.go
@@ -983,6 +983,8 @@ var clientURLResponseChPool sync.Pool
 
 func clientPostURL(dst []byte, url string, postArgs *Args, c clientDoer) (statusCode int, body []byte, err error) {
 	req := AcquireRequest()
+	defer ReleaseRequest(req)
+
 	req.Header.SetMethod(MethodPost)
 	req.Header.SetContentTypeBytes(strPostArgsContentType)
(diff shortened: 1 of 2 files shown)


panjf2000:
@erikdubbelboer PTAL

client.go
```diff
@@ -2376,6 +2377,8 @@ func (c *pipelineConnClient) DoDeadline(req *Request, resp *Response, deadline t
 	}
 
 	w := acquirePipelineWork(&c.workPool, timeout)
+	defer releasePipelineWork(&c.workPool, w)
```
erikdubbelboer, line 2380: This will go wrong in the last ErrTimeout case where w might still be in use by the worker.

erikdubbelboer (changes requested):

client.go
panjf2000: You're right, reverted it.

panjf2000: But still, we need to handle this case and make sure to put it back into the pool, I'll send another PR for it later.

erikdubbelboer:
Thanks!

PR #1434, opened 2022-11-15
Use time.Until(deadline) instead of -time.Since(deadline)

panjf2000:
As title.


panjf2000:
@erikdubbelboer PTAL

erikdubbelboer:
Thanks!

PR #1435, opened 2022-11-16
Wait for the response of pipelineWork in background and return it to pool

panjf2000:
Updates #1433

### DIFF
--- a/client.go
+++ b/client.go
@@ -2411,6 +2411,11 @@ func (c *pipelineConnClient) DoDeadline(req *Request, resp *Response, deadline t
 		releasePipelineWork(&c.workPool, w)
 	case <-w.t.C:
 		err = ErrTimeout
+		go func() {
+			// Wait for the response in background.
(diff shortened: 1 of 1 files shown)


panjf2000:
@erikdubbelboer PTAL

erikdubbelboer (changes requested):
I don't think this is worth the effort. The goroutine probably has a bigger overhead than just letting the allocation go to waste and having to allocate a new w next time don't you think?

panjf2000:
> I don't think this is worth the effort. The goroutine probably has a bigger overhead than just letting the allocation go to waste and having to allocate a new w next time don't you think?

Well, the memory waste caused by the case of timeout is indeed rare, I guess we can live with that, also thanks for your reply.

Tristin-C:
Big man, is that going to lead to a leak? Causing the number of coroutines to soar? Because I seem to have that problem

PR #1436, opened 2022-11-16
Make acquirePipelineWork() and releasePipelineWork() the member functions of pipelineConnClient

panjf2000:
Since the `workPool` is the member field of `pipelineConnClient`, there is no need to pass the `workPool` explicitly as the parameter every time calling `acquirePipelineWork()` and `releasePipelineWork()`.


panjf2000:
@erikdubbelboer PTAL

erikdubbelboer:
Thanks!

PR #1441, opened 2022-11-20
clean tmp file for SaveMultipartFile

laoshanxi:
https://github.com/valyala/fasthttp/issues/1439

### DIFF
--- a/server.go
+++ b/server.go
@@ -1053,16 +1053,17 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
 		return
 	}
 
+	var tmpfile string
 	var ok bool
 	if ff, ok = f.(*os.File); ok {
 		// Windows can't rename files that are opened.
 		if err = f.Close(); err != nil {
 			return
 		}
-
+		tmpfile = ff.Name()
 		// If renaming fails we try the normal copying method.
 		// Renaming could fail if the files are on different devices.
-		if os.Rename(ff.Name(), path) == nil {
+		if os.Rename(tmpfile, path) == nil {
 			return nil
 		}
 
@@ -1074,6 +1075,9 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
 
 	defer func() {
 		e := f.Close()
+		if e == nil && tmpfile != "" {
+			e = os.Remove(tmpfile)
+		}
 		if err == nil {
 			err = e
 		}


server.go
```diff
@@ -1053,16 +1053,17 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
...
 
+	var tmpfile string
 	var ok bool
 	if ff, ok = f.(*os.File); ok {
 		// Windows can't rename files that are opened.
 		if err = f.Close(); err != nil {
 			return
 		}
-
+		tmpfile = ff.Name()
```
li-jin-gou, line 1063: Is it possible to add unit tests?

li-jin-gou (changes requested):

server.go
laoshanxi: not possible, the case happened when save file to another logic volumn.

erikdubbelboer (changes requested):
I'm looking into this and I don't think this is the right solution. The problem I'm seeing is that `req.RemoveMultipartFormFiles()` is not always called. Calling this function should remove all temporary files. I'm traveling right now and can't look into this further, maybe you can look into this?

laoshanxi:
> I'm looking into this and I don't think this is the right solution. The problem I'm seeing is that `req.RemoveMultipartFormFiles()` is not always called. Calling this function should remove all temporary files. I'm traveling right now and can't look into this further, maybe you can look into this?

I saw that, `ctx.Request.RemoveMultipartFormFiles()` is in the upper layer of `fasthttp.SaveMultipartFile()`, I need try this function first.

laoshanxi:
Tried `ctx.Request.RemoveMultipartFormFiles()` could clean the tmp file, but I still think expose such a function for Request is not a good idea, not all the request is multipart, we should keep atomic `fasthttp.SaveMultipartFile()` instead of use a upper layer function clean the tmp file.

erikdubbelboer:
`ctx.Request.RemoveMultipartFormFiles()` should automatically be called by `Server` when your handler returns.
See [`RequestCtx.FormFile`](https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx.FormFile):
```
The file is automatically deleted after returning from RequestHandler, so either move or copy uploaded file
into new place if you want retaining it. 
```
but apparently it doesn't do this anymore.

So someone has to add this back.

erikdubbelboer:
I have looked into this and I can't find what could be causing it.

`Server.serveConn()` always calls [`ctx.Request.Reset()`](https://github/valyala/fasthttp/blob/49951353c83ac7b3ca9a78570dbea0c970b6d801/server.go#L2447) which calls `Request.resetSkipHeader()`, which calls `Request.ResetBody()`, which calls `Request.RemoveMultipartFormFiles()`. So all files should always be removed.
`SaveMultipartFile()` only creates the file at the path you specify, it doesn't create any tmp files.

How often does this happen to you? Could it for example be that these files are left from the process being killed?

laoshanxi:
I am running my application in Docker container and mount /data from host OS to container in order to save file for fasthttp.

we can see `Request.RemoveMultipartFormFiles()` will first try rename file to another position if I also use target location from `/tmp` that means here we tried to `move` file but use `copy` when `move` failed because inside container can not move `/tmp` to mounted `/data`.

when I use with this way, it really works:
https://github.com/laoshanxi/app-mesh/blob/main/src/sdk/agent/restagent.go#L251

And also when I use this way, it also works:
https://github.com/laoshanxi/app-mesh/commit/7b8b826975d875f006fb79a6e17215cbe40dd6df#diff-787317aba97c76198e13c8447f2649bba285a2a7ab621d292c6199678f3606f1L229

So I am talking how `fasthttp.SaveMultipartFile` and `ctx.Request.RemoveMultipartFormFiles()` are designed here, and whether we can remove `ctx.Request.RemoveMultipartFormFiles()`.

laoshanxi:
> But like I showed `RemoveMultipartFormFiles` will always be called when your handler returns. So you shouldn't have to call it manually. If you remove that `defer ctx.Request.RemoveMultipartFormFiles()` call, do these tmp files start showing up immediately. How often does it happen?

Yes, if I remove `defer ctx.Request.RemoveMultipartFormFiles()`, the tmp files was always left, you can see the issue I loged.
In order to release docker image container, I already removed the docker container, but I think every time it will happen.

PR #1444, opened 2022-11-25
client.go fix addMissingPort()

stokito:
Currently it has a minor bug when the default port is not added for IPv6 addreses.
Likely in practice it doesn't affect the system.
Anyway please accept the PR that improves a performance a little bit and also exposes the addMissingPort() so that we can use it our apps code (now I had to copy the func).

### DIFF
--- a/client.go
+++ b/client.go
@@ -495,7 +495,7 @@ func (c *Client) Do(req *Request, resp *Response) error {
 	hc := m[string(host)]
 	if hc == nil {
 		hc = &HostClient{
-			Addr:                          addMissingPort(string(host), isTLS),
+			Addr:                          AddMissingPort(string(host), isTLS),
 			Name:                          c.Name,
 			NoDefaultUserAgentHeader:      c.NoDefaultUserAgentHeader,
 			Dial:                          c.Dial,
@@ -1968,7 +1968,7 @@ func dialAddr(addr string, dial DialFunc, dialDualStack, isTLS bool, tlsConfig *
 		} else {
 			dial = Dial
 		}
-		addr = addMissingPort(addr, isTLS)
+		addr = AddMissingPort(addr, isTLS)
 	}
 	conn, err := dial(addr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -2006,16 +2005,37 @@ func (c *HostClient) getClientName() []byte {
 	return clientName
 }
 
-func addMissingPort(addr string, isTLS bool) string {
-	n := strings.Index(addr, ":")
-	if n >= 0 {
+func AddMissingPort(addr string, isTLS bool) string {
```
li-jin-gou, line 2008:
```
// AddMissingPort adds a port to a host if it is missing.
// A literal IPv6 address in hostport must be enclosed in square
// brackets, as in "[::1]:80", "[::1%lo0]:80".
func AddMissingPort(addr string, isTLS bool) string {
	if strings.IndexByte(addr, ':') >= 0 {
		endOfV6 := strings.IndexByte(addr, ']')
		// we do not care about the validity of the address, just check if it has more bytes after ']'
		if endOfV6 < len(addr)-1 {
			return addr
		}
	}
	if !isTLS {
		return addr + ":80"
	}
	return addr + ":443"
}

```
This is a much cleaner implementation and remember to add comments and Unit tests to AddMissingPort.

li-jin-gou (changes requested):

client.go
stokito: Thank you, your version is cleaner. However it won't work for IPv6 without brackets i.e. `::1` but we don't know if the plain IPv6 addresses are already in use.
My version may be not so clean or efficient but more safe.

li-jin-gou: > Thank you, your version is cleaner. However it won't work for IPv6 without brackets i.e. `::1` but we don't know if the plain IPv6 addresses are already in use. My version may be not so clean or efficient but more safe.

However, in the case of web development scenarios, according to RFC https://www.ietf.org/rfc/rfc2732.txt, [] is definitely present,

stokito: Yes, totally agree. Seems like we safe here. Even if someone creates manually the HostClient with the ipv6 without brackets it will work fine as previously. I'll change the pr later to your version.

erikdubbelboer (changes requested):
Looks good but your test seems to be failing.

stokito:
@erikdubbelboer now the pr is ready to be merged. 
@li-jin-gou thank you for your contribution. 👍

erikdubbelboer:
Thanks!

PR #1449, opened 2022-11-30
add optional simulated addresses to pipeconn and inmemorylistener

tobikris:
I have a use case where I need the code using this `InmemoryListener` to access the simulated address of the connection endpoints.
This PR adds optional addresses to the `PipeConns` and `InmemoryListener`.
If the feature is not used, the code behaves like before.

Example:

```go
func main() {
	ln := NewInmemoryListener()

	listenerAddr := &net.TCPAddr{IP: net.IPv4(127, 0, 0, 1), Port: 12345}

	ln.SetLocalAddr(listenerAddr)

	go client(ln)

	lc, err := ln.Accept()
	if err != nil {
		log.Fatalf("unexpected error: %v", err)
	}

	log.Println(lc.LocalAddr(), lc.RemoteAddr())
}

func client(ln *InmemoryListener) {
	clientAddr := &net.TCPAddr{IP: net.IPv4(127, 0, 0, 2), Port: 65432}

	c, err := ln.DialWithLocalAddr(clientAddr)
	if err != nil {
		log.Fatalf("unexpected error: %v", err)
	}

	log.Println(c.LocalAddr(), c.RemoteAddr())
}
```

### DIFF
--- a/fasthttputil/inmemory_listener.go
+++ b/fasthttputil/inmemory_listener.go
@@ -18,6 +18,7 @@ type InmemoryListener struct {
 	closed       bool
 	conns        chan acceptConn
 	listenerAddr net.Addr
+	addrLock     sync.RWMutex
 }
 
 type acceptConn struct {
@@ -34,6 +35,9 @@ func NewInmemoryListener() *InmemoryListener {
 
 // SetLocalAddr sets the (simulated) local address for the listener.
 func (ln *InmemoryListener) SetLocalAddr(localAddr net.Addr) {
+	ln.addrLock.Lock()
+	defer ln.addrLock.Unlock()
+
 	ln.listenerAddr = localAddr
 }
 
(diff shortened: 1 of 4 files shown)


fasthttputil/inmemory_listener.go
```diff
@@ -60,12 +66,23 @@ func (ln *InmemoryListener) Close() error {
...
+	return "InmemoryListener"
+}
+
 // Addr implements net.Listener's Addr.
 func (ln *InmemoryListener) Addr() net.Addr {
-	return &net.UnixAddr{
-		Name: "InmemoryListener",
-		Net:  "memory",
+	if ln.listenerAddr != nil {
+		return ln.listenerAddr
```
li-jin-gou, line 82: Do `SetLocalAddr()` and `Addr()` need to be considered for concurrency?

tobikris: You could argue that it needs to be. However, I thought the only meaningful way to set the address would be to do it before the listener is used in any way. I just did not add it to `NewInmemoryListener()` to leave this completely optional.

It is probably better to guard against data races, though and do not leave this to the caller. I will implement it, thanks!

erikdubbelboer:
Thanks!

PR #1453, opened 2022-12-08
feat: support custom formvalue function

li-jin-gou:
Some gin or net/http services have a different formvalue  function when migrating to fasthttp.
it maybe causes some bugs.

net/http:
```
// FormValue returns the first value for the named component of the query.
// POST and PUT body parameters take precedence over URL query string values.
// FormValue calls ParseMultipartForm and ParseForm if necessary and ignores
// any errors returned by these functions.
// If key is not present, FormValue returns the empty string.
// To access multiple values of the same key, call ParseForm and
// then inspect Request.Form directly.
```

The user can customize a formvalue function to avoid this problem, but it would be better if the framework provided this switch.

### DIFF
--- a/server.go
+++ b/server.go
@@ -406,6 +406,12 @@ type Server struct {
 	// instead.
 	TLSConfig *tls.Config
 
+	// FormValueFunc, which is used by RequestCtx.FormValue and support for customising
+	// the behaviour of the RequestCtx.FormValue function.
+	//
+	// NetHttpFormValueFunc gives a FormValueFunc func implementation that is consistent with net/http.
+	FormValueFunc FormValueFunc
+
 	nextProtos map[string]ServeHandler
 
 	concurrency      uint32
@@ -604,6 +610,7 @@ type RequestCtx struct {
 
 	hijackHandler    HijackHandler
 	hijackNoResponse bool
+	formValueFunc    FormValueFunc
 }
 
 // HijackHandler must process the hijacked connection c.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -1108,6 +1108,12 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
 //
 // The returned value is valid until your request handler returns.
 func (ctx *RequestCtx) FormValue(key string) []byte {
+	return defaultFormValue(ctx, key)
+}
+
+type FormValue func(*RequestCtx, string) []byte
```
erikdubbelboer, line 1114:
```suggestion
type FormValueFunc func(*RequestCtx, string) []byte
```

server.go
```diff
@@ -1126,6 +1132,34 @@ func (ctx *RequestCtx) FormValue(key string) []byte {
...
 }
 
+// SetCustomFormValueFunc sets FormValue function implementation to get form value.
+func SetCustomFormValueFunc(f FormValue) {
+	defaultFormValue = f
+}
+
+// SetStandardFormValueFunc sets FormValue function implementation to get form value.
+// Consistent behaviour with net/http. POST and PUT body parameters take precedence over URL query string values.
+func SetStandardFormValueFunc() {
```
erikdubbelboer, line 1142:
```suggestion
func SetNetHttpFormValueFunc() {
```

erikdubbelboer (changes requested):
Do you have a real world example of this going wrong or is it something you think might go wrong?

server.go
li-jin-gou: done

server.go
li-jin-gou: done

li-jin-gou:
> Do you have a real world example of this going wrong or is it something you think might go wrong?

There was a problem with the code in the real environment and I just noticed the difference. the front end passes the same parameters and different values in the http body and query and result in an unintended bug...😩

server.go
```diff
@@ -412,6 +412,7 @@ type Server struct {
 	concurrencyCh    chan struct{}
 	perIPConnCounter perIPConnCounter
 	serverName       atomic.Value
+	FormValueFunc    FormValueFunc
```
erikdubbelboer, line 415: This field needs some documentation since it's exposed.

server.go
```diff
@@ -1108,23 +1110,54 @@ func SaveMultipartFile(fh *multipart.FileHeader, path string) (err error) {
...
+		}
+		return nil
 	}
-	mf, err := ctx.MultipartForm()
-	if err == nil && mf.Value != nil {
-		vv := mf.Value[key]
-		if len(vv) > 0 {
-			return []byte(vv[0])
+
+	// NetHttpFormValueFunc Consistent behaviour with net/http. POST and PUT body parameters take precedence over URL query string values.
```
erikdubbelboer, line 1141:
```suggestion
	// NetHttpFormValueFunc gives consistent behavior with net/http. POST and PUT body parameters take precedence over URL query string values.
```

erikdubbelboer (changes requested):

server.go
li-jin-gou: thanks

server.go
li-jin-gou: thanks and done

erikdubbelboer:
Thanks!

Issue #1219, opened 2022-02-16
TimeoutHandler doesn't cancel the context


dimovnike:
The context doesnt become Done() on the timeout.

for example, this code:

```
func main() {
	fasthttp.ListenAndServe(":8080", fasthttp.TimeoutHandler(func(ctx *fasthttp.RequestCtx) {
		log.Println("begin")

		tm := time.NewTimer(10 * time.Second)

		select {
		case <-ctx.Done():
			log.Println("context cancelled")
			return
		case <-tm.C:
			log.Println("process finished ok")
		}

		log.Println("end")
	}, time.Second*2, "timeout"))
}
```

prints:
```
2022/02/16 12:52:56 begin
2022/02/16 12:53:06 process finished ok
2022/02/16 12:53:06 end
```

The client receives the timeout after 2 seconds, and the handler finishes after 10 seconds yet. It would be nice if the context became Done() in this case. If this would break existing code expectations then it can be made optional (via setting or a new TimeoutHandlerWithShutdown or something like it).

Will you accept a PR for this?

dimovnike:
note that the het/http behaves as (intuitively) expected:

```
func main() {
	srv := http.Server{
		Addr: ":8080",
		Handler: http.TimeoutHandler(http.HandlerFunc(
			func(w http.ResponseWriter, req *http.Request) {
				log.Println("begin")

				tm := time.NewTimer(10 * time.Second)

				select {
				case <-req.Context().Done():
					log.Println("context cancelled")
					return
				case <-tm.C:
					log.Println("process finished ok")
				}

				log.Println("end")
			},
		), 1*time.Second, "Timeout!\n"),
	}

	if err := srv.ListenAndServe(); err != nil {
		fmt.Printf("Server failed: %s\n", err)
	}
}
```

prints


```
2022/02/16 13:08:25 begin
2022/02/16 13:08:26 context cancelled
```

erikdubbelboer:
[...]

PR #1454, opened 2022-12-18
doc: optimize the comment of the Request.Done method

li-jin-gou:
PR for https://github.com/valyala/fasthttp/issues/1219

### DIFF
--- a/server.go
+++ b/server.go
@@ -2697,6 +2697,9 @@ func (ctx *RequestCtx) Deadline() (deadline time.Time, ok bool) {
 // Done returns a channel that's closed when work done on behalf of this
 // context should be canceled. Done may return nil if this context can
 // never be canceled. Successive calls to Done return the same value.
+//
+// Note: Because creating a new channel for every request is just too expensive, so
+// RequestCtx.s.done is only closed when the server is shutting down
 func (ctx *RequestCtx) Done() <-chan struct{} {
 	return ctx.s.done
 }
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -2697,6 +2697,9 @@ func (ctx *RequestCtx) Deadline() (deadline time.Time, ok bool) {
 // Done returns a channel that's closed when work done on behalf of this
 // context should be canceled. Done may return nil if this context can
 // never be canceled. Successive calls to Done return the same value.
+//
+// Note: Because creating a new channel for every request is just too expensive so
```
erikdubbelboer, line 2701:
```suggestion
// Note: Because creating a new channel for every request is just too expensive,
```

server.go
```diff
@@ -2707,6 +2710,9 @@ func (ctx *RequestCtx) Done() <-chan struct{} {
 // If Done is closed, Err returns a non-nil error explaining why:
 // Canceled if the context was canceled (via server Shutdown)
 // or DeadlineExceeded if the context's deadline passed.
+//
+// Note: Because creating a new channel for every request is just too expensive so
```
erikdubbelboer, line 2714:
```suggestion
// Note: Because creating a new channel for every request is just too expensive,
```

erikdubbelboer (changes requested):
Good comment, but just one small change.

li-jin-gou:
> Good comment, but just one small change.

done

erikdubbelboer:
Thanks!

Issue #1455, opened 2022-12-20
fasthttp.Server GetOnly should also support HEAD requests


TheKigen:
Hi,

While GetOnly does suggest only GET requests should be supported I think it'd be way more useful of a setting if HEAD requests were also allowed with GetOnly set.  HEAD requests are functionally similar to GET with the exception that the body of the reply is not expected.  HEAD requests are useful for monitoring when you just want to see if the HTTP status is 200.  And the main purpose I see behind GetOnly is to block DoS via pushing request bodies and random methods that various bots spam HTTP servers with.

TheKigen:
Closing this due to the PR being merged.

PR #1456, opened 2022-12-21
optimize: allow http head when use getonly

li-jin-gou:
pr for https://github.com/valyala/fasthttp/issues/1455

### DIFF
--- a/http.go
+++ b/http.go
@@ -1129,7 +1129,7 @@ func (req *Request) readLimitBody(r *bufio.Reader, maxBodySize int, getOnly bool
 	// Do not reset the request here - the caller must reset it before
 	// calling this method.
 
-	if getOnly && !req.Header.IsGet() {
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
I guess it makes sense. But please also change the documentation here to reflect this: https://github.com/valyala/fasthttp/blob/2a572e08ef547b02bec1328c150ed31deec31d17/server.go#L292-L299

li-jin-gou:
> I guess it makes sense. But please also change the documentation here to reflect this:
> 
> https://github.com/valyala/fasthttp/blob/2a572e08ef547b02bec1328c150ed31deec31d17/server.go#L292-L299

thanks and done

erikdubbelboer:
Thanks!

TheKigen:
Thank you.

PR #1457, opened 2022-12-22
fix adapter support http2

pedia:
Since the [adapter](fasthttpadaptor/request.go) hardcode `HTTP/1.1`:
```go
	r.Proto = "HTTP/1.1"
	r.ProtoMajor = 1
```

Maybe this is better?
```go
	r.Proto = b2s(ctx.Request.Header.Protocol())
	if r.Proto == "HTTP/2" {
		r.ProtoMajor = 2
	} else {
		r.ProtoMajor = 1
	}
```

With this change, compose  `fasthttp` and `grpc` together is possible.

Usage:
```go
// start server
ws := &fasthttp.Server{...}
http2.ConfigureServer(ws)
ws.ServeTLS(...)

// transform to grpc request
if bytes.Equal(ctx.Request.Header.ContentType(), grpcContentType) {
    fh := fasthttpadaptor.NewFastHTTPHandler(a.Rpc)
    fh(ctx)
    return
}
```

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -109,3 +109,5 @@ func (w *netHTTPResponseWriter) WriteHeader(statusCode int) {
 func (w *netHTTPResponseWriter) Write(p []byte) (int, error) {
 	return w.w.Write(p)
 }
+
+func (w *netHTTPResponseWriter) Flush() {}
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -25,8 +25,12 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 	}
 
 	r.Method = b2s(ctx.Method())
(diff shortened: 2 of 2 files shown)


pedia:
Agree, This is a small step.

erikdubbelboer:
How would this work? In your example you use `http2.ConfigureServer(ws)` but [`http2.ConfigureServer`](https://pkg.go.dev/golang.org/x/net/http2#ConfigureServer) expects a `http.Server` as argument, not a `fasthttp.Server`.

A fasthttp server isn't be able to receive `HTTP/2` requests, so how would that `fasthttp.RequestCtx` with a `Proto` of `HTTP/2` ever be created?

pedia:
in `http2.ConfigureServer(ws)`, `http2` come from https://github.com/dgrr/http2

I think http2.ConfigureServer patched a little.

Testing with `curl --http2`, it passed.

erikdubbelboer:
Thanks!

PR #1461, opened 2022-12-28
fix: client may remove hostclient incorrectly

Duslia:
The cleaner may remove hostclient unexpectedly.
I find that client will start cleaner when the first hostclient created.
image
The cleaner will remove hostclient if there is no conn in the hostclient. And cleaner will start judge logic immediately. At the same time, hostclient will create conn. So if client get the connsLock before hostclient create conn, the hostclient will be removed.
image
 It may cause resource leak.

### DIFF
--- a/client.go
+++ b/client.go
@@ -569,6 +569,7 @@ func (c *Client) mCleaner(m map[string]*HostClient) {
 	}
 
 	for {
+		time.Sleep(sleep)
 		c.mLock.Lock()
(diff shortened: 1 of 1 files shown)


li-jin-gou (changes requested):
Can you describe the PR in more detail?

li-jin-gou (approved):

erikdubbelboer:
Thanks, good find!

PR #1463, opened 2022-12-28
[WIP]optimize: pooled HeaderScanner and HeaderValueScanner

li-jin-gou:


li-jin-gou:
performance is not improved, select off.

PR #1466, opened 2023-01-01
client.go Simplify default UA logic

### LINKED ISSUE #1458: client.go: getClientName() may be removed
When client performs a request it sets a UA if it wasn't set:
```go
userAgentOld := req.Header.UserAgent()
if len(userAgentOld) == 0 {
	req.Header.userAgent = append(req.Header.userAgent[:0], c.getClientName()...)
}
```
The getClientName() checks if !NoDefaultUserAgentHeader then returns the Client.Name field.
But it also saves it to atomic field `clientName`:
```go
func (c *pipelineConnClient) getClientName() []byte {
	v := c.clientName.Load()
	var clientName []byte
	if v == nil {
		clientName = []byte(c.Name)
		if len(clientName) == 0 && !c.NoDefaultUserAgentHeader {
			clientName = defaultUserAgent
		}
		c.clientName.Store(clientName)
	} else {
		clientName = v.([]byte)
	}
	return clientName
}
```
From what I see the clientName is basically just not needed. It never changes.
We can safely simplify to something like:
```go
userAgentOld := req.Header.UserAgent()
if len(userAgentOld) == 0 {
	userAgent := c.Name
	if userAgent == "" && !c.NoDefaultUserAgentHeader {
		userAgent = defaultUserAgent
	}
	if userAgent != "" {
		req.Header.userAgent = append(req.Header.userAgent[:], userAgent...)
	}
}
```

Is there any reason to have that atomic or I get something wrong?
@kiyonlin you was a last who changed that code, please participate
https://github.com/valyala/fasthttp/commit/02e0722fb73c6237818b8e5af55957eb919a7334

stokito:
The getClientName() checks if !NoDefaultUserAgentHeader then returns the Client.Name field. But it also saves it to atomic field clientName. This is not needed and logic can be simplified. Previously the clientName vas a byte slice that was copied from c.Name and cached. See 02e0722fb73c6237818b8e5af55957eb919a7334

Fix #1458

### DIFF
--- a/client.go
+++ b/client.go
@@ -788,7 +788,6 @@ type HostClient struct {
 	// Connection pool strategy. Can be either LIFO or FIFO (default).
 	ConnPoolStrategy ConnPoolStrategyType
 
-	clientName  atomic.Value
 	lastUseTime uint32
 
 	connsLock  sync.Mutex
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Thanks!

PR #1467, opened 2023-01-03
server.go Simplify default Server name logic

stokito:
The serverName atomic.Value field is used as a cache. This is not needed and logic can be simplified. See related #1458


erikdubbelboer:
Thanks!

PR #1477, opened 2023-01-28
fix: handling non-disc cases that do not begin with /

li-jin-gou:
Windows processing logic only considers non-disk characters and empty strings.
such as `"a/b"` will not add `'/'`


li-jin-gou:
Close 1477 and code to 1478

PR #1482, opened 2023-02-05
update `github.com/klauspost/compress`, drop 1.16

efectn:
github.com/klauspost/compress no longer supports 1.16 and 1.16 is not exists on Fasthttp workflows.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.20
 
 require (
 	github.com/andybalholm/brotli v1.0.4
-	github.com/klauspost/compress v1.15.9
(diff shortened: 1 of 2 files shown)


li-jin-gou (approved):

erikdubbelboer:
Why does the title include `drop 1.16`?

efectn:
> Why does the title include `drop 1.16`?

Currently it's possible to use fasthttp on 1.16, after this update it won't be possible anymore. I wanted to mention it

erikdubbelboer:
Ok clear. Thanks!

PR #1497, opened 2023-02-16
Immediately return ErrTimeout if deadline is already reached.

iv-menshenin:
I found this while developing unit tests for my application.

If we call DoDeadline without first checking that the deadline has expired, then we don't get the expected behavior of returning ErrTimeout.

The implementation (c *pipelineConnClient) of DoDeadline already contains the necessary check

```
	if req.timeout < 0 {
		return ErrTimeout
	}
```

I also added it for
 - Client
 - HostClient

### DIFF
--- a/client.go
+++ b/client.go
@@ -374,6 +374,7 @@ func (c *Client) Post(dst []byte, url string, postArgs *Args) (statusCode int, b
 //
 // ErrTimeout is returned if the response wasn't returned during
 // the given timeout.
+// Immediately returns ErrTimeout if timeout value is negative.
 //
 // ErrNoFreeConns is returned if all Client.MaxConnsPerHost connections
 // to the requested host are busy.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -1152,6 +1161,9 @@ func ReleaseResponse(resp *Response) {
 // try setting a ReadTimeout.
 func (c *HostClient) DoTimeout(req *Request, resp *Response, timeout time.Duration) error {
 	req.timeout = timeout
+	if req.timeout < 0 {
```
li-jin-gou, line 1164: add unit test for DoTimeout

client.go
```diff
@@ -387,6 +388,9 @@ func (c *Client) Post(dst []byte, url string, postArgs *Args) (statusCode int, b
 // try setting a ReadTimeout.
 func (c *Client) DoTimeout(req *Request, resp *Response, timeout time.Duration) error {
 	req.timeout = timeout
+	if req.timeout < 0 {
```
erikdubbelboer, line 391: Why not `<=`?

client.go
iv-menshenin: i will

client.go
iv-menshenin: `timeout == 0` this is a special case, it is considered that then the timeout is not set  
this behavior is consistent with the standard http driver

erikdubbelboer:
Thanks!

Issue #1522, opened 2023-03-18
Don't panic for non-critical errors


mpldr:
Grepping the codebase, I noticed a number of panics in non-critical codepaths. This might be annoying as in #1504, or it might open users up to DoS attacks if someone finds a way to trigger something like [server.go:1947](https://github.com/valyala/fasthttp/blob/53e376f82282fd6437b5e0189e80dff0add50fe3/server.go#L1947) reliably.

This is *not* a vulnerability. It's just a design flaw, that might lead to a vulnerability when combined with bugs or rare behaviour.

Personally, I think it would be a better™ approach to log them and move on. "Connection is nil? The connection is invalid and gets dropped", "IP count is negative? Looks like a race condition. Nothing endangering data integrity".

```console
$ rg 'panic\("BUG:'
bytesconv.go
82:		panic("BUG: dst must not be nil")
129:		panic("BUG: int must be positive")
284:		panic("BUG: int must be positive")

client.go
1314:		panic("BUG: req cannot be nil")
1317:		panic("BUG: resp cannot be nil")
1997:		panic("BUG: DialFunc returned (nil, nil)")

compress.go
69:		panic("BUG: zlib.Reader doesn't implement zlib.Resetter???")

fs.go
605:		panic("BUG: negative fsFile.readersCount!")
1398:			panic("BUG: path must start with slash")

http.go
966:		panic("BUG: form boundary cannot be empty")
2200:		panic("BUG: expected zero-length buffer")

lbclient.go
87:		panic("BUG: LBClient.Clients cannot be empty")

peripconn.go
30:		panic("BUG: perIPConnCounter.Register() wasn't called")

server.go
1933:				panic("BUG: net.Listener returned non-nil conn and non-nil error")
1947:			panic("BUG: net.Listener returned (nil, nil)")
2581:		panic("BUG: cannot write timed out response")
2615:		panic("BUG: Reader must return at least one byte")
2790:	panic("BUG: unexpected Read call")
2794:	panic("BUG: unexpected Write call")
2798:	panic("BUG: unexpected Close call")
2803:		panic("BUG: cannot release timed out RequestCtx")

stackless/func.go
23:		panic("BUG: f cannot be nil")

timer.go
13:		panic("BUG: active timer trapped into initTimer()")

workerpool.go
50:		panic("BUG: workerPool already started")
75:		panic("BUG: workerPool wasn't started")
```

erikdubbelboer:
[...]

Issue #1504, opened 2023-02-25
Panic on server shutdown


mpldr:
I encounter [this panic](https://github.com/valyala/fasthttp/blob/v1.44.0/peripconn.go#L35) when shutting down the server.

The steps are:
- start server
- get index page with firefox
- send sigint (invalidates the context)
- see panic

Server setup/shutdown:
```go
	lis, err := net.Listen("tcp", "localhost:8080")
	if err != nil {
		glog.Errorf("failed to listen on %s: %v", ":8080", err)
		return
	}
	srv := &fasthttp.Server{
		Handler:                      handler.RootHandler,
		MaxConnsPerIP:                1,
		MaxRequestBodySize:           260 * 1024 * 1024,
		DisablePreParseMultipartForm: true,
		StreamRequestBody:            true,
		TLSConfig:                    &tls.Config{},
	}
	httpErr := make(chan error)
	go func() {
		httpErr <- srv.Serve(lis)
	}()

	select {
	case err = <-httpErr:
		if err != nil {
			glog.Errorf("server encoutered an error. not good: %v", err) // TODO: might want to rephrase this
		}
		errchan <- err
		fmt.Println(err)
	case <-ctx.Done():
		ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
		srv.ShutdownWithContext(ctx)
		cancel()
		return
	}

```

mpldr:
Not sure if the first even has to be a panic, to be honest. Looks like nothing but a sanity check to me.

li-jin-gou:
hello @mpldr  Can you provide a complete runnable code ? I would like to try to reproduce this problem in my computer.

mpldr:
Sure, here you go: http://0x0.st/Hs5T.tzst

It's not yet part of any repo, so a tarball is the best I can offer you right away :)

erikdubbelboer:
The problem is that `Shutdown` calls `closeIdleConns`. Which calls `Close` on all connections here: https://github.com/valyala/fasthttp/blob/559d536fd42f8d6099fa88bf788e2cc7dbb65d83/server.go#L2903
But then when the worker goroutine stops it also calls `Close` on the connection here: https://github.com/valyala/fasthttp/blob/404c8a896896943715f8fb3906a8d054fae17d3e/workerpool.go#L238

[...]

PR #1526, opened 2023-03-20
get rid of some panics

mpldr:
Addresses #1522 and #1504. The latter is not fixed, but only referenced as the underlying double-close is not resolved.

Fixes: #1522

### DIFF
--- a/brotli.go
+++ b/brotli.go
@@ -138,10 +138,7 @@ func nonblockingWriteBrotli(ctxv interface{}) {
 	ctx := ctxv.(*compressCtx)
 	zw := acquireRealBrotliWriter(ctx.w, ctx.level)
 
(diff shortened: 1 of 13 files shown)


erikdubbelboer (changes requested):
Can you fix the linting errors.

mpldr:
On Sun Mar 26, 2023 at 10:28 PM CEST, Erik Dubbelboer wrote:
> Can you fix the linting errors.

Will do, still not completely through, I also want to add commentary
explaining why the remaining panics should stay.
-- 
Moritz Poldrack
https://moritz.sh

mpldr:
/shrug Can't do anything about the tests taking too long

erikdubbelboer:
Thanks!

PR #1533, opened 2023-03-31
format : update some codes style

sunhailin-Leo:
Update some codes style, maybe improve a little performance.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1617,8 +1617,7 @@ func (c *HostClient) dialConnFor(w *wantConn) {
 	}
 
 	cc := acquireClientConn(conn)
-	delivered := w.tryDeliver(cc, nil)
-	if !delivered {
+	if !w.tryDeliver(cc, nil) {
 		// not delivered, return idle connection
 		c.releaseConn(cc)
 	}
@@ -1637,8 +1636,8 @@ func (c *HostClient) CloseIdleConnections() {
 	c.conns = c.conns[:0]
 	c.connsLock.Unlock()
 
-	for _, cc := range scratch {
-		c.closeConn(cc)
+	for i := range scratch {
+		c.closeConn(scratch[i])
 	}
 }
 
--- a/client.go
+++ b/client.go
@@ -1636,8 +1636,8 @@ func (c *HostClient) CloseIdleConnections() {
 	c.conns = c.conns[:0]
 	c.connsLock.Unlock()
 
-	for i := range scratch {
-		c.closeConn(scratch[i])
+	for _, cc := range scratch {
+		c.closeConn(cc)
 	}
 }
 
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -1637,8 +1636,8 @@ func (c *HostClient) CloseIdleConnections() {
 	c.conns = c.conns[:0]
 	c.connsLock.Unlock()
 
-	for _, cc := range scratch {
-		c.closeConn(cc)
+	for i := range scratch {
+		c.closeConn(scratch[i])
```
erikdubbelboer, line 1640: I prefer the original here, which I also think is slight faster.

fasthttpadaptor/adaptor.go
```diff
@@ -64,8 +64,8 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 				haveContentType = true
 			}
 
-			for _, v := range vv {
-				ctx.Response.Header.Add(k, v)
+			for i := range vv {
+				ctx.Response.Header.Add(k, vv[i])
```
erikdubbelboer, line 68: Same here.

erikdubbelboer (changes requested):

fasthttpadaptor/adaptor.go
sunhailin-Leo: `[]string` is not the same as `[]*clientConn`, and I think the former produces a copy of the value, which may affect some performance in some extreme scenarios.

tylitianrui: can  you  give an example about  an extreme scenario？

sunhailin-Leo: It may lead to excessive copying of values in some scenarios where many request headers need to be returned. Therefore, fetching by index can reduce the cost of copying this part of the value.

erikdubbelboer: But making a copy of a string is cheap as it's just 16 bytes. (pointer to data, length). It doesn't actually copy the value of the string. So I don't see that being an issue here.

fasthttpproxy/http.go
```diff
@@ -51,7 +51,7 @@ func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.Dia
 			return nil, err
 		}
 
-		req := fmt.Sprintf("CONNECT %s HTTP/1.1\r\nHost: %s\r\n", addr, addr)
+		req := "CONNECT " + addr + "HTTP/1.1\nHost: " + addr + "\r\n"
```
erikdubbelboer, line 54:
```suggestion
		req := "CONNECT " + addr + " HTTP/1.1\r\nHost: " + addr + "\r\n"
```

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1534, opened 2023-04-04
chore:remove redundant code

tylitianrui:
the same effect can be achieved if  the  destination slice do not  be reset  in `func (a *Args) CopyTo(dst *Args)` .


erikdubbelboer:
Thanks!

Issue #1529, opened 2023-03-22
Confusion about timeout


kukayiyi:
I want to use fasthttp.client.DoTimeout() to implement timeout control. It should return immediately after the given timeout and proceed to the next step. After searching various sources, I only found that I can modify the client’s dial to achieve this, but this is not elegant and loses the meaning of using DoTimeout(). My requirement is to easily set different timeouts for each request. After checking the history versions, I found that this problem might have occurred after #1346 . As a last resort, I had to downgrade my fasthttp to version 1.38.0 in my project to meet my needs. Am I using it wrong?

kukayiyi:
I have checked the source code and tried to modify readTimeout, writeTimeout and other max* parameters, but none of them work. No matter how other parameters are set, DoTimeout() will follow the time of the client’s dial.

kukayiyi:
I used the official sample code https://github.com/valyala/fasthttp/blob/master/examples/client/client.go and still have the same problem. Also, GetTimeout has a similar problem. [When the timeout parameter set is less than 3 seconds, it can normally return a timeout, but when it is greater than 3 seconds, it still fails to work

erikdubbelboer:
I'm not sure I fully understand, can you post some sample code how you are using fasthttp?

kukayiyi:
> I'm not sure I fully understand, can you post some sample code how you are using fasthttp?

`
var client  = &fasthttp.Client{
	MaxConnsPerHost:           1024, // MaxConnsPerHost  default is 512, increase to 16384
	ReadTimeout:               15 * time.Second,
	WriteTimeout:              15 * time.Second,

	// I set dial as a workaround, but that changes all timeouts
	//Dial: func(addr string) (net.Conn, error) {
	//	return fasthttp.DialTimeout(addr, time.Duration(config.Conf.Server.RequestTimeout)*time.Millisecond)
	//},
	//
}

req := fasthttp.AcquireRequest()
// build req
resp := fasthttp.AcquireResponse()
err := client.DoTimeout(req, resp, 15*time.Second)
`
I use it like this, but when the access is slow, it will return after 3 seconds, not 15 seconds, unless I set the dial

PR #1535, opened 2023-04-11
Fix issue #1529 : DoTimeout timeout failure

kukayiyi:
As discussed in #1529 , the main problem with the timeout failure is that req.timeout is not passed to dial to pause the dialing in time. One solution is to create a temporary dial. In order not to affect the original dial of HostClient, I choose to add the dialTimeout parameter in the dialHostHard function to control the dialing time. It will fail if HostClient.dial has been set, but general users should not set dial. In fact, I think the most appropriate solution is to make HostClient have a Dialer with a timeout attribute so that timeout can be set separately without affecting other settings and use HostClient.Dialer.dial instead of HostClient.dial, but obviously this will change a lot of code.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1613,7 +1613,7 @@ func (c *HostClient) acquireConn(reqTimeout time.Duration, connectionClose bool)
 		go c.connsCleaner()
 	}
 
-	conn, err := c.dialHostHard()
+	conn, err := c.dialHostHard(reqTimeout)
 	if err != nil {
 		c.decConnsCount()
 		return nil, err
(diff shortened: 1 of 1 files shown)


kukayiyi:
I've looked at other code, and when the connection is successfully established, the timeout is affected by readTimeout and writeTimeout as expected, so this change doesn't change the timeout control normally, only when the connection fails. I think the connection establish time should be able to be set, just like python request

erikdubbelboer:
I agree changing the `Dail` on HostClient is probably not a good idea as it isn't backwards compatible. If users set their own `Dial` then they can handle the timeouts in there as well.

I like the change this way. Thanks!

PR #1536, opened 2023-04-11
Fixed transfer-encoding for empty chunked payload

asaio001:
In release [v1.32.0](https://github.com/valyala/fasthttp/releases/tag/v1.32.0), there was a backwards incompatible change introduced as a (seemingly unintended) side-effect of [adding trailer support](https://github.com/valyala/fasthttp/pull/1165/commits/b2a0c32465dba2c024d127552733c9d10effa2b6).

The relevant scenario is in the case of a chunk-encoded request containing only a terminating chunk as its body payload (`0/r/n/r/n`). Previously, the[ content length would be set to the length of the body](https://github.com/valyala/fasthttp/pull/1165/commits/b2a0c32465dba2c024d127552733c9d10effa2b6#diff-6971e2713fe98b79ea18e9748bf68928f6ebcd4d93e7a27f10e7472ce6d78a48L1154) (0, in this case) whereas afterwards it [stopped doing that for chunked transfers](https://github.com/valyala/fasthttp/pull/1165/commits/b2a0c32465dba2c024d127552733c9d10effa2b6#diff-6971e2713fe98b79ea18e9748bf68928f6ebcd4d93e7a27f10e7472ce6d78a48R1164). 

The consequence of this for the above scenario is that the `Transfer-Encoding: chunked` header that would previously have been removed ([when setting the CL](https://github.com/valyala/fasthttp/blob/master/header.go#L295)) persists. If this request is then (re)used, it will result in a desync, since the header will be sent with no body at all.

### DIFF
--- a/http.go
+++ b/http.go
@@ -1268,7 +1268,7 @@ func (req *Request) ContinueReadBody(r *bufio.Reader, maxBodySize int, preParseM
 		return err
 	}
 
-	if req.Header.ContentLength() == -1 {
+	if contentLength == -1 {
 		err = req.Header.ReadTrailer(r)
 		if err != nil && err != io.EOF {
 			return err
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #1539, opened 2023-04-17
Refactor client's example code

Imm0bilize:
- Use *errors.Is*  for to compare errors
- Reducing the level of nesting for easier code perception ( using switch and removing some else)

### DIFF
--- a/examples/client/client.go
+++ b/examples/client/client.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


examples/client/client.go
```diff
@@ -72,54 +73,60 @@ func sendPostRequest() {
...
-			if err == io.EOF || err == nil {
-				fmt.Printf("DEBUG Parsed Response: %v\n", respEntity)
-			} else {
-				fmt.Fprintf(os.Stderr, "ERR failed to parse response: %v\n", err)
-			}
-		} else {
-			fmt.Fprintf(os.Stderr, "ERR invalid HTTP response code: %d\n", statusCode)
-		}
-	} else {
+	defer func() { fasthttp.ReleaseResponse(resp) }()
```
erikdubbelboer, line 80:
```suggestion
	defer fasthttp.ReleaseResponse(resp)
```
You don't have to wrap this in a closure.

erikdubbelboer (changes requested):

examples/client/client.go
Imm0bilize: Really, my mistake
Thanks for the comment :)

erikdubbelboer:
Thanks!

PR #1550, opened 2023-05-04
add concurrency for client's HostClient map

xuxiao415:


erikdubbelboer:
What is the use case for this?

xuxiao415:
> What is the use case for this?

I have a api gateway app. I just want to add concurrency for client's HostClient map. what's wrong with the code？

client.go
```diff
@@ -303,13 +303,24 @@ type Client struct {
...
 
+// GetHostClients return the map of Http HostClient
+func (c *Client) GetHostClients() map[string]*HostClient {
+	return c.m
+}
+
+// GetHttpsHostClients return the map of Https HostClient
+func (c *Client) GetHttpsHostClients() map[string]*HostClient {
+	return c.ms
+}
```
erikdubbelboer, line 322: Why add these methods? Since `m` and `ms` are protected by `mLock` it's impossible to use the return value of these functions in a safe way?

erikdubbelboer (changes requested):

client.go
xuxiao415: This is test code. I will delete it

client.go
```diff
@@ -485,68 +486,69 @@ func (c *Client) Do(req *Request, resp *Response) error {
 		return fmt.Errorf("unsupported protocol %q. http and https are supported", uri.Scheme())
 	}
 
+	c.mOnce.Do(func() {
+		c.m = make(map[string]*HostClient)
+		c.ms = make(map[string]*HostClient)
```
erikdubbelboer, line 491:
```suggestion
		c.mLock.Lock()
		c.m = make(map[string]*HostClient)
		c.ms = make(map[string]*HostClient)
		c.mLock.Unlock()
```
I know it's very unlikely but lets add this just to be safe in case someone calls `.Do()` and `.CloseIdleConnections()` at exactly the same time.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1562, opened 2023-05-12
fix the problem: HostClient maybe delete when pendingClientRequests > 0

xuxiao415:
fix the problem: HostClient maybe delete when pendingClientRequests > 0


erikdubbelboer:
Nice find!

PR #1569, opened 2023-05-29
rename interface{} to any

yuyang-ok:
golang has define the any type , this pr improve the readability

### DIFF
--- a/args.go
+++ b/args.go
@@ -32,7 +32,7 @@ func ReleaseArgs(a *Args) {
 }
 
 var argsPool = &sync.Pool{
-	New: func() interface{} {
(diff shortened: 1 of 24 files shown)


erikdubbelboer:
As you can see in the test results we are still supporting Go 1.17 which doesn't have `any`. As there is no clear guarantee on which version fasthttp supports I'm ok with dropping Go 1.17 support in this pull request. If you remove it please also update the readme as it still says we support Go 1.15.x 🙈

yuyang-ok:
@erikdubbelboer Sorry, I didn't notice this.

I think I'am going to drop this.

Issue #1572, opened 2023-06-03
Ordering of HTTP headers


antter:
Hi, I have a bit of an odd complaint. I originally turned to this package because you can control the case and ordering of headers, which is not possible with `net/http`.

As I had come to find out, even when `SetCanonical` is used, the `User-Agent` header always comes before `Host`. Headers re-ordering is a problem for proxies avoiding bot detection. In my case I am working on a proxy for intercepting traffic for web debugging. I end up sending out requests that starts with
```
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:109.0) Gecko/20100101 Firefox/113.0
Host: google.com
```
Even though Firefox would never put the `User-Agent` before `Host`.

erikdubbelboer:
I'm afraid that this is not possible with fasthttp. fasthttp stores some headers into internal variables and will always write them in the same order as seen here: https://github.com/valyala/fasthttp/blob/f0865d4aabbbea51a81d56ab31a3de2dfc5a9b05/header.go#L2473-L2492
There is currently no way to change that. If you want to make a pull request to add support for this it would be welcome, but only if it doesn't affect performance.

PR #1573, opened 2023-06-04
add DisableSpecialHeaders option

antter:
This is a PR is response to #1572 

Thanks for your quick response, and for your work in maintaining this package overall. Here is a PR to entirely disable fasthttp setting special headers for you. This way the user has complete control over headers and the burden is on them to make sure things work correctly.

Feedback is welcome

### DIFF
--- a/header_test.go
+++ b/header_test.go
@@ -364,24 +364,28 @@ func TestRequestDisableSpecialHeaders(t *testing.T) {
 	if err := h.Read(br); err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	// Host and User-Agent are parsed
-	if string(h.Host()) != "foobar" {
-		t.Fatalf("unexpected: %q. Expecting %q", h.Host(), "foobar")
+	// Host and User-Agent are not parsed
+	if h.Host() != nil {
+		t.Fatalf("unexpected: %q. Expecting nil", h.Host())
 	}
-	if string(h.UserAgent()) != "ua" {
-		t.Fatalf("unexpected: %q. Expecting %q", h.UserAgent(), "ua")
+	if h.Host() != nil {
+		t.Fatalf("unexpected: %q. Expecting nil", h.UserAgent())
 	}
 	if strings.Contains(h.String(), "Host") {
 		t.Fatalf("special header Host in headers: %q", h.String())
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):

antter:
made your suggested changes, you can take another look

erikdubbelboer (changes requested):
Tests are failing can you have a look?

antter:
Ahhh made some silly errors, should be good now

header_test.go
```diff
@@ -348,6 +348,47 @@ func TestRequestRawHeaders(t *testing.T) {
...
+	var h RequestHeader
+	h.DisableSpecialHeader()
+
+	s := "GET / HTTP/1.0\r\n" + kvs
+	br := bufio.NewReader(bytes.NewBufferString(s))
+	if err := h.Read(br); err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	// Host and User-Agent are not parsed
+	if h.Host() != nil {
```
erikdubbelboer, line 368: I don't understand why `Host()` should return `nil` here? `Host:` is present in the headers?

header_test.go
```diff
@@ -348,6 +348,47 @@ func TestRequestRawHeaders(t *testing.T) {
...
+		t.Fatalf("unexpected: %q. Expecting nil", h.Host())
+	}
+	if h.Host() != nil {
```
erikdubbelboer, line 371: You're checking this twice.

erikdubbelboer (changes requested):
I'm wondering it `DisableSpecialHeader` should affect parsing? It makes things a lot more confusing. Wouldn't it be simpler if it only affects how headers are written? Since that's the main use case for this?

header_test.go
antter: ah this is because for Host and User-Agent they are saved as separate bytes, not present in the headers `h.h`. When we call `Host()` with special headers disabled it tries to read it from `h.h`. I guess technically parsing is currently unaffected, but I agree this behavior is confusing. I'll push another commit that makes sure to parse all special headers  as regular headers

erikdubbelboer:
Thanks!

PR #1575, opened 2023-06-05
initialize  rootFSHandler before request 

tylitianrui:
`sync.Once` is only really needed if what you are doing inside it is really required for the rest of the code to continue. if  `rootFSHandler` was  initialized  in function  `ServeFile`. IT might  cause  that  the  rest requests waiting for   `rootFSHandler` initialization  and then to  continue.

Putting the initialization process into the init function.   it can  avoid  that problem


erikdubbelboer:
The reason this is done inside `ServeFile` is because `NewRequestHandler` starts a goroutine that never terminates (for background cleanups). We don't want to start this goroutine for programs that never call `ServeFile`.

tylitianrui:
> The reason this is done inside `ServeFile` is because `NewRequestHandler` starts a goroutine that never terminates (for background cleanups). We don't want to start this goroutine for programs that never call `ServeFile`.

thanks  for your comment

PR #1585, opened 2023-06-25
Auto add 'Vary' header after compression

### LINKED ISSUE #1583: Should add 'Vary: Accept-Encoding' header when compress is used
When responses are compressed, and may be compressed by different methods, responses should has 'Vary: Accept-Encoding' header so clients can get decodable content.
I think it should be such a build-in feature in the function `CompressHandlerBrotliLevel`, maybe with a config to turn it off.

AutumnSun1996:
Add config `SetAddVaryHeaderForCompression` to enable 'Vary: Accept-Encoding' header when compression is used.

Added a config variable `addVaryHeaderForCompress` to control this feature, with setter `SetAddVaryHeaderForCompression`.
To remain backward compatible, it's false by default.
When enabled, 'Vary: Accept-Encoding' header will be added when compression is used.

fix #1583

### DIFF
--- a/header.go
+++ b/header.go
@@ -344,8 +344,8 @@ func (h *ResponseHeader) SetContentEncodingBytes(contentEncoding []byte) {
 	h.contentEncoding = append(h.contentEncoding[:0], contentEncoding...)
 }
 
-// AddVaryBytes add value to the 'Vary' header if it's not included
-func (h *ResponseHeader) AddVaryBytes(value []byte) {
+// addVaryBytes add value to the 'Vary' header if it's not included
+func (h *ResponseHeader) addVaryBytes(value []byte) {
 	v := h.peek(strVary)
 	if len(v) == 0 {
 		// 'Vary' is not set
--- a/header.go
+++ b/header.go
@@ -344,6 +344,18 @@ func (h *ResponseHeader) SetContentEncodingBytes(contentEncoding []byte) {
 	h.contentEncoding = append(h.contentEncoding[:0], contentEncoding...)
 }
 
+// AddVaryBytes add value to the 'Vary' header if it's not included
+func (h *ResponseHeader) AddVaryBytes(value []byte) {
+	v := h.peek(strVary)
+	if len(v) == 0 {
+		// 'Vary' is not set
+		h.SetBytesV(HeaderVary, value)
+	} else if !bytes.Contains(v, value) {
+		// 'Vary' is set and not contains target value
+		h.SetBytesV(HeaderVary, append(append(v, ','), value...))
+	} // else: 'Vary' is set and contains target value
+}
+
 // Server returns Server header value.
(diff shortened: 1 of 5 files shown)


erikdubbelboer:
I have thought about it and I think it's better to remove `SetAddVaryHeaderForCompression` and always set the Vary header. We also always set Content-Encoding. Can you change that, then I can merge it. Thanks.

AutumnSun1996:
Hi, the code is updated, please review again. Thanks.

http.go
```diff
@@ -1723,6 +1723,7 @@ func (resp *Response) brotliBody(level int) error {
 		resp.bodyRaw = nil
 	}
 	resp.Header.SetContentEncodingBytes(strBr)
+	resp.Header.Set("Vary", "Accept-Encoding")
```
erikdubbelboer, line 1726: Sorry, but I'm thinking now it's probably better to use this. In case the user already set the Vary header before themselves.
```go
	if v := resp.Header.Peek(HeaderVary); !bytes.Contains(v, strAcceptEncoding) {
		resp.Header.SetBytesV(HeaderVary, append(v, (","+HeaderAcceptEncoding)...))
	} else {
		resp.Header.Add(HeaderVary, HeaderAcceptEncoding)
	}
```

erikdubbelboer (changes requested):

AutumnSun1996:
See what you mean. But this leads to another question: should we also handle the case where the user has added 'Vary: Accept-Encoding'?

erikdubbelboer:
That's exactly what my example does.

AutumnSun1996:
oh sorry, I just missed the second check. going to update it now

AutumnSun1996:
I created such a function with the check and update the Header, instead of copy-paste. any more comments?
```go
// AddVaryBytes add value to the 'Vary' header if it's not included
func (h *ResponseHeader) AddVaryBytes(value []byte) 
```

header.go
```diff
@@ -344,6 +344,18 @@ func (h *ResponseHeader) SetContentEncodingBytes(contentEncoding []byte) {
 	h.contentEncoding = append(h.contentEncoding[:0], contentEncoding...)
 }
 
+// AddVaryBytes add value to the 'Vary' header if it's not included
+func (h *ResponseHeader) AddVaryBytes(value []byte) {
```
erikdubbelboer, line 348: This is such a specific function and our API surface is already huge. Lets not export this function.

erikdubbelboer (changes requested):

header.go
AutumnSun1996: My thoughts was, it can also be used to add other values into 'Vary' header in application code.
Any way, it's ok to me to change it to 'addVaryBytes'

erikdubbelboer:
Thanks!

Issue #1578, opened 2023-06-19
Using System IPv6


DenScre:
Hello. I tried to transfer the implementation from net/http, but there was a problem, can you tell me how to correctly use ipv6 (system) to create requests.

In this case, if you replace Dial with 'Dial: fasthttpproxy.FasthttpHTTPDialerTimeout("127.0.0.1:3128", time.Second*10),', everything will work. (proxy that leads to '2001:DB8::9') 

```
addr, err := net.ResolveTCPAddr("tcp", "[2001:DB8::9]:0")
if err!=nil{return}

client = &fasthttp.Client{
 Dial: (&fasthttp.TCPDialer{
  LocalAddr: addr,
 }).Dial,
//Dial: fasthttpproxy.FasthttpHTTPDialerTimeout("127.0.0.1:3128", time.Second*10),
DialDualStack: true,
}

req := fasthttp.AcquireRequest()
req.SetRequestURI("https://ifconfig.co/ip")
req.Header.SetMethod(fasthttp.MethodGet)
resp := fasthttp.AcquireResponse()
errs := client.Do(req, resp)

fmt.Printf("Error: %s\n", errs)

fasthttp.ReleaseRequest(req)
respBody := resp.Body()
fmt.Printf("Response: %s\n", respBody)
fasthttp.ReleaseResponse(resp)
```
Displays an error (current ipv6 for example)

```
Error: dial tcp4: address [2001:DB8::9]:0: no suitable address found
Response: 
```

erikdubbelboer:
I'm afraid that this is currently not possible. [`FasthttpHTTPDialerTimeout` uses `fasthttp.Dial`](https://github.com/valyala/fasthttp/blob/ffdf59d04ca2f02556308d60d37b8e77ea04484c/fasthttpproxy/http.go#LL46-L46) which always dials with DualStack disabled. Replacing it with `DialDualStack` should work. You could try copying `fasthttpproxy/http.go` into your own project and use this method instead. Let me know if that works and I'll change it in fasthttp as well.

Pluto-zZ-zZ:
I had the same problem，error : couldn't find DNS entries for the given domain. Try using DialDualStack

Pluto-zZ-zZ:
I had the same problem，error : couldn't find DNS entries for the given domain. Try using DialDualStack
[...]

PR #1597, opened 2023-07-19
fasthttpproxy support ipv6

Pluto-zZ-zZ:
It's ok for fasthttpproxy use ipv6


Pluto-zZ-zZ:
issues : https://github.com/valyala/fasthttp/issues/1578

Pluto-zZ-zZ:
For ipv6 judgments, I saw something similar in net/ipsock.go, looking at whether the first character starts with a "["
6eeef7ae-f0d6-48d6-ac03-3567a0870886

erikdubbelboer:
Thanks!

PR #1602, opened 2023-08-03
Abstracts the RoundTripper interface and provides a default implement

### LINKED ISSUE #1601: Support default Transport implemention for HostClient
Transport does not provide a default implementation, but after some processing when I implement it myself, I need to call the remaining logic, but the remaining logic contains many private methods.
I would like the transport logic to be wrapped into a `DefaultTransport` for extension.

```go

	if c.Transport != nil {
		err := c.Transport(req, resp)
		return err == nil, err
	}

       // remaining logic
	cc, err := c.acquireConn(req.timeout, req.ConnectionClose())
	if err != nil {
		return false, err
	}
	conn := cc.c

	resp.parseNetConn(conn)

	if c.WriteTimeout > 0 {
		// Set Deadline every time, since golang has fixed the performance issue
		// See https://github.com/golang/go/issues/15133#issuecomment-271571395 for details
		currentTime := time.Now()
		if err = conn.SetWriteDeadline(currentTime.Add(c.WriteTimeout)); err != nil {
			c.closeConn(cc)
			return true, err
		}
	}
```

My Demo
```go
import (
	"github.com/valyala/fasthttp"
)

func init() {
	client := &fasthttp.Client{
		ConfigureClient: configureClient,
	}
	client.DoTimeout()
}

func configureClient(hc *fasthttp.HostClient) error {
	hc.Transport = Transport{hc: hc}.roundTrip
	return nil
}

type Transport struct {
	hc *fasthttp.HostClient
}

func (t Transport) roundTrip(req *fasthttp.Request, resp *fasthttp.Response) error {
	// my logic add trace-id for all client request
	req.Header.Set("trace-id", xxxx)

	// remain logic
	c := t.hc
	cc, err := c.acquireConn(req.timeout, req.ConnectionClose())
	if err != nil {
		return false, err
	}
	conn := cc.c

	//....
	
	// print logs after resp
}

```

timandy:
close #1601

### DIFF
--- a/client.go
+++ b/client.go
@@ -628,8 +628,10 @@ type DialFunc func(addr string) (net.Conn, error)
 // Request argument passed to RetryIfFunc, if there are any request errors.
 type RetryIfFunc func(request *Request) bool
 
-// TransportFunc wraps every request/response.
-type TransportFunc func(*Request, *Response) error
+// RoundTripper wraps every request/response.
+type RoundTripper interface {
+	RoundTrip(hc *HostClient, req *Request, resp *Response) (retry bool, err error)
+}
 
 // ConnPoolStrategyType define strategy of connection pool enqueue/dequeue
 type ConnPoolStrategyType int
@@ -791,7 +793,7 @@ type HostClient struct {
 	RetryIf RetryIfFunc
 
 	// Transport defines a transport-like mechanism that wraps every request/response.
-	Transport TransportFunc
+	Transport RoundTripper
 
 	// Connection pool strategy. Can be either LIFO or FIFO (default).
 	ConnPoolStrategy ConnPoolStrategyType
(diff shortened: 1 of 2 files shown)


timandy:
@erikdubbelboer PTAL,  We want to use this feature, Thanks a lot!

timandy:
code style problem be resolved

client.go
```diff
@@ -1343,119 +1345,13 @@ func (c *HostClient) doNonNilReqResp(req *Request, resp *Response) (bool, error)
...
-		tmpWriteDeadline := time.Now().Add(c.WriteTimeout)
-		if writeDeadline.IsZero() || tmpWriteDeadline.Before(writeDeadline) {
-			writeDeadline = tmpWriteDeadline
-		}
-	}
 
-	if err = conn.SetWriteDeadline(writeDeadline); err != nil {
-		c.closeConn(cc)
-		return true, err
+	// exec
```
erikdubbelboer, line 1349:
```suggestion
```

client.go
```diff
@@ -628,8 +628,10 @@ type DialFunc func(addr string) (net.Conn, error)
 // Request argument passed to RetryIfFunc, if there are any request errors.
 type RetryIfFunc func(request *Request) bool
 
-// TransportFunc wraps every request/response.
-type TransportFunc func(*Request, *Response) error
+// RoundTripper wraps every request/response.
+type RoundTripper interface {
+	RoundTrip(hc *HostClient, req *Request, resp *Response) (retry bool, err error)
```
erikdubbelboer, line 633: It feels weird to have `*HostClient` in there, why is that needed? We're going to have people who try to implement the RoundTripper by modifying something and then trying to call `hc.Do` for example. Which will result in an infinite loop.

client_test.go
```diff
@@ -3182,3 +3192,61 @@ func Test_AddMissingPort(t *testing.T) {
...
+	req.Header.Set("trace-id", "123")
+	*t.reqLog = req.String()
+	retry, err := t.base.RoundTrip(hc, req, resp)
+	*t.respLog = resp.String()
+	return retry, err
+}
+
+func TestClientTransportEx(t *testing.T) {
+	var reqLog string
+	var respLog string
```
erikdubbelboer, line 3212: Why not create `TransportEx` here, change it to `func (t *TransportEx) RoundTrip` and set `hc.Transport = &tp` so you don't have to use string pointers?

erikdubbelboer (changes requested):

client.go
timandy: > It feels weird to have *HostClient in there, why is that needed?

`RoundTripe` relies on `hc`'s fields and methods to create and release conns. If move these settings into the `Transport` struct, ay cause forward compatibility issues.

> We're going to have people who try to implement the RoundTripper by modifying something and then trying to call hc.Do for example. Which will result in an infinite loop.

The `TransportWrapper` testcase shows how to implement a coustom `RoundTripper` like go `net` sdk.

client_test.go
timandy: I have rewrite the testacase. Please check again. Thanks!
Using interfaces is more extensible than using methods, and it is easier to add methods to RoundTripper in the future.

client.go
timandy: I add a transport() method for HostClient to simplify the code.

client.go
```diff
@@ -2909,3 +2807,121 @@ func (c *pipelineConnClient) PendingRequests() int {
 }
 
 var errPipelineConnStopped = errors.New("pipeline connection has been stopped")
+
+var DefaultTransport RoundTripper = &Transport{}
+
+type Transport struct{}
+
+func (t *Transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (retry bool, err error) {
```
erikdubbelboer, line 2815:
```suggestion
var DefaultTransport RoundTripper = &transport{}

type transport struct{}

func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (retry bool, err error) {
```
I don't think we need to expose this. We already have a huge API surface so I would like to keep this private.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1603, opened 2023-08-11
Allow connection close for custom streams

becheran:
We do use fasthttp for our project as a proxy between an uplink connection and a target. We discovered the following scenario where we run into an issue that we can't solve without a minor change to the fasthttp client.

The scenario that we have is the following:

From the upsteam connection we get a request which is replied with a steam of data from the target. All works fine using fasthttp and the customStream body. We simply copy from the target reader to the upstream writer. Now the upstream suddenly closes the TCP connection and we get an error while trying to copy our data from the target.

Now we definitelly want to close the still open connection to the target. The only thing we can do from within the handler is set `fastRequest.SetConnectionClose()` and `fastResponse.SetConnectionClose()`. Though, the TCP connection will not be closed since it is too late when we set those values. The TCP connection is released and eventually re-used by the next upstream connection which would then result into an unexpected error.

If the closeConn check is done again in the newCloseReader wrapper function we could set the ConnectionClose() property from the outside and the connection will be closed once we release our request handler.

### DIFF
--- a/client.go
+++ b/client.go
@@ -2893,8 +2893,8 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 
 	br := hc.acquireReader(conn)
 	err = resp.ReadLimitBody(br, hc.MaxResponseBodySize)
-	hc.releaseReader(br)
 	if err != nil {
+		hc.releaseReader(br)
 		hc.closeConn(cc)
 		// Don't retry in case of ErrBodyTooLarge since we will just get the same again.
 		needRetry := err != ErrBodyTooLarge
@@ -2905,6 +2905,7 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 	if customStreamBody && resp.bodyStream != nil {
 		rbs := resp.bodyStream
 		resp.bodyStream = newCloseReader(rbs, func() error {
+			hc.releaseReader(br)
 			if r, ok := rbs.(*requestStream); ok {
 				releaseRequestStream(r)
 			}
@@ -2916,6 +2917,8 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 			return nil
 		})
 		return false, nil
+	} else {
+		hc.releaseReader(br)
 	}
 
 	if closeConn {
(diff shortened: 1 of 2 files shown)


http.go
```diff
@@ -1106,8 +1106,8 @@ func (resp *Response) Reset() {
 	if responseBodyPoolSizeLimit >= 0 && resp.body != nil {
 		resp.ReleaseBody(responseBodyPoolSizeLimit)
 	}
-	resp.Header.Reset()
 	resp.resetSkipHeader()
+	resp.Header.Reset()
```
erikdubbelboer, line 1110: This change is not described in your pull request, why did you change this order?

becheran: This change is required because the headers need to be reset after the body stream is cleaned up. Otherwise the information whether the connection shall be closed would not be available

erikdubbelboer (changes requested):
Can you have a look why this makes `TestAllocationClient` fail? It seems like it causes an allocation that wasn't there before.

becheran:
Interesting. Don't have access to a computer right now. Will have a deeper look next week.

client.go
```diff
@@ -2893,8 +2893,8 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 
 	br := hc.acquireReader(conn)
 	err = resp.ReadLimitBody(br, hc.MaxResponseBodySize)
-	hc.releaseReader(br)
 	if err != nil {
+		hc.releaseReader(br)
```
becheran, line 2897: @erikdubbelboer this is another change that would be required to avoid data races if the connection can be closed from outside. The reader shall only be released directly if it is not a stream connection. Otherwise do it in the cleanup method called via the closer interface.

The reason is that if the reader is realeased earlier, it would then be available for the next request and might cause a data race.

becheran:
@erikdubbelboer I did fix the `TestAllocationClient` but for the next run the `TestStreamBodyRequestContentLength` test failed. Though, I can't reproduce this locally and am also not sure how I could have possibly broken this? It is in the server side which I did not touch with the PR. Is this some test that is known to be flaky?

erikdubbelboer:
Thanks! Some tests can be flaky, especially on windows. It all looks good now.

PR #1605, opened 2023-08-15
increase ReadBufferSize & WriteBufferSize

klikpaspor:
From this isssue [https://github.com/gofiber/fiber/issues/2577]([https://github.com/gofiber/fiber/issues/2577](https://github.com/klikpaspor/fiber/pull/url)) i'm create a change pull request

### DIFF
--- a/server.go
+++ b/server.go
@@ -2572,8 +2572,8 @@ func writeResponse(ctx *RequestCtx, w *bufio.Writer) error {
 }
 
 const (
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
I'm sorry but this doesn't make any sense just like changing the setting wouldn't cause `panic: runtime error: invalid memory address or nil pointer dereference`.

PR #1607, opened 2023-08-24
fix round2_32

Duncaen:
There was a missing import in `round2_32.go`.

I split the tests into per int size files too, since `unsafe.Sizeof(int(0)) == 4` runtime does not allow to pass 64 bit constants to 32 bit functions at compile time. Not sure if there is a better solution.

### DIFF
--- a/http_test.go
+++ b/http_test.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"math"
 	"mime/multipart"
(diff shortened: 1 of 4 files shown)


Duncaen:
`round2_64` is missing `x |= x >> 32` not sure if this is intentional because of the 100MB cut-off.

erikdubbelboer:
Thanks!

PR #1634, opened 2023-10-21
BUGFIX: HostClient.DialDualStack not work when using DoDeadline

cloudfly:
If `HostClient` has `DialDualStack=true` and there is no custom `Dial` method specified, the `DialDualStack` setting will not work when sending requests using `DoDeadline` or `DoTimeout`. 

This issue was introduced in [ this commit ](https://github.com/valyala/fasthttp/commit/87cb886157ae414e178fd5bc14a57d5e1c8fcadb)


erikdubbelboer:
Thanks! Good find!

PR #1638, opened 2023-10-26
Allow redirect URI path to not be normalized.

### LINKED ISSUE #1637: Bug: Redirect location is always normalized
The location specified in a redirect response is always normalized, regardless of how the `DisablePathNormalizing` field is configured on the original request URI.

This behavior can cause issues if the server expects the redirect path to not be escaped. This problem was found while working with Google Drive URIs, where the redirect location contained an asterisk (`*`) that was then escaped to `%2A`, leading to an HTTP 500 response.

skidder:
Fix #1637 

Propagates the `DisablePathNormalizing` setting on a request URI to the redirect URI. This fixes an issue where redirect URIs were being normalized regardless of the setting on the request URI. This could effectively break the redirect URI, depending on the expectations of the server.

Preserves the current default behavior, which is for `DisablePathNormalizing` to be `false`.

### DIFF
--- a/http.go
+++ b/http.go
@@ -71,6 +71,15 @@ type Request struct {
 
 	// Use Host header (request.Header.SetHost) instead of the host from SetRequestURI, SetHost, or URI().SetHost
 	UseHostHeader bool
+
+	// Redirect path values are sent as-is without normalization
+	//
+	// Disabled redirect path normalization may be useful for proxying incoming requests
+	// to servers that are expecting paths to be forwarded as-is.
+	//
+	// By default redirect path values are normalized, i.e.
+	// extra slashes are removed, special characters are encoded.
+	DisableRedirectPathNormalizing bool
 }
 
 // Response represents HTTP response.
@@ -1080,6 +1089,7 @@ func (req *Request) Reset() {
 	req.resetSkipHeader()
 	req.timeout = 0
 	req.UseHostHeader = false
+	req.DisableRedirectPathNormalizing = false
 }
 
 func (req *Request) resetSkipHeader() {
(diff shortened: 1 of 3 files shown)


kirillDanshin:
I would consider this approach a breaking change though.
The issue is that users already got used to the existing meaning of this parameter, so I would rather this be done via a new option.
What do you think?
/cc @erikdubbelboer

skidder:
> I would consider this approach a breaking change though. The issue is that users already got used to the existing meaning of this parameter, so I would rather this be done via a new option. What do you think? /cc @erikdubbelboer

@kirillDanshin  Ah, good point. It sounds like you're proposing we signal to not normalize redirect locations using a new field on the request, is that right?

http.go
```diff
@@ -71,6 +71,15 @@ type Request struct {
...
 	// Use Host header (request.Header.SetHost) instead of the host from SetRequestURI, SetHost, or URI().SetHost
 	UseHostHeader bool
+
+	// Redirect path values are sent as-is without normalization
+	//
+	// Disabled redirect path normalization may be useful for proxying incoming requests
+	// to servers that are expecting paths to be forwarded as-is.
+	//
+	// By default redirect path values are normalized, i.e.
+	// extra slashes are removed, special characters are encoded.
```
erikdubbelboer, line 81:
```suggestion
	// DisableRedirectPathNormalizing disables redirect path normalization when used with DoRedirects.
	//
	// By default redirect path values are normalized, i.e.
	// extra slashes are removed, special characters are encoded.
```
In Go, comments for fields or functions should start the the name of that field or function.

erikdubbelboer (changes requested):

PR #1640, opened 2023-11-02
Add support to fs.fs on serve static files

### LINKED ISSUE #974: Ability to serve embed.FS files
Will there be a native fasthttp way (like .ServeFile) to share filesystems of embedded files, support for which came in go 1.16?

https://golang.org/doc/go1.16#library-embed

peczenyj:
this is the following patch

https://patch-diff.githubusercontent.com/raw/valyala/fasthttp/pull/1374.patch

from pr #1374 

however with smaller differences, lets continue this development?

should fix #974

### DIFF
--- a/compress.go
+++ b/compress.go
@@ -4,7 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"os"
+	"io/fs"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Thanks! And also thanks to @Aoang!

PR #1655, opened 2023-11-10
Feature: support HostClient.MaxGreedyConns

bingoohuang:
support HostClient.MaxGreedyConns to create new connections in greedy instead reusing until the MaxGreedyConns value


erikdubbelboer:
Can you explain the use case for this? And can you add a test to make sure it works and will keep working correctly in the future.

bingoohuang:
We use fasthttp client to a target server, and we want to known to how many connections can be established in max to the remote https server, without considering the request-response processing performance. The current fasthttp HostClient's implementation will try to reuse the existing idle connections at first, which is not easy for us to achieve the above scenario.

I will try to add a test case to this MaxGreedyConns. Thanks.

erikdubbelboer:
If you want to know how many connections you can open why not just use `net` and open TCP sockets without sending anything? Why do you need fasthttp for this?

bingoohuang:
> Can you explain the use case for this? And can you add a test to make sure it works and will keep working correctly in the future.

a test case added.

bingoohuang:
> If you want to know how many connections you can open why not just use `net` and open TCP sockets without sending anything? Why do you need fasthttp for this?

In simple, it is as you said.

But in fact, I used fasthttp as the base http client library. In my program, I setup a customized http session to specified server, which means there is still some session setup post-processing after  http/https connection established. This works well for a long time until recent days, the customer want to know how many this specified sessions can be setup in max, and then this pull request came out to find the answer in a easy way.

erikdubbelboer:
Did you find the limit already? I'm wondering how useful it is to merge this into fasthttp (which already has a really big API service) when it seems like no-one else needed this feature so far. Maybe it's best to keep this pull request open for a while and see if other people would also be interested.

http.go
```diff
@@ -77,6 +77,12 @@ type Request struct {
 	// By default redirect path values are normalized, i.e.
 	// extra slashes are removed, special characters are encoded.
 	DisableRedirectPathNormalizing bool
+
+	// UsingProxy 是否使用代理
```
gaby, line 81: Why the non-english comments?

http.go
```diff
@@ -77,6 +77,12 @@ type Request struct {
...
+	UsingProxy bool
+
+	// ConnAcquiredCallback 连接获取后回调，用于获取连接信息（例如 *tls.Conn 中的证书信息等）
```
gaby, line 84: Same here

PR #1656, opened 2023-11-10
Lazy load stackless functions

Gusted:
- I noticed that fasthttp was taking up 1.8MB of heap memory, even though it wasn't being used. This turned out to be the stackless function: 1.80MB  github.com/valyala/fasthttp/stackless.NewFunc
- Lazy load the stackless functions with sync.Once, given this a simple atomic read, it shouldn't affect performance for the fast-path (I haven't seen benchmarks with compression enabled).


erikdubbelboer:
Thanks!

PR #1657, opened 2023-11-11
Add support for non buffered body server responses.

pablolagos:
Adds support for non-buffered responses.

Sometimes we want to respond by copying from a stream (files or remote network connections), but the current method RegisterStreamBody has 2 cons:

- Is executed after the handler function returns
- Create 2 go threads and 2 channels for each response

This feature allows us to stream directly from the handler context

**In server**

```
´func handle(ctx *fasthttp.RequestCtx) {
     // DisableBuffering modifies fasthttp to disable headers and body buffering for this request
     ctx.DisableBuffering()
     
     // Write headers
     ctx.Response.Header.Set(.....)
     ctx.Response.Header.Set(.....)
     ctx.Response.Header.Set(.....)
     ctx.SetStatusCode(StatusOK)
    
     // Write body. Headers will be sent and cannot be modified once the body begins to be dispatched.
     // src can be any io.Reader or io.ReaderFrom or any object implementing io.WriteTo
     // The process would be unbuffered, well, actually using at most a 32KB buffer 
     // according to the current io.Copy implementation.
     // Ideal for serving large files without storing them on RAM. 
     io.Copy(ctx, src)

     // Optionally we can call CLoseReponse() or it will be called automatically after handler returns
     ctx.CloseResponse()

     fmt.Printf("Total bytes sent including headers: %d",ctx.BytesSent())
}
```

It will automatically set the Transfer-Encoding header to "chunked" and chunk the content on the fly.

### DIFF
--- a/http.go
+++ b/http.go
@@ -120,6 +120,8 @@ type Response struct {
 	raddr net.Addr
 	// Local TCPAddr from concurrently net.Conn
 	laddr net.Addr
+
+	headersWritten bool
 }
 
 // SetHost sets host for the request.
@@ -1122,6 +1124,7 @@ func (resp *Response) Reset() {
 	resp.laddr = nil
 	resp.ImmediateHeaderFlush = false
 	resp.StreamBody = false
+	resp.headersWritten = false
 }
 
 func (resp *Response) resetSkipHeader() {
--- a/server.go
+++ b/server.go
@@ -608,6 +608,47 @@ type RequestCtx struct {
 	hijackHandler    HijackHandler
 	hijackNoResponse bool
 	formValueFunc    FormValueFunc
+
+	disableBuffering  bool          // disables buffered response body
+	writer            *bufio.Writer // used to send response in non-buffered mode
+	bytesSent         int           // number of bytes sent to client (in non-buffered mode)
+	bodyChunkStarted  bool          // true if the response body chunk has been started
+	bodyLastChunkSent bool          // true if the last chunk of the response body has been sent
+}
+
+// DisableBuffering modifies fasthttp to disable body buffering for this request.
+// This is useful for requests that return large data or stream data.
+//
+// When buffering is disabled you must:
+//  1. Set response status and header values before writing body
+//  2. Set ContentLength is optional. If not set, the server will use chunked encoding.
+//  3. Write body data using methods like ctx.Write or  io.Copy(ctx,src), etc.
+//  4. Optionally call CloseResponse to finalize the response.
+//
+// CLosing the response will finalize the response and send the last chunk.
+// If the handler does not finish the response, it will be called automatically after handler returns.
+// Closing the response will also set BytesSent with the correct number of total bytes sent.
+func (ctx *RequestCtx) DisableBuffering() {
+	ctx.disableBuffering = true
+}
+
+// CloseResponse finalizes non-buffered response dispatch.
+// This method must be called after performing non-buffered responses
+// If the handler does not finish the response, it will be called automatically
+// after the handler function returns.
+func (ctx *RequestCtx) CloseResponse() {
+	if !ctx.disableBuffering || !ctx.bodyChunkStarted || ctx.bodyLastChunkSent {
+		return
+	}
+	if ctx.writer != nil {
+		// finalize chunks
+		if ctx.bodyChunkStarted && ctx.Response.Header.IsHTTP11() && !ctx.bodyLastChunkSent {
+			_, _ = ctx.writer.Write([]byte("0\r\n\r\n"))
+			_ = ctx.writer.Flush()
+			ctx.bytesSent += 5
+		}
+		ctx.bodyLastChunkSent = true
+	}
 }
 
 // HijackHandler must process the hijacked connection c.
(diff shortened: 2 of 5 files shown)


server_test.go
```diff
@@ -4237,6 +4237,53 @@ func TestServerChunkedResponse(t *testing.T) {
...
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			ctx.DisableBuffering()
+			ctx.SetStatusCode(StatusOK)
+			ctx.SetContentType("text/html; charset=utf-8")
+			reader := bytes.NewReader(expectedBody)
+			_, err := io.Copy(ctx, reader)
+			if err != nil {
+				t.Fatalf("Unexpected error when copying body: %v", err)
+			}
```
erikdubbelboer, line 4254: This doesn't really test if it works. This test would still pass if you don't call `ctx.DisableBuffering()`. You might for example test the number of allocations, or you could pause while writing the body and check if the receiving end did receive it already.

erikdubbelboer (changes requested):

server_test.go
erikdubbelboer: Just that check is not enough. I would like to see it block before the handler returns, then confirm the response body, and only when that has been received by the client return from the handler. That way you really check if it works the way it's supposed to work.

erikdubbelboer (changes requested):

server.go
```diff
@@ -1443,10 +1490,28 @@ func (ctx *RequestCtx) NotFound() {
...
+		ctx.unbufferedWriter = NewUnbufferedWriter(ctx)
+	}
+	return ctx.unbufferedWriter.Write(p)
+}
+
+// BytesSent returns the number of bytes sent to the client after non buffered operation.
+// Includes headers and body length.
+func (ctx *RequestCtx) BytesSent() int {
+	return ctx.bytesSent
+}
```
erikdubbelboer, line 1513: Why add this function?

unbuffered.go
```diff
@@ -0,0 +1,116 @@
...
+	"fmt"
+)
+
+type UnbufferedWriter interface {
+	Write(p []byte) (int, error)
+	WriteHeaders() (int, error)
+	Close() error
+}
+
+type UnbufferedWriterHttp1 struct {
```
erikdubbelboer, line 15: I don't think this type should be exposed, it can be kept internally. I also wouldn't add `Http1` to the name as everything in fastthttp is http1.

unbuffered.go
```diff
@@ -0,0 +1,116 @@
...
+// Ensure UnbufferedWriterHttp1 implements UnbufferedWriter.
+var _ UnbufferedWriter = &UnbufferedWriterHttp1{}
+
+// NewUnbufferedWriter
+//
+// Object must be discarded when request is finished
+func NewUnbufferedWriter(ctx *RequestCtx) *UnbufferedWriterHttp1 {
+	writer := acquireWriter(ctx)
+	return &UnbufferedWriterHttp1{ctx: ctx, writer: writer}
+}
```
erikdubbelboer, line 34: This function can also be kept internal then.

unbuffered.go
```diff
@@ -0,0 +1,116 @@
...
+
+type UnbufferedWriterHttp1 struct {
+	writer            *bufio.Writer
+	ctx               *RequestCtx
+	bodyChunkStarted  bool
+	bodyLastChunkSent bool
+}
+
+var ErrNotUnbuffered = errors.New("not unbuffered")
+var ErrClosedUnbufferedWriter = errors.New("closed unbuffered writer")
```
erikdubbelboer, line 23: Make it something like `use of closed unbuffered writer`

http.go
```diff
@@ -120,6 +120,8 @@ type Response struct {
 	raddr net.Addr
 	// Local TCPAddr from concurrently net.Conn
 	laddr net.Addr
+
+	headersWritten bool
```
erikdubbelboer, line 124: I don't think this state needs to be kept here, you can just add a `headersWritten bool` to `UnbufferedWriterHttp1`.

unbuffered.go
```diff
@@ -0,0 +1,116 @@
...
+	if !uw.ctx.Response.headersWritten {
+		if uw.ctx.Response.Header.contentLength == 0 && uw.ctx.Response.Header.IsHTTP11() {
+			if uw.ctx.Response.SkipBody {
+				uw.ctx.Response.Header.SetContentLength(0)
+			} else {
+				uw.ctx.Response.Header.SetContentLength(-1) // means Transfer-Encoding = chunked
+			}
+		}
+		h := uw.ctx.Response.Header.Header()
+		n, err := uw.writer.Write(h)
```
erikdubbelboer, line 80: Probably faster and more efficient to use [`uw.ctx.Response.Header.WriteTo(uw.writer)`](https://pkg.go.dev/github.com/valyala/fasthttp?utm_source=godoc#ResponseHeader.WriteTo)

unbuffered.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// write headers if not already sent (e.g. if there is no body written)
+	if !uw.ctx.Response.headersWritten {
+		// skip body, as we are closing without writing body
+		uw.ctx.Response.SkipBody = true
+		_, err := uw.WriteHeaders()
+		if err != nil {
+			return fmt.Errorf("error writing headers: %w", err)
+		}
+	}
```
erikdubbelboer, line 103: Can you add a test to make sure this works, calling `CloseResponse()` without writing anything to the body.

erikdubbelboer (changes requested):

PR #1661, opened 2023-11-14
Handle request URI containing query parameters but no path.

### LINKED ISSUE #1659: Issue: Default Path Not Set in FastHTTP Client for URIs with Only Query Parameters
### Problem Description
The FastHTTP client fails to set a default path (`/`) for request URIs that include only query parameters but no path. This behavior diverges from common web clients (like cURL, Chrome, Firefox) and does not align with the HTTP 1.1 specification.

### Expected Behavior 
An HTTP GET request for a URI like `https://example.com?foo=bar` should generate a request header in the format:
```
GET /?foo=bar
```
This follows the standard outlined in the [HTTP 1.1 spec, section 5.3.1](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3.1), which states that clients must treat an empty path as `/`.

### Actual Behavior
Currently, FastHTTP generates a header like:
```
GET ?foo=bar
```
This can lead to failures with servers that do not default to a path of `/` when one is not specified.

### Suggested Fix
Modify FastHTTP's [`RequestURI` function in `http.go`](https://github.com/valyala/fasthttp/blob/8ecfc989d9c4c6c30232a6465ec17d5c6d85fc5f/header.go#L679-L686) to check for an empty path and set it to `/` if necessary. This ensures compatibility with a wider range of HTTP servers and adherence to the HTTP 1.1 specification.

skidder:
Fixes #1659 

When returning a request URI, check whether it begins with a query (i.e. `?`), in which case we should apply a default path of `/`). This is required to conform to the [HTTP 1.1 specification](https://datatracker.ietf.org/doc/html/rfc7230#section-5.3.1).

>"If the target URI's path component is empty, the client MUST send "/" as the path within the origin-form of request-target."

### DIFF
--- a/header.go
+++ b/header.go
@@ -681,6 +681,8 @@ func (h *RequestHeader) RequestURI() []byte {
 	requestURI := h.requestURI
 	if len(requestURI) == 0 {
 		requestURI = strSlash
+	} else if requestURI[0] == '?' {
+		requestURI = append(strSlash, requestURI...)
 	}
 	return requestURI
 }
--- a/header.go
+++ b/header.go
@@ -681,8 +681,6 @@ func (h *RequestHeader) RequestURI() []byte {
 	requestURI := h.requestURI
 	if len(requestURI) == 0 {
 		requestURI = strSlash
-	} else if requestURI[0] == '?' {
-		requestURI = append(strSlash, requestURI...)
 	}
 	return requestURI
 }
@@ -691,6 +689,11 @@ func (h *RequestHeader) RequestURI() []byte {
 // RequestURI must be properly encoded.
 // Use URI.RequestURI for constructing proper RequestURI if unsure.
 func (h *RequestHeader) SetRequestURI(requestURI string) {
+	if len(requestURI) > 0 && requestURI[0] == '?' {
+		h.requestURI = append(h.requestURI[:0], strSlash...)
+		h.requestURI = append(h.requestURI, requestURI...)
+		return
+	}
 	h.requestURI = append(h.requestURI[:0], requestURI...)
 }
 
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -681,6 +681,8 @@ func (h *RequestHeader) RequestURI() []byte {
 	requestURI := h.requestURI
 	if len(requestURI) == 0 {
 		requestURI = strSlash
+	} else if requestURI[0] == '?' {
+		requestURI = append(strSlash, requestURI...)
```
erikdubbelboer, line 685: This would cause an allocation, which we always try to avoid in fasthttp. I'm wondering if it isn't better to add the `/` here when needed: https://github.com/valyala/fasthttp/blob/699af40ee434bdefa692ff30da20d428d3b6fceb/header.go#L2815

erikdubbelboer (changes requested):

header.go
skidder: Ah, great point. I've updated the change to modify the request URI when set through either the `SetRequestURI` or `SetRequestURIBytes` functions.

erikdubbelboer (changes requested):
I'm sorry but I don't think `SetRequestURI` and `SetRequestURIBytes` need to be fixed. Users have control over these and can append the `/` themselves.

Your issue was with clients sending a `GET ?foo=bar` request. In that case this is not going to fix it. You need to fix it here in the request parsing code: https://github.com/valyala/fasthttp/blob/699af40ee434bdefa692ff30da20d428d3b6fceb/header.go#L2815

skidder:
> I'm sorry but I don't think `SetRequestURI` and `SetRequestURIBytes` need to be fixed. Users have control over these and can append the `/` themselves.

Ah, I see. I'm only using the FastHTTP client, not the FastHTTP server. My expectation was that the FastHTTP client would handle URLs using this format since that's the behavior of the built-in `net/http` client. I've been able to implement a work-around in my application, but thought this might be worth covering in FastHTTP.

erikdubbelboer:
I'm a little bit afraid to make backwards breaking changes like this. I'll leave this pull request open for now and see if anyone else runs into this.

PR #1669, opened 2023-11-25
add support for custom dial function with timeouts

avivcarmis:


client.go
```diff
@@ -185,7 +185,15 @@ type Client struct {
...
-	// Default Dial is used if not set.
+	// Default DialTimeout is used if not set.
+	DialTimeout DialFuncWithTimeout
+
+	// Callback for establishing new connections to hosts.
+	//
+	// Note that if Dial is set instead of DialTimeout, Dial will ignore Request timeout.
+	// If you want the tcp dial process to account for request timeouts, use DialTimeout instead.
+	//
+	// If not set, DialTimeout is used.
```
avivcarmis, line 196: we may consider deprecating this one if you feel like it.
I went with a less aggressive approach and just documented the difference between the two.
it may seem a bit confusing in terms of the difference between them though...

client.go
```diff
@@ -1839,16 +1872,6 @@ func (c *HostClient) dialHostHard(dialTimeout time.Duration) (conn net.Conn, err
 		n = 1
 	}
 
-	dial := c.Dial
```
avivcarmis, line 1842: this block is removed in favor of the new [callDialFunc](https://github.com/valyala/fasthttp/pull/1669/files#diff-4b667feae66c9d46b21b9ecc19e8958cf4472d162ce0a47ac3e8386af8bbd8cfR1965) function

client.go
```diff
@@ -2591,7 +2626,7 @@ func (c *pipelineConnClient) init() {
 
 func (c *pipelineConnClient) worker() error {
 	tlsConfig := c.cachedTLSConfig()
-	conn, err := dialAddr(c.Addr, c.Dial, c.DialDualStack, c.IsTLS, tlsConfig, c.WriteTimeout)
+	conn, err := dialAddr(c.Addr, c.Dial, nil, c.DialDualStack, c.IsTLS, tlsConfig, 0, c.WriteTimeout)
```
avivcarmis, line 2629: to not overcomplicate the commit, i left PipelineClient as is - meaning although it does implement a DoTimeout function, it does not support passing it to the dial process and thus does not expose a DialWithTimeout param. Adding it could be quite complex due to the async nature of PipelineClient, the dial process might serve several different requests so to support it we might need to further design how exactly we want to act when we have several different timeouts on a single dial call.

If we do want to support it in PipelineClient, we can either address it in this PR or postpone it to later. I will need to understand how to support it here though

erikdubbelboer (changes requested):
Please fix the linting issues.

avivcarmis:
ugh, my bad.
done.

erikdubbelboer:
Thanks!

PR #1672, opened 2023-12-01
bug: Flush the write buffer before putting it to the pool

Jille:
A few lines later we check if `s.ReduceMemoryUsage && hijackHandler == nil` and call releaseWriter. We need to flush the buffer before returning it to the pool to avoid the data getting lost.


Jille:
This is a theoretical bugfix. I've not observed any problems in practice, I was just reading through the code.
I'm not experienced with fasthttp at all, so there's a solid chance I'm missing something.

erikdubbelboer:
Thanks! I haven't observed it either but from the code it makes sense that it could happen.

PR #1673, opened 2023-12-01
copyZeroAlloc: Try WriteTo and ReadFrom before acquiring a buffer

Jille:
These are the same statements at the beginning of io.CopyBuffer, but by doing them ourselves first we trade off a little cpu for not holding the 4kb buffer during the write.


erikdubbelboer:
Thanks! Makes sense.

PR #1674, opened 2023-12-01
writeBodyFixedSize: Only do an early flush if the reader is an *os.File

Jille:
or an *io.LimitedReader of an *os.File (because that's also supported by https://cs.opensource.google/go/go/+/refs/tags/go1.21.4:src/bufio/bufio.go;l=784)

I think that having to flush less often outweighs the overhead of the extra check. The appended data is known to be large, but it might still save us a syscall by allowing it to buffer more.

### DIFF
--- a/http.go
+++ b/http.go
@@ -2104,10 +2104,19 @@ func limitedReaderSize(r io.Reader) int64 {
 
 func writeBodyFixedSize(w *bufio.Writer, r io.Reader, size int64) error {
 	if size > maxSmallFileSize {
-		// w buffer must be empty for triggering
-		// sendfile path in bufio.Writer.ReadFrom.
-		if err := w.Flush(); err != nil {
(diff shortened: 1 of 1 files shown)


Jille:
I only ran go test, no other testing nor benchmarks.

erikdubbelboer:
It should hit this so you're right: https://github.com/golang/go/blob/de5b418bea70aaf27de1f47e9b5813940d1e15a4/src/net/sendfile_linux.go#L23-L33

erikdubbelboer:
Thanks!

Jille:
> It should hit this so you're right: https://github.com/golang/go/blob/de5b418bea70aaf27de1f47e9b5813940d1e15a4/src/net/sendfile_linux.go#L23-L33

Exactly there indeed :)

PR #1680, opened 2023-12-25
Add IsError and IsSuccess methods for Response struct

MBHuman:
# Copy IsSuccess and IsError from Resty

Resty: https://github.com/go-resty/resty

Code with implementation:

https://github.com/go-resty/resty/blob/105f7185be505f070bfb19df4c3a4535058e0f84/response.go#L150

### DIFF
--- a/http.go
+++ b/http.go
@@ -178,6 +178,11 @@ func (resp *Response) SetConnectionClose() {
 	resp.Header.SetConnectionClose()
 }
 
+// IsSuccess method returns true if HTTP status `code >= 200 and <= 299` otherwise false.
+func (resp *Response) IsSuccess() bool {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
What exactly is the reason to add these? The user can easily do this themselves. We already have a really big API surface.

MBHuman:
> What exactly is the reason to add these? The user can easily do this themselves. We already have a really big API surface.

My company migrated from rest to fasthttp and in part of services using IsSuccess() . The reason is to reduce conditions in code and improve readability

erikdubbelboer:
Thanks for the contribution but I'm sorry I'm not going to merge it. The API surface of fasthttp is already much too big. These are just one-liners that users can write themselves.

PR #1685, opened 2023-12-31
feat: add function to parse HTTP header parameters

nickajacks1:
The implementation is based on RFC-9110 5.6.6.

https://www.rfc-editor.org/rfc/rfc9110#name-parameters

The primary use case for parsing header parameters is for content negotiation, including multipart type boundaries.

The functor passed to VisitHeaderParams has a bool return value to allow the user to stop processing parameters early.
Examples of this being useful are to find a specific parameter, stop parsing after the q value, or to put a hard limit on the number of parameters a client can include.

I had originally written this for a different project using fasthttp, but I figured it was generic enough that it may warrant being merged in for others to use as well.  If not, no feelings will be hurt.

The implementation of validHeaderFieldByte is a little silly looking, but it's as fast as I could get it.
The Go standard library has its own implementation here, which is a bit slower: https://github.com/golang/go/blob/master/src/net/textproto/reader.go#L663
If the table solution is too arcane, we can copy the standard library instead.

Side note: VisitHeaderParams could also be used in MultipartFormBoundary to make fewer assumptions about how the client formatted their Content-Type header, but it would slow the function down quite a bit.
On my machine, it's 20 ns/op today vs 40 ns/op using VisitHeaderParams.

### DIFF
--- a/header.go
+++ b/header.go
@@ -545,6 +545,104 @@ func (h *ResponseHeader) AddTrailerBytes(trailer []byte) error {
 	return err
 }
 
+// validHeaderFieldByte returns true if c is a valid tchar as defined
+// by section 5.6.2 of [RFC9110].
+func validHeaderFieldByte(c byte) bool {
+	return c < 128 && tcharTable[c]
+}
+
+//	 tchar  = "!" / "#" / "$" / "%" / "&" / "'" / "*"
+//			   / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~"
+//			   / DIGIT / ALPHA
+//
+// See: https://www.rfc-editor.org/rfc/rfc9110#tokens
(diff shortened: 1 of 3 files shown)


erikdubbelboer (changes requested):
It's very hard to see if `VisitHeaderParams` has any bugs that might cause panics for example. Can you add a `FuzzVisitHeaderParams` to fuzz it and see if it crashes?

nickajacks1:
I see that existing fuzz tests use fuzzit, but given that the CI only builds on 1.18 and above and that there are no docs mentioning fuzzit, I just created a Fuzz test in header_test.go instead.

Ran a short 14 min fuzz:
`fuzz: elapsed: 14m5s, execs: 27008124 (33905/sec), new interesting: 34 (total: 50)`

PR #1686, opened 2024-01-02
chore: move cookie fuzz test to go 1.18 fuzzing

nickajacks1:
While it's possible to check in the fuzz that only the expected `error` types are returned, this would probably be a nightmare to maintain.


cookie_test.go
```diff
@@ -15,6 +15,22 @@ func TestCookiePanic(t *testing.T) {
...
+		`xxx=yyy; expires=Tue, 10 Nov 2009 23:00:00 GMT; domain=foobar.com; path=/a/b`,
+		" \n\t\"",
+	}
+	for _, input := range inputs {
+		f.Add([]byte(input))
+	}
+	c := AcquireCookie()
+	defer ReleaseCookie(c)
+	f.Fuzz(func(t *testing.T, cookie []byte) {
+		_ = c.ParseBytes(cookie)
```
erikdubbelboer, line 30: Might as well test `c.WriteTo` as well just like the original test did.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1689, opened 2024-01-06
refactor: move manually created tchar table to bytesconv_table_gen

nickajacks1:
The toUpper and toLower tables were also updated by `go generate`, I can revert those lines if that's not desirable.
Used a table of size 128 instead of 256 to lower impact on cache (AKA futile premature optimization)


erikdubbelboer:
Thanks!

ReneWerner87 (approved):

PR #1702, opened 2024-02-04
add DisableDNSResolution for TCPDialer. 

xuxiao415:
Sometimes, users do not need to use DNS resolution because they have already determined that the requested address is a list of IP addresses.


erikdubbelboer:
Thanks!

PR #1704, opened 2024-02-06
Implement `GetRejectedConnectionsCount` function

### LINKED ISSUE #1568: How to monitor requests dropped by  fasthttp.Server{Concurrency}?
Hello! 

We want to monitor requests that are dropped by the http.Server{Concurrency} setting and to count the number of rejected requests and write it to prometheus.

Could you make a handler so that something can be done when the request is rejected by ConcurrentLimit here https://github.com/valyala/fasthttp/blob/master/server.go#L1834 ?

mopeneko:
Closes #1568 

## Overview

- Add `rejectedRequestsCount` field
- Increment `rejectedRequestsCount` when connection is rejected
- Implement `GetRejectedConnectionsCount` function that returns `rejectedRequestCount`
- Write test

### DIFF
--- a/server.go
+++ b/server.go
@@ -431,6 +431,8 @@ type Server struct {
 	open int32
 	stop int32
 	done chan struct{}
+
+	rejectedRequestsCount uint32
 }
 
 // TimeoutHandler creates RequestHandler, which returns StatusRequestTimeout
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #1719, opened 2024-02-10
Prevent request smuggling

erikdubbelboer:
Prevent request smuggling when fasthttp is behind a reverse proxy that might interprets headers differently by being stricter. Should also prevent request smuggling when fasthttp is used as the reverse proxy.

### DIFF
--- a/header.go
+++ b/header.go
@@ -3029,6 +3029,8 @@ func (h *ResponseHeader) parseHeaders(buf []byte) (int, error) {
 func (h *RequestHeader) parseHeaders(buf []byte) (int, error) {
 	h.contentLength = -2
 
+	contentLengthSeen := false
+
 	var s headerScanner
 	s.b = buf
 	s.disableNormalizing = h.disableNormalizing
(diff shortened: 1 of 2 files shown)


header.go
```diff
@@ -3072,7 +3079,17 @@ func (h *RequestHeader) parseHeaders(buf []byte) (int, error) {
 				}
 			case 't':
 				if caseInsensitiveCompare(s.key, strTransferEncoding) {
-					if !bytes.Equal(s.value, strIdentity) {
+					isIdentity := bytes.Equal(s.value, strIdentity)
+					isChunked := bytes.Equal(s.value, strChunked)
```
kenballus, line 3083: These comparisons should also be case-insensitive. From RFC 9112, setion 7:
> All transfer-coding names are case-insensitive and ought to be registered within the HTTP Transfer Coding registry, as defined in [Section 7.3](https://www.rfc-editor.org/rfc/rfc9112.html#transfer.coding.registry). They are used in the [Transfer-Encoding](https://www.rfc-editor.org/rfc/rfc9112.html#field.transfer-encoding) ([Section 6.1](https://www.rfc-editor.org/rfc/rfc9112.html#field.transfer-encoding)) and TE ([Section 10.1.4](https://www.rfc-editor.org/rfc/rfc9110#section-10.1.4) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9112.html#HTTP)]) header fields (the latter also defining the "transfer-coding" grammar).

PR #1746, opened 2024-04-02
Add macos-14 platform, enable shuffle of unit-tests

gaby:
- Add `macos-14` platform which uses Apple M1 cpu's (ARM64).
- Add `-shuffle=on` to enable shuffling the unit-tests order. Each run will run tests in different order to catch tests depending on others or using data from other tests.

We recently enabled both of these options in gofiber with great success. https://github.com/gofiber/fiber/blob/main/.github/workflows/test.yml#L37


gaby:
@erikdubbelboer The test that failed is known to fail?

erikdubbelboer:
Thanks! Indeed some of our tests are a bit flakey.

PR #1763, opened 2024-04-22
feat: add address to tryDial errors

mdenushev:
It's helpful to know upstream when receive dial timeout.

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -360,6 +360,14 @@ func (d *TCPDialer) tryDial(
 var ErrDialTimeout = errors.New("dialing to the given TCP address timed out")
 
 // ErrDialWithUpstream wraps dial error with upstream info.
+//
+// Should use errors.As to get upstream information from error:
+//
+// var dialErr *fasthttp.ErrDialWithUpstream
+//
+//	if errors.As(err, &dialErr) {
+//	  upstream = dialErr.Upstream
+//	}
 type ErrDialWithUpstream struct {
 	Upstream string
 	wrapErr  error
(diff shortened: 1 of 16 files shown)


tcpdialer.go
```diff
@@ -346,15 +347,39 @@ func (d *TCPDialer) tryDial(
...
+		return nil, wrapDialWithUpstream(err, addr)
 	}
-	return conn, err
+	return conn, nil
 }
 
 // ErrDialTimeout is returned when TCP dialing is timed out.
 var ErrDialTimeout = errors.New("dialing to the given TCP address timed out")
 
+// ErrDialWithUpstream wraps dial error with upstream info.
```
erikdubbelboer, line 362: Can you add some documentation on how this type can be used to check the errors from the dialer and see which upstream caused it?

tcpdialer.go
```diff
@@ -346,15 +347,47 @@ func (d *TCPDialer) tryDial(
...
+//
+// Should use errors.As to get upstream information from error:
+//
+// var dialErr *fasthttp.ErrDialWithUpstream
+//
+//	if errors.As(err, &dialErr) {
+//	  upstream = dialErr.Upstream
+//	}
```
erikdubbelboer, line 370:
```suggestion
//	hc := fasthttp.HostClient{Addr: "foo.com,bar.com"}
//	err := hc.Do(req, res)
//	
//	var dialErr *fasthttp.ErrDialWithUpstream
//	if errors.As(err, &dialErr) {
//		upstream = dialErr.Upstream // foo.com or bar.com
//	}
```
Isn't this a better example of the usage? Isn't this what you build it for?

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #1778, opened 2024-05-17
fs: fix GenerateIndexPages when DirFS or embed.FS is used

efectn:
In current implementation of Fasthttp FS, GenerateIndexPages doesn't work properly with DirFS or embed.FS because of `fs.ValidPath` function. 

When you open `""` path using with DirFS it returns `invalid path error`. When using embed.FS, it returns `file does not exist` error. 

So, if the path is `""` both cases need to covered properly. 

**Note:** I ain't also sure this is a good solution.  Perhaps need smarter solution to cover more edge cases.


erikdubbelboer:
Thanks!

PR #1779, opened 2024-05-18
fs: fix openIndexFile when dirPath is empty string

efectn:
openIndexFile overrides the index path to make it usable with subdirs. This causes wrong index path like `/index.html` when the dirPath is empty string.


erikdubbelboer:
Thanks!

efectn:
Hi @erikdubbelboer do you have any plan for new release to include these bugfixes?

erikdubbelboer:
I'll do a release on Monday.

Issue #1785, opened 2024-06-02
`\r` is improperly permitted in header names.


kenballus:
From RFC 9112:
> A sender MUST NOT generate a bare CR (a CR character not immediately followed by LF) within any protocol elements other than the content. A recipient of such a bare CR MUST consider that element to be invalid or replace each bare CR with SP before processing the element or forwarding the message.

FastHTTP does not enforce this rule in the context of header names.

To see this for yourself, run [this FastHTTP server](https://github.com/narfindustries/http-garden/blob/main/images/fasthttp/app.go), and send the following request:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
\rtest: test\r\n
\r\n
```

Its response, after base64-decoding the pieces, should look something like this:
```
HTTPRequest(
    method=b'GET', uri=b'/', version=b'1.1',
    headers=[
        (b'\rtest2', b'test2'),
        (b'host', b'a'),
    ],
    body=b'',
)
```

The `\r` at the beginning of the first header name is invalid, and should cause the message to be rejected.

erikdubbelboer:
@kenballus please check https://github.com/valyala/fasthttp/pull/1789

PR #1789, opened 2024-06-09
Don't allow \r in header names

erikdubbelboer:
From RFC 9112:
```
  A sender MUST NOT generate a bare CR (a CR character not immediately
  followed by LF) within any protocol elements other than the content.
  A recipient of such a bare CR MUST consider that element to be invalid
  or replace each bare CR with SP before processing the element or forwarding
  the message.
```
net/http seems to completely error on this, so let's do the same.

Fixes https://github.com/valyala/fasthttp/issues/1785

See: https://go.dev/play/p/EmMgzC_yLvN


kenballus:
This patch looks good, but it could be improved by rejecting more characters than just `\t`, `\r`, and ` `.

The RFCs define header names with the following grammar rules:
```
field-name = token
token = 1*tchar
tchar = "!" / "#" / "$" / "%" / "&" / "'" / "*" / "+" / "-" / "." / "^" / "_" / "`" / "|" / "~" / DIGIT / ALPHA
```

Thus, all of the following characters should also be prohibited in field-names, in addition to `\r`, ` `, and `\t`:
- `\x00`-`\x08` inclusive,
- `\x0b`,
- `\x0c`,
- `\x0e`-`\x1f` inclusive,
- `"`,
- `(`,
- `)`,
- `,`,
- `/`,
- `;`,
- `<`,
- `=`,
- `>`,
- `?`,
- `@`,
- `[`,
- `\`,
- `]`,
- `{`,
- `}`,
- `\x7f`-`\xff` inclusive

erikdubbelboer:
Good point, I have no changed it to check all characters.

PR #1791, opened 2024-06-12
fix:  add timeout to proxy connection reading and writing

Sniper91:
fast http client may hang forever when reading from proxy connection. 
code to reproduce the bug:

proxy server
```
package main

import (
	"log"
	"net/http"
	"time"
)

func main() {
	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		log.Printf("Received request %s %s %s\n", r.Method, r.Host, r.RemoteAddr)
		if r.Method == http.MethodConnect {
			time.Sleep(1 * time.Hour) // proxy server may hang a long time for some reason
		}
	})
	log.Fatal(http.ListenAndServe(":8181", handler))
}

```

fast http client  hangs and ignores the timeout because proxy server doesn't response
```
package main

import (
	"fmt"
	"os"
	"time"

	"github.com/valyala/fasthttp"
	"github.com/valyala/fasthttp/fasthttpproxy"
)

func main() {
	client := fasthttp.Client{
		ReadTimeout:  5 * time.Second,
		WriteTimeout: 5 * time.Second,
		Dial:         fasthttpproxy.FasthttpHTTPDialerTimeout("127.0.0.1:8181", 5*time.Second),
	}
	req.SetRequestURI("https://localhost:8080/")
	req.Header.SetMethod(fasthttp.MethodGet)
	err := client.Do(req, resp)
	fasthttp.ReleaseRequest(req)
	if err == nil {
		fmt.Printf("DEBUG Response: %s\n", resp.Body())
	} else {
		fmt.Fprintf(os.Stderr, "ERR Connection error: %v\n", err)
	}
	fasthttp.ReleaseResponse(resp)
}
```

http client from standard library behaves normally
```
package main

import (
	"fmt"
	"io"
	"net/http"
	"net/url"
	"os"
	"time"
)

func main() {
	uu, _ := url.Parse("http://localhost:8181")
	client := &http.Client{
		Timeout: 5 * time.Second,
		Transport: &http.Transport{
			Proxy: http.ProxyURL(uu),
		},
	}
	resp, err := client.Get("https://localhost:8080/")
	if err != nil {
		fmt.Println(err)
	}
	io.Copy(os.Stdout, resp.Body)
	resp.Body.Close()
}

```

The implementation refers code from  https://github.com/golang/go/blob/fe9b3c339978f37aad53875d9e6d2df35a1996ad/src/net/http/transport.go#L1725

### DIFF
--- a/fasthttpproxy/http.go
+++ b/fasthttpproxy/http.go
@@ -42,6 +42,7 @@ func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.Dia
 	return func(addr string) (net.Conn, error) {
 		var conn net.Conn
 		var err error
+		start := time.Now()
 
 		if strings.HasPrefix(proxy, "[") {
 			// ipv6
@@ -63,6 +64,13 @@ func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.Dia
 			return nil, err
 		}
 
+		if timeout > 0 {
+			if err = conn.SetDeadline(start.Add(timeout)); err != nil {
+				conn.Close()
+				return nil, err
+			}
+		}
+
 		req := "CONNECT " + addr + " HTTP/1.1\r\nHost: " + addr + "\r\n"
 		if auth != "" {
 			req += "Proxy-Authorization: Basic " + auth + "\r\n"
@@ -70,6 +78,7 @@ func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.Dia
 		req += "\r\n"
 
 		if _, err := conn.Write([]byte(req)); err != nil {
+			conn.Close()
 			return nil, err
 		}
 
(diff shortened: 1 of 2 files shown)


Sniper91:
> Why not just set a deadline on the connection? https://pkg.go.dev/net#Conn.SetDeadline 
The connection will be read or write after function return. Calling net#Conn.SetDeadline here may cause unexpected  os.ErrDeadlineExceeded error.

erikdubbelboer:
You can call `conn.SetDeadline(time.Time{})` before returning from the function to remove the deadline again.

Sniper91:
> You can call `conn.SetDeadline(time.Time{})` before returning from the function to remove the deadline again.

Oh, It's more concise. I've already reimplemented this PR.

fasthttpproxy/http.go
```diff
@@ -86,6 +99,10 @@ func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.Dia
 			conn.Close()
 			return nil, fmt.Errorf("could not connect to proxy: %s status code: %d", proxy, res.Header.StatusCode())
 		}
+		if err := unsetDeadline(conn); err != nil {
+			conn.Close()
+			return nil, err
+		}
```
erikdubbelboer, line 105: I think it's nice to also just do this in a `if timeout > 0 {` instead of the `unsetDeadline` construction you have now.

erikdubbelboer (changes requested):

fasthttpproxy/http.go
Sniper91: https://github.com/valyala/fasthttp/blob/926da8124a40a39d2a810d6502a48bfbf7866798/fasthttpproxy/http.go#L67

```unsetDeadline``` will do nothing if timeout is zero.

Sniper91: > Yeah but it's better to just add `if timeout > 0 {` around this instead of using that closure variable.

OK

erikdubbelboer:
Thanks!

Issue #1794, opened 2024-06-19
`\r` and `\x00` are improperly permitted in header values


kenballus:
From RFC 9110, section 5.5:
> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message.

FastHTTP does not enforce this rule for `\r` (CR) and `\x00` (NUL).

To see this for yourself, run [this FastHTTP server](https://github.com/narfindustries/http-garden/blob/main/images/fasthttp/app.go), and send the following request:
```
GET / HTTP/1.1\r\n
Host: whatever\r\n
test: te\r\x00st\r\n
\r\n
```

The headers in its response, after base64-decoding, should look something like this:
```
('content-length', '0')
('host', 'whatever')
('test', 'te\r\x00st')
```

In order to comply with the RFC, the `\r` and `\x00` within the last header value should have either
1. caused the message to be rejected, or
2. been translated to spaces.

erikdubbelboer:
@kenballus please check https://github.com/valyala/fasthttp/pull/1796

PR #1796, opened 2024-06-24
Validate header values

erikdubbelboer:
Fixes https://github.com/valyala/fasthttp/issues/1794


bytesconv_table_gen.go
```diff
@@ -100,23 +100,50 @@ func main() {
...
+		//
+		//	HTAB           =  %x09
+		//	SP             =  %x20
+		//	VCHAR          =  %x21-7E
+		var table [256]byte
+		for c := 0; c < 256; c++ {
+			if (c >= 0x21 && c <= 0x7E) || // VCHAR
+				c == 0x20 || // SP
+				c == 0x09 || // HTAB
+				c >= 0x80 { // obs-text
```
kenballus, line 142: RFC 9110 does allow for you to be a little less strict here, if you'd like:
> Field values containing CR, LF, or NUL characters are invalid and dangerous, due to the varying ways that implementations might parse and interpret those characters; a recipient of CR, LF, or NUL within a field value MUST either reject the message or replace each of those characters with SP before further processing or forwarding of that message. Field values containing other CTL characters are also invalid; however, recipients MAY retain such characters for the sake of robustness when they appear within a safe context (e.g., an application-specific quoted string that will not be processed by any downstream HTTP parser).

erikdubbelboer: I think it's probably better to just reject these values like `net/http` does. Parsing the value to see if it's in a safe context will be complicated.

PR #1798, opened 2024-06-30
handle root properly in virtual file systems

ehsannm:
ServeFS does not handle `root` properly, this is a fix. I did not understood why we have separate logic for `osFS` and other implementations anyway.


erikdubbelboer:
Fixed in #2145

PR #1806, opened 2024-07-16
Fix possible race condition on request ctx done #1662

byte0o:
Fix possible race condition on request ctx done #1662

### DIFF
--- a/server.go
+++ b/server.go
@@ -2755,7 +2755,7 @@ func (ctx *RequestCtx) Done() <-chan struct{} {
 // RequestCtx.s.done is only closed when the server is shutting down
 func (ctx *RequestCtx) Err() error {
 	select {
-	case <-ctx.Done(): //fix  Use unified functions instead of reference variables to converge fetching into one place
+	case <-ctx.Done(): // fix  Use unified functions instead of reference variables to converge fetching into one place
 		return context.Canceled
 	default:
 		return nil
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Almost all tests are failing.

byte0o:
> Almost all tests are failing.

@erikdubbelboer  
I tested it on this machine and the test failed due to timeout. The code I submitted will not have this problem because the previous pull request was also a test timeout.

byte0o:
> Almost all tests are failing.

@erikdubbelboer  This is caused by the test taking too long. I will check the reason.

byte0o:
> Almost all tests are failing.

@erikdubbelboer Handled, it is a deadlock caused by using lock

server.go
```diff
@@ -2745,7 +2753,7 @@ func (ctx *RequestCtx) Done() <-chan struct{} {
 // RequestCtx.s.done is only closed when the server is shutting down
 func (ctx *RequestCtx) Err() error {
 	select {
-	case <-ctx.s.done:
+	case <-ctx.Done(): // fix  Use unified functions instead of reference variables to converge fetching into one place
```
erikdubbelboer, line 2756:
```suggestion
	case <-ctx.Done():
```
No need to add this as a comment.

erikdubbelboer (changes requested):

server.go
byte0o: Already processed

erikdubbelboer:
Thanks!

Issue #1808, opened 2024-07-17
FastHTTP doesn't strip tabs from header values


kenballus:
When FastHTTP receives a header value suffixed or prefixed with tabs, they should be stripped according to the RFCs. This is what nearly all other HTTP implementations do, including AIOHTTP, Apache httpd, Cheroot, Go net/http, H2O, HAProxy, Hyper, Hypercorn, Jetty, Libsoup, Lighttpd, Mongoose, Netty, Node.js, LiteSpeed, Passenger, Tomcat, Tornado, Twisted, Unicorn, Uvicorn, Waitress, and WEBrick.

Instead, FastHTTP allows the tabs to persist into the header value.

This can be confirmed by
1. running a FastHTTP server that echos header values (like [this one](https://github.com/narfindustries/http-garden/blob/main/images/fasthttp/app.go)),
2. sending it a request with a header value prefixed and suffixed with tabs, and extracting the echoed header value:
```sh
printf 'GET / HTTP/1.1\r\nHost: whatever\r\nTest: \t abc \t\r\n\r\n' \
    | nc localhost 80 | grep '"headers"' \
    | jq '.["headers"][1][1]' \
    | xargs echo \
    | base64 -d  \
    | xxd
```
```
00000000: 0920 6162 6320 09                        . abc .
```

Notably, spaces *are* correctly stripped. Presumably, the patch should change [this function](https://github.com/valyala/fasthttp/blob/7760a5b5062313206a83d4dfa3b2ebc8fbc52fdc/header.go#L3385C1-L3393C2) to strip tabs in addition to spaces.

erikdubbelboer:
Fixed in https://github.com/valyala/fasthttp/issues/1808

PR #1810, opened 2024-07-22
Fix header parser (#1808)

ksw2000:
See issue #1808. When FastHTTP receives a header value suffixed or prefixed with tabs, they should be stripped.

### DIFF
--- a/header.go
+++ b/header.go
@@ -3358,10 +3358,6 @@ func (s *headerScanner) next() bool {
 		s.value, s.b, s.hLen = normalizeHeaderValue(s.value, oldB, s.hLen)
 	}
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
Can you also add a test that shows that gets the value and shows the tabs are stripped?

ksw2000:
I have added tests for the parser of `ResponseHeader` and `RequestHeader`. Before making code changes, `TestIssue1808` could not pass. Additionally, my ground truth is compared against the parsing results of Go's built-in `net/http` package, which means our parser handles tabs in the same way as Go's built-in parser.

When I made the pull request, there was one test that did not pass, but in my opinion, it is unrelated to the changes I made.

erikdubbelboer:
Thanks!

PR #1818, opened 2024-08-05
Make RetryIfFunc handle request errors

### LINKED ISSUE #1744: Propagate request error to RetryIfFunc
Hi may I check is it possible to change the interface for the `RetryIfFunc` to pass the error for better retry strategy?

ksw2000:
Fix #1744
To maintain compatibility with the previous `RetryIfFunc` as much as possible, I think it can be relaxed into an `interface{}` type. 
I believe that directly adding an extra field might confuse developers. 
If we were to directly use generics introduced in Go 1.18 for type constraints, it would involve too many changes. 
I think the most balanced approach for now is to handle it using interface{} and type assertion.


erikdubbelboer (changes requested):
I have been thinking about this, and I think it's too much of a backwards incompatible change also also makes it less clear on how to use it. I would suggest we add a `RetryIfError func(request *Request) bool` and mark `RetryIf` as deprecated.

erikdubbelboer:
Thanks!

ksw2000:
Oops, I found that I only added the new field for `HostClient` but not for `Client`.

Issue #1816, opened 2024-08-05
BUG: content-length smaller than prefetched body size will lead requestStream Read() PANIC


DrakeOu:
**Steps to Reproduce:**
1. Enable StreamRequestBody to use StreamBody() for reading the body stream data.
2. Set MaxRequestBodySize to the lowest possible value (1).
3. Send a PUT HTTP request with a Content-Length smaller than the actual body size being transported, but ensure the Content-Length is smaller than 1024.
4. Implement a reader to read out the stream data until io.EOF. This will cause a panic.

Demo Code:
```go

package main

import (
	"fmt"
	"github.com/valyala/fasthttp"
)

func main() {
	// Server request handler
	requestHandler := func(ctx *fasthttp.RequestCtx) {
		stream := ctx.RequestBodyStream()

		buf := make([]byte, 10)
		for {
			n, err := stream.Read(buf)
			if err != nil {
				break
			}
			fmt.Printf("Read %d bytes: %s\n", n, buf[:n])
		}
	}

	// Start the server
	server := fasthttp.Server{
		Handler:            requestHandler,
		MaxRequestBodySize: 1,
		StreamRequestBody:  true,
	}
	if err := server.ListenAndServe("localhost:20020"); err != nil {
		fmt.Printf("Error in ListenAndServe: %s", err)
	}
}
```

Related Code in requestStream Class:

```go
func (rs *requestStream) Read(p []byte) (int, error) {
	var (
		n   int
		err error
	)
	if rs.header.ContentLength() == -1 {
		// ... chunk logic
	}
	if rs.totalBytesRead == rs.header.ContentLength() {
		return 0, io.EOF
	}
	prefetchedSize := int(rs.prefetchedBytes.Size())
	if prefetchedSize > rs.totalBytesRead {
		left := prefetchedSize - rs.totalBytesRead
		if len(p) > left {
			p = p[:left]
		}
		n, err := rs.prefetchedBytes.Read(p)
		rs.totalBytesRead += n
		if n == rs.header.ContentLength() {
			return n, io.EOF
		}
		return n, err
	}
	left := rs.header.ContentLength() - rs.totalBytesRead
	if len(p) > left {
		p = p[:left]
	}
	n, err = rs.reader.Read(p)
	rs.totalBytesRead += n
	if err != nil {
		return n, err
	}

	if rs.totalBytesRead == rs.header.ContentLength() {
		err = io.EOF
	}
	return n, err
}
```

**Issue Description:**

[...]

PR #1819, opened 2024-08-05
StreamRequestBody shouldn't read more data than actual need.

newacorn:
The StreamRequestBody feature on the server side should not read content that does not belong to the current request body.This is more logical and consistent with the result of not using the StreamRequestBody feature.
Fixes: [#1816](https://github.com/valyala/fasthttp/issues/1816). 
The reason for addressing this issue is that the Read method of requestStream assumes it should only read up to the Content-Length number of bytes.

### DIFF
--- a/http.go
+++ b/http.go
@@ -2283,11 +2283,13 @@ func readBodyWithStreaming(r *bufio.Reader, contentLength, maxBodySize int, dst 
 		readN = 8 * 1024
 	}
 
-	if contentLength >= 0 && maxBodySize >= contentLength {
-		b, err = appendBodyFixedSize(r, dst, readN)
-	} else {
-		b, err = readBodyIdentity(r, readN, dst)
-	}
+	// A fixed-length pre-read function should be used here; otherwise,
+	// it may read content beyond the request body into areas outside
+	// the br buffer. This could affect the handling of the next request
+	// in the br buffer, if there is one. The original two branches can
+	// be handled with this single branch. by the way,
+	// fix issue: https://github.com/valyala/fasthttp/issues/1816
+	b, err = appendBodyFixedSize(r, dst, readN)
 
 	if err != nil {
 		return b, err
--- a/http.go
+++ b/http.go
@@ -2290,7 +2290,6 @@ func readBodyWithStreaming(r *bufio.Reader, contentLength, maxBodySize int, dst 
 	// be handled with this single branch. by the way,
 	// fix issue: https://github.com/valyala/fasthttp/issues/1816
 	b, err = appendBodyFixedSize(r, dst, readN)
-
 	if err != nil {
 		return b, err
 	}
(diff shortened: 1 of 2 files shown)


newacorn:
https://github.com/valyala/fasthttp/blob/1fb34531652380677219b7c8c9c3d02da007a80b/http.go#L2286
Here, contentLength is guaranteed to be greater than or equal to 0 because the ContinueReadBodyStream function has already filtered out cases where contentLength is less than 0. It is assumed that a negative Content-Length value other than -1 should not be processed：
https://github.com/valyala/fasthttp/blob/1fb34531652380677219b7c8c9c3d02da007a80b/http.go#L1372

Below code will also handle cases where maxBodySize is less than contentLength.
https://github.com/valyala/fasthttp/blob/1fb34531652380677219b7c8c9c3d02da007a80b/http.go#L2295

erikdubbelboer (changes requested):
Thanks, the fix looks good. But the new file needs to be formatted. And http.go needs some more newlines:
```
  http.go:2287: line is 147 characters (lll)
  		A fixed-length pre-read function should be used here; otherwise, it may read content beyond the request body into areas outside of the br buffer.
  http.go:2288: line is 160 characters (lll)
  	        This could affect the handling of the next request in the br buffer, if there is one. The original two branches can be handled with this single branch.
```

http.go
```diff
@@ -2283,11 +2283,13 @@ func readBodyWithStreaming(r *bufio.Reader, contentLength, maxBodySize int, dst
...
-	} else {
-		b, err = readBodyIdentity(r, readN, dst)
-	}
+	// A fixed-length pre-read function should be used here; otherwise,
+	// it may read content beyond the request body into areas outside
+	// the br buffer. This could affect the handling of the next request
+	// in the br buffer, if there is one. The original two branches can
+	// be handled with this single branch. by the way,
+	// fix issue: https://github.com/valyala/fasthttp/issues/1816
+	b, err = appendBodyFixedSize(r, dst, readN)
```
erikdubbelboer, line 2293:
```suggestion
```
Sorry, our strict linter doesn't like this empty line here.

server_test.go
```diff
@@ -4358,3 +4358,51 @@ func (cl *testLogger) Printf(format string, args ...any) {
...
+		err = resp.Read(bufio.NewReader(cliCon))
+		if err != nil {
+			t.Error(err)
+		}
+		err = cliCon.Close()
+		if err != nil {
+			t.Error(err)
+		}
+	}()
+	//
```
erikdubbelboer, line 4390:
```suggestion

```

erikdubbelboer (changes requested):

newacorn:
@erikdubbelboer  May I ask what tools you use to lint the fasthttp codebase?

erikdubbelboer:
@newacorn you can see that here: https://github.com/valyala/fasthttp/blob/38a91cda6fa7e351834c1a7d97db9b89feec78c2/.github/workflows/lint.yml#L26-L28

erikdubbelboer:
Thanks!

PR #1826, opened 2024-08-10
Reduce the size of ResponseHeader and RequestHeader

ksw2000:
Replace the field `bufKV` whose size is **56** bytes in ResponseHeader and RequestHeader with two `[]byte` fields only **48** bytes.

+ Reduce ResponseHeader from **320** to **312** bytes.
+ Reduce RequestHeader from **360** to **352** bytes.
+ In the benchmark tests, although there is no significant performance improvement, it theoretically reduces memory usage by **2.2%** to **2.5%**.

### DIFF
--- a/header.go
+++ b/header.go
@@ -39,7 +39,9 @@ type ResponseHeader struct {
 	trailer []argsKV
 
 	cookies []argsKV
-	bufKV   argsKV
+	bufK    []byte
+	bufV    []byte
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Thanks, I didn't even realize we were using an `argsKV` without using any of the methods on an `argsKV`.

Issue #1827, opened 2024-08-10
The fasthttp fs module does not provide a separate cache container for the Zstandard (zst) compression algorithm.


newacorn:
## Issue Test
```go
func TestAcceptEncodingIsOnlyZstd(t *testing.T) {
	const fileName = "test.txt"
	f, err := os.Create("/tmp/" + fileName)
	if err != nil {
		t.Fatal(err)
	}
	defer func() {
		os.Remove("/tmp/" + fileName)
	}()
	_, err = f.Write(bytes.Repeat([]byte("1"), 1000))
	if err != nil {
		t.Fatal(err)
	}
	f.Close()
	fs := FS{Root: "/tmp", Compress: true, CacheDuration: time.Hour * 20}
	h := fs.NewRequestHandler()
	server := Server{
		Handler: h,
	}
	pcs := fasthttputil.NewPipeConns()
	cliCon, serCon := pcs.Conn1(), pcs.Conn2()
	go func() {
		defer func() {
			_ = cliCon.Close()
		}()
		req := AcquireRequest()
		resp := AcquireResponse()
		defer func() {
			ReleaseRequest(req)
			ReleaseResponse(resp)
		}()
		req.SetRequestURI("http://127.0.0.1:7070/" + fileName)
		req.Header.Set(HeaderAcceptEncoding, "zstd")
		_, err = req.WriteTo(cliCon)
		if err != nil {
			t.Error(err)
			return
		}
		err = resp.Read(bufio.NewReader(cliCon))
		if err != nil {
			t.Error(err)
			return
		}
		if !bytes.Equal(resp.Header.ContentEncoding(), []byte("zstd")) {
			t.Error("not zstd encoding")
			return
		}
		resp.Reset()
		req.Header.Del(HeaderAcceptEncoding)
		_, err = req.WriteTo(cliCon)
		if err != nil {
			t.Error(err)
			return
		}
		err = resp.Read(bufio.NewReader(cliCon))
		if err != nil {
			t.Error(err)
			return
		}
		if len(resp.Header.ContentEncoding()) != 0 {
			t.Error("has content encoding header")
			return
		}
		if utf8.Valid(resp.Body()) != true {
			t.Error("invalid utf8 bytes sequence")
			return
		}
		err = cliCon.Close()
		if err != nil {
			t.Error(err)
		}
	}()
	err = server.ServeConn(serCon)
	if err != nil {
		t.Fatal(err)
	}
```
## Result
```
[...]

PR #1828, opened 2024-08-11
Add a dedicated cache container for the zstd compression algorithm

newacorn:
Add a dedicated cache container for the zstd compression algorithm to prevent discrepancies between the response content and the implied Content-Encoding in certain scenarios.Fix: https://github.com/valyala/fasthttp/issues/1827#issue-2459271452


erikdubbelboer:
Thanks!

Issue #1822, opened 2024-08-08
How to set proxy and tcpdialer at the same time


0xfocu5:
```
client = &fasthttp.Client{
		ReadTimeout:         time.Second * 5,
		WriteTimeout:        time.Second * 5,
		MaxIdleConnDuration: time.Second * 5,
		//Dial: (&fasthttp.TCPDialer{
		//	Concurrency:      0,
		//	DNSCacheDuration: time.Hour,
		//}).Dial,
		Dial: fasthttpproxy.FasthttpSocksDialer("socks5://127.0.0.1:7890")
	}
```

I want these work together
```
Dial: (&fasthttp.TCPDialer{
			Concurrency:      0,
			DNSCacheDuration: time.Hour,
		}).Dial,
```
```
Dial: fasthttpproxy.FasthttpSocksDialer("socks5://127.0.0.1:7890")
```

erikdubbelboer:
`fasthttpproxy.FasthttpSocksDialer` is very simple and passes `proxy.Direct` to `golang.org/x/net/proxy.FromURL` which is what you want to change. I would copy the implementation of `fasthttpproxy.FasthttpSocksDialer` and try to pass a `fasthttp.TCPDialer` as second argument so it uses that instead if `net.Dial`.

0xfocu5:
> [`fasthttpproxy.FasthttpSocksDialer`](https://github.com/valyala/fasthttp/blob/9fe0bc2e6fafad8a28d1cb64cc07c7953f32abd7/fasthttpproxy/socks5.go#L19) is very simple and passes `proxy.Direct` to `golang.org/x/net/proxy.FromURL` which is what you want to change. I would copy the implementation of `fasthttpproxy.FasthttpSocksDialer` and try to pass a `fasthttp.TCPDialer` as second argument so it uses that instead if `net.Dial`.

Looking forward to your good news

newacorn:
## Implement based on fasthttpproxy
support socks5 and http.
```go
package fasthttpproxy

import (
	"bufio"
	"encoding/base64"
	"errors"
	"fmt"
	"github.com/valyala/fasthttp"
	"golang.org/x/net/proxy"
	"net"
	"net/url"
	"strings"
	"time"
)

type DialerFunc func(network, addr string) (net.Conn, error)

func (d DialerFunc) Dial(network, addr string) (net.Conn, error) {
	return d(network, addr)
}

type HttpProxyDialer interface {
	ConnectTimeout() time.Duration
[...]

PR #1829, opened 2024-08-12
Merge the various proxy dialer implementations from the fasthttpproxy into a single struct.

newacorn:
https://github.com/valyala/fasthttp/issues/1822
In this issue, I noticed that the existing implementation doesn't leverage fasthttp's dialer mechanism, so I created this pull request.
At the same time, provide more comprehensive support for proxy environment variables and allow configuration of the fasthttp dialer.
No tests were committed, and the logic is built on top of the existing fasthttpproxy foundation. I noticed that the original fasthttpproxy also didn't include any test code.
The proxy address is not universal, but I did some local debugging and didn't find any issues.

### DIFF
--- a/fasthttpproxy/dialer.go
+++ b/fasthttpproxy/dialer.go
@@ -1,0 +1,263 @@
+package fasthttpproxy
+
+import (
+	"bufio"
+	"encoding/base64"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/valyala/fasthttp"
+	"golang.org/x/net/http/httpproxy"
+	"golang.org/x/net/proxy"
+)
+
+var (
+	// Used for caching authentication information when using an HTTP proxy,
+	// it helps avoid re-encoding the authentication details when the ProxyURL
+	// changes along with the request URL.
+	authCache    = sync.Map{}
+	colonTLSPort = ":443"
+	tmpURL       = &url.URL{Scheme: httpsScheme, Host: "example.com"}
+)
+
+// Dialer embeds both fasthttp.TCPDialer and httpproxy.Config, allowing it
+// to take advantage of the optimizations provided by fasthttp for dialing while also
+// utilizing the finer-grained configuration options offered by httpproxy.
+type Dialer struct {
+	fasthttp.TCPDialer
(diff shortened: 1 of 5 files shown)


erikdubbelboer (changes requested):
I like it. I think it makes the old method obsolete, but we can't remove them for backwards compatibility of course. So I think we should rewrite the old methods to use this new `Dialer` instead. What do you think?

newacorn:
> I like it. I think it makes the old method obsolete, but we can't remove them for backwards compatibility of course. So I think we should rewrite the old methods to use this new `Dialer` instead. What do you think?

You are correct;I agree with your point. I forgot about backward compatibility and deduplication. Next, I will try to rewrite the existing functions.

fasthttpproxy/example_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Timeout:        time.Second * 30,
+		ConnectTimeout: time.Second * 20,
+		Config: httpproxy.Config{
+			HTTPProxy:  "http://127.0.0.1:8005",
+			HTTPSProxy: "http://127.0.0.1:8005",
+		},
+		DialDualStack: true,
+	}
+	// proxy information comes from environment
+	dialer, err := p.GetDialFunc(false)
```
erikdubbelboer, line 25: Shouldn't this be `true` if you want the config to come from the env?

fasthttpproxy/example_test.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	if statusCode != 200 {
+		log.Fatal(err)
+	}
+	_ = resp
+}
+
+func ExampleDialer_GetDialFunc_three() {
```
erikdubbelboer, line 80: What is the difference between this and the second example? Besides `NoProxy: "example.com",`

erikdubbelboer (changes requested):

fasthttpproxy/example_test.go
newacorn: My thinking was somewhat flawed; I should have combined the three examples together. Thank you.

fasthttpproxy/example_test.go
newacorn: My thinking was somewhat flawed; I should have combined the three examples together. Thank you.

erikdubbelboer:
Thanks!

Issue #1815, opened 2024-07-31
Bug: When deadline from HeaderReceived，panic assert may occur in logic.  


newacorn:
https://github.com/valyala/fasthttp/blob/ed6a27a88e9f759283a1783be68d469dbb4078b4/server.go#L2260
If I understand correctly, the panic assertion here assumes that this situation should logically never occur. However, in reality, it can indeed happen.

root reason: set deadline on a closed net.Conn return a error.
## A condition for occurring
let readtimeout for example
Server.ReadTimeout is zero, but Server.HeaderReceived.ReadTimeout is not.
net.Conn state change window from idle to active. at the same time server shutdown called.
## Test
```go
func TestSeverSetReadDeadlineUseHeaderReceivedBug(t *testing.T) {
	const clientConnCount = 500
	var cliConnSecReqNotify = make(chan struct{})
	s := fasthttp.Server{HeaderReceived: func(header *fasthttp.RequestHeader) fasthttp.RequestConfig {
		return fasthttp.RequestConfig{ReadTimeout: time.Hour}
	}}
	go func() {
		s.Handler = func(ctx *fasthttp.RequestCtx) {
			defer func() {
				if recover() != nil {
					t.Error(recover())
				}
			}()
			_, _ = ctx.WriteString("hello world")
		}
		err := s.ListenAndServe(":7070")
		if err != nil {
			log.Println(err)
		}
		time.Sleep(time.Hour)
	}()
	time.Sleep(time.Millisecond * 100)
	wg := sync.WaitGroup{}
	fistRqSend := sync.WaitGroup{}
	for i := 0; i < clientConnCount; i++ {
		wg.Add(1)
		fistRqSend.Add(1)
		time.Sleep(time.Millisecond * 10)
		go func() {
			defer wg.Done()
			cliCon, err := net.Dial("tcp", ":7070")
			if err != nil {
				log.Println(err)
				fistRqSend.Done()
				return
			}
			httpReqStr := "GET / HTTP/1.1\r\n\r\n"
			_, err = cliCon.Write([]byte(httpReqStr))
			if err != nil {
				log.Println(err)
			}
			resp := AcquireResponse()
			defer func() { ReleaseResponse(resp) }()
			buf := bufio.NewReader(cliCon)
			err = resp.Read(buf)
			if err != nil {
				log.Println(err)
			}
			assert.Eq(t, resp.Body(), []byte("hello world"))
			fistRqSend.Done()
[...]

PR #1835, opened 2024-08-21
We cannot avoid continuing to call SetReadDeadline and similar methods after the connection is closed.

newacorn:
In the current implementation, there are several places where it's assumed that SetReadDeadline and similar functions logically won't affect a closed connection. However, these assumptions may not hold in certain specific situations. Therefore, instead of asserting that such errors will never occur, simply return the errors encountered during the execution of these methods.

The current implementation of ShutDown can, under certain conditions, invalidate these assumptions.

## Case 1
https://github.com/valyala/fasthttp/blob/d29a2b90a63494f2007471d0653d5d85ead61397/server.go#L1600-L1609

We cannot assume that when the program reaches this point, the connection state has not yet transitioned from StateNew to the actual StateIdle. Even if this transition takes 5 seconds, we cannot make assumptions about the scheduler without synchronization primitives.
## Case 2
https://github.com/valyala/fasthttp/blob/d29a2b90a63494f2007471d0653d5d85ead61397/server.go#L2136-L2146
What ensures that getNextProto doesn't take longer than 5 seconds to execute? 
## Case 3
https://github.com/valyala/fasthttp/blob/d29a2b90a63494f2007471d0653d5d85ead61397/server.go#L2262-L2267
If neither Server.ReadTimeout nor Server.IdleTimeout are set,  and the timeout is configured through Server.HeaderReceived, there is a possibility that the connection might be closed by ShutDown during the transition from idle to active state.
See Issue: https://github.com/valyala/fasthttp/issues/1815
## Case 4
https://github.com/valyala/fasthttp/blob/d29a2b90a63494f2007471d0653d5d85ead61397/server.go#L2403-L2414
If there's buffered data in the underlying connection or in br, and bw.Flush hasn't been executed before, it's not guaranteed that the underlying connection's closure will be observed by the code executed previously.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1597,14 +1597,14 @@ func (s *Server) NextProto(key string, nph ServeHandler) {
 func (s *Server) getNextProto(c net.Conn) (proto string, err error) {
 	if tlsConn, ok := c.(connTLSer); ok {
 		if s.ReadTimeout > 0 {
-			if err := c.SetReadDeadline(time.Now().Add(s.ReadTimeout)); err != nil {
-				panic(fmt.Sprintf("BUG: error in SetReadDeadline(%v): %v", s.ReadTimeout, err))
+			if err = c.SetReadDeadline(time.Now().Add(s.ReadTimeout)); err != nil {
+				return
 			}
 		}
 
 		if s.WriteTimeout > 0 {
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Makes sense, thanks for the change!

PR #1836, opened 2024-08-21
How to handle TCP keepalive-related configurations more effectively?

newacorn:
Currently, when Server.TCPKeepalive is set to false, if the net.Listener is created by the package itself, it does not override the default behavior of net.ListenConfig, which keeps TCP keepalive enabled.

I also added TCPKeepalive and TCPKeepalivePeriod fields to the client to make its behavior consistent with the server. 

**The goal of this pull request is to ensure that whether TCP keepalive is enabled is entirely controlled by the configuration fields provided by the package when net.TCPConn is created by the package itself.**
**If net.TCPConn is created by the user, it means that on the server side, the user provides their own net.Listener, and on the client side, the user provides their own Dial and DialTimeout functions. The TCP keepalive mechanism is managed by the user in this case, When TCPKeepalive field is false.**

If the user provides their own net.Listener, even after this pull request is accepted, the  implementation of fasthttp means that Server.TCPKeepalive is only meaningful when set to true. If set to false, it does not change the existing TCP keepalive behavior of the net.TCPConn created by the user-provided net.Listener.
https://github.com/valyala/fasthttp/blob/43c7b83ee54b9e78ecb79ae1a6346e9a82342f03/server.go#L1956-L1967

To avoid unnecessary and redundant system calls, I think this behavior is acceptable. At the same time add some relevant comments to make the meaning clearer.
```go
	// Whether the operating system should send tcp keep-alive messages on the tcp connection.
	//
	// When this field is set to false, it is only effective if the net.Listener
	// is created by this package itself. i.e. listened by ListenAndServe, ListenAndServeUNIX,
	// ListenAndServeTLS and  ListenAndServeTLSEmbed method.
	//
	// If you have created a net.Listener yourself and configured TCP keepalive,
	// this field should be set to false to avoid redundant additional system calls.
	TCPKeepalive bool
```
The root of the problem is that the net package defaults to having TCP keepalive enabled, while the fasthttp package defaults to having TCP keepalive disabled. Clearly, the behavior of the fasthttp package is more reasonable.

The changes brought by this pull request makes the semantics of the configuration fields related to TCP keepalive clearer and more precise.

This is a preliminary personal thought; discussions are welcome.

### DIFF
--- a/client.go
+++ b/client.go
@@ -228,6 +228,13 @@ type Client struct {
 	// By default connection duration is unlimited.
 	MaxConnDuration time.Duration
 
+	// TCPKeepalivePeriod is the duration the connection needs to
+	// remain idle before TCP starts sending keepalive probes.
+	//
+	// This field is only effective when TCPKeepalive is set to true.
+	// A zero value indicates that the operating system's default setting will be used.
+	TCPKeepalivePeriod time.Duration
+
 	// Maximum number of attempts for idempotent calls.
 	//
 	// DefaultMaxIdemponentCallAttempts is used if not set.
@@ -315,6 +322,14 @@ type Client struct {
 
 	// StreamResponseBody enables response body streaming.
 	StreamResponseBody bool
+
+	// Whether the operating system should send tcp keep-alive messages on the tcp connection.
+	//
+	// By default tcp keep-alive connections are disabled.
+	//
+	// This option is used only if default TCP dialer is used,
+	// i.e. if Dial and DialTimeout are blank.
+	TCPKeepalive bool
 }
 
 // Get returns the status code and body of url.
@@ -527,6 +542,8 @@ func (c *Client) Do(req *Request, resp *Response) error {
 				MaxConns:                      c.MaxConnsPerHost,
 				MaxIdleConnDuration:           c.MaxIdleConnDuration,
 				MaxConnDuration:               c.MaxConnDuration,
+				TCPKeepalivePeriod:            c.TCPKeepalivePeriod,
+				TCPKeepalive:                  c.TCPKeepalive,
 				MaxIdemponentCallAttempts:     c.MaxIdemponentCallAttempts,
 				ReadBufferSize:                c.ReadBufferSize,
 				WriteBufferSize:               c.WriteBufferSize,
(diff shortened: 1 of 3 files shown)


createTcpListener_new.go
```diff
@@ -0,0 +1,10 @@
+//go:build go1.23
```
erikdubbelboer, line 1: I can already see us forgetting to update this when a new version of Go is release (like we did here: https://github.com/search?q=repo%3Avalyala%2Ffasthttp+%22go%3Abuild+go1.20%22&type=code). I think it would be better to do this:
```
//go:build !go1.19 && !go1.20 && !go1.21 && !go1.22
```
That would automatically make it work for a future go1.24 etc.

erikdubbelboer (changes requested):

createTcpListener_new.go
newacorn: Sorry, I misunderstood. Go 1.23 did not change the default behavior of net.ListenConfig. If the KeepAlive field is 0, TCP keepalive will still be enabled. The new KeepAliveConfig just provides more granular control. To disable the keepalive mechanism, you still need to set the KeepAlive field to a negative value.

newacorn: The current build tags for Go versions seem to be correct(b2s and s2b functions). I found a related question on Stack Overflow.
[Golang build tags for a particular Go version possible?](https://stackoverflow.com/questions/38439275/golang-build-tags-for-a-particular-go-version-possible)

> Put the code that depends on contexts in a file with the following build tag. This code is compiled for Go 1.7 and onward.
// +build go1.7
Put the backwards compatibility code in a file with the following build tag. This code is used for all versions of Go before 1.7.
// +build !go1.7

erikdubbelboer: You're right, I learned something new there.

erikdubbelboer:
I'm thinking we should treat this as a backwards incompatible change as it changes the behavior of the library.

What do you think about deprecating `TCPKeepalive` and introducing a new `TCPKeepaliveDisabled`. Then the default will stay the same and upgrading won't affect any servers.

newacorn:
> I'm thinking we should treat this as a backwards incompatible change as it changes the behavior of the library.
> 
> What do you think about deprecating `TCPKeepalive` and introducing a new `TCPKeepaliveDisabled`. Then the default will stay the same and upgrading won't affect any servers.

I agree with your view; I will rework it.

newacorn:
> I'm thinking we should treat this as a backwards incompatible change as it changes the behavior of the library.
> 
> What do you think about deprecating `TCPKeepalive` and introducing a new `TCPKeepaliveDisabled`. Then the default will stay the same and upgrading won't affect any servers.

I think that even with the addition of a TCPKeepaliveDisabled field, and maintaining backward compatibility, it is difficult to accurately convey the semantics. I believe that the program's runtime behavior should strictly adhere to the configuration field API. But TCP keepalive seems unlikely to affect the correctness of the program and might only have a minor impact on performance, it may be acceptable to address it in the next major release. Let's leave this issue to be resolved in the next major release.

Let's keep this pull request open; we may need more input from others.

Issue #1840, opened 2024-08-25
The new workerChanPool implementation does not recycle workerChan when the workerPool stops.


newacorn:
The new implementation:
https://github.com/valyala/fasthttp/blob/19c50cdc44e4cc209362d70e47e82c34e54fdf99/workerpool.go#L140-L144
The old implementation:
https://github.com/valyala/fasthttp/blob/b06f4e21d918faa84ae0aa12c9e4dc7285b9767e/workerpool.go#L195-L198
From
```go
go wp.workerFunc(ch)
```
To
```go
go func() {
			wp.workerFunc(ch)
			wp.workerChanPool.Put(vch)
		}()
```
And remove the release of workerChan during cleanup.
Seem more consistent with the original implementation.
It seems more reasonable to release workerChan only here. The current implementation seems only release workerChan during cleanup, but does not release it when the workerPool stops, which is inconsistent with the original mutex implementation.

newacorn:
@erikdubbelboer What do you think?

newacorn:
@erikdubbelboer  
Do you have any plans to release fasthttp2? The current fasthttp seems to lack vitality, while the codebases built on top of it are quite active. With the continuous release of new versions of Go, does fasthttp need to be refactored to make its semantics clearer and the code more concise? Additionally, to make fasthttp more vibrant, should we consider creating an official router similar to net/http? Making fasthttp ready-to-use out of the box would be crucial for its future development. Otherwise, many people may only focus on the higher-level frameworks built on fasthttp, without addressing its existing shortcomings and potential issues. The more contributors, the better. What do you think?

erikdubbelboer:
@newacorn I currently don't have any plans, I also don't have the time for such a big project. I think with all the performance improvements in Go itself, fasthttp is becoming less and less relevant. The lack of http2/3 also contributes to this.

newacorn:
Your response seems to make some sense. In complex environments, performance depends on many factors. Overall, it isn't as robust as the `net/http` implementation.

PR #1841, opened 2024-08-25
Make new workerChanPool implementation consistent with the original mutex implementation

newacorn:
The current implementation seems only release workerChan during cleanup, but does not release it when the workerPool stops, which is inconsistent with the original mutex implementation.
See Issue: https://github.com/valyala/fasthttp/issues/1840


newacorn:
Close Because：https://github.com/valyala/fasthttp/pull/1846

Issue #1839, opened 2024-08-25
fasthttp currently exhibits some strange behavior in its tests.


newacorn:
```bash
~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 35s
❯ go test .
ok  	github.com/valyala/fasthttp	7.009s

~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 8s
❯ go test -count=2 .
ok  	github.com/valyala/fasthttp	18.702s

~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 19s
❯ go test -count=3 .
--- FAIL: TestAllocationClient (9.22s)
    allocation_test.go:67: expected 0 allocations, got 4.000000
FAIL
FAIL	github.com/valyala/fasthttp	34.606s
FAIL

~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 35s
❯ go test -count=4 .
--- FAIL: TestAllocationClient (9.19s)
    allocation_test.go:67: expected 0 allocations, got 4.000000
--- FAIL: TestAllocationClient (13.79s)
    allocation_test.go:67: expected 0 allocations, got 5.000000
FAIL
FAIL	github.com/valyala/fasthttp	55.522s
FAIL
```
We can see that as the count flag value increases, the time taken to complete the test does not grow linearly, and test failures have occurred.
```
~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 56s
❯ go test -race --count=2 .
--- FAIL: TestZstdBytesConcurrent (1.03s)
    zstd_test.go:22: timeout
--- FAIL: TestZstdCompressConcurrent (1.02s)
    zstd_test.go:68: timeout
--- FAIL: TestServeFileDirectoryRedirect (0.00s)
    fs_test.go:926: Unexpected status code 200 for directory '/fasthttputil' without trailing slash. Expecting 302.
--- FAIL: TestZstdBytesConcurrent (1.00s)
    zstd_test.go:22: timeout
FAIL
FAIL	github.com/valyala/fasthttp	16.511s
FAIL
```
When the -race flag is added, the completion time is incorrect (which is reasonable, as enabling the race detector slightly impacts performance), and an additional error also appears.

I believe that without using the -race parameter, the test completion time should increase linearly as the count value increments.
The occurrence of test failures should not be affected by the count flag.

erikdubbelboer:
[...]

PR #1842, opened 2024-08-25
Fix issues with tests interfering with each other in certain situations.

newacorn:
In some cases, the goroutines started by one test do not terminate smoothly before the next round of tests begins, causing interference between tests.

Performance Impact: This results in test completion times not increasing linearly with the count value.

Correctness Impact: It affects the accuracy of memory allocation test cases.

Fix partial issues of this issue.: https://github.com/valyala/fasthttp/issues/1839


erikdubbelboer:
Thanks!

PR #1843, opened 2024-08-25
Fix Different request paths share the same `fs` cache is some cases.

newacorn:
For requests with paths that do not have a trailing slash, we always redirect to the corresponding path with a trailing slash. This approach avoids assuming that users will definitely be redirected to the trailing-slash path and instead ensures that we return a proper redirection, checking for an index file or generating a directory listing. We do this because we should not make assumptions about user behavior. Similarly, the cache for `dir/` requests should not be used for `dir` requests.
This issue is also reflected in this issue: https://github.com/valyala/fasthttp/issues/1839
```bash
~/GolandProjects/cbrotli_bug/fasthttp fasthttp_test_issue2 56s
❯ go test -race --count=2 .
--- FAIL: TestZstdBytesConcurrent (1.03s)
    zstd_test.go:22: timeout
--- FAIL: TestZstdCompressConcurrent (1.02s)
    zstd_test.go:68: timeout
--- FAIL: TestServeFileDirectoryRedirect (0.00s)
    fs_test.go:926: Unexpected status code 200 for directory '/fasthttputil' without trailing slash. Expecting 302.
--- FAIL: TestZstdBytesConcurrent (1.00s)
    zstd_test.go:22: timeout
FAIL
FAIL	github.com/valyala/fasthttp	16.511s
FAIL
```

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1022,7 +1022,6 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 		path = ctx.Path()
 	}
 	hasTrailingSlash := len(path) > 0 && path[len(path)-1] == '/'
-	path = stripTrailingSlashes(path)
 
 	if n := bytes.IndexByte(path, 0); n >= 0 {
 		ctx.Logger().Printf("cannot serve path with nil byte at position %d: %q", n, path)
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Nice find, thanks!

PR #1844, opened 2024-08-25
When `StreamRequestBody` is set to `true`, we cannot safely release `br`

newacorn:
For example, when using chunked encoding, it's possible that `br` has only read the request headers.Even br.Buffered() == 0.


erikdubbelboer:
Thanks!

PR #1833, opened 2024-08-16
Using atomic instead of mutex and delete scratch slice

NikoMalik:

PR #1846, opened 2024-08-26
Revert "Using atomic instead of mutex and delete scratch slice"

newacorn:
Reverts valyala/fasthttp#1833
The new implementation of the workerChan queue has the following issues:

1. The original implementation used push operations for enqueuing and pop operations for dequeuing, while the new implementation uses unshift operations for enqueuing and shift operations for dequeuing. This difference doesn't align with the original design, although this issue might not be significant on its own.

2. When cleaning up the workerChan queue, the new implementation does not support pop operations. Instead, it uses unshift operations to inspect whether a workChan is active. As a result, old workChan instances cannot be removed unless the youngest one in the queue has also expired.

The solution for issue one related to enqueuing has led to the problem where the cleanup task in issue two cannot be correctly implemented.

A single-headed unidirectional queue cannot fully replicate the original mutex-based workChan queue. When addressing the workChan queue issue with a lock-free linked list, we may need to consider more aspects carefully.

I am sorry to submitted a pull request opposing the previous one because the potentially buggy code has already been merged into the master branch. We need a better solution to the existing issue before considering merging it back into the master branch.


erikdubbelboer:
@NikoMalik what do you think, it was your pull request. Do you think it would be possible to fix this problem with cleanup? Otherwise I think we should indeed revert, the performance difference doesn't have such a big impact as it's only for new connections and doesn't impact performance of keep-alive connections.

erikdubbelboer:
@newacorn can you please rebase it on master, then I can merge it.

newacorn:
> @newacorn can you please rebase it on master, then I can merge it.

Have done

PR #1847, opened 2024-08-26
Fix potential data race reports when the `-race` flag is enabled.

newacorn:
The `testClientDoTimeoutError` and `testClientGetTimeoutSuccess` test function does not immediately return or fatal when the test fails. Instead, it continues to stringify the byte slice that was passed as the `dst` parameter to the `Client.GetTimeout` method. If the test fails due to a timeout, the request may still be ongoing, and there might be a data race between writing to `dst` and the stringification operation.

Other test cases also continue to access the body after a request failure, but they are not related to timeouts. I believe this should not be an issue.


erikdubbelboer:
Thanks!

PR #1850, opened 2024-08-27
RetryIf might need more information and greater control over the execution flow.

newacorn:
Another implementation of RetryIfErr

    * Support for delayed retry logic
    * Support for resetting the request timeout before retrying
    * Users can decide whether to retry based on the request error and the number of retries.
    Estimating the time required for a request to complete is relatively straightforward, but when retries are factored in, it becomes challenging to gauge this time accurately.

At the same time, comment out the logic for retrying non-idempotent requests：
```go
	if !retry {
			break
			// Retry non-idempotent requests if the server closes
			// the connection before sending the response.
			//
			// This case is possible if the server closes the idle
			// keep-alive connection on timeout.
			//
			// Apache and nginx usually do this.
			//
			// This assertion cannot ensure that the server has not read our request.
			// Even if part of the request has been read, retrying for a non-idempotent
			// request is still unsafe. It is better to leave this decision to the user.
			//
			// if err != io.EOF {
			//	break
			// }
		}
```
This logic only covers one case where the error is not io.EOF; in fact, there are many other situations and edge cases. For non-idempotent requests, we should not make assumptions for the user.

### DIFF
--- a/client.go
+++ b/client.go
@@ -199,7 +199,16 @@ type Client struct {
 	// RetryIf controls whether a retry should be attempted after an error.
 	//
 	// By default will use isIdempotent function.
+	//
+	// Deprecated: Use RetryIfErr instead.
+	// This field is only effective when the `RetryIfErr` field is not set.
 	RetryIf RetryIfFunc
+
+	// When the client encounters an error during a request, the behavior—whether to retry,
+	// whether to retry later, or whether to reset the request timeout—should be determined
+	// based on the return value of this field.
+	// This field is only effective within the range of MaxIdemponentCallAttempts.
+	RetryIfErr RetryIfErrFunc
 
 	// ConfigureClient configures the fasthttp.HostClient.
 	ConfigureClient func(hc *HostClient) error
@@ -537,6 +546,7 @@ func (c *Client) Do(req *Request, resp *Response) error {
 				DisablePathNormalizing:        c.DisablePathNormalizing,
 				MaxConnWaitTimeout:            c.MaxConnWaitTimeout,
 				RetryIf:                       c.RetryIf,
+				RetryIfErr:                    c.RetryIfErr,
 				ConnPoolStrategy:              c.ConnPoolStrategy,
 				StreamResponseBody:            c.StreamResponseBody,
 				clientReaderPool:              &c.readerPool,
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -659,9 +669,18 @@ type DialFuncWithTimeout func(addr string, timeout time.Duration) (net.Conn, err
...
+// whether to retry later, or whether to reset the request timeout—should be determined
+// based on the return value of this interface.
+//
+// retryCount indicates which retry attempt this is, with retry counting starting from 1.
+// The first request execution is not considered a retry.
+// err represents the error encountered during the previous request.
+// sleepDuration if is greater than 0, the retry will be executed after this interval.
+// resetReqTime indicates whether to reset the request timeout to its initial value before executing this retry.
+// retry indicates whether to perform this retry. If false, the request function will return immediately.
+type RetryIfErrFunc func(request *Request, retryCount int, err error) (sleepDuration time.Duration, resetReqTime bool, retry bool)
```
erikdubbelboer, line 683: What is the reason for returning the `sleepDuration`? The user of this function can more easily just call `time.Sleep()` in the `RetryIfErrFunc` itself.

client.go
```diff
@@ -1329,31 +1357,40 @@ func (c *HostClient) Do(req *Request, resp *Response) error {
...
-				break
-			}
+			//
+			// This assertion cannot ensure that the server has not read our request.
+			// Even if part of the request has been read, retrying for a non-idempotent
+			// request is still unsafe. It is better to leave this decision to the user.
+			//
+			// if err != io.EOF {
+			//	break
+			// }
```
erikdubbelboer, line 1387: I agree, lets just remove the comment to make the code cleaner.

client.go
```diff
@@ -659,9 +669,18 @@ type DialFuncWithTimeout func(addr string, timeout time.Duration) (net.Conn, err
...
-// RetryIfErrFunc defines the signature of the retry if error function.
-// request and error passed to RetryIfErrFunc, if there are any request errors.
-type RetryIfErrFunc func(request *Request, err error) bool
+// RetryIfErrFunc defines an interface used for implementing the following functionality:
+// When the client encounters an error during a request, the behavior—whether to retry,
+// whether to retry later, or whether to reset the request timeout—should be determined
+// based on the return value of this interface.
+//
+// retryCount indicates which retry attempt this is, with retry counting starting from 1.
+// The first request execution is not considered a retry.
```
erikdubbelboer, line 678: Maybe it's easier to call this `attempts`?

erikdubbelboer (changes requested):

client.go
newacorn: ok

client.go
newacorn: Makes sense.

erikdubbelboer:
Thanks!

Issue #1859, opened 2024-09-08
NewFastHTTPHandler may cause improper memory reuse


sigmundxia:
Hi developers,

I encountered a strange problem when using the middleware of [fiber](https://github.com/gofiber/fiber), a downstream software that depends on fasthttp. I think this problem may be related to fasthttp.

To better illustrate this problem, I added a test for fasthttp to simulate the process of fiber using middleware. As a result, this test failed and reported an error in line 177:

https://github.com/sigmundxia/fasthttp/blob/a165213f6370f59e5d6414f4f573eb6519c472ca/fasthttpadaptor/adaptor_test.go#L142-L187

As I mentioned in the code comments, this test can pass as long as line 172 is commented out, but line 172 is not related to cookies, which indicates that fasthttp may have experienced improper memory reuse in it.

I think there are two possible directions of the solution:

1. Fix the improper memory reuse problem in fasthttp. I have basically located the code that causes this problem. If you are sure that this is indeed a bug, I am happy to submit a PR to fix this problem.

2. Confirm that the way to use middleware in fiber is inappropriate. Since the fiber implementation ignores `ResponseWriter` and `r` in the middleware and restarts from `ctx`, this may be inappropriate. If you are sure that this is the reason for the error, I will try to communicate with the fiber developers.

Thank you for your attention! If you have any questions or would like to discuss further with me, please feel free to contact me.

Best wishes.

newacorn:
The issue is related to fasthttp, where shared memory is used when copying request cookies from fasthttp to net/http. Your pull request is welcome.

sigmundxia:
> The issue is related to fasthttp, where shared memory is used when copying request cookies from fasthttp to net/http. Your pull request is welcome.

Thanks! I've submitted a [pull request](https://github.com/valyala/fasthttp/pull/1860) that will hopefully fix this.

sigmundxia:
Since the pull request has been merged, I think this issue can be closed :)

PR #1860, opened 2024-09-08
Fix improper memory reuse in NewFastHTTPHandler

sigmundxia:
Hi Developers,

This pull request fixes the improper memory reuse in NewFastHTTPHandler reported in this [issue](https://github.com/valyala/fasthttp/issues/1859).

If there are any further adjustments or improvements you'd like me to make, please don't hesitate to reach out :)

Best regards!

### DIFF
--- a/header.go
+++ b/header.go
@@ -1286,8 +1286,9 @@ func (h *RequestHeader) VisitAll(f func(key, value []byte)) {
 
 	h.collectCookies()
 	if len(h.cookies) > 0 {
-		h.bufV = appendRequestCookieBytes(h.bufV[:0], h.cookies)
-		f(strCookie, h.bufV)
+		buf := make([]byte, 0)
+		buf = appendRequestCookieBytes(buf[:0], h.cookies)
+		f(strCookie, buf)
 	}
 	visitArgs(h.h, f)
 	if h.ConnectionClose() {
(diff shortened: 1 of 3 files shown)


header.go
newacorn, line 1: https://github.com/valyala/fasthttp/blob/c3050516d957b5bf75b353683c50302e4f955c6b/fasthttpadaptor/request.go#L58-L62
I think performing deep copying through a case branch would be more appropriate here, as the comment in Request.VisitAll mentions that key and value cannot be referenced after the function f finishes executing. Directly modifying VisitAll may affect its efficiency.

sigmundxia: Agree :) I have revised and made a new commit

newacorn: Following your initial approach, it would be better to perform a deep copy only for the cookie header. The other references likely don't have any issues.If a deep copy of the value is needed here, the key should also be copied. It seems that either solution could work; it's just a trade-off between performance and memory safety.From a personal perspective. what do you think? @erikdubbelboer

sigmundxia: That makes sense, I added a new commit to limit deep copy to the cookie case only. If there are further results on the tradeoff between performance and memory safety, I'll be happy to continue to make changes

gaby (commented):
Looks good to me

erikdubbelboer:
Thanks!

PR #1863, opened 2024-09-11
fix: type assertion to perIPTLSConn instead of erroneous perIPConn

KSpaceer:
If using MaxConnsPerIP setting along with HTTPS (TLS) server, the connection accepting leads to panic because of wrong type assertion.

This pull request fixes that problem


erikdubbelboer:
Thanks!

Issue #1862, opened 2024-09-11
header.Set does not reset but adds cookies


sigmundxia:
Hello developers,

I noticed that the header.Set function seems to have a problem when handling cookies. According to the definition of Set, it seems that when operating cookies, all current cookies should be cleared and then new values ​​should be added. However, when the following code actually performs the operation, the original cookies are still retained, making the actual effect become adding cookies:

https://github.com/valyala/fasthttp/blob/65e989e8b8bcf2541d840fcf27dcf8a11931b9db/header.go#L1474-L1477

https://github.com/valyala/fasthttp/blob/65e989e8b8bcf2541d840fcf27dcf8a11931b9db/cookie.go#L505-L516

This also caused [a downstream issue](https://github.com/gofiber/fiber/issues/3089).

May I ask if I have misunderstood the meaning of the operation, and this effect is actually expected by fasthttp? Or is there indeed some problem? If there is indeed a problem, I am happy to submit a pull request to fix it.

Best wishes!

sigmundxia:
Thanks for your detailed reply! I'll try to make a pull request.

sigmundxia:
After the discussion in pull request, a better solution would be to add a note about the behavior of cookies in `header.Set`. Since the pull request was merged, I think this issue can be closed :)

PR #1864, opened 2024-09-11
Add a note to clarify header.Set behavior on cookies

sigmundxia:
Hi Developers,

As described in [this issue](https://github.com/valyala/fasthttp/issues/1862), currently `header.Set` does not reset but adds cookies. This pull request attempts to fix this so that `header.Set resets` cookies instead of adding them.

Feel free to let me know if there are any further changes you would like me to make.

Best regards!

### DIFF
--- a/header.go
+++ b/header.go
@@ -1509,9 +1509,9 @@ func (h *RequestHeader) setNonSpecial(key, value []byte) {
 // Multiple headers with the same key may be added with this function.
 // Use Set for setting a single header for the given key.
 //
-// the Content-Type, Content-Length, Connection, Server, Set-Cookie,
-// Transfer-Encoding and Date headers can only be set once and will
-// overwrite the previous value.
+// the Content-Type, Content-Length, Connection, Server, Transfer-Encoding
+// and Date headers can only be set once and will overwrite the previous value,
+// while Set-Cookie will not clear previous cookies.
 //
 // If the header is set as a Trailer (forbidden trailers will not be set, see AddTrailer for more details),
 // it will be sent after the chunked response body.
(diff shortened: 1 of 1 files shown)


header.go
```diff
@@ -1473,7 +1473,7 @@ func (h *RequestHeader) setSpecialHeader(key, value []byte) bool {
 			return true
 		case caseInsensitiveCompare(strCookie, key):
 			h.collectCookies()
-			h.cookies = parseRequestCookies(h.cookies, value)
+			h.cookies = parseRequestCookies(make([]argsKV, 0), value)
```
newacorn, line 1476: The setSpecialHeader method is used in both the Set and Add methods of Header, so modifying it will affect its usage in Add. To reduce code duplication, an additional parameter can be added to setSpecialHeader, or h.Cookies can be reset before calling setSpecialHeader based on the parent method. Resetting h.Cookies can be done using RequestHeader.DelAllCookies's zero slice strategy, as using make may result in unnecessary memory allocation.

Such a pull request may not necessarily be accepted by @erikdubbelboer, but personally, I would accept it. Why not give it a try? 😊

newacorn (commented):
The setSpecialHeader method is used in both the Set and Add methods of Header, so modifying it will affect its usage in Add. To reduce code duplication, an additional parameter can be added to setSpecialHeader, or h.Cookies can be reset before calling setSpecialHeader based on the parent method. Resetting h.Cookies can be done using RequestHeader.DelAllCookies's zero slice strategy, as using make may result in unnecessary memory allocation.

Such a pull request may not necessarily be accepted by @erikdubbelboer, but personally, I would accept it. Why not give it a try? 😊

header.go
newacorn, line 1: The setSpecialHeader method is used in both the Set and Add methods of Header, so modifying it will affect its usage in Add. To reduce code duplication, an additional parameter can be added to setSpecialHeader, or h.Cookies can be reset before calling setSpecialHeader based on the parent method. Resetting h.Cookies can be done using RequestHeader.DelAllCookies's zero slice strategy, as using make may result in unnecessary memory allocation.

Such a pull request may not necessarily be accepted by @erikdubbelboer, but personally, I would accept it. Why not give it a try? 😊

sigmundxia: Thank you very much! I didn't notice that `Add` also uses the `setSpecialHeader` method, sorry for that oversight. I reimplemented the process of resetting cookies based on your suggestion.

I also noticed that ResponseHeader should also have corresponding changes, so I changed it accordingly. In addition, I added tests for each case.

erikdubbelboer:
This would be a backwards incompatible change for anyone using `header.Set` with `Cookie`. I don't think we should just change it because of that. I think a better solution would be to really make this clear in the documentation that you need to delete the 'Cookie' header before setting it if you don't want to add.

sigmundxia:
Indeed, changing the behavior of this method may cause compatibility issues for downstream software. I will try to make direct improvements to gofiber downstream. So, should this pull request be closed? Or should I undo all changes here and add a note about cookies in the code comments?

erikdubbelboer:
Thanks, it's good!

Issue #1745, opened 2024-04-01
Expose `timeout` field in Request


TAYTS:
I want to retrieve the timeout with custom `RoundTripper` without creating another wrapper for the `Request` to tweak how the timeout is being handled or is it possible we could support user custom values similar to `RequestCtx.userValues` to allow the flexibility to inject value to the request?
The use case for this is that the current HostClient timeout is set for the whole request and for stream response I would like to fail the request if the response header is slow but wait for the response body for slightly longer time or a constant deadline extend if there are incoming data from stream

erikdubbelboer:
I would be open to a pull request to add a method to expose this. But it needs some good documentation to make clear what it's for.

gaby:
@erikdubbelboer @TAYTS this can be closed as fixed.

PR #1878, opened 2024-10-01
feat (request): expose req timeout

aarcex3:
# Summary

Implemented `GetRequestTimeout` func to get the timeout attribute of the request.

# Related Issue
#1745

pd: is this what you guys had in mind?. I still need to document this btw.

### DIFF
--- a/http.go
+++ b/http.go
@@ -192,9 +192,9 @@ func (req *Request) SetConnectionClose() {
 	req.Header.SetConnectionClose()
 }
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Why as a string?

erikdubbelboer:
I think just returning a `time.Duration` is good. The most important part is the description of the function, it should describe exactly what this is for.

erikdubbelboer:
Thanks!

Issue #1879, opened 2024-10-04
Context is canceled on a newly initialized fasthttp.RequestCtx in 1.56.0


oschwald:
Given the following code, a "context canceled" error is printed out in 1.56.0 but not 1.55.0.

```go
package main

import (
	"fmt"

	"github.com/valyala/fasthttp"
)

func main() {
	var r fasthttp.Request

	r.Header.SetMethod("GET")
	r.SetRequestURI("http://example/")

	var requestCtx fasthttp.RequestCtx
	requestCtx.Init(&r, nil, nil)
	fmt.Println(requestCtx.Err())
}
```

I was able to bisect this to a7d488a9. 

We use code similar to this in our tests to create a `fasthttp.RequestCtx` to pass to the handler. This code has been working unchanged for many years.

dza89:
I am seeing the same issue. Reverting to 1.55 solves the issue.

```golang
done <- struct{}{}
```
This line signals the task is done, resulting in a cancelled context

Fix:
```golang
done := ctx.s.done

	if done == nil {
		done = make(chan struct{})
	}
	return done
```

unfamiliar with the original issue though.

@erikdubbelboer 
@byte0o

byte0o:
> 我遇到了同样的问题。恢复到1.55即可解决问题。
> 
> ```go
> done <- struct{}{}
> ```
> 
> 此行表示任务已完成，导致上下文被取消
> 
> 使固定：
> 
> ```go
> done := ctx.s.done
> 
> 	if done == nil {
> 		done = make(chan struct{})
> 	}
> 	return done
> ```
> 
> 尽管不熟悉原始问题。
> 
> @erikdubbelboer @byte0o



> I am seeing the same issue. Reverting to 1.55 solves the issue.
> 
> ```go
> done <- struct{}{}
> ```
> 
> This line signals the task is done, resulting in a cancelled context
> 
> Fix:
> 
> ```go
> done := ctx.s.done
> 
> 	if done == nil {
> 		done = make(chan struct{})
> 	}
> 	return done
> ```
> 
> unfamiliar with the original issue though.
> 
> @erikdubbelboer @byte0o

@dza89  This change is to address the https://github.com/valyala/fasthttp/pull/1662 issue,Now the semantics of the Err function need to be reconsidered.

erikdubbelboer:
Should be fixed in https://github.com/valyala/fasthttp/pull/1890, I'll publish a release tomorrow.

oschwald:
[...]

PR #1890, opened 2024-10-26
fix RequestCtx is canceled (#1879)

ksw2000:
In my opinion, the root cause of issue #1879 is that the `done` channel in `ctx.s` is nil. We can simply create `done` channel in `fakeServe`r during the initialization of RequestCtx.

Original
```
requestCtx.Init()
→ ctx.Init2
→ ctx.s = fakeServer        (the ctx.s.done is nil)

requestCtx.Err()
→ ctx.Done()        (generate the new buffered channel and send struct{}{} since the ctx.s.done is nil)
→ error is returned
```

Fix

```
requestCtx.Init()
→ ctx.Init2
→ ctx.s = fakeServer        (the ctx.s.done is initialized)

requestCtx.Err()
→ ctx.Done()        (return ctx.s.done since it is non nil)
→ error is nil
```

### DIFF
--- a/server.go
+++ b/server.go
@@ -2789,6 +2789,7 @@ func (ctx *RequestCtx) Value(key any) any {
 }
 
 var fakeServer = &Server{
+	done: make(chan struct{}),
 	// Initialize concurrencyCh for TimeoutHandler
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks, I like it!

PR #1901, opened 2024-11-14
fix: correct typo in MaxIdempotentCallAttempts variable name

tanakornon:


erikdubbelboer:
Thank you for the pull request. But I'm afraid this is not something we can change without breaking backwards compatibility. Too bad it's spelled wrong, but that's just the way it will have to be...

Issue #1900, opened 2024-11-13
Incorrect handling of Cookie.SetMaxAge with value 0 or less


anthonydresser:
Per the spec that was reference during the cookie max-age implementation (https://github.com/valyala/fasthttp/issues/184), https://datatracker.ietf.org/doc/html/rfc6265#section-5.2.2, a max-age value of 0 or less is valid and is interpreted as forcing the cookie to expire immediately. 

In the current implementation, setting max-age to 0 or less causes the max age attribute to be completely ignored, making it impossible to delete a cookie via the max-age attribute. https://github.com/valyala/fasthttp/blob/master/cookie.go#L281

Given 0 is the default value of int in go, probably doesn't make sense to support `max-age=0`, but supporting setting maxAge to <0 should be possible by changing the `maxAge > 0` to `maxAge != 0`.

``` go
	if c.maxAge != 0 {
		dst = append(dst, ';', ' ')
		dst = append(dst, strCookieMaxAge...)
		dst = append(dst, '=')
		dst = AppendUint(dst, c.maxAge)
	} else if !c.expire.IsZero() {
		c.bufV = AppendHTTPDate(c.bufV[:0], c.expire)
		dst = append(dst, ';', ' ')
		dst = append(dst, strCookieExpires...)
		dst = append(dst, '=')
		dst = append(dst, c.bufV...)
	}
```

Alternatively, the maxAge property could be changed to a `*int` to support setting max-age to 0 as well and check for `nil` during the creation of the cookie header.

ksw2000:
In golang `net/http` package. They set `Max-Age` to `0` when the `c.MaxAge < 0`.

|                    |       net/http           |        fasthttp           |
|--------------|----------------------|---------------------|
|maxAge = 0 | unset                      |         unset             |
|maxAge > 0 | Max-Age=maxAge | Max-Age=maxAge| 
|<mark>maxAge < 0</mark> | <mark>Max-Age=0</mark>            |         <mark>unset</mark>             |

https://github.com/golang/go/blob/2eac154b1c8b51d05fa5b110ae065d3610a61e06/src/net/http/cookie.go#L265-L270

```go
if c.MaxAge > 0 {
	b.WriteString("; Max-Age=")
	b.Write(strconv.AppendInt(buf[:0], int64(c.MaxAge), 10))
} else if c.MaxAge < 0 {
	b.WriteString("; Max-Age=0")
}
```

erikdubbelboer:
[...]

PR #1902, opened 2024-11-14
fix: incorrect handling of negative cookie.maxAge #1900

ksw2000:
This PR fixes issue #1900 

According to [RFC6265](https://datatracker.ietf.org/doc/html/rfc6265#section-5.2.2), a max-age value of 0 or less is valid and is interpreted as forcing the cookie to expire immediately.

In the Go `net/http` package, they set `Max-Age` to `0` when the `cookie.MaxAge < 0`. 

Before the fix

|                    |       net/http           |        fasthttp           |
|--------------|----------------------|---------------------|
|maxAge = 0 | unset                      |         unset             |
|maxAge > 0 | Max-Age=maxAge | Max-Age=maxAge| 
|<mark>maxAge < 0</mark> | <mark>Max-Age=0</mark>            |         <mark>unset</mark>             |

After the fix

|                    |       net/http           |        fasthttp           |
|--------------|----------------------|---------------------|
|maxAge = 0 | unset                      |         unset             |
|maxAge > 0 | Max-Age=maxAge | Max-Age=maxAge| 
|<mark>maxAge < 0</mark> | <mark>Max-Age=0</mark>            |         <mark>Max-Age=0</mark>             |

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -77,6 +77,9 @@ type Cookie struct {
 	bufK []byte
 	bufV []byte
 
+	// maxAge=0 means no 'max-age' attribute specified.
+	// maxAge<0 means delete cookie now, equivalently 'max-age=0'
+	// maxAge>0 means 'max-age' attribute present and given in seconds
 	maxAge int
 
 	sameSite    CookieSameSite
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks! Didn't see this before I replied to the issue : 🙈

PR #1919, opened 2024-12-13
fix: "identity" has been deprecated #1909

### LINKED ISSUE #1909: Interoperability issues with reverse proxy (Transfer-Encoding: identity)
Recently we switched our reverse proxy servers to haproxy. Our users started reporting that they see increase in http 502 responses. After a bit of digging I was able to narrow it down to this on haproxy side:
```
<0>2024-11-29T15:51:19.584121+01:00 [12|h1|0|mux_h1.c:1900] Unknown transfer-encoding : [B,RUN] [MSG_DONE, MSG_DONE]
<0>2024-11-29T15:51:19.584132+01:00 [12|h1|0|mux_h1.c:2133] parsing or not-implemented error : [B,RUN] [MSG_DONE, MSG_DONE]
```

This happens when client uses HEAD method, and there is no content-length set. fasthttp inserts `Transfer-Encoding: identity` into responses. I think the problem lies in ReadBody() function.

It looks like "identity" has been deprecated for quite a long time (2008?):
https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.1
https://datatracker.ietf.org/doc/html/rfc9112#name-transfer-encoding

And here is related issue on haproxy side (they will not revert that change): https://github.com/haproxy/haproxy/issues/2623

ksw2000:
Fixes #1909 

According to modern HTTP/1.1 specifications RFC 7230 `identity` as a value for `Transfer-Encoding` was removed in the errata to RFC 2616. See section "HTTP Transfer Coding Registry" in [https://www.iana.org/assignments/http-parameters/http-parameters.xhtml](https://www.iana.org/assignments/http-parameters/http-parameters.xhtml)

In this PR, I still follow the implementation in [net/http](https://cs.opensource.google/go/go/+/master:src/net/http/transfer.go;l=612;drc=fafd4477f3d19f2c11a628e6e407ecf9924309c1) to retain the logic for handling `Transfer-Encoding: identity`. However, `identity` will not be included in the output headers.

### DIFF
--- a/header.go
+++ b/header.go
@@ -252,15 +252,13 @@ func (h *ResponseHeader) SetContentLength(contentLength int) {
 	if contentLength >= 0 {
 		h.contentLengthBytes = AppendUint(h.contentLengthBytes[:0], contentLength)
 		h.h = delAllArgsBytes(h.h, strTransferEncoding)
-	} else {
+		return
+	} else if contentLength == -1 {
 		h.contentLengthBytes = h.contentLengthBytes[:0]
-		value := strChunked
-		if contentLength == -2 {
-			h.SetConnectionClose()
-			value = strIdentity
-		}
-		h.h = setArgBytes(h.h, strTransferEncoding, value, argsHasValue)
+		h.h = setArgBytes(h.h, strTransferEncoding, strChunked, argsHasValue)
+		return
 	}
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #1920, opened 2024-12-13
fix: Method-preserving doRequestFollowRedirects

### LINKED ISSUE #1916: Method-preserving doRequestFollowRedirects in client.go
Redirecting in HTTP/1.1 typically preserves methods, but conventionally changes the POST method to GET for 301/302 responses. Browsers such as Chrome also follow this behavior. However, doRequestFollowRedirects in client.go always preserves methods, sometimes causing problems.

I've been having this problem. Can you add comment or support this? I think this should also be implemented in the redirect function.

rfc9110:

> Note: For historical reasons, a user agent MAY change the request method from POST to GET for the subsequent request. If this behavior is undesired, the 307/308 status code can be used instead.

prepaser:
Closes #1916

### DIFF
--- a/client.go
+++ b/client.go
@@ -1147,6 +1147,10 @@ func doRequestFollowRedirects(
 			break
 		}
 		url = getRedirectURL(url, location, req.DisableRedirectPathNormalizing)
+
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks!

PR #1929, opened 2024-12-19
Migrate valyala/tcplisten to this repo #1926

### LINKED ISSUE #1926: Migrate tcplisten to this repo
@erikdubbelboer I was looking at the code for `prefork` and notice it uses `tcplisten` from https://github.com/valyala/tcplisten but that repo is not receiving any updates at all.

It might be worth moving it under this repo to be able to apply improvements like https://github.com/valyala/tcplisten/pull/6 , https://github.com/valyala/tcplisten/pull/10/files and https://github.com/valyala/tcplisten/pull/8/files

Thoughts?

ksw2000:
Resolve #1926 

This PR moves the [valyala/tcplisten](https://github.com/valyala/tcplisten) package to fasthttp and also migrates [PR#6](https://github.com/valyala/tcplisten/pull/6) and [PR#10](https://github.com/valyala/tcplisten/pull/10) together. Additionally, it upgrades the deprecated `ioutil` package and replaces the `+build` directive with `go:build`. Furthermore, it fixes the warning in `tcplisten_test.go` regarding "the goroutine calls T.Fatalf, which must be called in the same goroutine as the test".


ksw2000:
I think the error is caused by `DeferAccept` and `FastOpen` not being implemented in **darwin** 🤔

gaby:
@ksw2000 Can these be added too?

- https://github.com/valyala/tcplisten/pull/8
- https://github.com/valyala/tcplisten/issues/3

ksw2000:
@gaby 

> @ksw2000 Can these be added too?
> 
> * [z/OS Compatibility tcplisten#8](https://github.com/valyala/tcplisten/pull/8)
> * [Add TCP keepalive support? tcplisten#3](https://github.com/valyala/tcplisten/issues/3)


When calling `NewListener` for `tcplisten.Config`, it internally calls `net.FileListener` to generate a `net.Listener` struct/interface, which contains the private field `lc net.ListenConfig`. I don't have a graceful solution to implement `keepalive` at the moment. 

However, there is an alternative method to set keepAlive after calling `ln.Accept()`.

```go
func main() {
	cfg := tcplisten.Config{
		ReusePort: true,
	}
	ln, err := cfg.NewListener("tcp", "0.0.0.0:8082")
	if err != nil {
		panic(err)
	}
	for {
		conn, err := ln.Accept()
		if err != nil {
			panic(err)
		}
		if conn, ok := conn.(*net.TCPConn); ok {
			conn.SetKeepAlive(true)
			conn.SetKeepAlivePeriod(time.Second)
		}
		go func() {
			defer conn.Close()
			fmt.Println("Accepted connection from", conn.RemoteAddr())
		}()
	}
}
```

Of course, if we still want to add the keepAlive feature, we could also wrap the `net.Listener` generated by `net.FileListener` and override the `Accept()` method, set keepalive before returning `conn`. But, in my opinion, this approach might confuse the original usage patterns.

gaby:
@ksw2000 Agree, it does sound confusing.

gaby:
@erikdubbelboer Thoughts on merging this?

erikdubbelboer:
Looks good to me, but needs to be rebased on master, can you do that @ksw2000?

erikdubbelboer:
Thanks!

PR #1936, opened 2025-01-10
client: Modify some interfaces to be exportable

zhangyongding:


erikdubbelboer:
What is the benefit of this?

zhangyongding:
When customizing Transport, it is necessary to use the private interface of HostClient

erikdubbelboer:
Can you show example code so we have an example of a custom implementation here?

zhangyongding:
Referring to the Default Transport in https://github.com/valyala/fasthttp/blob/master/client.go, it is necessary to use the private interface of HostClient

erikdubbelboer:
I get it, the default implementation uses all these methods [1]. I guess it's currently only possible to build your own `RoundTrip` function is you do your own connection management etc.

1. https://github.com/valyala/fasthttp/blob/321b998f311084e5260eb7688bc790e2c3e78150/client.go#L2967-L3080

PR #1945, opened 2025-01-28
perf: split delAllArgs into delAllArgs and delAllArgsStable

ksw2000:
The original `delAllArgs` method is stable but requires more copying operations to maintain the order of elements.

- Renamed the original `delAllArgs` method to `delAllArgsStable` to maintain stable behavior.
- Added a new `delAllArgs` method for non-stable functionality, improving runtime efficiency.

Depending on the specific use case:

+ Use `delAllArgsStable` when stability (order preservation) is required.
+ Use `delAllArgs` when stability is not necessary.

In the case of handling headers in `h.h`, the order of headers is not important. Refer to net/http [header.go](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/http/header.go;l=80-82;drc=4f0408a3a205a88624dced4b188e11dd429bc3ad), where the Del function calls [textproto.MIMEHeader.Del](https://cs.opensource.google/go/go/+/refs/tags/go1.23.5:src/net/textproto/header.go;l=54-56;drc=1d45a7ef560a76318ed59dfdb178cecd58caf948). This function is implemented using a map, and the map delete operation is inherently unstable.

Benchmark compare `delAllArgsStable` and `delAllArgs`

```go
func BenchmarkArgsDel(b *testing.B) {
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		var a []argsKV
		for j := 1; j < 10; j++ {
			a = appendArg(a, fmt.Sprintf("item%d", j), "val", argsHasValue)
		}
		for j := 1; j < 10; j++ {
			a = appendArg(a, fmt.Sprintf("item%d", j), "val", argsHasValue)
		}
		b.StartTimer()
		for j := 1; j < 10; j++ {
			a = delAllArgs(a, fmt.Sprintf("item%d", j))
		}
	}
}

func BenchmarkArgsDelStable(b *testing.B) {
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		var a []argsKV
		for j := 1; j < 10; j++ {
			a = appendArg(a, fmt.Sprintf("item%d", j), "val", argsHasValue)
		}
		for j := 1; j < 10; j++ {
			a = appendArg(a, fmt.Sprintf("item%d", j), "val", argsHasValue)
		}
		b.StartTimer()
		for j := 1; j < 10; j++ {
			a = delAllArgsStable(a, fmt.Sprintf("item%d", j))
		}
	}
}
```

Benchmark result:

```
$ go test --run=^$ --bench BenchmarkArgsDel -benchtime=100000x
goos: linux
goarch: amd64
cpu: AMD EPYC 7763 64-Core Processor                
BenchmarkArgsDel-4                100000              2072 ns/op
BenchmarkArgsDelStable-4          100000              2370 ns/op
PASS
ok      github.com/valyala/fasthttp     8.312s
```

### DIFF
--- a/args.go
+++ b/args.go
@@ -155,12 +155,12 @@ func (a *Args) WriteTo(w io.Writer) (int64, error) {
 
 // Del deletes argument with the given key from query args.
 func (a *Args) Del(key string) {
-	a.args = delAllArgs(a.args, key)
+	a.args = delAllArgsStable(a.args, key)
 }
 
 // DelBytes deletes argument with the given key from query args.
 func (a *Args) DelBytes(key []byte) {
-	a.args = delAllArgs(a.args, b2s(key))
+	a.args = delAllArgsStable(a.args, b2s(key))
 }
 
 // Add adds 'key=value' argument.
@@ -392,11 +392,7 @@ func copyArgs(dst, src []argsKV) []argsKV {
 	return dst
 }
 
-func delAllArgsBytes(args []argsKV, key []byte) []argsKV {
-	return delAllArgs(args, b2s(key))
-}
-
-func delAllArgs(args []argsKV, key string) []argsKV {
+func delAllArgsStable(args []argsKV, key string) []argsKV {
 	for i, n := 0, len(args); i < n; i++ {
 		kv := &args[i]
 		if key == string(kv.key) {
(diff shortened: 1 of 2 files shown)


ksw2000:
@erikdubbelboer It's been 3 weeks, what do you think?

erikdubbelboer:
Thanks I somehow missed it.

PR #1950, opened 2025-02-09
fix: compression priority

inetol:
Before this PR, fasthttp would never use zstd if clients sent their `Accept-Encoding` header including gzip because zstd was the last option and would only be used if zstd was the only encoding available in `Accept-Encoding` header. This is really an issue when serving precompressed sidecars of zstd and gzip where the server preferred the latter.

The following order has been established; first brotli, then zstd, and finally gzip. It takes into account whether `CompressBrotli` and the new `CompressZstd` option are enabled to select which compression to use.

It also fixes the missing Vary header on compressed assets.

~~There is also an issue that has been revealed with the rework of these tests that should be addressed in a separate PR~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,8 @@
 tags
 *.pprof
-*.fasthttp.gz
 *.fasthttp.br
+*.fasthttp.zst
+*.fasthttp.gz
 .idea
 .vscode
 .DS_Store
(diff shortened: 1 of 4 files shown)


erikdubbelboer:
Nice, I like the change. It makes sense to try better algorithms before falling back to gzip.

Can you have a look are the linting issues and failing tests?

erikdubbelboer:
Thanks!

PR #1953, opened 2025-02-12
fix: accept invalid headers with a space

### LINKED ISSUE #1917: Ignoring parsing errors of parseHeaders in client.go
I have to deal with the wrong response header (`Referrer Policy`, not `Referrer-Policy` maybe a typo?) sent by the server, but fasthttp causes an error, the moment I receive the wrong header field, and everything goes down. To deal with this, there is nothing I can do but change the fasthttp code. Simply deleting the validHeaderFieldByte in parseHeaders works fine. Browsers also generally seem to just ignore those fields.

Could you please provide a way to ignore the wrong header field? I think it would be enough to optionally disable header field validation, or provide a way to modify validHeaderFieldByteTable.

ksw2000:
Fix #1917 

Make behavior consistent with `net/http` by allowing header keys and trailers containing spaces without canonicalizing them.

Change the behavior of `headerScanner.next()`. When parsing headers, we do not normalize the header key. 
In addition, before normalizing the key, check whether the key contains a space. If it does, disable normalization forcibly.

### DIFF
--- a/header.go
+++ b/header.go
@@ -2689,7 +2689,7 @@ func (h *RequestHeader) parse(buf []byte) (int, error) {
 	return m + n, nil
 }
 
-func addTrailerBytes(src []byte, buf []byte, trailers [][]byte, disableNormalizing bool) ([]byte, [][]byte, error) {
+func addTrailerBytes(src, buf []byte, trailers [][]byte, disableNormalizing bool) ([]byte, [][]byte, error) {
 	var err error
 	for i := -1; i+1 < len(src); {
 		src = src[i+1:]
--- a/header.go
+++ b/header.go
@@ -2736,35 +2736,32 @@ func parseTrailer(src []byte, dest []argsKV, disableNormalizing bool) ([]argsKV,
 
 	var s headerScanner
 	s.b = src
-	var err error
+
 	for s.next() {
 		if len(s.key) > 0 {
-			// We accept invalid headers with a space before the
-			// colon, but must not canonicalize them.
-			// See: https://github.com/valyala/fasthttp/issues/1917
-			for _, b := range s.key {
-				switch b {
-				case '\t':
-					err = fmt.Errorf("invalid trailer key %q", s.key)
-					continue
-				case ' ':
-					disableNormalizing = true
+			disable := disableNormalizing
+			for _, ch := range s.key {
+				if !validHeaderFieldByte(ch) {
+					// We accept invalid headers with a space before the
+					// colon, but must not canonicalize them.
+					// See: https://github.com/valyala/fasthttp/issues/1917
+					if ch == ' ' {
+						disable = true
+						continue
+					}
+					return dest, 0, fmt.Errorf("invalid trailer key %q", s.key)
 				}
 			}
 			// Forbidden by RFC 7230, section 4.1.2
 			if isBadTrailer(s.key) {
-				err = fmt.Errorf("forbidden trailer key %q", s.key)
-				continue
+				return dest, 0, fmt.Errorf("forbidden trailer key %q", s.key)
 			}
-			normalizeHeaderKey(s.key, disableNormalizing)
+			normalizeHeaderKey(s.key, disable)
 			dest = appendArgBytes(dest, s.key, s.value, argsHasValue)
 		}
 	}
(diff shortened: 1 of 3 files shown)


ksw2000:
I think the fuzzy test run failure is independent of my changes since the previous commit passed. The latest commit is a refactor to fix linter problems. In addition, the failed test - `fuzzResponseReadLimitBody` is irrelevant to the header parser

erikdubbelboer:
No, what the fuzzer found is caused by your changes. Here is the test case from the fuzzer that passes on master but panics on your branch:
```go
func TestReadLimitBodyExample(t *testing.T) {
	body := []byte{0xa1, 0xfd, 0xdf, 0x77, 0x20, 0x35, 0xa, 0x54, 0x72, 0x61, 0x6e, 0x73, 0x66, 0x65, 0x72, 0x2d, 0x45, 0x6e, 0x63, 0x6f, 0x64, 0x69, 0x6e, 0x67, 0x3a, 0xb0, 0xa, 0xa, 0x30, 0x52, 0xd, 0xa, 0x3a, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x89, 0x3a, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x89, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0xa, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0xa, 0x20, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0x20, 0x3a, 0x3a, 0xa, 0xa, 0x20, 0x20, 0xa, 0x20, 0xa, 0x20, 0x3a, 0x20, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x20, 0x3a, 0x20, 0xa, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2e}

	res := AcquireResponse()
	defer ReleaseResponse(res)

	_ = res.ReadLimitBody(bufio.NewReader(bytes.NewReader(body)), 24922)
}
```

ksw2000:
This bug is caused by two issues:

1. When refactoring `parseTrailer`, the usage of [continue](https://github.com/valyala/fasthttp/blob/4f90454a39b5b5cad778f7703873a2761da97072/header.go#L2749) jumped to the wrong loop, and there was also a logical error in using [disableNormalizing](https://github.com/valyala/fasthttp/blob/4f90454a39b5b5cad778f7703873a2761da97072/header.go#L2751). This is not the main problem.
2. In the original implementation, if the key in the trailer was `" \n"`, the program would return an error because of the space. However, after the modification, spaces were allowed, but since the initial logic did not block `"\n"`, the modified program would mistakenly allow this error. In the new implementation, `parseTrailer` follows the approach of `parseHeaders` by calling `validHeaderFieldByte` to check for unexpected characters.

I am trying to minimize the test you provided and write it back as a unit test.

erikdubbelboer:
Thanks!

Issue #1960, opened 2025-02-21
Building for linux-386 is failing with error math.MaxUint32 (untyped int constant 4294967295) overflows int


vivek-yadav:
In Go, the math.MaxUint32 constant is an untyped integer constant that defaults to type int in contexts where a typed value is required. On a 32-bit architecture like linux-386, the int type is 32-bit signed, and math.MaxUint32 (which is 4294967295) exceeds the maximum value that can be represented by a 32-bit signed integer, causing an overflow error.

When building for Linux 32bit - linux-386 platform, I am getting this error:
`github.com/valyala/fasthttp@v1.59.0/tcplisten/tcplisten.go:187:9: math.MaxUint32 (untyped int constant 4294967295) overflows int`

This change:
https://github.com/valyala/fasthttp/pull/1929/files#diff-7736b8b9d7fd04be372c289a070fe0dbc9b6f16c3a6ee6ed1dcf5861c27f7546

```go
if i > math.MaxUint32 {
	return 0, errors.New("value exceeds uint32 max value")
}
```

It is blocking a lot of my services and CLI from building as I have pipeline for `linux-386` also.

To avoid this error, you can explicitly convert math.MaxUint32 to an unsigned integer type like `uint32` when using it in comparisons or assignments. This ensures that the constant is treated as an unsigned integer, which can handle the full range of MaxUint32 without overflow.

PR #1961, opened 2025-02-21
fix: explicit uint32 conversion so it works in linux-386 platform build

vivek-yadav:
Fix for issue: https://github.com/valyala/fasthttp/issues/1960

In Go, the math.MaxUint32 constant is an untyped integer constant that defaults to type int in contexts where a typed value is required. On a 32-bit architecture like linux-386, the int type is 32-bit signed, and math.MaxUint32 (which is 4294967295) exceeds the maximum value that can be represented by a 32-bit signed integer, causing an overflow error.

To avoid this error, you can explicitly convert math.MaxUint32 to an unsigned integer type like uint32 when using it in comparisons or assignments. This ensures that the constant is treated as an unsigned integer, which can handle the full range of MaxUint32 without overflow.

### DIFF
--- a/tcplisten/tcplisten.go
+++ b/tcplisten/tcplisten.go
@@ -184,7 +184,7 @@ func safeIntToUint32(i int) (uint32, error) {
 	if i < 0 {
 		return 0, errors.New("value is negative, cannot convert to uint32")
 	}
-	if i > math.MaxUint32 {
+	if uint32(i) > uint32(math.MaxUint32) {
 		return 0, errors.New("value exceeds uint32 max value")
 	}
 	return uint32(i), nil


erikdubbelboer:
Seems like `gosec` thinks these changes are insecure, can you have a look at that?

guillemj:
I've applied the patch from the branch at https://github.com/guillemj/golang-valyala-fasthttp/tree/pu/fix-32-bit-builds in the Debian package for this project (https://tracker.debian.org/news/1622205/accepted-golang-github-valyala-fasthttp-11590-2-source-into-unstable/), which I think is the more correct fix, and does not trigger a new security report (although the existing cast does trigger a security report, so that should be fixed independently, perhaps by marking it as a non-issue in the repo or something). The test results at https://github.com/guillemj/golang-valyala-fasthttp/actions/runs/13576028312/job/37952351606.

Cacsjep:
also need this =)

erikdubbelboer:
Fixed in: https://github.com/valyala/fasthttp/commit/30b09beff11be4282c37ef6b7d8854589d565447

Issue #1975, opened 2025-03-16
Drop in performance at Tech Empower benchmark


NBM1977:
Hi, what happened to fasthttp scores at Tech Empower benchmark, it used to be one of the best framework, but the performance is dropping like a rock at this benchmark since 2022. It was some kind of security bug?

This is the plaintext benchmark:

2020  fasthttp	     5,962,266	     81.2%
2021  fasthttp       6,671,354         95.1%
2022  fasthttp       929,880           13.2%
2023  fasthttp       367,504            5.2%
2025  fasthttp	     569,793	     2.0%

Thanks.

erikdubbelboer:
I have no idea what could be causing this. Lots of changes have been made over the years, but none should affect performance this much. Has the benchmark changed?

If I have some time I'll try to run my own benchmark against older code and see if I can find anything.

NBM1977:
I am not sure if they changed something, In 2021 I decided to use  fasthttp in my project after I saw the great performance at their benchmark. From time to time I check that website for updates and saw this sharp drop in performance in 2022. I thought it was some error in their benchmark but later the results continued bad, now I am kind of worried about this.

erikdubbelboer:
I have tested all versions and only found this specific commit that caused a ~15% slowdown in handling requests: https://github.com/valyala/fasthttp/commit/3b117f8f1e822d8069eb2fb3794a69a5e237c223
I'm trying to find a fix for this. It's using the Mutex too often.

Here is the result of my running a wrk benchmark on all versions (printing requests per second):
```
v1.10.0: 362579.16
v1.11.0: 359667.31
v1.12.0: 364035.49
v1.13.0: 363075.05
v1.14.0: 363688.84
v1.15.0: 364585.41
v1.16.0: 365185.00
v1.17.0: 358476.87
v1.18.0: 363085.11
v1.19.0: 364730.17
v1.20.0: 360243.44
v1.21.0: 364885.29
v1.22.0: 360594.45
v1.23.0: 360707.36
v1.24.0: 366293.64
v1.25.0: 364223.53
v1.26.0: 368327.07
v1.27.0: 360010.40
v1.28.0: 362552.11
v1.29.0: 361046.06
v1.30.0: 361533.39
v1.31.0: 363819.79
v1.32.0: 312502.24 <-- this is where the rps dropped
v1.33.0: 314383.42
v1.34.0: 310461.47
v1.35.0: 304573.11
v1.36.0: 310698.89
[...]

PR #1986, opened 2025-04-01
Remove idleConns mutex for every request

erikdubbelboer:
Locking and unlocking a mutex multiple times per request is a major slowdown that we can avoid with clever use of atomics.

Before:
```
BenchmarkServerGet100ReqPerConn10KClients-12    	 84167428	       867.7 ns/op
```
After:
```
BenchmarkServerGet100ReqPerConn10KClients-12    	187397954	       386.3 ns/op
```

See: https://github.com/valyala/fasthttp/issues/1975

### DIFF
--- a/server.go
+++ b/server.go
@@ -219,7 +219,7 @@ type Server struct {
 
 	concurrencyCh chan struct{}
 
-	idleConns map[net.Conn]time.Time
+	idleConns map[net.Conn]*atomic.Int64
 	done      chan struct{}
 
 	// Server name for sending in response headers.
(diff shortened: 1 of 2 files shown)


mxmauro:
Hi @erikdubbelboer, how much simultaneous connections do you usually see on a heavy loaded backend?

I say this because a new allocation is done per idle connection (again, assuming a heavy loaded backend) and GC becomes a CPU killer a some point.

Regards,
Mauro.

erikdubbelboer:
@mxmauro which allocation do you mean? This change shouldn't introduce any new allocations, we have tests to make sure of that.

mxmauro:
@erikdubbelboer in the change, you have the `v = &atomic.Int64{}` located at line 2143 that fills the pool. This involves an allocation per entry.

Of course at some point, they will be reused but I make emphasis on the heavily loaded instance because there can exists 10k active entries at the same time. Well, depending on how many simultaneous connections are being handled.

One workaround I have in mind is to allocate a slice of int64s with an index of the next-free element and return it `v = &currentArray[nextFreeIndex]`. When `nextFreeIndex` reaches the size of the array, you can allocate a new slice with a fixed amount or double amount of entries. Here you'll have a single allocation per, let's say for example, 1024 entries.

erikdubbelboer:
There are many more allocations per connection for all kinds of buffers. The important part is that the object is reused if a connection is closed and another is opened.

You can try building your slice solution as it might be more efficient. I don't see how this would work without a Mutex if you have to dynamically resize it, but I'm curious if you find a way.

mxmauro:
Hi @erikdubbelboer sorry for the late reply. I'll think on something.

Have a nice week.

PR #1995, opened 2025-04-11
Fix round robin addresses in dual stack dialing

raviqqe:
This PR fixes round robin addresses in dual stack dialing.

# Question

What and where is the best way to unit-test this fix?

# References

- https://github.com/raviqqe/muffet/issues/449

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -299,7 +299,7 @@ func (d *TCPDialer) dial(addr string, dualStack bool, timeout time.Duration) (ne
 	}
 	var conn net.Conn
 	n := uint32(len(addrs)) // #nosec G115
-	for n > 0 {
+	for range n {
 		conn, err = d.tryDial(network, addrs[idx%n].String(), deadline, d.concurrencyCh)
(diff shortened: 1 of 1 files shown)


tcpdialer.go
```diff
@@ -299,7 +299,7 @@ func (d *TCPDialer) dial(addr string, dualStack bool, timeout time.Duration) (ne
 	}
 	var conn net.Conn
 	n := uint32(len(addrs)) // #nosec G115
-	for n > 0 {
+	for range n {
 		conn, err = d.tryDial(network, addrs[idx%n].String(), deadline, d.concurrencyCh)
```
raviqqe, line 303: In this statement, the variable `n` should not change over different iterations of the loop while we want to try `n` addresses.

For example, in the previous implementation, if `n` is 2 and the initial `idx` is 2,

1. In the first iteration, `idx % n = 2 % 2 = 0`.
1. In the second iteration, `idx % n = 3 % 1 = 0`.

erikdubbelboer:
Nice find!

Issue #1992, opened 2025-04-08
Support HTTP Early Hints


pjebs:
Early hints allows the server to hint what other resources the page would need and then the browser can fetch it while the server is processing the main request. 

That way, the browser can start downloading CSS/jss/images etc with a head start. 

All major browsers support it

https://developer.chrome.com/docs/web-platform/early-hints

Early hints are technically a http2+ feature but all major browsers still understand it (or safely ignore) with http1.1

https://github.com/whatwg/fetch/issues/1698#issuecomment-1703594215

```go
HTTP/1.1 103 Early Hints
Link: </style.css>; rel=preload; as=style
Link: </script.js>; rel=preload; as=script

HTTP/1.1 200 OK
Date: Fri, 26 May 2017 10:02:11 GMT
Content-Length: 1234
Content-Type: text/html; charset=utf-8
Link: </style.css>; rel=preload; as=style
Link: </script.js>; rel=preload; as=script

[… rest of the response body is omitted from the example …]

```

pjebs:
```go
func main() {
    helloHandler := func(w http.ResponseWriter, req *http.Request) {
        w.Header().Add("Link", "</style.css>; rel=preload; as=style")
        w.Header().Add("Link", "</script.js>; rel=preload; as=script")

        w.WriteHeader(103)

        // do your heavy tasks such as DB or remote APIs calls here

        w.WriteHeader(200)

        io.WriteString(w, "<!doctype html>\n[... rest of the response body is omitted from the example ...]")
    }

    http.HandleFunc("/hello", helloHandler)
    log.Fatal(http.ListenAndServeTLS(":443", "cert.pem", "key.pem", nil))
}
```

pjebs:
Early hints allows the server to hint what other resources the page would need any send it to the server while processing the main request. 

That way, the browser can start downloading CSS/jss/images etc as a head start. 

All major browsers support it

https://developer.chrome.com/docs/web-platform/early-hints

-

pjebs:
Perhaps it will never be possible with fasthttp:

```go
package main

import (
	"fmt"
	"github.com/valyala/fasthttp"
	"log"
	"time"
	"sync"
	"net/http"
)

func main() {
	wg := sync.WaitGroup{}
	wg.Add(2)

	go func() { // fasthttp
		requestHandler := func(ctx *fasthttp.RequestCtx) {
[...]

PR #1996, opened 2025-04-16
early hint functionality

pjebs:
Draft of Early Hints functionality.

Fixes: https://github.com/valyala/fasthttp/issues/1992

### DIFF
--- a/server.go
+++ b/server.go
@@ -623,6 +623,72 @@ type RequestCtx struct {
 	hijackNoResponse bool
 }
 
+// EarlyHints allows the server to hint to the browser what resources a page would need
+// so the browser can preload them while waiting for the server's full response. Only Link
+// headers already written to the response will be transmitted as Early Hints.
+//
+// This is a HTTP/2+ feature but all browsers will either understand it or safely ignore it.
+//
+// NOTE: Older HTTP/1.1 non-browser clients may face compatibility issues.
+//
+// See: https://developer.chrome.com/docs/web-platform/early-hints and
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link#syntax
+//
+// Example:
+//
+//	func(ctx *fasthttp.RequestCtx) {
+//	   ctx.Response.Header.Add("Link", "<https://fonts.google.com>; rel=preconnect")
+//	   ctx.EarlyHints()
+//	   time.Sleep(5*time.Second) // some time-consuming task
+//	   ctx.SetStatusCode(fasthttp.StatusOK)
+//	   ctx.SetBody([]byte("<html><head></head><body><h1>Hello from Fasthttp</h1></body></html>"))
+//	}
+func (ctx *RequestCtx) EarlyHints() error {
+	links := ctx.Response.Header.PeekAll(b2s(strLink))
+	if len(links) > 0 {
+		c := acquireWriter(ctx)
+		defer releaseWriter(ctx.s, c)
+		_, err := c.Write(strEarlyHints)
+		if err != nil {
+			return err
+		}
+		for _, l := range links {
+			if len(l) == 0 {
+				continue
+			}
+			_, err = c.Write(strLink)
+			if err != nil {
+				return err
+			}
+			_, err = c.Write(strColon)
+			if err != nil {
+				return err
+			}
+			_, err = c.Write(strSpace)
+			if err != nil {
+				return err
+			}
+			_, err = c.Write(l)
+			if err != nil {
+				return err
+			}
+			_, err = c.Write(strCRLF)
+			if err != nil {
+				return err
+			}
+		}
+		_, err = c.Write(strCRLF)
+		if err != nil {
+			return err
+		}
+		err = c.Flush()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
 // HijackHandler must process the hijacked connection c.
 //
 // If KeepHijackedConns is disabled, which is by default,
--- a/server_test.go
+++ b/server_test.go
@@ -1919,6 +1919,49 @@ func TestServerExpect100Continue(t *testing.T) {
 	}
 }
 
+func TestServerExpect103EarlyHints(t *testing.T) {
+	t.Parallel()
+
+	s := &Server{
+		NoDefaultContentType:  true,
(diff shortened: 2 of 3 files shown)


server.go
```diff
@@ -623,6 +623,42 @@ type RequestCtx struct {
...
+			hint = k + b2s(strSemiColon) + v // k;v
+		} else {
+			hint = b2s(strAngleLeft) + k + b2s(strAngleRight) + b2s(strSemiColon) + v // <k>;v
+		}
+		c.Write(strLink)   //nolint:errcheck // no way to handle this error anyway
+		c.Write(strColon)  //nolint:errcheck // no way to handle this error anyway
+		c.Write(strSpace)  //nolint:errcheck // no way to handle this error anyway
+		c.Write(s2b(hint)) //nolint:errcheck // no way to handle this error anyway
+		c.Write(strCRLF)   //nolint:errcheck // no way to handle this error anyway
+		ctx.Response.Header.AddBytesKV(strLink, s2b(hint))
```
erikdubbelboer, line 657: Why is this also needed?

erikdubbelboer (changes requested):
The `hints map[string]string` prevents me from generating for example these headers:
```
Link: <https://fonts.google.com>; rel=preconnect
Link: </main.css>; rel=preload; as=style
Link: </common.js>; rel=preload; as=script
Link: </experimental.3eab3290.css>; rel=preload; as=style
```

Can you also add a test case that shows the Early Hints response and makes sure the normal response after it also works fine.

pjebs:
> The `hints map[string]string` prevents me from generating for example these headers:
> 
> ```
> Link: <https://fonts.google.com>; rel=preconnect
> Link: </main.css>; rel=preload; as=style
> Link: </common.js>; rel=preload; as=script
> Link: </experimental.3eab3290.css>; rel=preload; as=style
> ```
> 
> Can you also add a test case that shows the Early Hints response and makes sure the normal response after it also works fine.

```go
ctx.EarlyHints(map[string]string{
    "https://fonts.google.com": "rel=preconnect",
    "/main.css": "rel=preload; as=style",
    "/common.js": "rel=preload; as=script",
    "/experimental.3eab3290.css": "rel=preload; as=style",
}) 

```

server.go
pjebs: It was the fastest way I could think of the concat `[]byte`

erikdubbelboer: But now you are both sending the Link in the Early Hints and in the normal response after it, that isn't needed then right?

erikdubbelboer:
I'm wondering if this is the best API. Another way might be that this function has no arguments and just sends the headers set in the response already and then resets them. Not sure if an Early Hints response can also have other headers?

pjebs:
> I'm wondering if this is the best API. Another way might be that this function has no arguments and just sends the headers set in the response already and then resets them. Not sure if an Early Hints response can also have other headers?

The point of early hints is that the Link Headers are sent before the processing of the full response.
In some cases, the full response may take "considerable" time to determine what the headers are and then write them to Response.

server.go
pjebs: The article: https://developer.chrome.com/docs/web-platform/early-hints

At the bottom says to do that:

pjebs:
@erikdubbelboer Is there a sample test that I can use as a template that tests something similar?

erikdubbelboer:
The API I was thinking of is just:
```go
ctx.Response.Header.Add("Link", "...")
ctx.EarlyHints()
```
Which then only send the Link header and not any other header that might already have been set.

This design is more in line with what we already have and makes it easier to have no memory allocations.

For an example of a test you can take the 100 continue test: https://github.com/valyala/fasthttp/blob/48f3a2f423f103cd67e504a51f3ec3a1381a5620/server_test.go#L1881

pjebs:
> The API I was thinking of is just:
> 
> ```go
> ctx.Response.Header.Add("Link", "...")
> ctx.EarlyHints()
> ```
> 
> Which then only send the Link header and not any other header that might already have been set.
> 
> This design is more in line with what we already have and makes it easier to have no memory allocations.
> 
> For an example of a test you can take the 100 continue test:
> 
> https://github.com/valyala/fasthttp/blob/48f3a2f423f103cd67e504a51f3ec3a1381a5620/server_test.go#L1881

But I still have to iterate over the headers to determine if the header is a `Link` (I.e by doing a comparison). That is slower. 

There is no guarantee the developer will only add Links to header before calling `EarlyHints`.

erikdubbelboer:
The amount of headers will be so low that its probably around the same speed as the map. A map makes it really hard not to generate garbage which would make that much slower.

pjebs:
@erikdubbelboer Done

erikdubbelboer (changes requested):

server.go
```diff
@@ -623,6 +623,36 @@ type RequestCtx struct {
...
+// This is a HTTP/2+ feature but all browsers will either understand it or safely ignore it.
+//
+// NOTE: Older HTTP/1.1 non-browser clients may face compatibility issues.
+//
+// See: https://developer.chrome.com/docs/web-platform/early-hints and
+// https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Link#syntax
+func (ctx *RequestCtx) EarlyHints() {
+	links := ctx.Response.Header.PeekAll(b2s(strLink))
+	if len(links) > 0 {
+		c := acquireWriter(ctx)
```
erikdubbelboer, line 639: Don't forget to release it again after the flush.

erikdubbelboer (changes requested):

pjebs:
@erikdubbelboer Done

erikdubbelboer (changes requested):
Sorry, 2 more changes.
1. Can you make the function return an error so all the `nolint:errcheck` can go away and can instead be turned into error returns. Also check the error returned by Flush and `defer` the `releaseWriter` so it also gets released on error.
2. Can you add a simple example to the documentation of the function.

pjebs:
What's the difference between c.Conn() and using that compared to acquire/release?

erikdubbelboer:
Conn() returns the raw connection. Writing directly to it could cause each Write() to be a separate TCP packet. The writer is buffered and will send everything in a single packet on Flush().

PR #1997, opened 2025-04-20
WIP: Improved idle connections timeout handling.

mxmauro:
Hi @erikdubbelboer, hope you are doing well.

This is a proposal for a change in how idle connections timeouts are handled.

As I suspected, the `v ;= &atomic.Int64{}` does an allocation but the Go test framework does not show it because the `testing.AllocsPerRun` does an integer division (the why is documented) and, because no parallel execution, that allocated atomic value is reused on each run (because of the pool).

Because all the serving is enclosed in a single function my proposal is the following:

1. A fixed set of double-linked list were added to the `Server` struct, in order to reduce the lock time on the `closeIdleConns` loop.
2. An entry is allocated on the stack with the connection and atomic int for the timestamp, and then added to the list. I'm using unsafe pointers on purpose to avoid the compiler assume the list item will escape.
3. The entry is removed from the list before the function exits.
4. I'm not removing the entry if the idle time times out. Closing the connection should be enough to break the main serve routine and, on exit, it will be removed. Also allows to avoid the extra "already removed" check.

I'm keeping the PR a work-in-progress until your feedback,

Kind regards,
Mauro.

### DIFF
--- a/server.go
+++ b/server.go
@@ -15,7 +15,10 @@ import (
 	"sync"
 	"sync/atomic"
 	"time"
+	"unsafe"
 )
+
+const idleConnListCount = 16 // Keep this a multiple of 2.
 
 var errNoCertOrKeyProvided = errors.New("cert or key has not provided")
 
@@ -219,8 +222,9 @@ type Server struct {
 
 	concurrencyCh chan struct{}
 
-	idleConns map[net.Conn]*atomic.Int64
-	done      chan struct{}
+	idleConns [idleConnListCount]idleConnList
+
+	done chan struct{}
 
 	// Server name for sending in response headers.
 	//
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
There is a bunch of tests failing.

mxmauro:
Yes :( I'm trying to figure why. I could reproduce locally using the shuttle and race flags else didn't happen. I'll let you now when fixed.

mxmauro:
Figured out the problem is stack variables can change their position on stack grow. I'll close it until I have a better approach now I also know how to run the CI tests.

erikdubbelboer:
Ah yeah and the `uintptr` doesn't get updated then. I also didn't see that issue. Too bad escape analysis isn't smart enough to understand it doesn't escape.

PR #2000, opened 2025-04-28
Add support for streaming identity-encoded or unknown length response bodies

osxtest:
This change adds support for streaming response bodies when the `Transfer-Encoding` is identity (or when `Content-Length` is not provided).


erikdubbelboer:
Thanks!

PR #2005, opened 2025-05-06
modify `acceptConn` for `RIO`

wamshawn:
* add connKeepAliveer interface{}.
* use connKeepAliveer insteadof *net.TCPConn to set `TCPKeepalive` and `TCPKeepalivePeriod`.

I create a [rio](https://github.com/brickingsoft/rio) project, which is based on `IOURING`.

Because `rio.TCPConn` implmenets `net.Conn`, and has same functions as `net.TCPConn` had. 
For `rio` can be used as a network library for `FastHTTP`, so I create `connKeepAliveer` interface, use it to set `TCPKeepalive` and `TCPKeepalivePeriod`.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1979,6 +1979,12 @@ func (s *Server) ShutdownWithContext(ctx context.Context) (err error) {
 	}
 }
 
+type connKeepAliveer interface {
+	SetKeepAlive(keepalive bool) error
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
Thanks!

PR #2008, opened 2025-05-09
fix: removed resolved issue link from readme file, issue no longer ex…

viralkansarav:
The readme file contains a link of issue which is not found, the issue might be resolved or may be closed. Hence there is no need of adding link of the issue in the main documentation.


erikdubbelboer:
Thanks!

Issue #22, opened 2015-12-21
URI().Path bytes to string


iowelinux:
Hi,

It's told to avoid conversion between []byte and string 'cause Fasthttp API provides functions for both.
However I've looked through the code and haven't found a function that would return a string for Path.
There is URI().String() but I don't see the same for URI().Path()

It seems the only way to sort this out is to convert from bytes to string but it's not recommended by you :)
Is there any other way to get Path as a string?

Thanks.

valyala:
Hi @iowelinux ,

fasthttp API doesn't provide `Path().String()` because it cannot be implemented reliably without memory allocation. There is [a hack](http://stackoverflow.com/a/27614755/274937), which allows converting `[]byte` to `string` without memory allocation, but it may break in new go versions, so fasthttp API doesn't use it. So just use `string(URI().Path())` if you cannot use `[]byte` result. This will allocate memory for `string` and copy `[]byte` contents into `string`. Or use the aforementioned hack on your own risk :)

valyala:
@iowelinux , see the update for `string` conversion exceptions.

PR #2009, opened 2025-05-16
Removed wrong information from main documentation as these functions does not exist

viralkansarav:
There are no such functions to manually convert []bytes to string or vice versa, confirmed by [valyala](https://github.com/valyala) as per closed [issue](https://github.com/valyala/fasthttp/issues/22) . We can later add "unsafe" practices that users can attain in case of critical performance like using unsafe method to convert []byte to string without memory allocation. But we can't recommend these into best practices section , we can have a different section to that states attain over-clocked performance using certain hacks.


erikdubbelboer:
Closing this as it has 0 files changed now.

PR #2027, opened 2025-06-22
refact: eliminate duplication in Request/Response via struct embedding

ksw2000:
Reduces code duplication by embedding the `header` struct into both `RequestHeader` and `ResponseHeader` to share common fields.

I found that there are many duplicate methods in `header.go`. We can reduce this repetition by using struct embedding. This PR improves maintainability by allowing shared logic between `RequestHeader` and `ResponseHeader` to be implemented in a single method with a single comment.

Additionally, by embedding the Header struct, it implicitly communicates to users which methods on `RequestHeader` and `ResponseHeader` share the exact same underlying implementation. 

During the refactoring process, I also discovered that the SetProtocol methods in RequestHeader and ResponseHeader are inconsistent.

```go
// SetProtocol sets HTTP request protocol.
func (h *RequestHeader) SetProtocol(method string) {
	h.protocol = append(h.protocol[:0], method...)
	h.noHTTP11 = !bytes.Equal(h.protocol, strHTTP11)
}

// SetProtocolBytes sets HTTP request protocol.
func (h *RequestHeader) SetProtocolBytes(method []byte) {
	h.protocol = append(h.protocol[:0], method...)
	h.noHTTP11 = !bytes.Equal(h.protocol, strHTTP11)
}

// SetProtocol sets response protocol bytes.
func (h *ResponseHeader) SetProtocol(protocol []byte) {
	h.protocol = append(h.protocol[:0], protocol...)
}
```

The same function name `SetProtocol` in `RequestHeader` and `ResponseHeader` have different inputs, and the `ResponseHeader.SetProtocol()` does not update the `noHTTP11`.

### DIFF
--- a/header.go
+++ b/header.go
@@ -693,14 +693,14 @@ func (h *header) Protocol() []byte {
 }
 
 // SetProtocol sets HTTP request protocol.
-func (h *RequestHeader) SetProtocol(method string) {
-	h.protocol = append(h.protocol[:0], method...)
+func (h *RequestHeader) SetProtocol(protocol string) {
+	h.protocol = append(h.protocol[:0], protocol...)
 	h.noHTTP11 = !bytes.Equal(h.protocol, strHTTP11)
 }
 
 // SetProtocolBytes sets HTTP request protocol.
-func (h *RequestHeader) SetProtocolBytes(method []byte) {
-	h.protocol = append(h.protocol[:0], method...)
+func (h *RequestHeader) SetProtocolBytes(protocol []byte) {
+	h.protocol = append(h.protocol[:0], protocol...)
 	h.noHTTP11 = !bytes.Equal(h.protocol, strHTTP11)
 }
 


header.go
```diff
@@ -411,22 +344,7 @@ func (h *RequestHeader) SetContentEncodingBytes(contentEncoding []byte) {
...
-	b = append(b, boundary...)
-	h.bufV = b
-
-	h.SetContentTypeBytes(h.bufV)
-}
-
-// SetMultipartFormBoundaryBytes sets the following Content-Type:
-// 'multipart/form-data; boundary=...'
-// where ... is substituted by the given boundary.
-func (h *RequestHeader) SetMultipartFormBoundaryBytes(boundary []byte) {
```
erikdubbelboer: This function is only removed, it's not added back to `header`.

header.go
```diff
@@ -723,23 +670,23 @@ func (h *RequestHeader) SetMethodBytes(method []byte) {
...
+func (h *header) Protocol() []byte {
+	if len(h.protocol) == 0 {
 		return strHTTP11
 	}
-	return h.proto
+	return h.protocol
 }
 
 // SetProtocol sets HTTP request protocol.
 func (h *RequestHeader) SetProtocol(method string) {
```
erikdubbelboer: Can you change the param name to `protocol` while you're making changes here anyways 😄

header.go
```diff
@@ -723,23 +670,23 @@ func (h *RequestHeader) SetMethodBytes(method []byte) {
...
-	h.proto = append(h.proto[:0], method...)
-	h.noHTTP11 = !bytes.Equal(h.proto, strHTTP11)
+	h.protocol = append(h.protocol[:0], method...)
+	h.noHTTP11 = !bytes.Equal(h.protocol, strHTTP11)
 }
 
 // SetProtocolBytes sets HTTP request protocol.
 func (h *RequestHeader) SetProtocolBytes(method []byte) {
```
erikdubbelboer: Can you change the param name to `protocol` while you're making changes here anyways 😄

erikdubbelboer (changes requested):

header.go
ksw2000: I think I mistook `SetMultipartFormBoundaryBytes` for `func (h *ResponseHeader) SetMultipartFormBoundary`. I've reverted the changes.

header.go
ksw2000: Done!

header.go
ksw2000: Done!

erikdubbelboer:
Thanks!

PR #2028, opened 2025-06-23
add helper functions to allow set time or conntime (useful in tests)

peczenyj:
On my programs, sometimes I need to fetch the fasthttp.RequestCtx Time (or ConnTime) and if I need to test it properly I need to inject some fake time.

Since time is a private field, I add two helper functions that can swap the content of each attribute.

I call it `RequestCtxSwapTime` because it return the previous value, to be restored if needed.

I think the usage of this feature will be very limited, but it will work for me :)

### DIFF
--- a/server.go
+++ b/server.go
@@ -3047,3 +3047,23 @@ var stateName = []string{
 func (c ConnState) String() string {
 	return stateName[c]
 }
+
+// RequestCtxSwapTime is a helper function that allows swap the request time,
+// returning the old value.
+func RequestCtxSwapTime(ctx *RequestCtx, time time.Time) time.Time {
+	old := ctx.Time()
+
+	ctx.time = time
+
+	return old
+}
+
+// RequestCtxSwapConnTime is a helper function that allows swap the connection time,
+// returning the old value.
+func RequestCtxSwapConnTime(ctx *RequestCtx, connTime time.Time) time.Time {
+	old := ctx.ConnTime()
+
+	ctx.connTime = connTime
+
+	return old
+}
--- a/server_test.go
+++ b/server_test.go
@@ -4471,3 +4471,45 @@ func TestRequestCtxInitShouldNotBeCanceledIssue1879(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestRequestCtxSwapTime(t *testing.T) {
+	var r Request
+	var requestCtx RequestCtx
+
+	requestCtx.Init(&r, nil, nil)
+
+	now := time.Date(2025, 6, 20, 0, 0, 0, 0, time.UTC)
+
+	previousRequestCtxTime := RequestCtxSwapTime(&requestCtx, now)
+
+	if !previousRequestCtxTime.IsZero() {
+		t.Fatalf("request ctx time after init should be zero, instead %v", previousRequestCtxTime)
+	}
+
+	requestCtxTime := requestCtx.Time()
+
+	if requestCtxTime != now {
+		t.Fatalf("request ctx time should be 2025/06/20, instead %v", requestCtxTime)
+	}
+}
+
+func TestRequestCtxSwapConnTime(t *testing.T) {
+	var r Request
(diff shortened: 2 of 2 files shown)


erikdubbelboer:
Is it possible to show how you use this in your tests?

peczenyj:
sure, 

I was not aware of this https://github.com/valyala/fasthttp/pull/379 (and this is quite old) so I was considering that `fctx.Time()` could be faster than `time.Now()`. And it is, but in one scenario where I need the current time several times, and perhaps this is not a big deal, I must check with proper benchmark.

one use case is, consider the following handler:

```go
package foo

import (
	"net/http"
	"time"

	"github.com/valyala/fasthttp"
)

type Clock interface {
	Since(t time.Time) time.Duration
}

type Application struct {
	Clock Clock
}

func New(c Clock) *Application {
	return &Application{Clock: c}
}

func (a *Application) Handler(fctx *fasthttp.RequestCtx) {
	start := fctx.Time()

	// computation

	elapsed := a.Clock.Since(start)

	fctx.SetStatusCode(http.StatusOK)

	// here we set in headers, but in reality it can be metrics
	fctx.Response.Header.Add("X-Start", start.String())
	fctx.Response.Header.Add("X-Elapsed", elapsed.String())

	// this is real code
	fctx.Response.Header.Set(
		fasthttp.HeaderLastModified, 
		fctx.Time().UTC().Format(http.TimeFormat))
}
```

I have some tests like

```go
package foo_test

import (
	"net"
	"net/http"
	"testing"
	"time"

	"github.com/valyala/fasthttp"

	"example.com/go/foo"
)

const (
	defaultTestPort   = 8080
	defaultRemoteIP   = "8.8.8.8"
	defaultUserAgent  = "Chrome"
	defaultRequestURI = "/endpoint"
	defaultMethod     = http.MethodGet
	defaultHTTPStatus = http.StatusContinue
)

func TestHandler(t *testing.T) {
	t.Parallel()

	var fctx fasthttp.RequestCtx

	setup(&fctx)

	foo.New(clock{}).Handler(&fctx)

	if fctx.Response.StatusCode() != http.StatusOK {
		t.Fatalf("unexpected response status code (got %v)", fctx.Response.StatusCode())
	}

	elapsed := fctx.Response.Header.Peek(`X-Elapsed`)

	if string(elapsed) != "1s" {
		t.Fatalf("unexpected x-elapsed header")
	}

	t.Logf("response headers: %#v", fctx.Response.Header.String())
}

func setup(fctx *fasthttp.RequestCtx) {
	var addr net.TCPAddr

	addr.IP = net.ParseIP(defaultRemoteIP).To4()
	addr.Port = defaultTestPort

	var r fasthttp.Request

	r.SetRequestURI(defaultRequestURI)
	r.Header.SetUserAgent(defaultUserAgent)
	r.Header.SetMethod(defaultMethod)

	fctx.Init(&r, &addr, nil)
}

type clock struct{}

func (clock) Since(time.Time) time.Duration { return time.Second }
```

as you can see, I _still_ need to inject a "clock" object if I want to ensure some delta of time, and on this case the absolute value of `fctx.Time()` is irrelevant. BUT if I need to do something with the `fctx.Time()` then I need to _assume_ that this value will be zero time.Time. As you can see I can assert the value of the elapsed time, but not the `X-Start` (or Last Modify) because this is an internal detail of how fasthttp works.

However, if for some reason I need to use `fctx.ConnTime()` then it is no longer zero time.Time, but `time.Now()` and I have even less control over this.

In summary: I think I am using `fctx.Time()` more than I really need, but in some cases it seems acceptable and I am not confortable with consider it being zero value. I would prefer have some control.

erikdubbelboer:
I'm always very cautious with adding new methods as the API is already much too big. So I'm wondering, won't the new [synctest](https://pkg.go.dev/testing/synctest) in Go 1.25 make this not needed anymore?

peczenyj:
yep I agree, lets close it for now. Thanks

PR #2040, opened 2025-07-24
Avoid allocations in Method helpers

gaby:
- Avoid string allocation by comparing bytes directly


gaby:
Turns out the compiler takes care of this already. Closing

PR #2046, opened 2025-08-02
Add reuseport support for Solaris

jwntree:
Copied from reuseport_aix.go to reuseport_solaris.go

SO_REUSEPORT is works on Solaris, but fails on illumos as intended.
I choose to avoid using tcplisten, because the unix.Listen from golang.org/x/sys/unix is currently broken.

### DIFF
--- a/reuseport/reuseport_solaris.go
+++ b/reuseport/reuseport_solaris.go
@@ -1,0 +1,31 @@
+//go:build solaris
+
+package reuseport
+
+import (
+	"context"
+	"net"
+	"syscall"
+
+	"golang.org/x/sys/unix"
+)
+
+const (
+	_SO_REUSEPORT = 0x100e
+)
+
+var listenConfig = net.ListenConfig{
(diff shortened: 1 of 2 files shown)


reuseport/reuseport_solaris.go
```diff
@@ -0,0 +1,32 @@
...
+const (
+	_SO_REUSEPORT = 0x100e
+)
+
+var listenConfig = net.ListenConfig{
+	Control: func(network, address string, c syscall.RawConn) (err error) {
+		return c.Control(func(fd uintptr) {
+			err = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEADDR, 1)
+			if err == nil {
+				//err = unix.SetsockoptInt(int(fd), unix.SOL_SOCKET, unix.SO_REUSEPORT, 1)
```
erikdubbelboer:
```suggestion
```

erikdubbelboer:
Thanks!

Issue #1765, opened 2024-04-23
It is not safe to read all stream body to memery without max size limit


Quons:
https://github.com/valyala/fasthttp/blob/57b9352ad1cc93a0aaaa72b2130e03ace8a5b118/http.go#L427

byte0o:
If you want to add a read data size limit to the Body function, can you abandon the Body function and only support BodyStream? BodyStream can satisfy the caller's control of the read data size.

gaby:
@byte0o That would be a backward incompatible change, right?

byte0o:
@gab Yes, you can first mark the Body function as deprecated

erikdubbelboer:
@dojutsu-user maybe you can have a look at changing it so that `Request.Body()` prints a warning using [Server.Logger](https://pkg.go.dev/github.com/valyala/fasthttp#Server.Logger) (so only when used in a server context).

dojutsu-user:
@erikdubbelboer We'll have to inject the Server.Logger instance in the request, right?
Currently, there's no logger that I can fine in the req.

erikdubbelboer:
Yeah but keep it private, we don't want to expose that.

PR #2070, opened 2025-09-18
Add Body() warnings for streaming bodies in server context

aabishkaryal:
Addresses issue #1765 by logging warnings when Request.Body() or Response.Body() methods read from bodyStream in server context, which can cause OOM attacks by reading unlimited stream data into memory via copyZeroAlloc().

Changes:
- Add private serverLogger field to Request and Response structs
- Inject server logger reference in Server.acquireCtx() for context detection
- Add warnings in Response.Body() and Request.bodyBytes() when bodyStream != nil
- Clear logger references in Reset() methods for proper cleanup
- Add TestServerBodyWarning test to verify streaming body warnings

The warnings only appear when:
1. Body() is called in server context (serverLogger reference exists)
2. Body() attempts to read entire stream into memory (bodyStream != nil + copyZeroAlloc)

Normal body usage and client-side calls remain unaffected. This encourages safer BodyStream() usage without breaking compatibility.

### DIFF
--- a/http.go
+++ b/http.go
@@ -81,6 +81,10 @@ type Request struct {
 	// By default redirect path values are normalized, i.e.
 	// extra slashes are removed, special characters are encoded.
 	DisableRedirectPathNormalizing bool
+
+	// serverLogger is used internally to log warnings when Body() is called
+	// in server context. This is set by the server when processing requests.
+	serverLogger Logger
 }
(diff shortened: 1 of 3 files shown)


gaby:
This shouldn't be added, the docs should be updated instead.

PR #2072, opened 2025-09-21
Add DNS cache management methods for TCPDialer

### LINKED ISSUE #2066: Force Fresh DNS Lookups
Is it possible to provide a method to force refresh the dns cache?
I want to call the `tcpAddrsClean` method in the `tcpdialer.go` file once if I need, so I can set the cache time to be longer.

aabishkaryal:
Resolves #2066

This commit introduces new methods for managing DNS cache in TCPDialer:

1. FlushDNSCache() - Clears all cached DNS entries, forcing fresh lookups

Key changes:
- Add FlushDNSCache() method to TCPDialer
- Add global FlushDNSCache() function for default dialer
- Refactor tcpAddrsClean() to extract reusable cleanExpiredDNSEntries() method
- Add comprehensive tests with mock resolver to verify caching behavior

Use case: Users can now set longer cache durations (e.g., 30 minutes) and manually refresh DNS when needed, providing better control over DNS resolution timing while maintaining performance benefits of caching.

### DIFF
--- a/tcpdialer.go
+++ b/tcpdialer.go
@@ -271,6 +271,36 @@ func (d *TCPDialer) DialDualStackTimeout(addr string, timeout time.Duration) (ne
 	return d.dial(addr, true, timeout)
 }
 
+// FlushDNSCache clears all cached DNS entries, forcing fresh DNS lookups on subsequent dials.
+// This is useful when you want to ensure fresh DNS resolution, for example after network changes.
+func (d *TCPDialer) FlushDNSCache() {
+	d.tcpAddrsMap.Range(func(k, v any) bool {
+		d.tcpAddrsMap.Delete(k)
+		return true
+	})
+}
+
+// CleanDNSCache removes expired DNS cache entries based on DNSCacheDuration.
+// This is useful when you have set a longer DNSCacheDuration and want to manually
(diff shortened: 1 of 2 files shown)


tcpdialer.go
```diff
@@ -271,6 +271,36 @@ func (d *TCPDialer) DialDualStackTimeout(addr string, timeout time.Duration) (ne
...
+	d.tcpAddrsMap.Range(func(k, v any) bool {
+		d.tcpAddrsMap.Delete(k)
+		return true
+	})
+}
+
+// CleanDNSCache removes expired DNS cache entries based on DNSCacheDuration.
+// This is useful when you have set a longer DNSCacheDuration and want to manually
+// trigger cleanup of expired entries without waiting for the automatic cleanup.
+func (d *TCPDialer) CleanDNSCache() {
```
erikdubbelboer: Lets not expose this method. Expired DNS entries are already not used anymore. The only thing this method does is clean up some unused memory a bit earlier. I don't think that's worth increasing the API surface for.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2082, opened 2025-09-30
Reduce allocations in copyTrailer (saves ~43% time)

arturmelanchyk:
```bash
 benchstat old.txt new.txt 
goos: darwin
goarch: arm64
cpu: Apple M3 Max
               │   old.txt   │               new.txt               │
               │   sec/op    │   sec/op     vs base                │
CopyTrailer-16   260.6n ± 1%   147.5n ± 0%  -43.42% (p=0.000 n=20)

               │   old.txt    │               new.txt               │
               │     B/op     │     B/op      vs base               │
CopyTrailer-16   1.109Ki ± 0%   1.125Ki ± 0%  +1.41% (p=0.000 n=20)

               │   old.txt   │              new.txt               │
               │  allocs/op  │ allocs/op   vs base                │
CopyTrailer-16   16.000 ± 0%   1.000 ± 0%  -93.75% (p=0.000 n=20)
```

### DIFF
--- a/header.go
+++ b/header.go
@@ -3359,9 +3359,18 @@ func copyTrailer(dst, src [][]byte) [][]byte {
 		dst = append(dst[:0], src...)
 	}
 
+	var totalSrcLen int
 	for i := range dst {
-		dst[i] = make([]byte, len(src[i]))
-		copy(dst[i], src[i])
+		totalSrcLen += len(src[i])
+	}
+
+	// allocating single buffer for all trailer values
+	buf := make([]byte, totalSrcLen)
+	bufPos := 0
+	for i := range dst {
+		n := copy(buf[bufPos:], src[i])
+		dst[i] = buf[bufPos : bufPos+n : bufPos+n]
+		bufPos += n
 	}
 	return dst
 }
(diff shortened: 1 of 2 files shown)


arturmelanchyk:
Hi @erikdubbelboer 
the `dst` resizing logic in the beginning of the function was not affected by this change, `dst` can still be reused.
Important thing to note here is that `dst` is a slice of a byte slice.
The suggested change is not related to `dst` directly but rather to each `dst[i]` entry, that previously performed allocation on every iteration:
```go
	for i := range dst {
		dst[i] = make([]byte, len(src[i]))
		copy(dst[i], src[i])
```

Now it does exactly the same but allocates once and then inside a `for` loop it safely reslice by setting `cap`:
```go
    buf := make([]byte, totalSrcLen)
	bufPos := 0
	for i := range dst {
		n := copy(buf[bufPos:], src[i])
		dst[i] = buf[bufPos : bufPos+n : bufPos+n]
```

can we give it a try and run the tests?

erikdubbelboer:
Sorry I meant it can better like this: https://github.com/valyala/fasthttp/commit/19624506292b2f05ed08774d1f6c814340efb409
This way it reuses the memory instead of making a new slice every time.

PR #2096, opened 2025-11-10
Expose header parsing error variables

ReneWerner87:
# Summary

- expose exported error variables for header parsing to replace inline error constructions
- update header parsing logic and related utilities to rely on the exported errors
- adjust server tests to reference the new exported error values

### DIFF
--- a/header.go
+++ b/header.go
@@ -463,7 +463,22 @@ func (h *header) AddTrailer(trailer string) error {
 	return h.AddTrailerBytes(s2b(trailer))
 }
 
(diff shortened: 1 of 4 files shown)


gaby:
👍 for this

erikdubbelboer:
Thanks!

ReneWerner87:
Just wanted to prevent such situations, not that you change the error messages and we are then outdated.

https://github.com/gofiber/fiber/blob/19cc3f56f09fa758fc6863c67c3747d6639029c2/app.go#L1327

Issue #1699, opened 2024-01-20
question: Add user level value to connection?


stevedw:
Would it be possible to add an exposed user level value (something like a uint64) to the underlying connection on the request context.

This would make adding stuff like rate limit much easier without having to maintain another book keeping data structure keyed by conn id and then having to deal with connection timeouts and book-keeping etc in the user code.

Ideally it would be useful to have it as a direct value so the caller can decide on atomic or not.

Obviously, one could do it with user level attributes, but often the dos spam type stuff is just brute forcing at a conn level and rates are useful to apply at this level prior to identifying a user identity.

erikdubbelboer:
A pull request would be welcome!

PR #2100, opened 2025-11-20
 Issue #1699

Dostonlv:
Implement connection-level user value management in RequestCtx

- Added methods to set and retrieve connection-level user values, including support for uint64 values.
- Implemented reset functionality to clear connection user values.
- Introduced a new example demonstrating rate limiting using connection-level user values.

### DIFF
--- a/examples/conn_user_value/main.go
+++ b/examples/conn_user_value/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"sync/atomic"
+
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Sorry but I don't think expanding the API surface for this is a good idea. Unless you can show me some really good use case that can't be covered by using [`RequestCtx.ConnID()`](https://pkg.go.dev/github.com/valyala/fasthttp#RequestCtx.ConnID) and a `map`.

PR #2104, opened 2025-12-05
Add ConvertNetHttpRequestToFastHttpRequest adaptor function

aaydin-tr:
One of my personal project i need to convert net http request to fasthttp request instead of doing it directly in my own project i want to add this functionality to fasthttp for possible future uses and  i also looking for a review if its correct way to convert both request object.

### DIFF
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -96,11 +96,7 @@ func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.Reque
 	}
 
 	if r.Body != nil {
-		body, err := io.ReadAll(r.Body)
-		if err != nil {
-			return err
-		}
-		ctx.Request.SetBody(body)
+		ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))
 	}
 
 	if r.RemoteAddr != "" {
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -71,8 +71,23 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 	return nil
 }
 
-// ConvertNetHttpToFastHttp converts an http.Request to a fasthttp.RequestCtx.
-// The caller is responsible for the lifecycle of the fasthttp.RequestCtx.
+// ConvertNetHttpRequestToFastHttpRequest converts an http.Request to a fasthttp.RequestCtx.
+//
+// The caller is responsible for the lifecycle of the fasthttp.RequestCtx and the
+// underlying fasthttp.Request. The ctx (and its Request) must only be used for
+// the duration that fasthttp considers it valid (typically within a handler),
+// and must not be accessed after the handler has returned.
+//
+// The request body is not copied. If r.Body is non-nil, it is passed directly to
+// ctx.Request via SetBodyStream. This means:
+//   - r.Body must remain readable for as long as ctx may need to read it.
+//   - r.Body should not be read from, written to, or closed by the caller until
+//     fasthttp is done with ctx.
+//   - The same r.Body must not be reused concurrently in other goroutines while
+//     it is attached to ctx.Request.
+//
+// After calling this function, you should treat r.Body as effectively owned by
+// ctx.Request for the lifetime of that context.
 func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.RequestCtx) {
 	ctx.Request.Header.SetMethod(r.Method)
 
@@ -103,18 +118,21 @@ func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.Reque
 		addr := parseRemoteAddr(r.RemoteAddr)
 		ctx.SetRemoteAddr(addr)
 	}
-
 }
 
 func parseRemoteAddr(addr string) net.Addr {
-	host, port, err := net.SplitHostPort(addr)
-	if err != nil {
-		return &net.TCPAddr{IP: net.ParseIP(addr)}
+	if tcpAddr, err := net.ResolveTCPAddr("tcp", addr); err == nil {
+		return tcpAddr
 	}
-	return &net.TCPAddr{
-		IP:   net.ParseIP(host),
-		Port: parsePort(port),
+
+	if _, _, err := net.SplitHostPort(addr); err != nil {
+		if tcpAddr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(addr, "0")); err == nil {
+			return tcpAddr
+		}
 	}
+
+	host := strings.Trim(addr, "[]")
+	return &net.TCPAddr{IP: net.ParseIP(host)}
 }
 
 func parsePort(port string) int {
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -73,7 +73,7 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 
 // ConvertNetHttpToFastHttp converts an http.Request to a fasthttp.RequestCtx.
 // The caller is responsible for the lifecycle of the fasthttp.RequestCtx.
-func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.RequestCtx) error {
+func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.RequestCtx) {
 	ctx.Request.Header.SetMethod(r.Method)
 
 	if r.RequestURI != "" {
@@ -104,7 +104,6 @@ func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.Reque
 		ctx.SetRemoteAddr(addr)
 	}
 
-	return nil
 }
 
 func parseRemoteAddr(addr string) net.Addr {
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -3,8 +3,10 @@ package fasthttpadaptor
 import (
 	"bytes"
 	"io"
+	"net"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 
 	"github.com/valyala/fasthttp"
@@ -67,4 +69,63 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 	}
 
 	return nil
+}
+
+// ConvertNetHttpToFastHttp converts an http.Request to a fasthttp.RequestCtx.
+// The caller is responsible for the lifecycle of the fasthttp.RequestCtx.
+func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.RequestCtx) error {
+	ctx.Request.Header.SetMethod(r.Method)
+
+	if r.RequestURI != "" {
+		ctx.Request.SetRequestURI(r.RequestURI)
+	} else if r.URL != nil {
+		ctx.Request.SetRequestURI(r.URL.RequestURI())
+	}
+
+	ctx.Request.Header.SetProtocol(r.Proto)
+	ctx.Request.SetHost(r.Host)
+
+	for k, values := range r.Header {
+		for i, v := range values {
+			if i == 0 {
+				ctx.Request.Header.Set(k, v)
+			} else {
+				ctx.Request.Header.Add(k, v)
+			}
+		}
+	}
+
+	if r.Body != nil {
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			return err
+		}
+		ctx.Request.SetBody(body)
+	}
+
+	if r.RemoteAddr != "" {
+		addr := parseRemoteAddr(r.RemoteAddr)
+		ctx.SetRemoteAddr(addr)
+	}
+
+	return nil
+}
+
+func parseRemoteAddr(addr string) net.Addr {
+	host, port, err := net.SplitHostPort(addr)
+	if err != nil {
+		return &net.TCPAddr{IP: net.ParseIP(addr)}
+	}
+	return &net.TCPAddr{
+		IP:   net.ParseIP(host),
+		Port: parsePort(port),
+	}
+}
+
+func parsePort(port string) int {
+	p, err := strconv.Atoi(port)
+	if err != nil {
+		return 0
+	}
+	return p
 }
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -3,10 +3,10 @@ package fasthttpadaptor
 import (
 	"bytes"
 	"io"
+	"math"
 	"net"
 	"net/http"
 	"net/url"
-	"strconv"
 	"strings"
 
 	"github.com/valyala/fasthttp"
@@ -111,7 +111,12 @@ func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.Reque
 	}
 
 	if r.Body != nil {
-		ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))
+		contentLength := int(r.ContentLength)
+		if r.ContentLength >= int64(math.MaxInt) {
+			contentLength = -1
+		}
+
+		ctx.Request.SetBodyStream(r.Body, contentLength)
 	}
 
 	if r.RemoteAddr != "" {
@@ -133,12 +138,4 @@ func parseRemoteAddr(addr string) net.Addr {
 
 	host := strings.Trim(addr, "[]")
 	return &net.TCPAddr{IP: net.ParseIP(host)}
-}
-
-func parsePort(port string) int {
-	p, err := strconv.Atoi(port)
-	if err != nil {
-		return 0
-	}
-	return p
 }
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -71,7 +71,7 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 	return nil
 }
 
-// ConvertNetHttpRequestToFastHttpRequest converts an http.Request to a fasthttp.RequestCtx.
+// ConvertNetHTTPRequestToFastHTTPRequest converts an http.Request to a fasthttp.RequestCtx.
 //
 // The caller is responsible for the lifecycle of the fasthttp.RequestCtx and the
 // underlying fasthttp.Request. The ctx (and its Request) must only be used for
@@ -88,7 +88,7 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 //
 // After calling this function, you should treat r.Body as effectively owned by
 // ctx.Request for the lifetime of that context.
-func ConvertNetHttpRequestToFastHttpRequest(r *http.Request, ctx *fasthttp.RequestCtx) {
+func ConvertNetHTTPRequestToFastHTTPRequest(r *http.Request, ctx *fasthttp.RequestCtx) {
 	ctx.Request.Header.SetMethod(r.Method)
 
 	if r.RequestURI != "" {
--- a/fasthttpadaptor/request_test.go
+++ b/fasthttpadaptor/request_test.go
@@ -174,12 +174,13 @@ func TestConvertNetHttpRequestToFastHttpRequest(t *testing.T) {
 		t.Parallel()
 		bodyContent := []byte("test body content")
 		httpReq := &http.Request{
-			Method:     "POST",
-			RequestURI: "/",
-			Proto:      "HTTP/1.1",
-			Host:       "example.com",
-			Header:     http.Header{},
-			Body:       io.NopCloser(bytes.NewReader(bodyContent)),
+			Method:        "POST",
+			RequestURI:    "/",
+			Proto:         "HTTP/1.1",
+			Host:          "example.com",
+			Header:        http.Header{},
+			Body:          io.NopCloser(bytes.NewReader(bodyContent)),
+			ContentLength: int64(len(bodyContent)),
 		}
 
 		ctx := &fasthttp.RequestCtx{}
@@ -264,18 +265,24 @@ func TestConvertNetHttpRequestToFastHttpRequest(t *testing.T) {
 	t.Run("body read error", func(t *testing.T) {
 		t.Parallel()
 		httpReq := &http.Request{
-			Method:     "POST",
-			RequestURI: "/",
-			Proto:      "HTTP/1.1",
-			Host:       "example.com",
-			Header:     http.Header{},
-			Body:       io.NopCloser(errReader{}),
+			Method:        "POST",
+			RequestURI:    "/",
+			Proto:         "HTTP/1.1",
+			Host:          "example.com",
+			Header:        http.Header{},
+			Body:          io.NopCloser(errReader{}),
+			ContentLength: 10,
 		}
 
 		ctx := &fasthttp.RequestCtx{}
 		err := ConvertNetHttpRequestToFastHttpRequest(httpReq, ctx)
+		if err != nil {
+			t.Fatalf("unexpected error during conversion: %v", err)
+		}
+
+		_, err = io.ReadAll(ctx.RequestBodyStream())
 		if err == nil {
-			t.Fatal("expected error, got nil")
+			t.Fatal("expected error when reading body stream, got nil")
 		}
 	})
 }
--- a/fasthttpadaptor/request_test.go
+++ b/fasthttpadaptor/request_test.go
@@ -238,6 +238,106 @@ func TestConvertNetHttpRequestToFastHttpRequest(t *testing.T) {
 		}
 	})
 
+	t.Run("IPv6 remote address with port", func(t *testing.T) {
+		t.Parallel()
+		httpReq := &http.Request{
+			Method:     "GET",
+			RequestURI: "/",
+			Proto:      "HTTP/1.1",
+			Host:       "example.com",
+			Header:     http.Header{},
+			RemoteAddr: "[2001:db8::1]:8080",
+		}
+
+		ctx := &fasthttp.RequestCtx{}
+		ConvertNetHttpRequestToFastHttpRequest(httpReq, ctx)
+
+		remoteAddr := ctx.RemoteAddr().String()
+		if remoteAddr != "[2001:db8::1]:8080" {
+			t.Errorf("expected remote addr [2001:db8::1]:8080, got %s", remoteAddr)
+		}
+	})
+
+	t.Run("IPv6 remote address without port", func(t *testing.T) {
+		t.Parallel()
+		httpReq := &http.Request{
+			Method:     "GET",
+			RequestURI: "/",
+			Proto:      "HTTP/1.1",
+			Host:       "example.com",
+			Header:     http.Header{},
+			RemoteAddr: "2001:db8::1",
+		}
+
+		ctx := &fasthttp.RequestCtx{}
+		ConvertNetHttpRequestToFastHttpRequest(httpReq, ctx)
+
+		remoteAddr := ctx.RemoteAddr().String()
+		if remoteAddr != "[2001:db8::1]:0" {
+			t.Errorf("expected remote addr [2001:db8::1]:0, got %s", remoteAddr)
+		}
+	})
+
+	t.Run("IPv6 remote address with zone and port", func(t *testing.T) {
+		t.Parallel()
+		httpReq := &http.Request{
+			Method:     "GET",
+			RequestURI: "/",
+			Proto:      "HTTP/1.1",
+			Host:       "example.com",
+			Header:     http.Header{},
+			RemoteAddr: "[fe80::1%eth0]:9090",
+		}
+
+		ctx := &fasthttp.RequestCtx{}
+		ConvertNetHttpRequestToFastHttpRequest(httpReq, ctx)
+
+		remoteAddr := ctx.RemoteAddr().String()
+		if remoteAddr != "[fe80::1%eth0]:9090" {
+			t.Errorf("expected remote addr [fe80::1%%eth0]:9090, got %s", remoteAddr)
+		}
+	})
+
+	t.Run("IPv6 remote address with zone without port", func(t *testing.T) {
+		t.Parallel()
+		httpReq := &http.Request{
+			Method:     "GET",
+			RequestURI: "/",
+			Proto:      "HTTP/1.1",
+			Host:       "example.com",
+			Header:     http.Header{},
+			RemoteAddr: "fe80::1%eth0",
+		}
+
+		ctx := &fasthttp.RequestCtx{}
+		ConvertNetHttpRequestToFastHttpRequest(httpReq, ctx)
+
+		remoteAddr := ctx.RemoteAddr().String()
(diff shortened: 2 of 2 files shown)


fasthttpadaptor/request.go
```diff
@@ -68,3 +70,62 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+			if i == 0 {
+				ctx.Request.Header.Set(k, v)
+			} else {
+				ctx.Request.Header.Add(k, v)
+			}
+		}
+	}
+
+	if r.Body != nil {
+		body, err := io.ReadAll(r.Body)
```
gaby: This will be an issue for big files, it also needs to support streaming.

aaydin-tr: now i use SetBodyStream for avoid unnecessary memory allocation and now also support streaming i guess

erikdubbelboer:
I'm sorry I don't have the time to look into this closely. On an initial glance the function seems kinda small, are you sure it supports all fields and cases (such as the various way a fasthttp request can have a body)?

aaydin-tr:
> I'm sorry I don't have the time to look into this closely. On an initial glance the function seems kinda small, are you sure it supports all fields and cases (such as the various way a fasthttp request can have a body)?

```go
if r.Body != nil {
	ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))
}
```

this is how i pass net/http request body to fasthttp request body. indeed fasthttp request has various ways to set the body, but as far as i see, `ctx.Request.Body()` and all other functions to get the request body fall through to the `req.bodyBytes()` function, and it reads req.bodyStream if it’s not nil. based on that, i guess `ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))` will handle all cases. but i am not experienced with fasthttp, so if you have an edge case in mind, i can try to convert that too.

gaby:
@erikdubbelboer Can you trigger copilot to also check this?

fasthttpadaptor/request.go
```diff
@@ -68,3 +70,75 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+		if tcpAddr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort(addr, "0")); err == nil {
+			return tcpAddr
+		}
+	}
+
+	host := strings.Trim(addr, "[]")
+	return &net.TCPAddr{IP: net.ParseIP(host)}
+}
+
+func parsePort(port string) int {
```
gaby: This function is unused.

gaby:
> > I'm sorry I don't have the time to look into this closely. On an initial glance the function seems kinda small, are you sure it supports all fields and cases (such as the various way a fasthttp request can have a body)?
> 
> ```go
> if r.Body != nil {
> 	ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))
> }
> ```
> 
> this is how i pass net/http request body to fasthttp request body. indeed fasthttp request has various ways to set the body, but as far as i see, `ctx.Request.Body()` and all other functions to get the request body fall through to the `req.bodyBytes()` function, and it reads req.bodyStream if it’s not nil. based on that, i guess `ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))` will handle all cases. but i am not experienced with fasthttp, so if you have an edge case in mind, i can try to convert that too.

Body Streams larger than 4GB will fail with this, since those need an `uint64` instead of `int`.

aaydin-tr:
> > > I'm sorry I don't have the time to look into this closely. On an initial glance the function seems kinda small, are you sure it supports all fields and cases (such as the various way a fasthttp request can have a body)?
> > 
> > 
> > ```go
> > if r.Body != nil {
> > 	ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this is how i pass net/http request body to fasthttp request body. indeed fasthttp request has various ways to set the body, but as far as i see, `ctx.Request.Body()` and all other functions to get the request body fall through to the `req.bodyBytes()` function, and it reads req.bodyStream if it’s not nil. based on that, i guess `ctx.Request.SetBodyStream(r.Body, int(r.ContentLength))` will handle all cases. but i am not experienced with fasthttp, so if you have an edge case in mind, i can try to convert that too.
> 
> Body Streams larger than 4GB will fail with this, since those need an `uint64` instead of `int`.

```go
	if r.Body != nil {
		contentLength := int(r.ContentLength)
		if r.ContentLength >= int64(math.MaxInt) {
			contentLength = -1
		}

		ctx.Request.SetBodyStream(r.Body, contentLength)
	}
```

i make this change if `r.ContentLength` is larger than `math.MaxInt` contentLength is set `-1` which means unknown and body stream will read until its see EOF according to docs

gaby:
@copilot review

aaydin-tr:
@erikdubbelboer is there any update for this pr

fasthttpadaptor/request.go
```diff
@@ -68,3 +70,72 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	}
+
+	if r.RemoteAddr != "" {
+		addr := parseRemoteAddr(r.RemoteAddr)
+		ctx.SetRemoteAddr(addr)
+	}
+}
+
+func parseRemoteAddr(addr string) net.Addr {
+	if tcpAddr, err := net.ResolveTCPAddr("tcp", addr); err == nil {
```
erikdubbelboer: There is no timeout on this and it could potentially hang. We have this code here, I'm not sure if it should somehow be reused since it's not exposed: https://github.com/valyala/fasthttp/blob/3471acf23f1a39e83acd13beec99cd1e809762b1/tcpdialer.go#L479

fasthttpadaptor/request.go
```diff
@@ -68,3 +70,72 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+// ctx.Request via SetBodyStream. This means:
+//   - r.Body must remain readable for as long as ctx may need to read it.
+//   - r.Body should not be read from, written to, or closed by the caller until
+//     fasthttp is done with ctx.
+//   - The same r.Body must not be reused concurrently in other goroutines while
+//     it is attached to ctx.Request.
+//
+// After calling this function, you should treat r.Body as effectively owned by
+// ctx.Request for the lifetime of that context.
+func ConvertNetHTTPRequestToFastHTTPRequest(r *http.Request, ctx *fasthttp.RequestCtx) {
```
erikdubbelboer: The function doesn't reset `ctx.Request`. This should be documented that users need to do this themselves if needed.

fasthttpadaptor/request.go
```diff
@@ -68,3 +70,72 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	ctx.Request.Header.SetMethod(r.Method)
+
+	if r.RequestURI != "" {
+		ctx.Request.SetRequestURI(r.RequestURI)
+	} else if r.URL != nil {
+		ctx.Request.SetRequestURI(r.URL.RequestURI())
+	}
+
+	ctx.Request.Header.SetProtocol(r.Proto)
+	ctx.Request.SetHost(r.Host)
```
erikdubbelboer: `r.Host` can be empty in which case it should be `r.URL.Host`.

fasthttpadaptor/request.go
erikdubbelboer: Also why have a `fasthttp.RequestCtx` as argument? From the function name I would expect a `fasthttp.Request` as argument, which makes much more sense and makes it much more usable.

fasthttpadaptor/request.go
```diff
@@ -68,3 +70,72 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+		}
+
+		ctx.Request.SetBodyStream(r.Body, contentLength)
+	}
+
+	if r.RemoteAddr != "" {
+		addr := parseRemoteAddr(r.RemoteAddr)
+		ctx.SetRemoteAddr(addr)
+	}
+}
```
erikdubbelboer, line 313: You are missing some properties that aren't being copied right now. For example `r.TLS` and `r.Close`, `r.TransferEncoding`, `r.Trailer`, and `r.URL.Scheme` which are all used by `net/http`.

erikdubbelboer (changes requested):

aaydin-tr:
@erikdubbelboer sorry for the very late update, i couldn't find time to work on this earlier. i pushed the changes, all review points should be addressed now:

- replaced `net.ResolveTCPAddr` with `netip.ParseAddrPort`, it only parses the string so there is no DNS lookup and nothing can block. net/http always sets RemoteAddr as "IP:port" anyway, non IP values are just ignored. so i think we don't need `resolveTCPAddrs` from tcpdialer.go here
- documented that ctx.Request is not reset, users should call `ctx.Request.Reset()` themselves when reusing a ctx
- empty r.Host falls back to `r.URL.Host`
- r.Close, r.URL.Scheme, r.TLS and r.Trailer are converted now. r.TransferEncoding is handled through the content length, chunked requests have ContentLength == -1 which maps to SetBodyStream(body, -1) and fasthttp writes the chunked encoding itself. r.TLS can only affect the URI scheme since ctx gets the TLS state from the connection
- about the RequestCtx argument, i kept it because r.RemoteAddr lives on the ctx, there is no way to put it on a fasthttp.Request. my use case is passing the converted request to code that takes a RequestCtx. 

also fixed two things on the way: HTTP/2 Proto is normalized to HTTP/1.1 since fasthttp writes the protocol into the request line as is, and ContentLength == 0 with a non-nil body is treated as unknown length like net/http does, otherwise bodies from http.NewRequest with a plain reader would be silently dropped

fasthttpadaptor/request.go
```diff
@@ -68,3 +71,141 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	} else if r.URL != nil {
+		ctx.Request.SetRequestURI(r.URL.RequestURI())
+	}
+
+	for k, values := range r.Header {
+		if strings.EqualFold(k, fasthttp.HeaderHost) {
+			continue
+		}
+		for _, v := range values {
+			ctx.Request.Header.Add(k, v)
```
erikdubbelboer: `Content-Length`, `Transfer-Encoding`, and `Trailer` are controlled by the dedicated fields on `net/http.Request`; `Request.Write` excludes those names from `Header`. This loop still imports `Content-Length`, while the body block below ignores `r.TransferEncoding`.

I reproduced both mismatches on this head. A GET with `ContentLength == 0`, no body, and `Header["Content-Length"] = []string{"42"}` is written by `net/http` without a content length, but the converted request sends `Content-Length: 42` and no bytes. A POST with a four-byte body, `ContentLength == 4`, and `TransferEncoding = []string{"chunked"}` is written chunked by `net/http`, but the converted request sends a fixed length.

The first case can leave a peer waiting for bytes that will never arrive. Please skip the net/http-managed framing headers here and derive fasthttp framing from `Body`, `ContentLength`, `TransferEncoding`, and `Trailer` together.

fasthttpadaptor/request.go
```diff
@@ -68,3 +71,141 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+		}
+	}
+
+	for k, values := range r.Trailer {
+		if ctx.Request.Header.AddTrailer(k) != nil {
+			continue
+		}
+		if len(values) > 0 {
+			ctx.Request.Header.Set(k, strings.Join(values, ", "))
```
erikdubbelboer: For an incoming chunked request, `r.Trailer` initially contains only the keys with nil values; it is populated while `r.Body` is read to EOF. This snapshots the empty values before attaching that body to the fasthttp request.

I parsed a raw request ending in `X-Final: done`, converted it, and wrote the fasthttp request. After the body was drained, `r.Trailer.Get("X-Final")` was `done`, but the converted wire request contained `X-Final:` with an empty value. Please synchronize the trailer values when the body reaches EOF, likely through a body wrapper, and add a regression starting from `http.ReadRequest` rather than a pre-populated trailer map.

fasthttpadaptor/request.go
```diff
@@ -68,3 +71,141 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+			ctx.SetRemoteAddr(remoteAddr)
+		}
+	}
+
+	scheme := ""
+	if r.URL != nil {
+		scheme = r.URL.Scheme
+	}
+	if scheme == "" && r.TLS != nil {
+		scheme = "https"
```
erikdubbelboer, line 253: This changes the URI scheme, but it does not preserve the `RequestCtx` TLS state. After initializing the context with `ctx.Init` and converting a request with a non-nil TLS 1.3 state, `ctx.Request.URI().Scheme()` is `https` while `ctx.IsTLS()` is false and `ctx.TLSConnectionState()` is nil. Handler code using those methods for secure-transport or client-certificate decisions therefore sees different request state.

This means the earlier `r.TLS` / `RequestCtx` concern remains. Please either make the converted context expose the source TLS state, or keep the conversion request-only by accepting a `*fasthttp.Request` instead of a context that cannot represent all of the source state.

fasthttpadaptor/request.go
```diff
@@ -68,3 +71,141 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	}
+	if scheme != "" && scheme != "http" {
+		ctx.Request.URI().SetScheme(scheme)
+	}
+
+	if r.URL != nil && r.URL.User != nil {
+		uri := ctx.Request.URI()
+		uri.SetUsername(r.URL.User.Username())
```
erikdubbelboer: `fasthttp.Request.Write` turns URI userinfo into a Basic `Authorization` header after the explicit headers above have already been copied. With URL user/password credentials and `Authorization: Bearer explicit-token`, the converted wire request contains only the Basic credentials. `net/http.Client` gives an existing `Authorization` header precedence over URL credentials.

Please preserve that precedence; copying URL userinfo here must not overwrite an explicit `Authorization` header when the converted request is written.

aaydin-tr:
@erikdubbelboer thanks for the detailed review, i pushed the changes, all review points should be addressed now:

- the function now takes a `*fasthttp.Request` instead of a `*fasthttp.RequestCtx`. you were right that a ctx cannot represent the source TLS state, `IsTLS()` and `TLSConnectionState()` are derived from the underlying connection and there is no way to set them from outside, so the conversion is request-only now. a fasthttp.Request models only the wire message, r.RemoteAddr and the TLS connection state are out of scope, and r.TLS is only used to derive the URI scheme. callers that have a ctx can convert into `&ctx.Request` and set the remote address themselves via `ctx.SetRemoteAddr`
- the header loop now skips Content-Length, Transfer-Encoding and Trailer, the same set net/http excludes with reqWriteExcludeHeader, and framing is derived from Body, ContentLength, TransferEncoding and Trailer together: chunked TransferEncoding or trailers force SetBodyStream(body, -1). both of your repros are wire-level regression tests now. one deliberate difference to net/http: with a known ContentLength plus trailers net/http drops the trailers, i force chunked instead, because HTTP/2 requests carry a known length and trailers at the same time and dropping them on conversion would silently lose data
- trailer values are synchronized through a body wrapper when the body reaches EOF. net/http fills r.Trailer right before a body read returns io.EOF (both the HTTP/1.1 body and the HTTP/2 server do this), and fasthttp writes trailers only after draining the body stream, so the synced values are exactly what goes on the wire. the wrapper forwards Close so the body is still closed after writing, and it only syncs names AddTrailer accepted, so a forbidden name can never reach Header.Set. the regression test starts from http.ReadRequest with a raw chunked request ending in `X-Final: done`, as you asked
- URL userinfo is copied to the URI only when r.Header has no Authorization entry, the same precedence net/http.Client applies in send(). an explicit `Authorization: Bearer` header now survives the write unchanged, and plain URL credentials still produce the same Basic header net/http would send

erikdubbelboer:
Thanks, I re-reviewed the latest changes. The request-only API and the original framing, trailer sync, and Authorization cases are fixed.

I still found a few cases that need to be handled:

- An absolute-form request still loses an explicit Authorization header. Starting from `http.ReadRequest` with `GET https://url-user:url-pass@example.com/ HTTP/1.1` and `Authorization: Bearer explicit-token`, the converted request is written with the URL's Basic credentials. Calling `req.URI()` parses the userinfo from `RequestURI` before the later check skips copying `r.URL.User`, so the parsed credentials still overwrite the header.

- A body longer than `ContentLength` is written past the declared boundary before `Request.Write` returns its mismatch error. With `ContentLength == 4` and a body starting with `dataGET /smuggled ...`, the second request already reached the underlying writer. `net/http` limits the wire body to the declared length; this adaptor needs to do the same.

- Trailer sync only works while the body stays streamed. If code calls `req.Body()` after conversion and then writes the request, the output switches to a fixed `Content-Length`, keeps `Trailer: X-Final`, and never sends the final `X-Final` value.

- Unknown bodies are always made chunked. This turns an empty non-nil GET body into a chunked request even though `net/http` suppresses it, and it emits chunked transfer encoding for HTTP/1.0.

- A body without `Content-Type` is written with `Content-Type: application/octet-stream`, so the conversion adds a header that wasn't present on the source request.

I reproduced all of these against `37b8a49`. The full tests, the race test for `fasthttpadaptor`, vet, and a clean merge onto current master pass, but these cases aren't covered yet. Please address them and add wire-level regressions.

aaydin-tr:
@erikdubbelboer thanks for the re-review, i pushed fixes for three of the five cases with wire-level regression tests. the other two i traced down to the exact code paths but i want your input before going further, details below.

- the absolute-form Authorization case is fixed. the cause was what you said: `req.URI()` parses the target lazily and `uri.parse` pulls the userinfo out of the host part ([uri.go:304-318](https://github.com/valyala/fasthttp/blob/c2db56193f8b/uri.go#L304-L318)), then `Request.Write` rebuilds a Basic Authorization header from `uri.username` unconditionally ([http.go:1631-1648](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L1631-L1648)), so my guard on `r.URL.User` never covered it. i also found a second variant: with an absolute-form `http://` target the URI was never parsed, so the raw request line went out verbatim, credentials included. now an explicit Authorization header clears the parsed userinfo and the request line is written in origin-form; without the header, URL credentials still become the same Basic header net/http.Client would send ([client.go:251-257](https://github.com/golang/go/blob/go1.26.2/src/net/http/client.go#L251-L257)). tests start from `http.ReadRequest` with both schemes.

- the body/ContentLength case is fixed. `writeBodyFixedSize` copies the whole stream and only compares sizes afterwards ([http.go:2326-2330](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L2326-L2330)), so the excess reached the writer before the mismatch error. known-size bodies are now wrapped in a reader that caps reads at the declared length and fails the write if more data follows, like net/http writing through `io.LimitReader` and reporting the mismatch after ([transfer.go:372-393](https://github.com/golang/go/blob/go1.26.2/src/net/http/transfer.go#L372-L393)). the regression uses your `ContentLength=4` / `dataGET /smuggled` shape.

- the Content-Type case is fixed. the default is added at serialization keyed on the method, not body presence ([header.go:2561-2566](https://github.com/valyala/fasthttp/blob/c2db56193f8b/header.go#L2561-L2566)), so even a bodiless POST gained the header. when `r.Header` has no Content-Type entry the conversion now sets `SetNoDefaultContentType(true)` (the same flag [fs.go:1199](https://github.com/valyala/fasthttp/blob/c2db56193f8b/fs.go#L1199) uses). explicit values are preserved as is.

On the trailer-sync point (`req.Body()` before write): I dug into this one and it turned out not to be adaptor behavior — it reproduces in fasthttp core with no adaptor involved:

```go
var req fasthttp.Request
req.Read(bufio.NewReader(strings.NewReader(
    "POST /upload HTTP/1.1\r\nHost: example.com\r\n" +
    "Transfer-Encoding: chunked\r\nTrailer: X-Final\r\n\r\n" +
    "4\r\ndata\r\n0\r\nX-Final: done\r\n\r\n")))
_ = req.Body() // materialize
req.Write(bw)  // -> Content-Length: 4, dangling "Trailer: X-Final", value lost
```

The cause is structural: `writeTrailer` has exactly two call sites, both in the streamed chunked paths — [`Request.writeBodyStream`](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L2056) and [`Response.writeBodyStream`](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L2112) — while the buffered path in `Request.Write` ([http.go:1665-1684](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L1665-L1684)) has no trailer handling. But `RequestHeader.AppendBytes` still writes the `Trailer:` line whenever `h.trailer` is non-empty ([header.go:2586-2588](https://github.com/valyala/fasthttp/blob/c2db56193f8b/header.go#L2586-L2588)) and excludes the trailer-named values from the header block ([header.go:2571-2584](https://github.com/valyala/fasthttp/blob/c2db56193f8b/header.go#L2571-L2584)). So any buffered body + registered trailer produces a `Trailer:` announcement whose values never appear. net/http avoids the inconsistency with its "Sanitize Trailer" step ([transfer.go:144-147](https://github.com/golang/go/blob/go1.26.2/src/net/http/transfer.go#L144-L147): `if !chunked(t.TransferEncoding) { t.Trailer = nil }`), which both the `Trailer:` header write ([transfer.go:311-325](https://github.com/golang/go/blob/go1.26.2/src/net/http/transfer.go#L311-L325)) and the value write ([transfer.go:397-399](https://github.com/golang/go/blob/go1.26.2/src/net/http/transfer.go#L397-L399)) key off — it silently drops trailers on non-chunked messages instead.

Since the adaptor has no hook into `Body()`/`Write`, I can't fix this from the adaptor side. I have a small core fix ready in `Request.Write`'s buffered path — inserted right after the hasBody block ([http.go:1671](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L1671)), before the header write at [http.go:1673](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L1673). When trailers are registered and there's a body, write it chunked and append the trailers, the same composition the streamed path already uses at [http.go:2050-2057](https://github.com/valyala/fasthttp/blob/c2db56193f8b/http.go#L2050-L2057):

```go
if hasBody && len(req.Header.trailer) > 0 {
    req.Header.SetContentLength(-1)
    if err = req.Header.Write(w); err != nil {
        return err
    }
    if err = writeBodyChunked(w, bytes.NewReader(body)); err != nil {
        return err
    }
    return req.Header.writeTrailer(w)
}
```

the trailer values are already where `writeTrailer` reads them in both scenarios (the adaptor sync fires during the `Body()` drain, and for requests read from the wire `tryReadTrailer` parses them into `h.h`, [header.go:2194-2195](https://github.com/valyala/fasthttp/blob/c2db56193f8b/header.go#L2194-L2195)). but i didn't want to touch http.go in this PR without asking first. do you want this included here, as a separate PR (`Response.Write` has the same hazard if you want both), or would you rather handle it core-side?

On the chunking of unknown bodies: both sub-cases confirmed. net/http disambiguates the empty vs unknown body with a one-byte probe bounded by a 200ms timer, only for GET/HEAD/DELETE/OPTIONS/PROPFIND/SEARCH (`shouldSendChunkedRequestBody`/`probeRequestBody`, [transfer.go:170-248](https://github.com/golang/go/blob/go1.26.2/src/net/http/transfer.go#L170-L248)): immediate EOF suppresses the body entirely, a byte gets stitched back with `io.MultiReader`, a timeout falls through to chunked. and the HTTP/1.0+chunked combination is unrepresentable in net/http because the request line is always written as `HTTP/1.1` ([request.go:673](https://github.com/golang/go/blob/go1.26.2/src/net/http/request.go#L673)), while this adaptor keeps the source protocol, so the 1.0-correct fix would be buffering an unknown-length body to a real Content-Length. porting the probe (goroutine, timer, stitched readers) plus the 1.0 buffering adds a lot of machinery to a conversion function, and the timeout path can't fully match net/http anyway since `FlushHeaders` has no equivalent inside `Request.Write`'s single pass. before i add that complexity, is the bounded-probe port what you want here, or would you prefer something simpler for the adaptor (synchronous probe, or only fixing the HTTP/1.0 framing)?

erikdubbelboer (changes requested):
Thanks for tracing these in detail. I re-reviewed `f7c3a529`. The absolute-form Authorization, body/ContentLength boundary, and missing Content-Type fixes look good.

I'm still requesting changes for the remaining cases:

- URL credentials with an empty username are dropped. For `https://:secret@example.com/`, `net/http.Client` sends `Authorization: Basic OnNlY3JldA==`. The adapter stores the password in the fasthttp URI with an empty username, but `Request.Write` only synthesizes Basic authorization when the username is non-empty, so the converted wire request has no Authorization header. Please generate the Basic header directly from `r.URL.User` while clearing URI userinfo, and add a regression.
- I agree that losing trailers after a streamed body is materialized is a fasthttp core issue. Please open a separate core PR covering both Request and Response, merge that first, then rebase this PR and add the adaptor regression here.
- For unknown bodies, please don't port the timed probe into the converter: conversion shouldn't read the body or start a potentially stranded goroutine. Keep `nil`/`http.NoBody` as the explicit empty-body representation and document that a non-nil unknown-length body is streamed. The invalid HTTP/1.0 chunked output still needs to be fixed, though: normalize to HTTP/1.1 whenever conversion requires chunked framing, with wire-level tests.

aaydin-tr:
@erikdubbelboer thanks for the review, i pushed the fixes for the two remaining adaptor cases:

- empty-username credentials: the Basic header is now generated directly from `r.URL.User` instead of relying on `Request.Write`, which only synthesizes it for a non-empty username. userinfo parsed from an absolute-form target is cleared so it can't displace the generated header. regressions assert `Basic OnNlY3JldA==` on the wire for both origin-form and absolute-form targets.

- unknown-length bodies: no probe, as requested. nil/`http.NoBody` stay the explicit empty-body representation and the docs now state a non-nil unknown-length body is always streamed. the invalid HTTP/1.0 + chunked output is fixed by normalizing to HTTP/1.1 only when chunked framing is actually needed, with wire-level tests; a 1.0 request with a known length keeps its protocol. the implied close semantics survive via `r.Close`, which the conversion already turns into an explicit `Connection: close`.

the core trailer fix is open as #2378, covering both `Request.Write` and `Response.Write` against current master. once that lands i'll rebase this PR and add the adaptor regression for the materialized-body case.

erikdubbelboer (changes requested):
Thanks for adding the materialized-body trailer regression. I verified that trailers now survive both the streamed and buffered paths with #2378 included.

I found one remaining case with CONNECT requests. For a request created with:

```go
http.NewRequest("CONNECT", "http://example.com:443", nil)
```

`r.RequestURI` is empty, so the converter uses `r.URL.RequestURI()`, which returns `/`. The converted request is written as `CONNECT / HTTP/1.1`, while net/http writes `CONNECT example.com:443 HTTP/1.1`. This breaks proxy tunneling.

Could you preserve the authority-form target for CONNECT requests and add a wire-level regression test for this case?

aaydin-tr:
> Thanks for adding the materialized-body trailer regression. I verified that trailers now survive both the streamed and buffered paths with #2378 included.
> 
> I found one remaining case with CONNECT requests. For a request created with:
> 
> ```go
> http.NewRequest("CONNECT", "http://example.com:443", nil)
> ```
> 
> `r.RequestURI` is empty, so the converter uses `r.URL.RequestURI()`, which returns `/`. The converted request is written as `CONNECT / HTTP/1.1`, while net/http writes `CONNECT example.com:443 HTTP/1.1`. This breaks proxy tunneling.
> 
> Could you preserve the authority-form target for CONNECT requests and add a wire-level regression test for this case?

Fixed. CONNECT now uses the host as the request target when the URL carries no
path, same as net/http's Request.write.

One related case came up while testing: even when the authority is already
correct, Request.Write re-derives the target from the parsed URI whenever
something parses it (scheme, r.TLS, URL credentials), which turned it into
`CONNECT /example.com:443`. I set the target again at the end of the conversion
to drop the parsed URI.

Added wire-level tests for your case, the https variant, a CONNECT read from the
wire with TLS, and one with a path to confirm origin-form is kept there.

erikdubbelboer (changes requested):
Thanks for fixing the CONNECT authority-form target. The standard outbound, TLS, read-from-wire, and path-form cases now look good.

I found two related cases that still need to be handled:

- The final CONNECT target restoration reintroduces credentials that the earlier Authorization handling removed. Starting from a request read from `CONNECT url-user:url-pass@example.com:443 HTTP/1.1`, conversion correctly generates the Basic Authorization header and clears the URI userinfo, but the final `SetRequestURI(target)` restores the raw authority. The wire request therefore contains the credentials in both the request target and the Authorization header. Please restore a sanitized target instead and add a wire-level regression for this case.

- CONNECT requests with `URL.Opaque` still differ from `net/http.Request.Write`. With `URL{Scheme: "https", Opaque: "//example.com:443"}`, net/http writes `CONNECT //example.com:443`, while the converter writes `CONNECT https://example.com:443`. An opaque target with `RawQuery` also gains the query in the converted request, while net/http uses `URL.Opaque` directly. Please mirror the net/http CONNECT branch: when the path is empty, use the host normally, but use `URL.Opaque` itself when it is non-empty, and preserve that chosen target through the later URI parsing.

I reproduced both against `9ef06f2`. The PR otherwise merges cleanly with current master, and `go test ./...`, `go test -race ./fasthttpadaptor`, `go vet ./...`, and `git diff --check` all pass.

aaydin-tr:
> Thanks for fixing the CONNECT authority-form target. The standard outbound, TLS, read-from-wire, and path-form cases now look good.
> 
> I found two related cases that still need to be handled:
> 
> * The final CONNECT target restoration reintroduces credentials that the earlier Authorization handling removed. Starting from a request read from `CONNECT url-user:url-pass@example.com:443 HTTP/1.1`, conversion correctly generates the Basic Authorization header and clears the URI userinfo, but the final `SetRequestURI(target)` restores the raw authority. The wire request therefore contains the credentials in both the request target and the Authorization header. Please restore a sanitized target instead and add a wire-level regression for this case.
> * CONNECT requests with `URL.Opaque` still differ from `net/http.Request.Write`. With `URL{Scheme: "https", Opaque: "//example.com:443"}`, net/http writes `CONNECT //example.com:443`, while the converter writes `CONNECT https://example.com:443`. An opaque target with `RawQuery` also gains the query in the converted request, while net/http uses `URL.Opaque` directly. Please mirror the net/http CONNECT branch: when the path is empty, use the host normally, but use `URL.Opaque` itself when it is non-empty, and preserve that chosen target through the later URI parsing.
> 
> I reproduced both against `9ef06f2`. The PR otherwise merges cleanly with current master, and `go test ./...`, `go test -race ./fasthttpadaptor`, `go vet ./...`, and `git diff --check` all pass.

Both fixed.

The restore now drops the userinfo first. The same leak happened with an
explicit Authorization header — the sanitizer only checks for "://" in
RequestURI, so an authority-form target never matched it.

The URL branch now mirrors the net/http CONNECT branch: Opaque when set, host
otherwise, and Opaque verbatim since Request.Write writes it as-is.

Both came from touching req.URI(), which sets parsedURI and makes Write rebuild
the request line from the parsed URI. The old string check missed an Opaque
starting with "//", so it's now one decision made where the target is picked.

That also surfaced a third case, fixed here: absolute-form targets survived for
http but became origin-form for https, since only a non-http scheme is set on
the URI. Same for non-CONNECT opaque URLs.

Tests added.

PR #2113, opened 2025-12-22
fix: allow URIs with spaces in them

srabraham:


srabraham:
sorry, I meant to just have this hacky thing in my own fork...

PR #2135, opened 2026-02-06
feat: enhance performance

ReneWerner87:
## Performance Improvements Summary

This PR introduces benchmark-driven performance optimizations in core `fasthttp` hot paths and keeps only changes with statistically significant wins (`benchstat`, `p < 0.05`) and no allocation regressions.

### 1) Header parse hot-path: remove redundant key validation work
**File:** `header.go`

#### Changes
- Added `normalizeHeaderKeyValidated(...)` for call sites where header-key validity has already been verified.
- Replaced `normalizeHeaderKey(...)` with the validated variant in:
  - request header parsing
  - response header parsing
  - trailer parsing

#### Benchmark
`BenchmarkRequestHeaderRead`:
- **Before:** `100.90 ns/op`
- **After:** `94.56 ns/op`
- **Delta:** `-6.28%` (`p=0.026`)
- `B/op` and `allocs/op` remained `0`

---

### 2) Cookie parsing: major reductions in time and allocations
**File:** `cookie.go`  
**Related helper reuse:** `bytesconv.go`

#### Changes
- Added fast RFC1123-GMT date parsing path for cookie `Expires`.
- Switched cookie date parsing flow to:
  1. `ParseHTTPDate(...)` fast path
  2. legacy cookie date fallback (`Mon, 02-Jan-2006 15:04:05 MST`) for compatibility
- Reworked cookie scanner in `ParseBytes` to a no-copy parsing flow (`nextRaw`) for intermediate token processing.
- Added no-copy trim helper for cookie token parsing (`trimCookieArgNoCopy`), with copy only at final field assignment.

#### Benchmark
`BenchmarkCookieParseFull`:
- **Before (original):** `291.6 ns/op`, `163 B/op`, `4 allocs/op`
- **After:** `85.15 ns/op`, `0 B/op`, `0 allocs/op`
- **Net Delta:** `-70.8%` (about `-206 ns/op`), `-100% B/op`, `-100% allocs/op`

---

### 3) Status line formatting: much faster and fewer allocations
**File:** `status.go`

#### Changes
- Pre-grow destination buffer capacity before assembling the status line.
- Added fast 3-digit status-code append path (avoids generic integer conversion for common codes).
- Avoid repeated status text fallback evaluation in the formatter.

#### Benchmarks
- `BenchmarkStatusLine99`:
  - Before: `46.99 ns/op`, `120 B/op`, `4 allocs/op`
  - After: `19.49 ns/op`, `48 B/op`, `1 allocs/op`
  - Delta: `-58.52%` (`p=0.000`)
- `BenchmarkStatusLine200`:
  - Before: `26.38 ns/op`, `56 B/op`, `3 allocs/op`
  - After: `11.29 ns/op`, `24 B/op`, `1 allocs/op`
  - Delta: `-57.21%` (`p=0.000`)
- `BenchmarkStatusLine512`:
  - Before: `47.03 ns/op`, `120 B/op`, `4 allocs/op`
  - After: `18.91 ns/op`, `48 B/op`, `1 allocs/op`
  - Delta: `-59.79%` (`p=0.000`)

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
 	"strconv"
 	"sync"
 	"time"
@@ -122,7 +123,7 @@ func ParseHTTPDate(date []byte) (time.Time, error) {
 	if t, ok := parseRFC1123DateGMT(date); ok {
 		return t, nil
 	}
-	return time.Parse(time.RFC1123, b2s(date))
+	return time.Parse(http.TimeFormat, b2s(date))
 }
 
 func parseRFC1123DateGMT(b []byte) (time.Time, bool) {
--- a/bytesconv_test.go
+++ b/bytesconv_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"html"
 	"net"
+	"net/http"
 	"net/url"
 	"strconv"
 	"testing"
@@ -213,7 +214,7 @@ func TestParseHTTPDateCompatibility(t *testing.T) {
 		{name: "epoch", value: "Thu, 01 Jan 1970 00:00:00 GMT", roundTrip: true},
 		{name: "year-boundary", value: "Fri, 31 Dec 1999 23:59:59 GMT", roundTrip: true},
 		{name: "leap-year", value: "Mon, 29 Feb 2016 12:34:56 GMT", roundTrip: true},
-		{name: "utc-fallback", value: "Tue, 10 Nov 2009 23:00:00 UTC"},
+		{name: "utc-fallback", value: "Tue, 10 Nov 2009 23:00:00 UTC", hasError: true},
 		{name: "mixedcase-weekday-month", value: "tUe, 10 nOv 2009 23:00:00 GMT"},
 		{name: "day-zero", value: "Tue, 00 Nov 2009 23:00:00 GMT", hasError: true},
 		{name: "invalid-day", value: "Tue, 31 Feb 2009 23:00:00 GMT", hasError: true},
@@ -232,7 +233,7 @@ func TestParseHTTPDateCompatibility(t *testing.T) {
 			t.Parallel()
 
 			got, gotErr := ParseHTTPDate([]byte(tc.value))
-			want, wantErr := time.ParseInLocation(time.RFC1123, tc.value, time.UTC)
+			want, wantErr := time.Parse(http.TimeFormat, tc.value)
 
 			if (gotErr != nil) != (wantErr != nil) {
 				t.Fatalf("error mismatch for %q: ParseHTTPDate err=%v, ParseInLocation err=%v", tc.value, gotErr, wantErr)
@@ -267,7 +268,7 @@ func BenchmarkParseHTTPDate(b *testing.B) {
 		b.ReportAllocs()
 		s := string(date)
 		for range b.N {
-			_, _ = time.Parse(time.RFC1123, s)
+			_, _ = time.Parse(http.TimeFormat, s)
 		}
 	})
 }
@@ -296,20 +297,8 @@ func FuzzParseHTTPDate(f *testing.F) {
 	f.Fuzz(func(t *testing.T, s string) {
 		b := []byte(s)
 
-		// Reference: time.Parse is what ParseHTTPDate falls back to.
-		stdTime, stdErr := time.Parse(time.RFC1123, s)
-
-		// The fast path must never accept a string that time.Parse
-		// rejects, and when it accepts, the result must be identical.
-		fastTime, fastOK := parseRFC1123DateGMT(b)
-		if fastOK {
-			if stdErr != nil {
-				t.Fatalf("parseRFC1123DateGMT accepted %q but time.Parse rejected it: %v", s, stdErr)
-			}
-			if !fastTime.Equal(stdTime) {
-				t.Fatalf("time mismatch for %q: fast=%v std=%v", s, fastTime, stdTime)
-			}
-		}
+		// Reference: time.Parse with http.TimeFormat is what ParseHTTPDate falls back to.
+		stdTime, stdErr := time.Parse(http.TimeFormat, s)
 
 		// The public API must always agree with time.Parse.
 		got, gotErr := ParseHTTPDate(b)
--- a/cookie.go
+++ b/cookie.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"io"
+	"net/http"
 	"sync"
 	"time"
 )
@@ -402,18 +403,11 @@ func (c *Cookie) ParseBytes(src []byte) error {
 
 			case 'e': // "expires"
 				if caseInsensitiveCompare(strCookieExpires, k) {
-					exptime, err := ParseHTTPDate(v)
-					if err == nil {
-						c.expire = exptime
-					} else {
-						vs := b2s(v)
-						// Keep legacy cookie date compatibility used by net/http.
-						exptime, err = time.Parse("Mon, 02-Jan-2006 15:04:05 MST", vs)
-						if err != nil {
-							return err
-						}
-						c.expire = exptime
+					exptime, err := parseCookieExpires(v)
+					if err != nil {
+						return err
 					}
+					c.expire = exptime
 				}
 
 			case 'd': // "domain"
@@ -643,6 +637,23 @@ func trimCookieArgNoCopy(src []byte, skipQuotes bool) []byte {
 
 // caseInsensitiveCompare does a case insensitive equality comparison of
 // two []byte. Assumes only letters need to be matched.
+func parseCookieExpires(src []byte) (time.Time, error) {
+	if t, ok := parseRFC1123DateGMT(src); ok {
+		return t, nil
+	}
+
+	s := b2s(src)
+
+	// UTC-anchored RFC1123 parsing behavior for non-GMT.
+	t, err := time.ParseInLocation(http.TimeFormat, s, time.UTC)
+	if err == nil {
+		return t, nil
+	}
+
+	// Legacy cookie date compatibility used by net/http.
+	return time.Parse("Mon, 02-Jan-2006 15:04:05 MST", s)
+}
+
 func caseInsensitiveCompare(a, b []byte) bool {
 	if len(a) != len(b) {
 		return false


gaby:
@ReneWerner87 The security workflow probably needs to be updated. Since min go for fasthttp is not 1.20.x

gaby:
It should be using 1.24.x, the version of gosec also needs to be updated.

bytesconv_timing_test.go
```diff
@@ -187,3 +188,32 @@ func BenchmarkParseUfloat(b *testing.B) {
...
+	b.RunParallel(func(pb *testing.PB) {
+		var buf []byte
+		for pb.Next() {
+			buf = AppendHTTPDate(buf[:0], ts)
+			if string(buf) != want {
+				b.Fatalf("unexpected date: %q. Expecting %q", buf, want)
+			}
+		}
+	})
+}
```
erikdubbelboer: What are these benchmarks for? The code they call wasn't modified at all?

bytesconv.go
```diff
@@ -117,9 +117,130 @@ func AppendHTTPDate(dst []byte, date time.Time) []byte {
...
 // ParseHTTPDate parses HTTP-compliant (RFC1123) date.
 func ParseHTTPDate(date []byte) (time.Time, error) {
+	if t, ok := parseRFC1123DateGMT(date); ok {
+		return t, nil
+	}
 	return time.Parse(time.RFC1123, b2s(date))
 }
 
+func parseRFC1123DateGMT(b []byte) (time.Time, bool) {
+	// Expects "Mon, 02 Jan 2006 15:04:05 GMT".
```
erikdubbelboer, line 130: `b[0:3]` is never validated. I guess it doesn't really matter as it's not needed, but probably better to explain this in a comment.

fasthttpadaptor/request.go
```diff
@@ -68,3 +73,28 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	// Fast path for the common origin-form request URI that doesn't require unescaping.
+	if requestURI != "" && requestURI[0] == '/' && !strings.ContainsAny(requestURI, "%#") {
+		if r.URL == nil {
+			r.URL = &url.URL{}
+		} else {
+			*r.URL = url.URL{}
+		}
+		if n := strings.IndexByte(requestURI, '?'); n >= 0 {
+			r.URL.Path = requestURI[:n]
+			r.URL.RawQuery = requestURI[n+1:]
```
erikdubbelboer: This doesn't assign [`URL.ForceQuery`](https://pkg.go.dev/net/url#URL.ForceQuery) which `url.ParseRequestURI` does.

fasthttpadaptor/request.go
```diff
@@ -68,3 +73,28 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+		} else {
+			r.URL.Path = requestURI
+		}
```
erikdubbelboer: Don't forget about `r.URL.RawPath`

bytesconv.go
```diff
@@ -117,9 +117,130 @@ func AppendHTTPDate(dst []byte, date time.Time) []byte {
...
+	minute, ok := parse2Digits(b[20], b[21])
+	if !ok || minute > 59 {
+		return time.Time{}, false
+	}
+	second, ok := parse2Digits(b[23], b[24])
+	if !ok || second > 59 {
+		return time.Time{}, false
+	}
+
+	t := time.Date(year, month, day, hour, minute, second, 0, time.UTC)
```
erikdubbelboer: Using UTC here is not the same as GMT. Especially if this time is formatted again for another request for example.

ReneWerner87:
I'll look at your comments over the next few days.

bytesconv_timing_test.go
ReneWerner87: That's right, I tried to optimize the other function as well, but it was already optimized to the maximum extent possible.

fasthttpadaptor/request.go
ReneWerner87: fixed with https://github.com/valyala/fasthttp/pull/2135/changes/2d2fc57b38729ffc1dae6d138ab11d7d5cc30dc5

fasthttpadaptor/request.go
ReneWerner87: fixed with https://github.com/valyala/fasthttp/pull/2135/changes/2d2fc57b38729ffc1dae6d138ab11d7d5cc30dc5

bytesconv_timing_test.go
ReneWerner87: removed

bytesconv.go
ReneWerner87: fixed with https://github.com/valyala/fasthttp/pull/2135/changes/2d2fc57b38729ffc1dae6d138ab11d7d5cc30dc5

bytesconv.go
ReneWerner87: fixed with https://github.com/valyala/fasthttp/pull/2135/changes/2d2fc57b38729ffc1dae6d138ab11d7d5cc30dc5

ReneWerner87:
@erikdubbelboer  Can you check again? I'm happy to move things around or remove/adjust them.

ReneWerner87:
I made this https://github.com/valyala/fasthttp/pull/2135/commits/219a75baef1c25a1d57a1fd1fb672d4458547e5b change to make it more readable.

```md
- `parseMonth3` (current packed vs original branch-based):
  - current: `40.82 ns/op`
  - original: `42.89 ns/op`
  - improvement: -4.83% (about -2.07 ns/op), `p=0.000`
  - allocations: unchanged (`0 B/op`, `0 allocs/op`)

- `isWeekday3` (current packed vs original branch-based):
  - current: `25.45 ns/op`
  - original: `26.56 ns/op`
  - improvement: -4.18% (about -1.11 ns/op), `p=0.000`
  - allocations: unchanged (`0 B/op`, `0 allocs/op`)
```

ReneWerner87:
@erikdubbelboer Can you check again? Should I adjust anything else?

fasthttpadaptor/request.go
```diff
@@ -68,3 +73,29 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	if requestURI != "" && requestURI[0] == '/' && !strings.ContainsAny(requestURI, "%#") {
+		if r.URL == nil {
+			r.URL = &url.URL{}
+		} else {
+			*r.URL = url.URL{}
+		}
+		if n := strings.IndexByte(requestURI, '?'); n >= 0 {
+			r.URL.Path = requestURI[:n]
+			r.URL.RawQuery = requestURI[n+1:]
+			r.URL.ForceQuery = n == len(requestURI)-1
```
erikdubbelboer: I think this case also needs to set `RawPath`.

status.go
```diff
@@ -164,14 +164,61 @@ func StatusMessage(statusCode int) string {
...
+			byte('0'+statusCode/100),
+			byte('0'+(statusCode/10)%10),
+			byte('0'+statusCode%10),
+		)
+		return dst
+	}
+	return strconv.AppendInt(dst, int64(statusCode), 10)
+}
+
+func growBytes(dst []byte, n int) []byte {
```
erikdubbelboer: This function is only used once, can you just inline it?

fasthttpadaptor/request.go
```diff
@@ -68,3 +73,29 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 
 	return nil
 }
+
+func parseRequestURI(r *http.Request, requestURI string) error {
```
erikdubbelboer: I have thought about this and I'm sorry but it feels like the extra code is not worth the performance here. Especially since anyone using fasthttpadaptor isn't going for speed. I would rather have less code that is more future proof. Can you remove this change from this pull?

bytesconv.go
```diff
@@ -117,9 +119,133 @@ func AppendHTTPDate(dst []byte, date time.Time) []byte {
 
 // ParseHTTPDate parses HTTP-compliant (RFC1123) date.
 func ParseHTTPDate(date []byte) (time.Time, error) {
+	if t, ok := parseRFC1123DateGMT(date); ok {
+		return t, nil
+	}
 	return time.Parse(time.RFC1123, b2s(date))
 }
 
+func parseRFC1123DateGMT(b []byte) (time.Time, bool) {
```
erikdubbelboer, line 129: I'm sorry but I don't think the amount of extra code is worth the minor performance improvement. The code feels quite fragile. Can you remove it so I can merge the other performance improvements?

cookie.go
```diff
@@ -591,3 +654,7 @@ func caseInsensitiveCompare(a, b []byte) bool {
 	}
 	return true
 }
+
+func cookieTokenEqual(a, b []byte) bool {
+	return bytes.Equal(a, b) || caseInsensitiveCompare(a, b)
```
erikdubbelboer: This is of course only faster if headers have the exact same casing. If it's different casing it could in theory be slower. But I guess the most common other way to write headers is all lowercase which would immediately get into the or clause making the performance impact minimal.

bytesconv.go
ReneWerner87: This function is used in the Cookie Parser for the "expires" header and in IfModifiedSince which is then used in the fs handler

parseRFC1123DateGMT vs time.Parse(time.RFC1123, ...)
```
Fast path (parseRFC1123DateGMT): 31.08 ns/op
Standard (time.Parse): 230.95 ns/op
Delta: -86.5%

Memory/allocation:
Fast path: 0 B/op, 0 allocs/op
Standard parse: 163 B/op, 4 allocs/op
```

> minor performance improvement

I wouldn't say it's a minor improvement, but of course I can delete them.

fasthttpadaptor/request.go
ReneWerner87: ok, i will remove it
Just the numbers for history

Image

bytesconv.go
ReneWerner87: I've adjusted everything except for this code.
@erikdubbelboer 
Since it's such a massive boost, I'd like to wait for your opinion again after I've posted the numbers just for that.

.github/workflows/security.yml
```diff
@@ -8,7 +8,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.20.x]
+        go-version: [1.24.x]
```
gaby: This should be 1.25.x at this point

bytesconv.go
ReneWerner87: @erikdubbelboer ping

erikdubbelboer: I'm sorry but I don't think the performance gain is worth the complications as the code is right now. I can already see issues where if fasthttp is used as reverse proxy it doesn't validate the day of the week while the backend server does. Same for weird years. I have been seeing and fixing a lot of issues around fasthttp not parsing http correctly.

If you could write a Fuzz test and show that the parsing is always the same vs `time.Parse(time.RFC1123, ...)`, then I could accept this.

ReneWerner87:
@erikdubbelboer https://github.com/valyala/fasthttp/pull/2135/changes/3231343557467f8b2ff03c9fbe873caf516beafb#diff-a19d0a74cdecfd528924eedab5bb64ecd75edc67d8bdcba74ae00855a6c15149
the requested fuzz test

ReneWerner87:
image

is not related to my code changes , just flaky

bytesconv.go
```diff
@@ -117,9 +119,133 @@ func AppendHTTPDate(dst []byte, date time.Time) []byte {
 
 // ParseHTTPDate parses HTTP-compliant (RFC1123) date.
 func ParseHTTPDate(date []byte) (time.Time, error) {
+	if t, ok := parseRFC1123DateGMT(date); ok {
+		return t, nil
+	}
 	return time.Parse(time.RFC1123, b2s(date))
```
erikdubbelboer: Can you change this to `http.TimeFormat`. It's more accurate and makes things slightly simpler.

bytesconv_test.go
```diff
@@ -200,6 +200,128 @@ func TestAppendHTTPDate(t *testing.T) {
...
+	}
+	for _, s := range seeds {
+		f.Add(s)
+	}
+
+	f.Fuzz(func(t *testing.T, s string) {
+		b := []byte(s)
+
+		// Reference: time.Parse is what ParseHTTPDate falls back to.
+		stdTime, stdErr := time.Parse(time.RFC1123, s)
```
erikdubbelboer: This should be `http.TimeFormat`.

bytesconv_test.go
```diff
@@ -200,6 +200,128 @@ func TestAppendHTTPDate(t *testing.T) {
...
+
+		// The fast path must never accept a string that time.Parse
+		// rejects, and when it accepts, the result must be identical.
+		fastTime, fastOK := parseRFC1123DateGMT(b)
```
erikdubbelboer: It doesn't really make sense to call `parseRFC1123DateGMT(b)` if you're going to call `ParseHTTPDate(b)` right after anyways.

bytesconv.go
erikdubbelboer: Can you change this to `http.TimeFormat`.

`net/http` also only accepts `GMT` as timezone.

bytesconv_test.go
```diff
@@ -200,6 +200,128 @@ func TestAppendHTTPDate(t *testing.T) {
...
+		for range b.N {
+			_, _ = ParseHTTPDate(date)
+		}
+	})
+
+	b.Run("stdlib-only", func(b *testing.B) {
+		b.ReportAllocs()
+		s := string(date)
+		for range b.N {
+			_, _ = time.Parse(time.RFC1123, s)
```
erikdubbelboer: `http.TimeFormat`

cookie.go
```diff
@@ -379,88 +379,90 @@ func (c *Cookie) ParseBytes(src []byte) error {
...
 			case 'e': // "expires"
-				if caseInsensitiveCompare(strCookieExpires, c.bufK) {
-					v := b2s(c.bufV)
-					// Try the same two formats as net/http
-					// See: https://github.com/golang/go/blob/00379be17e63a5b75b3237819392d2dc3b313a27/src/net/http/cookie.go#L133-L135
-					exptime, err := time.ParseInLocation(time.RFC1123, v, time.UTC)
-					if err != nil {
-						exptime, err = time.Parse("Mon, 02-Jan-2006 15:04:05 MST", v)
+				if caseInsensitiveCompare(strCookieExpires, k) {
+					exptime, err := ParseHTTPDate(v)
```
erikdubbelboer: Cookies are supposed to be `GMT` as well, but they aren't always. `net/http` also supports other timezones here. If `ParseHTTPDate` is only going to support GMT, we need some fallbacks here. There is also the issue that other timezones should parse based on UTC, not based on the current time of the server (that's why it was using `time.ParseInLocation`).

I suggest to use this new function:
```go
func parseCookieExpires(src []byte) (time.Time, error) {
	if t, ok := parseRFC1123DateGMT(src); ok {
		return t, nil
	}

	s := b2s(src)

	// UTC-anchored RFC1123 parsing behavior for non-GMT.
	t, err := time.ParseInLocation(time.RFC1123, s, time.UTC)
	if err == nil {
		return t, nil
	}

	// Legacy cookie date compatibility used by net/http.
	return time.Parse("Mon, 02-Jan-2006 15:04:05 MST", s)
}
```

ReneWerner87:
image
not related

@erikdubbelboer the rest is changed as you requested

erikdubbelboer:
Thanks!

PR #2145, opened 2026-02-20
Honor Root for fs.FS and normalize fs-style roots

erikdubbelboer:
- Apply `fs.Root` in non-`os.FS` path resolution.
- Normalize `fs.FS` roots (`./`, trailing slash, leading slash, separators).
- Handle `PathRewrite` outputs without a leading slash.
- Add tests for `MapFS` and `DirFS` root enforcement.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -10,6 +10,7 @@ import (
 	"mime"
 	"net/http"
 	"os"
+	"path"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


vnykmshr:
Fix looks correct, root normalization also covers the PathRewrite (without-leading-slash) case which I had missed.

PR #2150, opened 2026-03-01
Replace mustDiscard panic with error return (#1442)

### LINKED ISSUE #1442: mustDiscard() panics and the whole server crashes
In fasthttp, in file `header.go`, this function:
```golang
func mustDiscard(r *bufio.Reader, n int) {
	if _, err := r.Discard(n); err != nil {
		panic(fmt.Sprintf("bufio.Reader.Discard(%d) failed: %v", n, err))
	}
}
```
sometimes, under high concurrency, panics with error `panic: bufio.Reader.Discard(517) failed: read tcp4 ADDR-1->:8081->ADDR-2:30116: i/o timeout` and the server crashes.

Why it is not handled by error? Because the panic is on the side of the worker, and the thrown panic cannot be recovered by any of the RouterPanic or Recovery middleware. What should be done here?

Nicolas0315:
## Summary

Replace `mustDiscard()` which panics on `bufio.Reader.Discard()` failure with `discardHeader()` that returns an error, preventing server crashes under high concurrency.

## Problem

Under high concurrency, `mustDiscard()` in `header.go` panics when a connection times out during header reading:

```
panic: bufio.Reader.Discard(517) failed: read tcp4 ...: i/o timeout
```

This panic occurs in the worker goroutine and cannot be recovered by any middleware (RouterPanic, Recovery, etc.), causing the entire server to crash.

## Fix

- Rename `mustDiscard()` to `discardHeader()` and return an `error` instead of panicking
- Update all three call sites in `header.go` to propagate the error
- All call sites already return `error`, so the change is straightforward

## Testing

```bash
go test -run TestResponseHeader -count=1  # PASS
go test -run TestRequestHeader -count=1   # PASS
```

Fixes #1442

### DIFF
--- a/header.go
+++ b/header.go
@@ -2170,7 +2170,9 @@ func (h *ResponseHeader) tryRead(r *bufio.Reader, n int) error {
 	if errParse != nil {
 		return headerError("response", err, errParse, b, h.secureErrorLogMessage)
 	}
-	mustDiscard(r, headersLen)
+	if err := discardHeader(r, headersLen); err != nil {
+		return err
+	}
 	return nil
 }
 
@@ -2226,7 +2228,9 @@ func (h *header) tryReadTrailer(r *bufio.Reader, n int) error {
 		}
 		return headerError("response", err, errParse, b, h.secureErrorLogMessage)
 	}
-	mustDiscard(r, headersLen)
+	if err := discardHeader(r, headersLen); err != nil {
+		return err
+	}
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


erikdubbelboer:
This looks like an LLM finding a problem that doesn't exist. Do you have an example test case that triggers this bug?

gaby:
@erikdubbelboer still a good idea to avoid calling panic in such a hot path.

PR #2158, opened 2026-03-15
fix: detect master process death in prefork children

meruiden:
## Summary

Prefork child processes have no mechanism to detect if the master process dies unexpectedly. When the master is killed (e.g., SIGKILL), the `defer` cleanup in `prefork()` never fires, and children become orphans — reparented to PID 1 and running silently with no supervision.

This is the inverse of the issue fixed in gofiber/fiber#4133. Fiber's children *did* watch the parent PID but used a hardcoded `== 1` check that broke in Docker. fasthttp children don't watch the parent PID at all, so they can't detect master death in any environment.

### Fix

Add a `watchMaster()` goroutine to child processes that:
- Stores the original parent PID at startup
- Polls `os.Getppid()` every 500ms
- Exits if the parent PID changes (indicating the master died and the child was reparented)

This is the same pattern used in the gofiber/fiber fix, and is Docker-safe because it compares against the actual initial PPID rather than a hardcoded value.

### Comparison

| | Before this fix | After this fix |
|---|---|---|
| Child polls PPID? | ❌ no | ✅ yes |
| Docker-safe? | ✅ (no false exit) | ✅ (no false exit) |
| Detects master death? | ❌ orphans forever | ✅ exits cleanly |

## Test plan

- [x] All existing prefork tests pass
- [x] Integration test: children exit when master is killed (`Test_watchMaster_detectsParentDeath`)

Related: gofiber/fiber#4133


meruiden:
Thanks for the feedback! I've made `watchMaster` configurable and off by default:

- Added an `OnMasterDeath func()` callback field to `Prefork`. When nil (the default), no master monitoring happens — preserving backwards compatibility.
- When set, it enables the `watchMaster` goroutine, which calls the callback when the master dies. This lets users do custom cleanup (e.g. graceful shutdown) instead of being forced into `os.Exit(1)`.
- Provided `DefaultOnMasterDeath()` as a convenience for the simple `os.Exit(1)` behavior.
- Also fixed the ticker leak flagged by Copilot.

Usage:
```go
p := prefork.New(server)
p.OnMasterDeath = prefork.DefaultOnMasterDeath // simple exit
// or
p.OnMasterDeath = func() {
    server.Shutdown() // custom cleanup
}
```

Will push the updated branch shortly.

prefork/prefork.go
```diff
@@ -94,6 +104,25 @@ func (p *Prefork) logger() Logger {
 	return defaultLogger
 }
 
+// DefaultOnMasterDeath is the default OnMasterDeath handler that logs and
+// exits the child process immediately.
+func DefaultOnMasterDeath() {
+	os.Exit(1)
+}
```
erikdubbelboer: Anything we export as `Default...` means it is the default if you don't do anything. That is different in this case. I would just remove thes function and document that it's recommended to do a function that does `os.Exit(1)`.

prefork/prefork.go
```diff
@@ -94,6 +104,25 @@ func (p *Prefork) logger() Logger {
...
+
+func (p *Prefork) watchMaster(masterPID int) {
+	ticker := time.NewTicker(500 * time.Millisecond)
+	defer ticker.Stop()
+
+	for range ticker.C {
+		if os.Getppid() != masterPID {
+			p.logger().Printf("master process died, exiting child\n")
```
erikdubbelboer: Remove the `exiting child` part since that is not true anymore.

prefork/watchmaster_test.go
```diff
@@ -0,0 +1,201 @@
...
+			_ = childProc.Kill()
+			t.Fatal("child process did not exit after master was killed")
+		case <-ticker.C:
+			if err := childProc.Signal(syscall.Signal(0)); err != nil {
+				// Process is gone, watchMaster detected parent death.
+				return
+			}
+		}
+	}
+}
```
erikdubbelboer: All these tests are a bit much and feel very AI generated verbose. Just remove them, the code is simple enough.

meruiden: Fair enough, mostly because its very tricky to test te master child stuff since it spawns multiple threads

prefork/prefork.go
meruiden: Agreed, that is confusing,. not entirely convinced about the no cleanup by default and hoping the user sees this function and its docs, but then again its also a pretty advanced feature so i think we'll be fine

prefork/prefork.go
meruiden: oops, thanks. done

prefork/prefork.go
meruiden: but personally id still also put it as default in 

```
func New(s *fasthttp.Server) *Prefork {
```

and then also just mention it in the docs as is now, and then the end user can always override if os.exit is too aggressive for them or want to control cleanup

erikdubbelboer:
Thanks!

PR #2175, opened 2026-04-06
feat: add ExpectHandler for richer Expect: 100-continue handling

miretskiy:
ContinueHandler only returns a bool, limiting the server to either accepting (100) or rejecting with 417. ExpectHandler allows returning any HTTP status code, and closes the connection on rejection since the client may have already started sending body data per RFC 9110.

ExpectHandler takes precedence when both handlers are set.

### DIFF
--- a/server.go
+++ b/server.go
@@ -205,8 +205,7 @@ type Server struct {
 	// client may have already started sending the request body.
 	//
 	// The ctx provides access to request headers and connection metadata (e.g.
-	// RemoteAddr for IP-based filtering). The response must not be modified —
-	// only the returned status code is used.
+	// RemoteAddr for IP-based filtering). The response must not be modified.
 	//
 	// If both ExpectHandler and ContinueHandler are set, ExpectHandler
 	// takes precedence.
--- a/server.go
+++ b/server.go
@@ -194,6 +194,25 @@ type Server struct {
 	// like they are normal requests.
 	ContinueHandler func(header *RequestHeader) bool
 
+	// ExpectHandler is called after receiving the Expect 100 Continue Header.
+	//
+	// https://www.rfc-editor.org/rfc/rfc9110.html#field.expect
+	//
+	// ExpectHandler provides more control than ContinueHandler by allowing
+	// the server to respond with any final status code. The handler should return
+	// StatusContinue (100) to accept the request and proceed to read the body,
+	// or any other status code to reject it. If StatusExpectationFailed (417) is
+	// returned, the response is sent and the connection is closed. For any other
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -194,6 +194,25 @@ type Server struct {
...
+	// returned, the response is sent and the connection is closed. For any other
+	// non-100 status code, the response is also sent and the connection is closed,
+	// since the client may have already started sending the request body.
+	//
+	// If both ExpectHandler and ContinueHandler are set, ExpectHandler
+	// takes precedence.
+	//
+	// The default behavior (when neither handler is set) is to automatically accept
+	// the request body.
+	ExpectHandler func(header *RequestHeader) int
```
erikdubbelboer: I can also see a case where someone would want to block this based on IP for example. One the one hand passing a `RequestCtx` instead would expose that kind of information. But it's weird to have a Reponse object on this then (unless that is then actually usable). What do you think?

miretskiy: Sure; I don't mind; I think comment is good enough.

miretskiy: @erikdubbelboer -- it's done; tests added; let me know

server.go
```diff
@@ -194,6 +194,29 @@ type Server struct {
...
+	//
+	// https://www.rfc-editor.org/rfc/rfc9110.html#field.expect
+	//
+	// ExpectHandler provides more control than ContinueHandler by allowing
+	// the server to respond with any final status code. The handler should return
+	// StatusContinue (100) to accept the request and proceed to read the body,
+	// or any other status code to reject it. If StatusExpectationFailed (417) is
+	// returned, the response is sent and the connection is closed. For any other
+	// non-100 status code, the response is also sent and the connection is closed,
+	// since the client may have already started sending the request body.
```
erikdubbelboer:
```suggestion
	// or any other status code to reject it and close the connection since the
	// client may have already started sending the request body.
```

server.go
```diff
@@ -194,6 +194,29 @@ type Server struct {
...
+	//
+	// The ctx provides access to request headers and connection metadata (e.g.
+	// RemoteAddr for IP-based filtering). The response must not be modified —
+	// only the returned status code is used.
```
erikdubbelboer:
```suggestion
	// RemoteAddr for IP-based filtering). The response must not be modified.
```
This is not strictly true, so lets word it like this.

erikdubbelboer (changes requested):

miretskiy:
All done.

erikdubbelboer:
Thanks!

PR #2176, opened 2026-04-07
feat(client): add RetryIfErrUpstream function to handle upstream information

mdenushev:
Currently, there is no way to get upstream information in RetryIfErr function, but upstream information is very important for logging purposes.

Follow up to #2105.

### DIFF
--- a/client.go
+++ b/client.go
@@ -216,7 +216,7 @@ type Client struct {
 	// This field is only effective within the range of MaxIdemponentCallAttempts.
 	RetryIfErr RetryIfErrFunc
 
-	// RetryIfErrUpstream works just like RetryIfErr but also provides information about upstream if known.
+	// RetryIfErrUpstream works just like RetryIfErr but also provides information about which upstream caused the error, if known.
 	// Upstream information is a <host>:<port> format.
 	RetryIfErrUpstream RetryIfErrUpstreamFunc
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -706,16 +715,7 @@ const (
...
-//
-// HostClient may be used for balancing load among multiple upstream hosts.
-// While multiple addresses passed to HostClient.Addr may be used for balancing
-// load among them, it would be better using LBClient instead, since HostClient
-// may unevenly balance load among upstream hosts.
-//
-// It is forbidden copying HostClient instances. Create new instances instead.
-//
-// It is safe calling HostClient methods from concurrently running goroutines.
+// HostClient represents a high-performance HTTP client optimized for low-level control and customization.
```
erikdubbelboer: Why change this unrelated comment? Please remove this.

client.go
```diff
@@ -216,6 +216,10 @@ type Client struct {
 	// This field is only effective within the range of MaxIdemponentCallAttempts.
 	RetryIfErr RetryIfErrFunc
 
+	// RetryIfErrUpstream works just like RetryIfErr but also provides information about upstream if known.
```
erikdubbelboer:
```suggestion
	// RetryIfErrUpstream works just like RetryIfErr but also provides information about which upstream caused the error, if known.
```

client.go
```diff
@@ -693,6 +698,10 @@ type RetryIfFunc func(request *Request) bool
 // the request function will immediately return with the `err`.
 type RetryIfErrFunc func(request *Request, attempts int, err error) (resetTimeout bool, retry bool)
 
+// RetryIfErrUpstreamFunc works just like a RetryIfErrFunc and also provides information about upstream caused problems if known.
```
erikdubbelboer:
```suggestion
// RetryIfErrUpstreamFunc works just like a RetryIfErrFunc and also provides information about which upstream caused the error, if known.
```

client.go
```diff
@@ -754,6 +754,10 @@ type HostClient struct {
 	// This field is only effective within the range of MaxIdemponentCallAttempts.
 	RetryIfErr RetryIfErrFunc
 
+	// RetryIfErrUpstream works just like RetryIfErr but also provides information about upstream if known.
```
erikdubbelboer:
```suggestion
	// RetryIfErrUpstream works just like RetryIfErr but also provides information about which upstream causes the error, if known.
```

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2180, opened 2026-04-12
feat(prefork): Enhance prefork management with WatchMaster, CommandProducer, and Windows support

ReneWerner87:
We've been working on reducing code duplication between Fiber and fasthttp's prefork package. Right now Fiber maintains its own prefork implementation (~200 lines) that reimplements most of what `fasthttp/prefork` already does - process spawning, child management, the recovery loop - just to hook into a few lifecycle events (logging when a child spawns, showing a startup message after all workers are up, etc.).

This PR adds a small set of optional callbacks to the `Prefork` struct that make it possible for consumers like Fiber to plug into the prefork lifecycle without copying the whole thing:

- **`OnChildSpawn(pid int) error`** - called when a new child is forked, so the caller can track PIDs or run hooks
- **`OnMasterReady(childPIDs []int) error`** - called once after all children are up, useful for startup messages or metrics
- **`OnChildRecover(pid int) error`** - called when a crashed child gets replaced, for logging/alerting
- **`CommandProducer(files []*os.File) (*exec.Cmd, error)`** - allows overriding how child commands are created, which is useful for testing (injecting dummy commands instead of re-executing the binary)

All fields are optional and nil-safe - existing code doesn't need to change. The zero value works exactly as before.

Additionally:
- `listenAsChild()` extracted to remove duplicated setup code across the three `ListenAndServe*` methods
- `watchMaster` now handles Windows correctly (using `FindProcess`+`Wait` instead of `Getppid` polling, which doesn't detect parent exit on Windows)
- Goroutine closures in the spawn/recovery loop pass `cmd` and `pid` as parameters instead of capturing the loop variable

You can see the practical usage in the Fiber PR that builds on this: https://github.com/gofiber/fiber/pull/4210 - it replaces Fiber's entire prefork implementation with ~50 lines that configure these callbacks. The rest is handled by this package.


ReneWerner87:
@erikdubbelboer WDYT?

prefork/prefork.go
```diff
@@ -104,6 +130,22 @@ func (p *Prefork) logger() Logger {
 }
 
 func (p *Prefork) watchMaster(masterPID int) {
+	if runtime.GOOS == "windows" {
+		// On Windows, os.Getppid() returns a static PID that doesn't change
+		// when the parent exits (no reparenting). Use FindProcess+Wait instead.
+		proc, err := os.FindProcess(masterPID)
+		if err == nil {
+			_, _ = proc.Wait()
```
erikdubbelboer: Can't something be done with these errors?

At least `p.logger().Print` them.

prefork/prefork.go
```diff
@@ -205,25 +278,47 @@ func (p *Prefork) prefork(addr string) (err error) {
...
-		}()
+		pid := cmd.Process.Pid
+		childProcs[pid] = cmd
+		childPIDs = append(childPIDs, pid)
+
+		// Call OnChildSpawn callback
+		if p.OnChildSpawn != nil {
+			if err = p.OnChildSpawn(pid); err != nil {
+				p.logger().Printf("OnChildSpawn callback failed for PID %d: %v\n", pid, err)
+				return err
```
erikdubbelboer, line 316: Shouldn't we kill the child here?

prefork/prefork.go
```diff
@@ -158,6 +218,19 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
 }
 
 func (p *Prefork) doCommand() (*exec.Cmd, error) {
+	// Use custom CommandProducer if provided
+	if p.CommandProducer != nil {
+		cmd, err := p.CommandProducer(p.files)
```
erikdubbelboer, line 230: What exactly is the use case of `CommandProducer`? Right now it doesn't document how to handle the args or the `FASTHTTP_PREFORK_CHILD=1` env variable, or what to do with stdout and stderr.

If it's only used for the executable maybe it's better to have it just return a string?

prefork/prefork.go
```diff
@@ -237,19 +332,27 @@ func (p *Prefork) prefork(addr string) (err error) {
...
-		childProcs[cmd.Process.Pid] = cmd
-		go func() {
-			sigCh <- procSig{pid: cmd.Process.Pid, err: cmd.Wait()}
-		}()
+		pid := cmd.Process.Pid
+		childProcs[pid] = cmd
+
+		// Call OnChildRecover callback and ignore its returned error.
+		if p.OnChildRecover != nil {
+			_ = p.OnChildRecover(pid)
```
erikdubbelboer: Why add the error return value when it's just ignored?

erikdubbelboer (changes requested):

ReneWerner87:
will check the hints and answer tomorrow, thx

prefork/prefork.go
ReneWerner87: Right, those errors are just swallowed. I'll log them and handle the case where FindProcess fails instead of calling OnMasterDeath blindly.

prefork/prefork.go
ReneWerner87: The defer iterates childProcs and kills everything when the function returns. 
https://github.com/valyala/fasthttp/pull/2180/changes/BASE..be2ef67270a32712e777078f8531b2dc373af0e8#diff-ee20e9cc4d9dce2b9266a37017103020dd1502faa0e6c0fef2c21496a435086dR283
The child is already in the map at that point, so it gets cleaned up. I'll add a comment.

prefork/prefork.go
ReneWerner87: Mainly needed for testing in Fiber. Without it, Fiber has to maintain its own package-level test variables to inject dummy commands instead of re-executing the real binary in unit tests. A simple string for the executable path wouldn't be enough since Fiber also needs to customize stdout/stderr/env in tests. Happy to improve the docs and describe the contract more clearly though.

prefork/prefork.go
ReneWerner87: Fair point, if the error is ignored anyway the return value doesn't make sense. I'll change it to func(pid int), simpler and more honest.

ReneWerner87:
I will implement the changes suggested in the comments

ReneWerner87:
ready for a recheck @erikdubbelboer

prefork/prefork.go
```diff
@@ -237,19 +341,26 @@ func (p *Prefork) prefork(addr string) (err error) {
...
-		childProcs[cmd.Process.Pid] = cmd
-		go func() {
-			sigCh <- procSig{pid: cmd.Process.Pid, err: cmd.Wait()}
-		}()
+		pid := cmd.Process.Pid
+		childProcs[pid] = cmd
+
+		if p.OnChildRecover != nil {
+			p.OnChildRecover(pid)
+		}
```
erikdubbelboer: Shouldn't `OnChildSpawn` be called here as well? Or the documentation of `OnChildSpawn` should change.

prefork/prefork.go
```diff
@@ -104,6 +133,26 @@ func (p *Prefork) logger() Logger {
 }
 
 func (p *Prefork) watchMaster(masterPID int) {
+	if runtime.GOOS == "windows" {
+		// On Windows, os.Getppid() returns a static PID that doesn't change
+		// when the parent exits (no reparenting). Use FindProcess+Wait instead.
+		proc, err := os.FindProcess(masterPID)
+		if err != nil {
+			p.logger().Printf("watchMaster: failed to find master process %d: %v\n", masterPID, err)
+			return
```
erikdubbelboer, line 228: Shouldn't this trigger `p.OnMasterDeath()`? Not being able to find the master process probably means it died right?

prefork/prefork.go
```diff
@@ -77,6 +77,35 @@ type Prefork struct {
...
+	// If nil, the default implementation re-executes the current binary
+	// with FASTHTTP_PREFORK_CHILD=1 in the environment.
+	//
+	// The producer receives the files to pass as ExtraFiles and must return
+	// an already started command (i.e. cmd.Start() must have been called).
+	// The caller is responsible for setting Stdout, Stderr, Env, and ExtraFiles
+	// on the command before starting it.
+	//
+	// This is primarily useful for testing (injecting dummy commands)
+	// or for frameworks that need custom child process setup.
```
erikdubbelboer: Doesn't the producer have to set `FASTHTTP_PREFORK_CHILD=1`? Otherwise `IsChild()` returns false and the listener won't be setup correctly.

Please also add a comment that this is isn't required and what the default does.

prefork/prefork.go
```diff
@@ -77,6 +77,35 @@ type Prefork struct {
...
+
+	// OnMasterReady is called in the master process after all child processes have been spawned.
+	// It receives a slice of all child process PIDs.
+	//
+	// If this callback returns an error, the prefork operation will be aborted.
+	OnMasterReady func(childPIDs []int) error
+
+	// OnChildRecover is called in the master process when a child process is restarted
+	// after a crash. It receives the PID of the newly recovered child process.
+	OnChildRecover func(pid int)
```
erikdubbelboer: Does it make sense to also provide the old pid as argument? That way any process management can know which pid was replaced?

erikdubbelboer (changes requested):

prefork/prefork.go
ReneWerner87: Makes sense, a recovered child is still a spawned child. I'll call OnChildSpawn here as well so the caller doesn't need to track both callbacks separately. OnChildRecover then becomes the one you use when you only care about the recovery case specifically.

prefork/prefork.go
ReneWerner87: Yeah, good point. If we can't find the process it's most likely gone. I'll call OnMasterDeath in that case too and just log the error.

prefork/prefork.go
ReneWerner87: Right, that's a requirement that's not documented. The producer has to include the env var otherwise IsChild() won't work in the child. I'll add that to the doc comment and also note that setting it to nil uses the default which handles all of that automatically.

prefork/prefork.go
ReneWerner87: That would be useful for process tracking. I'll change the signature to func(oldPid, newPid int) so the caller can update its PID mapping or log which process was replaced by which.

ReneWerner87:
@erikdubbelboer something else or ready for merge ?

prefork/prefork.go
```diff
@@ -205,25 +289,49 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+		childProcs[pid] = cmd
+		childPIDs = append(childPIDs, pid)
+
+		// Call OnChildSpawn callback.
+		// On error we return early — the child is already in childProcs
+		// and will be killed by the deferred cleanup above.
+		if p.OnChildSpawn != nil {
+			if err = p.OnChildSpawn(pid); err != nil {
+				p.logger().Printf("OnChildSpawn callback failed for PID %d: %v\n", pid, err)
+				return err
```
erikdubbelboer: The `go func(c *exec.Cmd, pid int) {` below actually needs to be above this line.

If `OnChildSpawn` returns an error and the deferred cleanup happens it will call `Kill()` on each child.
But after `Kill()` you always need to `Wait()` on the child to make sure it's cleaned up. If you don't `Wait()` it will hang as a zombie process.

prefork/prefork.go
```diff
@@ -237,19 +345,34 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+
+		if p.OnChildRecover != nil {
+			p.OnChildRecover(sig.pid, newPid)
+		}
+
+		// A recovered child is still a spawned child.
+		if p.OnChildSpawn != nil {
+			if err = p.OnChildSpawn(newPid); err != nil {
+				p.logger().Printf("OnChildSpawn callback failed for recovered PID %d: %v\n", newPid, err)
+				break
```
erikdubbelboer: Same here, if the `Wait()` isn't before this `break` it will create a zombie process.

prefork/prefork_test.go
```diff
@@ -224,3 +226,281 @@ func Test_ListenAndServeTLSEmbed(t *testing.T) {
...
+		t.Error("OnMasterDeath should not be nil")
+	}
+
+	p.OnMasterDeath()
+	if !called {
+		t.Error("OnMasterDeath was not called")
+	}
+}
+
+func Test_Prefork_Callbacks_NotNil(t *testing.T) {
```
erikdubbelboer: This whole test only tests if the Go compiler works 🤷‍♂️ it doesn't test any functionality at all.

prefork/prefork_test.go
```diff
@@ -224,3 +226,281 @@ func Test_ListenAndServeTLSEmbed(t *testing.T) {
...
+	}
+	if !readyCalled {
+		t.Error("OnMasterReady was not called")
+	}
+	if !recoverCalled {
+		t.Error("OnChildRecover was not called")
+	}
+}
+
+func Test_Prefork_Callbacks_Nil(t *testing.T) {
```
erikdubbelboer: Same here, what is this actually testing? These tests were obviously AI written, I'm not going to review tests you couldn't even bother to take the time to review yourself.

erikdubbelboer (changes requested):

prefork/prefork.go
ReneWerner87: Fixed — moved the Wait goroutine above OnChildSpawn, and added Wait() in the deferred cleanup after Kill() to properly reap children.

prefork/prefork.go
ReneWerner87: Same fix applied here — Wait goroutine now starts before callbacks.

prefork/prefork_test.go
ReneWerner87: You're right, removed it. Replaced with Test_Prefork_Lifecycle that runs the full prefork loop with a CommandProducer and checks callback order and arguments.

prefork/prefork_test.go
ReneWerner87: Removed. The new lifecycle test covers this properly.

ReneWerner87:
All feedback addressed:

- Wait goroutine now starts before OnChildSpawn (both in spawn and recovery loop) to avoid zombies
- Added Wait() in deferred cleanup after Kill()
- Removed the shallow callback tests, replaced with Test_Prefork_Lifecycle that runs the full prefork loop and verifies callback order/arguments

Ready for another look @erikdubbelboer

prefork/prefork_test.go
```diff
@@ -64,8 +71,22 @@ func Test_New(t *testing.T) {
...
+func Test_defaultRecoverThreshold_SingleCore(t *testing.T) {
+	prev := runtime.GOMAXPROCS(1)
+	t.Cleanup(func() {
+		runtime.GOMAXPROCS(prev)
+	})
+
+	if threshold := defaultRecoverThreshold(); threshold != 1 {
+		t.Errorf("defaultRecoverThreshold() == %d, want 1", threshold)
+	}
+}
```
erikdubbelboer: No need to test this, `defaultRecoverThreshold` is so simple it really doesn't need a test.

prefork/prefork_test.go
```diff
@@ -224,3 +245,161 @@ func Test_ListenAndServeTLSEmbed(t *testing.T) {
...
+		t.Error("Default logger should not be nil")
+	}
+
+	// Test custom logger
+	customLogger := &testLogger{}
+	p.Logger = customLogger
+	if p.logger() != customLogger {
+		t.Error("Custom logger should be returned")
+	}
+}
```
erikdubbelboer: No need to test this, this is just testing this statement only:
```
	if p.Logger != nil {
		return p.Logger
	}
```
I trust that this if statement works and that nobody will break this in the future.

prefork/prefork_test.go
```diff
@@ -64,8 +71,22 @@ func Test_New(t *testing.T) {
...
+		t.Errorf("defaultRecoverThreshold() == %d, want 1", threshold)
+	}
+}
+
 func Test_listen(t *testing.T) {
-	t.Parallel()
+	prev := runtime.GOMAXPROCS(0)
+	t.Cleanup(func() {
+		runtime.GOMAXPROCS(prev)
+	})
```
erikdubbelboer, line 102: This is a great catch! All tests running together and after this test would run on a single core. Never noticed!

prefork/prefork.go
```diff
@@ -237,19 +349,35 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+			sigCh <- procSig{pid: pid, err: c.Wait()}
+		}(cmd, newPid)
+
+		if p.OnChildRecover != nil {
+			p.OnChildRecover(sig.pid, newPid)
+		}
+
+		if p.OnChildSpawn != nil {
+			if err = p.OnChildSpawn(newPid); err != nil {
+				p.logger().Printf("OnChildSpawn callback failed for recovered PID %d: %v\n", newPid, err)
```
erikdubbelboer: Doesn't it make more sense to do `return err` here? That way `ListenAndServe` returns with this error and it can properly be handled?

Then you can also document in `OnChildSpawn` that returning an error will kill all childs and make `ListenAndServe` return that error.

prefork/prefork.go
```diff
@@ -237,19 +349,35 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+		childProcs[newPid] = cmd
+
+		// Start Wait goroutine before callbacks to avoid zombie processes
+		// if a callback fails and the deferred Kill() runs.
+		go func(c *exec.Cmd, pid int) {
+			sigCh <- procSig{pid: pid, err: c.Wait()}
+		}(cmd, newPid)
+
+		if p.OnChildRecover != nil {
+			p.OnChildRecover(sig.pid, newPid)
```
erikdubbelboer: You're calling `OnChildRecover` before `OnChildSpawn`. If `OnChildSpawn` returns an error than the child wasn't really recovered as everything will be shut down?

prefork/prefork.go
```diff
@@ -205,25 +293,49 @@ func (p *Prefork) prefork(addr string) (err error) {
 
 	goMaxProcs := runtime.GOMAXPROCS(0)
 	sigCh := make(chan procSig, goMaxProcs)
-	childProcs := make(map[int]*exec.Cmd)
+	childProcs := make(map[int]*exec.Cmd, goMaxProcs)
 
 	defer func() {
 		for _, proc := range childProcs {
 			_ = proc.Process.Kill()
+			_, _ = proc.Process.Wait() // avoid zombie processes after Kill
```
erikdubbelboer: Moving `sigCh <- procSig...` to before `OnChildSpawn` was already enough to make sure we don't leave any zombies. Adding this additional `Wait()` now makes each child Waited on twice, which is usually not advised to do.

See also: https://github.com/golang/go/blob/8594bf46218254ce5508e1500b92d4329f0a627c/src/os/exec_unix.go#L33-L35

erikdubbelboer (changes requested):

prefork/prefork_test.go
ReneWerner87: Fixed in db78ffe: removed the single-core helper test and the duplicate New() threshold assertion.

prefork/prefork_test.go
ReneWerner87: Fixed in db78ffe: removed the logger getter test.

prefork/prefork_test.go
ReneWerner87: Thanks, agreed. Kept the cleanup so this test no longer leaves the package running with GOMAXPROCS=1.

prefork/prefork.go
ReneWerner87: Fixed in db78ffe: recovered-child OnChildSpawn errors now return immediately from prefork()/ListenAndServe, and the OnChildSpawn docs describe that behavior.

prefork/prefork.go
ReneWerner87: Fixed in db78ffe: OnChildSpawn is called first for recovered children; OnChildRecover only runs after the recovered child spawn callback succeeds. Added a regression test for the error case.

prefork/prefork.go
ReneWerner87: Fixed in db78ffe: removed the extra Wait() from deferred cleanup. Each started child already has exactly one Wait goroutine before callbacks run.

erikdubbelboer:
Thanks! I'll tag a new release next week.

erikdubbelboer:
I now see that you pushed another commit right between me checking your previous commit and merging this. Github didn't show this 😠

ReneWerner87:
Oh, sorry, I didn't realize you were about to merge. Usually a little time passes before that happens.

Feel free to review the last commit again, and I'll fix whatever you notice.

erikdubbelboer:
I have reverted that commit on master: https://github.com/valyala/fasthttp/commit/e9208ecebf0c102176bb0635043c17333b10401d
Please make a new pull request for these changes.

ReneWerner87:
ok

PR #2181, opened 2026-04-19
Match net/http sensitive header redirect policy

erikdubbelboer:
Strip sensitive headers in DoRedirects matching net/http's redirect policy.

Reported by @vnykmshr


vnykmshr (approved):
Reviewed - all three DoRedirects surfaces (top-level, Client, HostClient) covered, boundary logic is sound.

PR #2182, opened 2026-04-20
Sanitize first-line header setters to prevent CRLF injection

erikdubbelboer:
Prevent request and response first-line setters from serializing embedded CR or LF bytes into the start line.

Route SetMethod, SetRequestURI, SetProtocol, and SetStatusMessage through the existing newline sanitization used by other header-value setters. This preserves behavior for valid inputs while preventing header injection through malformed first-line values.

Thanks to @vnykmshr for reporting this issue.


vnykmshr (approved):
Ran the PoC from the email thread against this branch. LGTM

PR #2194, opened 2026-04-28
fix: prevent zombie connsCleaner goroutine on dial failure

wucm667:
## Summary

When `dialHostHard()` fails, the `connsCleaner` goroutine (started before dial) becomes a zombie — it sleeps for `MaxIdleConnDuration` before checking if it should stop. With a large `MaxIdleConnDuration` (e.g., 5 minutes), each failed dial attempt leaves a zombie goroutine for that duration.

## Root Cause

In `AcquireConn()`, the cleaner goroutine is started at line 1735 *before* `dialHostHard()`:

```go
if startCleaner {
    go c.connsCleaner()  // started before dial
}

conn, err := c.dialHostHard(reqTimeout)
if err != nil {
    c.decConnsCount()  // connsCount drops, but cleaner is already sleeping
    return nil, err
}
```

The cleaner's loop checks `mustStop := c.connsCount == 0` only after waking from `time.Sleep(sleepFor)`, where `sleepFor` can be up to `MaxIdleConnDuration`.

## Fix

1. Move the `go c.connsCleaner()` start to **after** a successful dial
2. On dial failure, reset `connsCleanerRun = false` if `connsCount` drops to 0, so the next connection attempt can start the cleaner

```go
conn, err := c.dialHostHard(reqTimeout)
if err != nil {
    c.decConnsCount()
    if startCleaner {
        c.connsLock.Lock()
        if c.connsCount == 0 {
            c.connsCleanerRun = false
        }
        c.connsLock.Unlock()
    }
    return nil, err
}
cc = acquireClientConn(conn)

if startCleaner {
    go c.connsCleaner()
}
```

This handles concurrent scenarios correctly: if dial fails and count drops to 0, the next goroutine that needs a connection will see `connsCleanerRun = false` and start the cleaner itself.

## Testing

All existing client tests pass.

Fixes #2171

### DIFF
--- a/client.go
+++ b/client.go
@@ -1731,16 +1731,26 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
 		}
 	}
 
-	if startCleaner {
-		go c.connsCleaner()
-	}
-
 	conn, err := c.dialHostHard(reqTimeout)
 	if err != nil {
 		c.decConnsCount()
+		// Reset cleaner flag if connsCount dropped to 0, allowing the next
+		// connection attempt to start the cleaner. This prevents zombie
+		// goroutines when dial fails. See issue #2171.
+		if startCleaner {
+			c.connsLock.Lock()
+			if c.connsCount == 0 {
+				c.connsCleanerRun = false
+			}
+			c.connsLock.Unlock()
+		}
 		return nil, err
 	}
 	cc = acquireClientConn(conn)
+
+	if startCleaner {
+		go c.connsCleaner()
+	}
 
 	return cc, nil
 }


erikdubbelboer (changes requested):
I'm sorry but this is worse. Now there is as a possibility that the cleaner is never started while it should be.

Dial 1 starts. No other conns are open so it does `c.connsCleanerRun = true` and sets `startCleaner = true`. And then it starts dialing.

Dial 2 starts. `c.connsCleanerRun` is already `true` so `startCleaner = false` and it doesn't try to start a conns cleaner.

Dial 1 fails. `c.connsCount > 0` (from dial 2) so it doesn't set `c.connsCleanerRun = false` but it also doesn't start the connsCleaner.

No we have open connections but the cleaner never starts.


Also your statement of `each failed dial attempt leaves a zombie goroutine for that duration` sounds heavy, but is only true if no other dial attempts succeed within that duration. It also won't start new cleaners from other failed dial attempts within that duration. So at most you have one cleaner running that has nothing to clean.

wucm667:
Good catch, you're right about the race condition. I didn't consider the case where dial 1 fails while dial 2 is still in progress — that would indeed leave connections without a cleaner.

And you're also right that the zombie goroutine situation isn't as bad as I described. At most there's one idle cleaner goroutine sleeping, not a continuous leak.

I'll close this PR. Thanks for the thorough review.

PR #2199, opened 2026-05-02
feat(prefork): graceful shutdown, leak fixes, hook robustness (re-open of #2180 follow-up)

ReneWerner87:
Re-opens the follow-up commit that was applied to master right before #2180 was merged and then reverted in [e9208ec](https://github.com/valyala/fasthttp/commit/e9208ec) because @erikdubbelboer hadn't reviewed it. Splitting it back out so it can be reviewed on its own merits.

Original PR (merged): #2180
Reverted commit: 262ea09 -> e9208ec

---

## Summary

Addresses outstanding review concerns from #2180 plus several adjacent issues surfaced during a follow-up review pass.

### Lifecycle / supervision

- Track every per-child `Wait` goroutine via `sync.WaitGroup` and unblock pending `sigCh` sends through a `context.Cancel` so early-return paths (`OnChildSpawn` / `OnMasterReady` error, recovery `doCommand` error, `ErrOverRecovery`) can no longer leak goroutines or stall children.
- Replace the unconditional `SIGKILL` defer with a `SIGTERM`-then-`SIGKILL` sequence gated by a configurable `ShutdownGracePeriod` (defaults to 5s; the Windows path stays `SIGKILL` since `Process.Signal(SIGTERM)` is unsupported there).
- On shutdown, cancel the per-child Wait goroutines up front so a child that already exited while parked on its `RecoverInterval` backoff cannot delay teardown for a full interval (or the grace period). `cmd.Wait()` is independent of the context, so the graceful wait still tracks the real process exits.

### API

- Add `RecoverInterval` (optional per-child crash-loop backoff) and `ShutdownGracePeriod` fields. `RecoverInterval` is opt-in: a zero value keeps the previous immediate respawn. `ShutdownGracePeriod` defaults to 5s when zero, so teardown sends `SIGTERM` and waits up to the grace period before falling back to `SIGKILL` (this intentionally replaces the previous unconditional `SIGKILL`).
- Export `ErrCommandProducerNilCmd` and `ErrCommandProducerNotStarted` sentinel errors so callers can `errors.Is` them.
- Rename `oldPid`/`newPid` to `oldPID`/`newPID` per Go initialism convention.
- Add an explicit compile-time `var _ Logger = fasthttp.Logger(nil)` check so the local `Logger` interface stays in sync with `fasthttp.Logger`.

### Resource hygiene

- Master now closes both the original `tcpListener` and the duped fd in `p.files` when `prefork()` returns; previously the duped fd leaked once per call.
- `doCommand` wraps every error path with `%w` + `fmt.Errorf` so caller-side diagnostics keep stage context.
- Strip pre-existing `FASTHTTP_PREFORK_CHILD` entries before appending so child env never carries duplicate keys.
- Extract magic numbers as package constants (`inheritedListenerFD`, `masterPollInterval`, `defaultShutdownGracePeriod`, `preforkChildEnvValue`).
- Name the inherited listener fd via `os.NewFile` so `net.FileListener` errors are diagnosable.

### Tests

- Migrate to `t.Setenv` (drop the global `setUp/tearDown` helpers), fixing the env-mutation-vs-parallel race.
- Replace the `rand.Intn` port helper with `:0` + `Listener.Addr()` to remove port-collision flakes under `-count` / parallel runs.
- Collapse the three near-identical `Test_ListenAndServe*` tests into a single table-driven subtest that actually asserts the args forwarded to `ServeFunc` / `ServeTLSFunc` / `ServeTLSEmbedFunc`.
- Add coverage for the previously untested branches: `CommandProducer` returning err / nil cmd / unstarted cmd, initial `OnChildSpawn` error, `OnMasterReady` error, `RecoverInterval` enforcement, the `childEnv` marker dedup, and that shutdown does not block on a pending `RecoverInterval` backoff.
- `noopChildProducer` helper kills + waits any spawned child binaries during cleanup so failed tests no longer leave subprocesses around.

---

Local run on this branch: `go test -race ./prefork/...` and `golangci-lint run ./prefork/...` are both clean. Happy to split this into smaller PRs if you'd prefer to land the goroutine-leak fix and the API change separately from the new fields.


prefork/prefork.go
```diff
@@ -185,8 +262,10 @@ func (p *Prefork) listen(addr string) (net.Listener, error) {
 		return reuseport.Listen(p.Network, addr)
 	}
 
-	// File descriptor 3 is the first ExtraFiles entry passed by the master process.
-	return net.FileListener(os.NewFile(3, ""))
+	// fd inheritedListenerFD is the first ExtraFiles entry passed by the
+	// master process when Reuseport is false. Naming the file gives clearer
+	// errors from net.FileListener if the fd is invalid.
+	return net.FileListener(os.NewFile(inheritedListenerFD, "fasthttp-prefork-listener"))
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 257c1f9: `os.NewFile` is now bound to a local `f`, `net.FileListener(f)` runs, then `f.Close()` is called before returning. The returned listener owns the dup’d fd and is unaffected; the original wrapper no longer leaks. If FileListener fails *and* Close also fails, the close error is wrapped and returned.

prefork/prefork.go
```diff
@@ -213,62 +292,142 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
-		return err
+		return fmt.Errorf("prefork: listen tcp %s: %w", addr, err)
 	}
 
-	p.ln = tcplistener
+	p.ln = tcpListener
 
-	fl, err := tcplistener.File()
+	listenerFile, err := tcpListener.File()
 	if err != nil {
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 257c1f9: `tcpListener.File()` is now called *before* `p.ln` is assigned. If `File()` fails we `tcpListener.Close()` and return early, so neither the bound socket nor `p.ln` is left in a half-initialised state.

prefork/prefork.go
```diff
@@ -278,109 +437,157 @@ func (p *Prefork) prefork(addr string) (err error) {
...
 
-		// Start Wait goroutine before callbacks to avoid zombie processes
-		// if a callback fails and the deferred Kill() runs.
-		go func(c *exec.Cmd, pid int) {
-			sigCh <- procSig{pid: pid, err: c.Wait()}
-		}(cmd, newPid)
+			if p.RecoverInterval > 0 {
+				select {
+				case <-time.After(p.RecoverInterval):
+				case <-signalCh:
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 257c1f9: replaced with `time.NewTimer`. When the shutdown signal wins the select we now call `timer.Stop()` and drain the channel if Stop returned false, so the timer goroutine is reclaimed and no extra channel allocation lingers across crash-loop restarts.

prefork/prefork.go
```diff
@@ -140,20 +205,32 @@ func (p *Prefork) logger() Logger {
 	return defaultLogger
 }
 
+// invokeHook runs fn under a panic recovery, returning the panic as an error
+// so a misbehaving callback never tears down the master.
+func (p *Prefork) invokeHook(name string, fn func() error) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
+			err = fmt.Errorf("prefork: %s panicked: %v", name, r)
+			p.logger().Printf("%v", err)
```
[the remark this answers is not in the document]
ReneWerner87: Fixed in 257c1f9: removed the `p.logger().Printf(...)` line from inside `invokeHook`. The recover block now only sets `err = fmt.Errorf("prefork: %s panicked: %v", ...)`; logging is left to the call sites (which already log the returned error), so panics are reported exactly once. Comment on the function clarifies the contract.

prefork/prefork.go
```diff
@@ -76,25 +118,45 @@ type Prefork struct {
...
+	// replaced. It receives the PID of the old (crashed) process and the PID
+	// of its replacement.
+	//
+	// If this callback returns an error, all running children are killed and
+	// the prefork operation returns that error.
+	//
+	// Threading: invoked synchronously from the master goroutine, after
+	// OnChildSpawn for the new child. Panics are recovered and surfaced as
+	// the returned error.
+	OnChildRecover func(oldPID, newPID int) error
```
erikdubbelboer: What is the reason for this to be able to return an error?

prefork/prefork.go
```diff
@@ -213,62 +307,145 @@ func (p *Prefork) setTCPListenerFiles(addr string) error {
...
+	childProcs map[int]*exec.Cmd,
+	wg *sync.WaitGroup,
+	cancel context.CancelFunc,
+	grace time.Duration,
+) {
+	if grace <= 0 {
+		grace = defaultShutdownGracePeriod
+	}
+
+	if runtime.GOOS != "windows" {
```
erikdubbelboer: On windows the children don't get any signal that they have to shut down. But we're still waiting `grace` duration for nothing to happen. Does that make sense?

prefork/prefork.go
```diff
@@ -278,109 +455,164 @@ func (p *Prefork) prefork(addr string) (err error) {
...
-	}
+	// ctx cancels per-child Wait goroutines so they unblock from sigCh sends
+	// once the supervision loop is gone.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Catch SIGTERM/SIGINT in the master so we run our shutdown path instead
+	// of being killed by the OS without children getting a graceful chance.
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, syscall.SIGTERM, syscall.SIGINT)
```
erikdubbelboer: I'm not sure if this should be done by default. Maybe the program using fasthttp is already doing it's own thing on SIGTERM.

prefork/prefork.go
```diff
@@ -140,20 +205,32 @@ func (p *Prefork) logger() Logger {
 	return defaultLogger
 }
 
+// invokeHook runs fn under a panic recovery, returning the panic as an error
+// so a misbehaving callback never tears down the master. Diagnostic logging
+// is left to the call site so the same condition is not reported twice.
+func (p *Prefork) invokeHook(name string, fn func() error) (err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
erikdubbelboer: I would prefer not to do this. Let's leave it to the library user to decide what to do with panics.

prefork/prefork_test.go
```diff
@@ -53,51 +87,37 @@ func Test_New(t *testing.T) {
...
-
-	if reflect.ValueOf(p.ServeTLSEmbedFunc).Pointer() != reflect.ValueOf(s.ServeTLSEmbed).Pointer() {
-		t.Errorf("Prefork.ServeTLSFunc == %p, want %p", p.ServeTLSEmbedFunc, s.ServeTLSEmbed)
+	if p.ServeFunc == nil || p.ServeTLSFunc == nil || p.ServeTLSEmbedFunc == nil {
+		t.Error("New() did not wire one of ServeFunc/ServeTLSFunc/ServeTLSEmbedFunc")
 	}
 }
 
-func Test_listen(t *testing.T) {
+func Test_listen_Reuseport(t *testing.T) {
```
erikdubbelboer: Looking at this test again, it's not really testing anything. Please remove it.

prefork/prefork_test.go
```diff
@@ -53,51 +87,37 @@ func Test_New(t *testing.T) {
 	if p.Network != defaultNetwork {
```
erikdubbelboer: `Test_New` is also not really testing anything useful, please remove it.

prefork/prefork_test.go
```diff
@@ -108,144 +128,187 @@ func Test_setTCPListenerFiles(t *testing.T) {
...
-
-	addr := getAddr()
-
-	err := p.ListenAndServe(addr)
-	if err != nil {
-		t.Errorf("Unexpected error: %v", err)
-	}
-
-	p.ln.Close()
+func Test_setTCPListenerFiles_BadAddr(t *testing.T) {
```
erikdubbelboer: Please remove this test, its just testing if `net.ResolveTCPAddr` works.

prefork/prefork_test.go
```diff
@@ -108,144 +128,187 @@ func Test_setTCPListenerFiles(t *testing.T) {
...
+	tests := []struct {
+		name    string
+		produce func(files []*os.File) (*exec.Cmd, error)
+		wantErr error
+	}{
+		{
+			name: "producer returns error",
+			produce: func([]*os.File) (*exec.Cmd, error) {
+				return nil, errors.New("boom")
+			},
```
erikdubbelboer, line 307: Shouldn't this have a `wamtErr`?

prefork/prefork_test.go
```diff
@@ -362,56 +418,154 @@ func Test_Prefork_Lifecycle(t *testing.T) {
...
+	}
+
+	err := p.prefork(freeAddr(t))
+	if err == nil {
+		t.Fatal("expected error from panicking hook, got nil")
+	}
+}
+
+// Test_Prefork_RecoverInterval verifies the optional backoff delays the
+// respawn without relying on brittle wallclock assertions.
```
erikdubbelboer: This comment is not true, it does assert the wall time.

prefork/prefork_test.go
```diff
@@ -3,44 +3,78 @@ package prefork
...
+	"time"
 
 	"github.com/valyala/fasthttp"
 )
 
-func setUp() {
-	os.Setenv(preforkChildEnvVariable, "1")
+// freeAddr returns a TCP4 address that the OS believes is free at call time.
+// It avoids the rand-port collision flakes the previous helper exhibited.
+func freeAddr(t testing.TB) string {
```
erikdubbelboer: `freeAddr()` is not super safe, it doesn't lock the port so it could be used by another process in the meantime. I think you can replace most (if not all) uses of `freeAddr()` with just `127.0.0.1:0`.

erikdubbelboer (changes requested):

prefork/prefork.go
ReneWerner87: Reverted in b2c87c1: `OnChildRecover` no longer returns an error and is back to a notification-only callback.

prefork/prefork.go
ReneWerner87: Fixed in b2c87c1: the Windows shutdown path now skips the grace wait and kills children directly, since no SIGTERM is delivered there.

prefork/prefork.go
ReneWerner87: Fixed in b2c87c1: removed the default `signal.Notify` handling from prefork so applications keep ownership of their own SIGTERM/SIGINT policy.

prefork/prefork.go
ReneWerner87: Fixed in b2c87c1: removed hook panic recovery and deleted `invokeHook`, so panics are left to the library user as requested.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: removed `Test_listen_Reuseport`.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: removed `Test_New`.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: removed `Test_setTCPListenerFiles_BadAddr`.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: added the missing `wantErr` for the producer error case so the wrapped producer error is asserted with `errors.Is`.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: updated the RecoverInterval test comment so it no longer claims to avoid wallclock assertions.

prefork/prefork_test.go
ReneWerner87: Fixed in b2c87c1: removed `freeAddr()` and switched the remaining test callers to `127.0.0.1:0`.

ReneWerner87:
@erikdubbelboer can you check

prefork/prefork.go
```diff
@@ -76,25 +117,40 @@ type Prefork struct {
...
+	// OnMasterReady is called in the master process exactly once, after all
+	// initial children have been spawned and before the supervision loop runs.
+	// It receives a slice of all initial child PIDs.
+	//
+	// If this callback returns an error, the prefork operation aborts and
+	// returns that error after killing the children.
 	//
-	// If this callback returns an error, the prefork operation will be aborted.
+	// Threading: invoked synchronously from the master goroutine. The slice
+	// is owned by the caller after the call returns.
```
erikdubbelboer: Why? Doesn't it make much more sense that the callback copies the slice if it wants to keep a copy? Now you're always making a copy even if that isn't needed.

prefork/prefork.go
```diff
@@ -278,109 +448,126 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+			)
+			return ErrOverRecovery
 		}
 
-		var cmd *exec.Cmd
-		cmd, err = p.doCommand()
-		if err != nil {
-			break
+		if p.RecoverInterval > 0 {
+			time.Sleep(p.RecoverInterval)
```
erikdubbelboer: Does it make sense for this to block the whole loop? If I set this to 10 seconds and I have multiple childs restarting they will take 10 seconds in between restarts instead of actually restarting 10 seconds after they quit.

erikdubbelboer (changes requested):
You write:
```
Install signal.Notify(SIGTERM, SIGINT) in the master so deploy /
rolling-restart signals enter the shutdown path instead of killing
the master without graceful teardown.
```
But that doesn't happen.

Similar you write `OnChildRecover now returns error`, but that's also not true.

```
Add RecoverInterval (optional crash-loop backoff) and ShutdownGracePeriod
fields with safe zero-value defaults — both are opt-in, existing behaviour
is preserved when zero.
```
This is also no true, on zero it will wait for 5 seconds while existing behavior is to terminate immediately.

prefork/prefork.go
ReneWerner87: Fixed in 2554b7a: dropped the defensive copy, so OnMasterReady now receives the internal childPIDs slice directly and no allocation happens when it isn't needed. The doc now states the slice is only valid for the duration of the call, so a callback that wants to keep the PIDs copies them itself.

prefork/prefork.go
ReneWerner87: Fixed in 2554b7a: moved the RecoverInterval backoff out of the supervision loop into the per-child Wait goroutine, so it no longer blocks the loop. Each crashed child now waits RecoverInterval after it exits before its exit is reported and respawned, independently of the others. With your example (10s, multiple children restarting) they each restart ~10s after they quit instead of 10s apart. The wait is interruptible via ctx so a shutdown doesn't have to outlast a full interval.

ReneWerner87:
updated for the review hints

erikdubbelboer:
Can you please rebase this on master, I made some changes today and now there are merge conflicts.

ReneWerner87:
@erikdubbelboer refreshed, the test fail is not related to my changes

prefork/prefork.go
```diff
@@ -281,114 +456,141 @@ func (p *Prefork) prefork(addr string) (err error) {
...
+			result := childExit{pid: pid, err: cmd.Wait()}
+
+			// Apply the crash-loop backoff here, per child, before reporting
+			// the exit. Sleeping in the supervision loop instead would
+			// serialize the wait across all crashed children, so N children
+			// dying at once would restart RecoverInterval apart rather than
+			// each one RecoverInterval after it quit. The wait is interruptible
+			// so shutdown does not have to outlast a full interval.
+			if p.RecoverInterval > 0 {
+				timer := time.NewTimer(p.RecoverInterval)
```
erikdubbelboer, line 514: This can make normal shutdown wait for `RecoverInterval`. The wait goroutine is part of `wg`, but `shutdownChildren` waits for `wg` before calling `cancel()`. So if `prefork()` returns while a child has already exited and this goroutine is sleeping here, shutdown can block until this timer fires, or until `ShutdownGracePeriod` expires, even though the child is already gone.

erikdubbelboer (changes requested):

prefork/prefork.go
ReneWerner87: ok

ReneWerner87: Fixed in a5a00a0: `shutdownChildren` now calls `cancel()` up front, before waiting on `wg`. A child that already exited while parked on the `RecoverInterval` backoff (or a `sigCh` send) is unblocked immediately via `ctx.Done()`, so it no longer holds up shutdown for a full `RecoverInterval` or `ShutdownGracePeriod`. `cmd.Wait()` is independent of the context, so the graceful `SIGTERM` wait still tracks the real process exits. Added `Test_Prefork_ShutdownDoesNotBlockOnRecoverInterval` as a regression guard (it forces an immediate return via an `OnMasterReady` error while children sit in a 10s backoff and asserts shutdown completes well under that).

ReneWerner87:
@erikdubbelboer updated

erikdubbelboer:
Thanks!

PR #2201, opened 2026-05-06
client: add aggregate connection count metrics

### LINKED ISSUE #2200: Client: add aggregate connection pool metrics for observability
## Feature Request

The `fasthttp.Client` type is a multi-host wrapper that dynamically creates `HostClient` instances per hostname. While `HostClient` exposes `ConnsCount()` for per-host connection tracking, the top-level `Client` has no equivalent method to aggregate connection pool state across all managed hosts.

This makes it impossible to monitor the overall connection pool health when using `fasthttp.Client` — the most common entry point for production code.

**Current gap:**

```go
c := &fasthttp.Client{MaxConnsPerHost: 100}
// No way to get: total active connections, total idle connections, 
// per-host breakdown, or pool utilization across all hosts
c.CloseIdleConnections() // only method available, no metrics
```

**Why this matters:**
- **Capacity planning**: Without aggregate metrics, operators cannot determine if `MaxConnsPerHost` is appropriately sized
- **Leak detection**: Connection leaks are hard to spot without being able to monitor total connection counts over time
- **Auto-scaling**: Connection pool saturation is a valuable signal for scaling decisions
- **Debugging**: When `ErrNoFreeConns` occurs, there is no way to introspect which hosts are consuming connections

**Describe the solution you would like**

Add aggregate metrics methods to `fasthttp.Client`:

```go
// ConnsCount returns total active + idle connections across all HostClients.
func (c *Client) ConnsCount() int

// IdleConnsCount returns total idle connections across all HostClients.
func (c *Client) IdleConnsCount() int

// Optional: detailed per-host breakdown for debugging
type ClientPoolStats struct {
    TotalConns    int
    IdleConns     int
    HostStats     map[string]HostPoolStats
}

func (c *Client) PoolStats() ClientPoolStats
```

The implementation would iterate `c.m` (the `map[string]*HostClient`) under `mLock.RLock()`, sum up per-host `ConnsCount()` values, and optionally collect per-host details.

**Describe alternatives you have considered**
- Wrapping `Client` with custom dialer tracking (loses internal pool semantics)
- Using `HostClient` directly (forfeits the convenience of automatic per-host client management)
- External metrics via HTTP client middleware (cannot observe pool-internal state like idle vs active)

**Additional context**
This is complementary to existing per-host observability (`HostClient.ConnsCount()`) and would make `Client` production-observable without changing the user-facing API.

z9z:
Adds Client.ConnsCount() and Client.IdleConnsCount(), aggregating across all managed HostClients (both http and https   maps). Also adds HostClient.IdleConnsCount() since there was no public accessor for per-host idle count.

Closes #2200

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -74,6 +74,64 @@ func TestCloseIdleConnections(t *testing.T) {
 	}
 }
 
+func TestClientConnectionCounts(t *testing.T) {
+	t.Parallel()
+
+	ln := fasthttputil.NewInmemoryListener()
+
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+		},
+	}
+	go func() {
+		if err := s.Serve(ln); err != nil {
+			t.Error(err)
+		}
+	}()
+
+	c := &Client{
+		Dial: func(addr string) (net.Conn, error) {
+			return ln.Dial()
+		},
+	}
+
+	if n := c.ConnsCount(); n != 0 {
+		t.Errorf("unexpected connection count %d. Expecting %d", n, 0)
+	}
+	if n := c.IdleConnsCount(); n != 0 {
+		t.Errorf("unexpected idle connection count %d. Expecting %d", n, 0)
+	}
+
+	if _, _, err := c.Get(nil, "http://google.com"); err != nil {
+		t.Fatal(err)
+	}
+
+	if n := c.ConnsCount(); n != 1 {
+		t.Errorf("unexpected connection count %d. Expecting %d", n, 1)
+	}
+	if n := c.IdleConnsCount(); n != 1 {
+		t.Errorf("unexpected idle connection count %d. Expecting %d", n, 1)
+	}
+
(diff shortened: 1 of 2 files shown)


client_test.go
```diff
@@ -74,6 +74,64 @@ func TestCloseIdleConnections(t *testing.T) {
...
+
+	var c Client
+	if n := c.ConnsCount(); n != 0 {
+		t.Fatalf("unexpected connection count %d. Expecting %d", n, 0)
+	}
+	if n := c.IdleConnsCount(); n != 0 {
+		t.Fatalf("unexpected idle connection count %d. Expecting %d", n, 0)
+	}
+}
+
```
erikdubbelboer: This is already tested in the above test as well, please remove this test.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2206, opened 2026-05-17
client: add Client.DoContext and HostClient.DoContext

### LINKED ISSUE #2198: Client: add DoContext method for context.Context-aware request execution
## Feature Request

The fasthttp client provides `DoTimeout()` and `DoDeadline()`, but lacks a `DoContext(ctx context.Context, req, resp)` method. This means:

1. **No cancellation propagation**: If a parent context is cancelled (e.g., an HTTP handler's `ctx.Done()`), the in-flight fasthttp request continues until its own timeout or completion.
2. **No deadline inheritance**: If a caller has an existing context with a deadline (e.g., from gRPC or an upstream request), there's no way to pass it through.
3. **No metadata attachment**: Tracing systems (OpenTelemetry, etc.) rely on context to propagate trace IDs.

This forces users to either:
- Use `net/http` (which has `Client.Do(req.WithContext(ctx))`) and lose performance
- Manually create a `context.WithTimeout` and call `DoDeadline`, losing parent cancellation

**Describe the solution you'd like**

Add a `DoContext` method to both `Client` and `HostClient`:

```go
func (c *Client) DoContext(ctx context.Context, req *Request, resp *Response) error
func (c *HostClient) DoContext(ctx context.Context, req *Request, resp *Response) error
```

Behavior:
- If `ctx` is already cancelled, return `ctx.Err()` immediately
- If `ctx` has a deadline, use it (respecting any existing timeout set on the Client)
- If the context is cancelled during dial, reading, or writing, abort the operation and return `ctx.Err()`
- The implementation can reuse the existing timeout/deadline logic in the TCP dialer, which already uses `context.Context` internally

**Describe alternatives you've considered**

- Using `DoDeadline` with `ctx.Deadline()` — works but loses cancellation propagation (only deadline is respected, not `ctx.Done()`)
- Wrapping `DoTimeout` in a goroutine with `select { case <-ctx.Done() }` — racy, doesn't actually cancel the underlying I/O
- Using `net/http` — defeats the purpose of using fasthttp

**Additional context**

- The internal `tcpdialer` already uses `context.Context` for DNS resolution (`Resolver.LookupIPAddr(ctx, host)`)
- `Server.ShutdownWithContext(ctx)` already exists, showing the project embraces context for lifecycle management
- Other Go HTTP clients (net/http, resty, go-resty) all support context-aware requests

z9z:
Closes #2198.

Adds DoContext on Client and HostClient. Thin wrapper that honors ctx.Deadline() via req.timeout and returns ctx.Err() when the context is cancelled. No changes to the internal call chain. Cancellation is bounded by req.timeout (the inner Do completes naturally before DoContext returns).

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,6 +3,7 @@ package fasthttp
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -527,6 +528,13 @@ func (c *Client) Do(req *Request, resp *Response) error {
 	atomic.AddInt32(&hc.pendingClientRequests, 1)
 	defer atomic.AddInt32(&hc.pendingClientRequests, -1)
 	return hc.Do(req, resp)
+}
+
+// DoContext is like Do but uses ctx for deadline and cancellation.
+// If ctx is cancelled, the underlying connection is closed to interrupt
+// any in-flight read or write.
+func (c *Client) DoContext(ctx context.Context, req *Request, resp *Response) error {
+	return doContext(ctx, req, resp, c.Do)
 }
 
 func (c *Client) hostClient(host []byte, isTLS bool) (*HostClient, error) {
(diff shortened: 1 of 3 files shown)


erikdubbelboer (changes requested):
I'm not sure about this. The current implementation can't actually supports context cancellation in the usual Go sense.

It only copies `ctx.Deadline()` into `req.timeout` and checks `ctx.Err()` before/after `Do`. Calling `cancel()` from another goroutine won’t interrupt connection wait, dial, write, or read as expected. `DoContext` only returns after the underlying `Do` finishes naturally or hits a timeout.

It also overwrites any existing `req.timeout` instead of using the earlier of the request timeout and context deadline.

PR #2211, opened 2026-05-22
fix double release of streamed client response body

Mereng:
Fix double release of the underlying streamed response body when a client response `BodyStream()` is closed directly.

When `StreamResponseBody` is enabled, the transport wraps `resp.bodyStream` with a  `ReadCloserWithError` that releases the reader, closes or releases the backend connection, and releases the underlying `requestStream`.

If user code passes `resp.BodyStream()` elsewhere and that stream is closed directly with `CloseWithError`, the wrapper already releases the underlying `requestStream`. A later `ReleaseResponse(resp)` calls `Response.Reset()`, sees the same `resp.bodyStream`, and can release the same `requestStream` again.

This change clears `resp.bodyStream` when the wrapper is closed, making subsequent `ReleaseResponse` / `CloseBodyStream` calls no-op for that stream.

This can happen in proxy-style code that forwards `resp.BodyStream()` into another response writer and closes it when the downstream client disconnects.

### DIFF
--- a/client.go
+++ b/client.go
@@ -3205,7 +3205,11 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 	closeConn := resetConnection || req.ConnectionClose() || resp.ConnectionClose()
 	if customStreamBody && resp.bodyStream != nil {
 		rbs := resp.bodyStream
+		var closed atomic.Bool
 		resp.bodyStream = newCloseReaderWithError(rbs, func(wErr error) error {
+			if !closed.CompareAndSwap(false, true) {
+				return nil
+			}
 			hc.ReleaseReader(br)
 			if r, ok := rbs.(*requestStream); ok {
 				releaseRequestStream(r)
--- a/http_test.go
+++ b/http_test.go
@@ -3349,6 +3349,73 @@ func TestResponseBodyStream(t *testing.T) {
 			}
 		})
 
+		t.Run("direct close with error releases client stream", func(t *testing.T) {
+			t.Parallel()
+
+			client := Client{StreamResponseBody: true, MaxConnsPerHost: 1}
+			resp := AcquireResponse()
+			defer ReleaseResponse(resp)
+			request := AcquireRequest()
+			defer ReleaseRequest(request)
+			request.SetRequestURI(server.URL + "?chunked=true")
+			if err := client.Do(request, resp); err != nil {
+				t.Fatal(err)
+			}
+			stream, ok := resp.BodyStream().(ReadCloserWithError)
+			if !ok {
+				t.Fatalf("unexpected body stream type %T", resp.BodyStream())
+			}
+			if err := stream.CloseWithError(errors.New("client closed")); err != nil {
+				t.Fatalf("close body stream err: %v", err)
+			}
+			if err := stream.CloseWithError(errors.New("client closed again")); err != nil {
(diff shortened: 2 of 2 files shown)


erikdubbelboer (changes requested):
Thanks for the pull request but I think this needs one change before merging.

The new identity guard makes `CloseWithError` return before releasing the client reader/connection whenever `resp.bodyStream` has been replaced before the old wrapper is closed. That can happen through the existing compressed response paths: `gzipBody` / `deflateBody` / `brotliBody` / `zstdBody` replace `resp.bodyStream` with a `NewStreamReader`, then close the previous stream later.

A minimal regression test for that case:

```go
t.Run("compressed write releases client stream", func(t *testing.T) {
	t.Parallel()

	client := Client{StreamResponseBody: true, MaxConnsPerHost: 1}

	resp := AcquireResponse()
	defer ReleaseResponse(resp)
	request := AcquireRequest()
	defer ReleaseRequest(request)

	request.SetRequestURI(server.URL)
	if err := client.Do(request, resp); err != nil {
		t.Fatal(err)
	}

	bw := bufio.NewWriter(io.Discard)
	if err := resp.WriteGzip(bw); err != nil {
		t.Fatalf("write gzip response err: %v", err)
	}
	if err := bw.Flush(); err != nil {
		t.Fatalf("flush gzip response err: %v", err)
	}

	resp2 := AcquireResponse()
	defer ReleaseResponse(resp2)
	request2 := AcquireRequest()
	defer ReleaseRequest(request2)

	request2.SetRequestURI(server.URL)
	if err := client.Do(request2, resp2); err != nil {
		t.Fatalf("second request err: %v", err)
	}
})

**Comment by Mereng** [comment]
@erikdubbelboer Good catch, thanks.

  I made the identity guard less strict: it now only guards clearing `resp.bodyStream`, so compressed
  response paths can still close the previous stream and release the client reader/connection.

  I also kept an explicit `closed` guard because the wrapper owns those resources, and `CloseWithError`
  may be reached more than once through different close paths. With this, the original stream resources
  are released once, while `resp.bodyStream` is cleared only when it still points to the same wrapper.

**Review by erikdubbelboer — CHANGES_REQUESTED** [review]
You now have a race condition. I think something like this should probably work (haven't tested it, writing this on my phone):
```go
var closed atomic.Bool

resp.bodyStream = newCloseReaderWithError(rbs, func(wErr error) error {
	if !closed.CompareAndSwap(false, true) {
		return nil
	}

	hc.ReleaseReader(br)
		releaseRequestStream(r)
	}
	if closeConn || resp.ConnectionClose() || wErr != nil {
		hc.CloseConn(cc)
	} else {
		hc.ReleaseConn(cc)
	}
	return nil
})
```

Mereng:
Oh, my fault.  Yes, we only need to protect the wrapper-owned resources from being released twice, clearing bodyStream is not necessary.   I updated it to use atomic.Bool as suggested

Mereng:
@erikdubbelboer This looks unrelated to this PR. The failure is in `TestServerPipelineFlush`, which seems timing-
  sensitive: it expects the first pipelined response before the delayed second request completes.

  I reran that test locally with the same shuffle seed and `-count=20`, and it passes. I think this is
  likely a Windows/Go 1.26 flaky run. A rerun should confirm.

erikdubbelboer:
Thanks!

PR #2259, opened 2026-05-28
fix(fs): bound memory of filesLockMap via fixed-size shards

### LINKED ISSUE #2256: bug: filesLockMap grows unboundedly, leaking memory
## Description

`filesLockMap` in `fs.go:1929-1934` uses a `sync.Map` that stores a mutex for every unique compressed file path ever encountered. Over a long-running server serving many unique files, this map grows without bound:

```go
var filesLockMap sync.Map

func getFileLock(path string) *sync.Mutex {
    v, _ := filesLockMap.LoadOrStore(path, &sync.Mutex{})
    return v.(*sync.Mutex)
}
```

Entries are never removed, even after files are deleted or the compressed cache is cleaned.

## Impact

- Unbounded memory growth proportional to the number of unique file paths served
- Each entry is a `*sync.Mutex` (~8 bytes) + map overhead + key string

## Suggested Fix

Use an LRU-based cleanup, a fixed-size sharded lock map, or periodically evict old entries when the map exceeds a threshold.

## File

`fs.go:1929-1934`

u5surf:
Replaces the unbounded sync.Map keyed by absolute file path with a fixed-size pool of 1024 mutexes selected by FNV-1a hash of the path. This guards against long-running fasthttp servers leaking memory proportional to the number of unique compressed file paths ever served.

Distinct paths may share a mutex (1 in 1024), but contention only occurs on the cache-miss / compression-generation slow path, so the practical impact is negligible. Fixes #2256.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"html"
 	"io"
 	"io/fs"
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thank you for your pull request, but I'm sorry I didn't really like the pool of mutex solution as it would cause issues with hash collisions. I have applied a different fixed in #2286

PR #2260, opened 2026-05-28
fix(server): return ErrMissingFile when MultipartForm.File is nil

### LINKED ISSUE #2235: bug: FormFile returns (nil, nil) when MultipartForm.File is nil
## Description

In `server.go:1147-1148`, `FormFile` returns `(nil, err)` when `mf.File == nil`. But at this point `err` is `nil` (from the successful `MultipartForm()` call), so the function returns `(nil, nil)` — no file header and no error.

```go
mf, err := ctx.MultipartForm()
if err != nil {
    return nil, err
}
if mf.File == nil {
    return nil, err  // BUG: err is nil here
}
```

This was previously reported in #528 but as a panic. The panic was fixed, but the nil-error return was not.

## Impact

Callers checking `if err != nil` will proceed with a nil `*multipart.FileHeader`, likely causing a nil pointer dereference downstream.

## Suggested Fix

```go
if mf.File == nil {
    return nil, ErrMissingFile
}
```

u5surf:
RequestCtx.FormFile previously returned (nil, nil) when the parsed multipart form had a nil File map: the code returned the err variable from the preceding successful MultipartForm() call, which is always nil at that point. Callers checking `if err != nil` proceeded with a nil *multipart.FileHeader, causing nil pointer dereferences downstream.

Return ErrMissingFile instead, matching the existing semantics for the missing-key case immediately below. Fixes #2235.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1145,7 +1145,7 @@ func (ctx *RequestCtx) FormFile(key string) (*multipart.FileHeader, error) {
 		return nil, err
 	}
 	if mf.File == nil {
-		return nil, err
+		return nil, ErrMissingFile
 	}
 	fhh := mf.File[key]
 	if fhh == nil {


server_test.go
```diff
@@ -4890,3 +4890,29 @@ func TestRequestCtxInitShouldNotBeCanceledIssue1879(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// Regression test for #2235: FormFile must not return (nil, nil) when the
+// parsed multipart form has a nil File map. Previously the code returned the
+// nil error from the preceding MultipartForm() call, causing callers checking
+// `if err != nil` to proceed with a nil *multipart.FileHeader and nil-deref
+// downstream.
+func TestRequestCtxFormFileNilFileMapIssue2235(t *testing.T) {
```
erikdubbelboer: There's no need to add a test for such a simple change.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2262, opened 2026-06-01
fix(pprofhandler): use exact path matching to prevent prefix-based route confusion

### LINKED ISSUE #2258: security: pprofhandler imprecise path matching exposes debug data
## Description

`pprofhandler/pprof.go:27-38` uses `bytes.HasPrefix` for routing, which matches prefix substrings:

```go
case bytes.HasPrefix(ctx.Path(), []byte("/debug/pprof/cmdline")):
```

This means `/debug/pprof/cmdlineFoo` matches the `cmdline` handler. Similarly for other profile paths.

## Impact

- Debug data exposed under slightly malformed paths
- Affects production deployments that expose pprof without strict path matching

## Suggested Fix

Use exact match:

```go
case bytes.Equal(ctx.Path(), []byte("/debug/pprof/cmdline")):
```

Or check for trailing `/` or end-of-path.

## File

`pprofhandler/pprof.go:27-38`

toller892:
## Problem

PprofHandler uses  for matching pprof endpoint paths. This means paths like  incorrectly match the  handler,  matches , etc.

Fixes #2258

## Fix

Switch from  to  for the four named handlers (cmdline, profile, symbol, trace). For runtime profile names in the default case, check that the path either exactly matches or has a trailing slash boundary.

Changes:
- : Replace  with  for exact path matching
- : Add tests verifying that malformed paths with extra suffixes don't match specific handlers

## Testing

All existing functionality preserved — exact paths still resolve correctly. Malicious/malformed paths with extra suffixes no longer match specific handlers.

### DIFF
--- a/pprofhandler/pprof_test.go
+++ b/pprofhandler/pprof_test.go
@@ -1,0 +1,71 @@
+package pprofhandler
+
+import (
+	"net"
+	"testing"
+
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttputil"
+)
+
+func TestPprofHandler_PathMatching(t *testing.T) {
+	ln := fasthttputil.NewInmemoryListener()
+	defer ln.Close()
+
+	s := &fasthttp.Server{
+		Handler: PprofHandler,
+	}
+	go func() {
+		_ = s.Serve(ln)
+	}()
+
+	client := &fasthttp.Client{
+		Dial: func(addr string) (net.Conn, error) {
+			return ln.Dial()
+		},
+	}
+
+	doReq := func(path string) (int, []byte) {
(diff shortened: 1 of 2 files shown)


pprofhandler/pprof_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"net"
+	"testing"
+
+	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttputil"
+)
+
+func TestPprofHandler_PathMatching(t *testing.T) {
```
erikdubbelboer, line 11: Can you please remove the test. This is such a simple change, it doesn't need this.

erikdubbelboer (changes requested):

PR #2263, opened 2026-06-02
fix(fasthttpproxy): propagate GetDialFunc error instead of returning nil DialFunc

### LINKED ISSUE #2248: bug: fasthttpproxy dialers return nil DialFunc on error, causing panic
## Description

In `fasthttpproxy`, all four public dialer constructors discard the error from `GetDialFunc`:

```go
// http.go:33, proxy_env.go:37, socks5.go:18, socks5.go:32
dialFunc, _ := d.GetDialFunc(false)  // error ignored
return dialFunc                       // nil if GetDialFunc errored
```

If the proxy URL is malformed or uses an unknown scheme, `GetDialFunc` returns `(nil, err)`. The returned `fasthttp.DialFunc` is `nil`, and the next `Dial` call panics with a nil pointer dereference.

## Impact

- Server crash from malformed proxy configuration
- No error message to help diagnose the issue

## Suggested Fix

```go
dialFunc, err := d.GetDialFunc(false)
if err != nil {
    return func(addr string) (net.Conn, error) { return nil, err }
}
return dialFunc
```

Or change the API to return errors (breaking change).

## File

`fasthttpproxy/http.go:33`, `fasthttpproxy/proxy_env.go:37`, `fasthttpproxy/socks5.go:18,32`

sapirbaruch:
Fixes #2248.

## Problem

The five public dialer constructors in `fasthttpproxy` all ignore the error returned by `GetDialFunc`:

```go
// http.go, socks5.go, proxy_env.go
dialFunc, _ := d.GetDialFunc(false)
return dialFunc  // nil when GetDialFunc failed
```

`GetDialFunc` returns `(nil, err)` when the proxy URL is malformed or uses an unknown scheme. The silent `_` discard means callers receive a `nil` `fasthttp.DialFunc`, and any subsequent call to `Dial` panics with a nil pointer dereference.

## Fix

Wrap the error in a closure that propagates it on every `Dial` call — the same pattern the standard library uses for similar situations:

```go
dialFunc, err := d.GetDialFunc(false)
if err != nil {
    return func(addr string) (net.Conn, error) { return nil, err }
}
return dialFunc
```

This preserves the existing API (callers still get a non-nil `DialFunc`) while surfacing the configuration error at dial time rather than panicking.

**Affected functions:**
- `FasthttpHTTPDialerTimeout`
- `FasthttpHTTPDialerDualStackTimeout`
- `FasthttpSocksDialer`
- `FasthttpSocksDialerDualStack`
- `FasthttpProxyHTTPDialerTimeout`

### DIFF
--- a/fasthttpproxy/http.go
+++ b/fasthttpproxy/http.go
@@ -1,6 +1,7 @@
 package fasthttpproxy
 
 import (
+	"net"
 	"time"
 
 	"github.com/valyala/fasthttp"
@@ -30,7 +31,10 @@ func FasthttpHTTPDialer(proxy string) fasthttp.DialFunc {
 //	}
 func FasthttpHTTPDialerTimeout(proxy string, timeout time.Duration) fasthttp.DialFunc {
 	d := Dialer{Config: httpproxy.Config{HTTPProxy: proxy, HTTPSProxy: proxy}, Timeout: timeout, ConnectTimeout: timeout}
-	dialFunc, _ := d.GetDialFunc(false)
+	dialFunc, err := d.GetDialFunc(false)
+	if err != nil {
+		return func(addr string) (net.Conn, error) { return nil, err }
+	}
 	return dialFunc
 }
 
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Thank you for your pull request, but I'm afraid that this was now already fixed in #2279

PR #2264, opened 2026-06-02
fix(fs): remove temporary file when compression fails

### LINKED ISSUE #2239: bug: temporary file leak when file compression fails
## Description

In `fs.go:1586-1629`, if `os.Create(tmpFilePath)` succeeds but the compression step fails, the temporary file is not cleaned up:

```go
zf, err := os.Create(tmpFilePath)
// ... compression fails ...
// tmpFilePath remains on disk
```

## Impact

- Stale compressed temporary files accumulate on disk
- Over time, this can exhaust disk space
- Files contain partial compressed data

## Suggested Fix

Add cleanup in the error path:

```go
if err != nil {
    os.Remove(tmpFilePath)
    return err
}
```

## File

`fs.go:1586-1629`

u5surf:
compressFileNolock created a ".tmp" file but left it on disk when the compression copy, os.Chtimes, or os.Rename step failed, leaking partial files that accumulate over time. Remove the temporary file in each error path.

Fixes #2239

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1617,13 +1617,16 @@ func (h *fsHandler) compressFileNolock(
 	_ = zf.Close()
 	_ = f.Close()
 	if err != nil {
+		_ = os.Remove(tmpFilePath)
 		return nil, fmt.Errorf("error when compressing file %q to %q: %w", filePath, tmpFilePath, err)
 	}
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2265, opened 2026-06-02
fix(client): switch to GET on 303 redirect per RFC 9110

### LINKED ISSUE #2240: bug: HTTP 303 redirect does not switch POST to GET (RFC 7231 violation)
## Description

In `client.go:1224`, the redirect logic only converts POST→GET for 301 and 302 status codes:

```go
if string(req.Header.Method()) == string(MethodPost) && (statusCode == 301 || statusCode == 302) {
}
```

Per RFC 7231 Section 6.4.4, a **303 (See Other)** response MUST cause the client to change the method to GET. `StatusCodeIsRedirect` correctly recognizes 303 as a redirect, but the method switch is missing.

## Impact

- POST requests redirected with 303 will continue as POST to the new location
- Violates HTTP/1.1 specification
- May cause unexpected behavior with APIs that rely on 303 semantics

## Suggested Fix

```go
if string(req.Header.Method()) == string(MethodPost) && (statusCode == 301 || statusCode == 302 || statusCode == 303) {
}
```

Per RFC 9110 §15.4.4, 303 should also drop the request body.

## File

`client.go:1224`

u5surf:
RFC 9110 section 15.4.4, a 303 (See Other) response must be followed by a body-less GET regardless of the original method. 

https://datatracker.ietf.org/doc/html/rfc9110#name-303-see-other

Handle 303 explicitly: change any non-GET/HEAD method to GET, drop the body, and clear Content-Length/Content-Type. 301/302 keep their POST-to-GET behavior, and 307/308 still preserve the method and body.

Fixes #2240

### DIFF
--- a/client.go
+++ b/client.go
@@ -1221,7 +1221,29 @@ func doRequestFollowRedirects(
 		stripSensitiveHeadersOnRedirect(req, initialHost, redirectURI)
 		ReleaseURI(redirectURI)
 
-		if string(req.Header.Method()) == "POST" && (statusCode == 301 || statusCode == 302) {
+		switch {
+		case statusCode == StatusSeeOther:
+			// RFC 9110 section 15.4.4: a 303 (See Other) response redirects
+			// the user agent to retrieve the new URI with a GET (or HEAD)
+			// request, regardless of the original method, and without the
+			// original request body. Drop the body together with every header
+			// that frames it: per RFC 9112 a body is signaled by Content-Length
+			// or Transfer-Encoding, and a Trailer only applies to a chunked
+			// body, so all of them must go once the body is gone.
+			if !req.Header.IsGet() && !req.Header.IsHead() {
+				req.Header.SetMethod(MethodGet)
+			}
+			req.Header.Del(HeaderContentLength)
+			req.Header.Del(HeaderContentType)
+			req.Header.Del(HeaderTransferEncoding)
+			req.Header.Del(HeaderTrailer)
+			req.ResetBody()
+			req.postArgs.Reset()
+			req.parsedPostArgs = false
+		case req.Header.IsPost() && (statusCode == StatusMovedPermanently || statusCode == StatusFound):
+			// RFC 9110 sections 15.4.2/15.4.3 Note: for historical reasons a
+			// user agent MAY change the request method from POST to GET for a
+			// 301 (Moved Permanently) or 302 (Found) response.
 			req.Header.SetMethod(MethodGet)
 		}
 	}
--- a/client_test.go
+++ b/client_test.go
@@ -2024,6 +2024,139 @@ func TestClientFollowRedirects(t *testing.T) {
 	ReleaseResponse(resp)
 }
 
+func TestClientRedirectMethodSwitch(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


client.go
```diff
@@ -1221,7 +1221,24 @@ func doRequestFollowRedirects(
 		stripSensitiveHeadersOnRedirect(req, initialHost, redirectURI)
 		ReleaseURI(redirectURI)
 
-		if string(req.Header.Method()) == "POST" && (statusCode == 301 || statusCode == 302) {
+		switch {
+		case statusCode == StatusSeeOther:
+			// RFC 9110 section 15.4.4: a 303 (See Other) response redirects
+			// the user agent to retrieve the new URI with a GET (or HEAD)
+			// request, regardless of the original method, and without the
+			// original request body.
```
erikdubbelboer: This fails if the original was already a `GET`:
```go
func TestClientRedirect303DropsGetRequestBody(t *testing.T) {
	t.Parallel()

	s := &Server{
			switch string(ctx.Path()) {
			case "/redirect":
				ctx.Redirect("/landing", StatusSeeOther)
			case "/landing":
				ctx.SetBodyString(string(ctx.Method()) + "|" + string(ctx.PostBody()))
			default:
				ctx.Error("not found", StatusNotFound)
			}
		},
	}
	serverStopCh := make(chan struct{})
	go func() {
		if err := s.Serve(ln); err != nil && !errors.Is(err, fasthttputil.ErrInmemoryListenerClosed) {
			t.Errorf("unexpected error: %v", err)
		}
		close(serverStopCh)
	}()
	t.Cleanup(func() {
		if err := ln.Close(); err != nil && !errors.Is(err, fasthttputil.ErrInmemoryListenerClosed) {
			t.Errorf("unexpected error: %v", err)
		}
		select {
		case <-serverStopCh:
		case <-time.After(time.Second):
			t.Errorf("timeout")
		}
	})

	c := &HostClient{
		Addr: "xxx",
			return ln.Dial()
		},
	}

	req := AcquireRequest()
	resp := AcquireResponse()
	defer ReleaseRequest(req)
	defer ReleaseResponse(resp)

	req.SetRequestURI("http://xxx/redirect")
	req.SetBodyString("hello")

	if err := c.DoRedirects(req, resp, 16); err != nil {
		t.Fatalf("unexpected error: %v", err)
	}
	if statusCode := resp.StatusCode(); statusCode != StatusOK {
		t.Fatalf("unexpected status code: %d", statusCode)
	}
	if body := string(resp.Body()); body != "GET|" {
		t.Fatalf("unexpected response %q. Expecting %q", body, "GET|")
	}
}
```

erikdubbelboer (changes requested):

client.go
```diff
@@ -1221,7 +1221,24 @@ func doRequestFollowRedirects(
...
+			if !req.Header.IsGet() && !req.Header.IsHead() {
+				req.Header.SetMethod(MethodGet)
+			}
+			req.Header.Del(HeaderContentLength)
+			req.Header.Del(HeaderContentType)
```
erikdubbelboer, line 1237: According to the spec you also need to remove `Transfer-Encoding` and Trailers.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2266, opened 2026-06-02
fix(client): guard nil waiter in decConnsCount

### LINKED ISSUE #2242: bug: potential nil pointer dereference in decConnsCount
## Description

`decConnsCount` in `client.go:1896-1905` calls `q.popFront()` without checking for nil:

```go
if q := c.connsWait; q != nil && q.len() > 0 {
    for q.len() > 0 {
        w := q.popFront()
        if w.waiting() {  // nil dereference if popFront returns nil
            go c.dialConnFor(w)
            dialed = true
            break
        }
        c.connsWait.failedWaiters.Add(-1)
    }
}
```

`q.len()` uses the atomic `failedWaiters` counter, which is incremented in `AcquireConn` outside `connsLock`. This creates a window where `len()` returns a positive value but the queue is actually empty, causing `popFront()` to return nil.

## Impact

- Nil pointer dereference panic in the connection management goroutine
- Can crash the server under high concurrency with connection pool pressure

## Suggested Fix

```go
w := q.popFront()
if w == nil {
    break
}
if w.waiting() {
```

## File

`client.go:1896-1905`

SAY-5:
wantConnQueue.len() subtracts the atomic failedWaiters counter, which is incremented in AcquireConn when a waiter times out before that waiter is removed from the queue. This means len() can report a positive value while popFront() actually returns nil: decConnsCount then dereferences the nil in w.waiting() and panics the connection-management path under MaxConnWaitTimeout.

Break out of the loop when popFront returns nil, matching the nil check already used elsewhere when draining the queue. Added a regression test that reproduces the panic without the guard.

Fixes #2242.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1963,6 +1963,9 @@ func (c *HostClient) ReleaseConn(cc *clientConn) {
 	if q := c.connsWait; q != nil && q.len() > 0 {
 		for q.len() > 0 {
 			w := q.popFront()
+			if w == nil {
+				break
+			}
 			if w.waiting() {
 				delivered = w.tryDeliver(cc, nil)
 				// This is the last resort to hand over conCount sema.
--- a/client.go
+++ b/client.go
@@ -1755,7 +1755,6 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
 		case <-w.ready:
 			return w.conn, w.err
 		case <-tc.C:
-			c.connsWait.failedWaiters.Add(1)
 			if timeoutOverridden {
 				return nil, ErrTimeout
 			}
@@ -1893,18 +1892,14 @@ func (c *HostClient) decConnsCount() {
 	c.connsLock.Lock()
 	defer c.connsLock.Unlock()
 	dialed := false
-	if q := c.connsWait; q != nil && q.len() > 0 {
+	if q := c.connsWait; q != nil {
 		for q.len() > 0 {
 			w := q.popFront()
-			if w == nil {
-				break
-			}
 			if w.waiting() {
 				go c.dialConnFor(w)
 				dialed = true
 				break
 			}
-			c.connsWait.failedWaiters.Add(-1)
 		}
 	}
 	if !dialed {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -1896,6 +1896,9 @@ func (c *HostClient) decConnsCount() {
 	if q := c.connsWait; q != nil && q.len() > 0 {
 		for q.len() > 0 {
 			w := q.popFront()
+			if w == nil {
+				break
+			}
```
erikdubbelboer: This avoids the nil deref in `decConnsCount`, but the same `q.popFront()` -> `q.len() > 0`/`q.waiting()` is still in other places. Can you apply the same fix everywhere?

client_test.go
```diff
@@ -3862,3 +3862,28 @@ func TestClient_RetryIfErrUpstream(t *testing.T) {
 		}
 	})
 }
+
+func TestHostClientDecConnsCountStaleWaiters(t *testing.T) {
```
erikdubbelboer: Can you explain why you wrote this test?

It's a scenario that doesn't reflect the code?

erikdubbelboer (changes requested):

client.go
erikdubbelboer: Thanks, but I think the remaining problem is with `failedWaiters`, not just the missing nil guard.

It is still possible that `ReleaseConn` pops a waiter and successfully calls `tryDeliver`, but the waiting goroutine selects the timer case and increments `failedWaiters` for that waiter after it has already left the queue. That stale counter then makes `q.len()` underreport. If another real waiter is queued later, `q.len()` can return 0 and `ReleaseConn` will skip delivery, leaving the connection idle while the waiter remains blocked.

So I would rather fix the counter/queue accounting directly. For example, avoid subtracting `failedWaiters` from `len()` and treat `len()` as the physical queue length, then clean non-waiting entries while popping, similar to net/http’s `wantConnQueue`.

erikdubbelboer (changes requested):

PR #2268, opened 2026-06-04
Fix POST redirects after 303 responses

### LINKED ISSUE #2240: bug: HTTP 303 redirect does not switch POST to GET (RFC 7231 violation)

puneetdixit200:
Fixes #2240.

Summary:
- Treat a POST followed by `303 See Other` like the existing 301/302 redirect rewrite and continue with `GET`.
- Clear the original request body and stale `Content-Length` before the redirected request is sent.
- Add a regression test that verifies the redirected request reaches the target as `GET` without the POST body.

Tests:
- `go test ./ -run '^TestClientPostRedirectSeeOtherUsesGetWithoutBody$' -count=1 -timeout=30s`
- `go test ./ -run 'TestClient(FollowRedirects|PostRedirectSeeOtherUsesGetWithoutBody)$|TestShouldStripSensitiveHeadersOnRedirect$|Test_getRedirectURL$' -count=1`
- `go test ./... -count=1`
- `git diff --check`

### DIFF
--- a/client.go
+++ b/client.go
@@ -1221,8 +1221,11 @@ func doRequestFollowRedirects(
 		stripSensitiveHeadersOnRedirect(req, initialHost, redirectURI)
 		ReleaseURI(redirectURI)
 
-		if string(req.Header.Method()) == "POST" && (statusCode == 301 || statusCode == 302) {
+		if string(req.Header.Method()) == MethodPost &&
+			(statusCode == StatusMovedPermanently || statusCode == StatusFound || statusCode == StatusSeeOther) {
 			req.Header.SetMethod(MethodGet)
+			req.ResetBody()
+			req.Header.SetContentLength(0)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thank you for submitting this pull request, but I'm afraid this was now already fixed in #2265

PR #2269, opened 2026-06-04
fix(fasthttputil): add stop channel to InmemoryListener to avoid dead…

dima11223432:
## Description
Fixes a critical deadlock and potential panic in `InmemoryListener` under high concurrency or when `Close()` is called while `Dial()` calls are blocked.

### The Problem
1. **Deadlock**: In `DialWithLocalAddr`, the listener held `ln.lock` while performing blocking channel operations (`ln.conns <- req` and `<-accepted`). If the buffer was full, it blocked holding the mutex, preventing `Close()` from acquiring the same lock.
2. **Panic**: Attempting to fix this by simply unlocking early introduced a `panic: send on closed channel` because `Close()` explicitly closed `ln.conns` while concurrent writers (`Dial`) were randomly selected to execute the send case in a multi-case `select`.

### The Solution
- Introduced a dedicated signaling channel `stopCh` to handle listener termination.
- Removed `close(ln.conns)` from `Close()` to avoid panics on concurrent sends. The channel will be garbage collected normally.
- Updated `DialWithLocalAddr` and `Accept` to leverage `stopCh` inside non-blocking `select` blocks outside of the critical section (`ln.lock`).
- Added strict `ln.closed` guards at the entry of `Accept` and `DialWithLocalAddr` to comply with standard `net.Listener` behavior after closure.

## Verification
- All existing package tests pass successfully.
- Added a new concurrent stress test `TestInmemoryListener_Deadlock_Fix` which reliably reproduces the deadlock on the master branch and passes smoothly with this fix.

### DIFF
--- a/fasthttputil/inmemory_listener.go
+++ b/fasthttputil/inmemory_listener.go
@@ -19,6 +19,7 @@ type InmemoryListener struct {
 	addrLock     sync.RWMutex
 	lock         sync.Mutex
 	closed       bool
+	stopCh       chan struct{}
 }
 
 type acceptConn struct {
@@ -29,7 +30,8 @@ type acceptConn struct {
 // NewInmemoryListener returns new in-memory dialer<->net.Listener.
 func NewInmemoryListener() *InmemoryListener {
 	return &InmemoryListener{
-		conns: make(chan acceptConn, 1024),
+		conns:  make(chan acceptConn, 1024),
+		stopCh: make(chan struct{}),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thank you for your contribution, but this has now already been fixed in #2277

dima11223432:
It's really unpleasant, I did it 2 days before and you didn't even launch the pipeline :(

erikdubbelboer:
Sorry, I only noticed it now after I committed the fix I build a couple of days ago. The repo got a lot of AI attention (like your pull request), so I didn't have time to read everything.

PR #2287, opened 2026-06-08
fix(lbclient): prevent deadlock when all clients are removed

### LINKED ISSUE #2270: Found a bug in LBClient.get() that causes a permanent deadlock when cs becomes empty.
When using AddClient/RemoveClients for dynamic backend management, if all backends are removed, cc.cs becomes an empty slice. Any concurrent DoDeadline call at that point will call get(), which takes cc.mu.RLock() and then immediately accesses cs[0]. Since cs is empty, this panics with "index out of range [0] with length 0". Because RUnlock is not deferred, it never runs, and the RLock is leaked permanently.

The consequence is that any subsequent call to AddClient or RemoveClients tries to acquire cc.mu.Lock() and blocks forever. The lock is never released, so even if new backends are registered they never make it into cs.

We hit this in a Kubernetes setup where backends are pod IPs registered dynamically. During a pod restart, RemoveClients removes all backends, cs goes empty, the next incoming request panics in get(), RLock leaks, and the following AddClient for the new pod deadlocks. The service never recovers until the process is restarted.

u5surf:
`LBClient.get()` takes an `RLock` and then immediately reads `cs[0]` without checking for an empty slice:

Once every client is removed via `RemoveClients`, `cc.cs` becomes empty. The next concurrent `DoDeadline`/`Do` calls `get()`, panics on `cs[0]`, and because `RUnlock` is not deferred it is never reached — the read lock leaks permanently. Every subsequent `AddClient`/`RemoveClients` then blocks forever on `cc.mu.Lock()`, so the balancer never recovers (e.g. in a Kubernetes setup where backends are added/removed dynamically during a pod restart).

Fixes #2270.

### DIFF
--- a/lbclient.go
+++ b/lbclient.go
@@ -1,10 +1,15 @@
 package fasthttp
 
 import (
+	"errors"
 	"sync"
 	"sync/atomic"
 	"time"
 )
+
+// ErrNoAvailableClients is returned by LBClient methods when no clients are
+// available, for example after every client has been removed.
+var ErrNoAvailableClients = errors.New("no available clients")
 
 // BalancingClient is the interface for clients, which may be passed
 // to LBClient.Clients.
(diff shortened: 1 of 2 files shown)


lbclient.go
erikdubbelboer, line 133: If `rc` panics it will still result in a broken state because it did `cc.cs[idx] = nil`.

It's probably better to only do `cc.cs[idx] = nil` after.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2288, opened 2026-06-08
ci: re-enable gocritic deferInLoop check for non-test code

### LINKED ISSUE #2233: safety: re-enable deferInLoop gocritic check to catch defer-in-loop resource leaks
## Description

The `deferInLoop` check from `gocritic` is explicitly disabled in `.golangci.yml` (line 50):

```yaml
gocritic:
  disabled-checks:
    - deferInLoop   # <-- disabled
```

This check detects `defer` statements inside `for` loops. `defer` runs at **function exit**, not loop iteration, so resources (files, connections, locks) accumulate until the function returns — causing memory leaks and file descriptor exhaustion under load.

## Why This Matters

While the current codebase doesn't have defer-in-loop bugs (audited manually), disabling this check means **CI won't catch future regressions**.

fasthttp is a high-performance server where:
- Connection handling loops are common
- Resource leaks under load are catastrophic (fd exhaustion → server hangs)
- Contributors may not realize `defer` in a `for` loop doesn't run per iteration

## Example of the Bug This Prevents

```go
// This would NOT be caught by CI with deferInLoop disabled
func (s *Server) processConnections(conns []net.Conn) {
    for _, conn := range conns {
        defer conn.Close() // BUG: all conns stay open until function returns
        process(conn)
    }
}
```

## Suggested Fix

1. Remove `deferInLoop` from the `disabled-checks` list in `.golangci.yml`
2. If any existing code triggers the check, add `//nolint:gocritic // deferInLoop: extracted to func` annotations with justification
3. This ensures future contributors are warned when they add `defer` inside a loop

## Impact

- Medium severity — preventive measure, no current bugs
- Zero runtime cost — lint-only change
- Protects against a common Go mistake that is silent until production load exposes it

u5surf:
deferInLoop was disabled in .golangci.yml, so CI could not catch defer statements added inside for loops. Such defers run only at function exit, not per iteration, leaking file descriptors, connections, or locks under load.

Re-enable the check, but exclude _test.go files: deferring cleanup to the end of a short-lived test is a common and harmless pattern, while the leak matters in the long-running server and connection-handling loops of the non-test code, which is now guarded against regressions.

Fixes #2233

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -47,7 +47,6 @@ linters:
   settings:
     gocritic:
       disabled-checks:
-        - deferInLoop
         - importShadow
         - sloppyReassign
         - unnamedResult
--- a/client_test.go
+++ b/client_test.go
@@ -2943,12 +2943,20 @@ func TestClientHTTPSConcurrent(t *testing.T) {
 func TestClientManyServers(t *testing.T) {
 	t.Parallel()
 
+	servers := make([]*testEchoServer, 0, 10)
 	addrs := make([]string, 0, 10)
 	for range 10 {
 		s := startEchoServer(t, "tcp", "127.0.0.1:")
-		defer s.Stop()
+		servers = append(servers, s)
 		addrs = append(addrs, s.Addr())
 	}
+	// All servers must stay up until the test ends, so stop them with a single
+	// deferred loop rather than deferring inside the loop above.
+	defer func() {
+		for _, s := range servers {
+			s.Stop()
+		}
+	}()
 
 	var wg sync.WaitGroup
 	for i := range 4 {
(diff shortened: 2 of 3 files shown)


.golangci.yml
```diff
@@ -83,6 +82,10 @@ linters:
       - linters:
           - lll
         path: _test\.go
+      - linters:
+          - gocritic
+        text: deferInLoop
+        path: _test\.go
```
erikdubbelboer: There's only a single defer in a loop. If you create a closure there it should be fine and you don't need to exclude all test files.

erikdubbelboer (changes requested):

erikdubbelboer:
Thanks!

PR #2294, opened 2026-06-15
reject duplicate Content-Length in response header parser

alhudz:
`Repro`: a response with two differing `Content-Length` lines.

`HTTP/1.1 200 OK\r\nContent-Length: 10\r\nContent-Length: 20\r\n\r\n`

`Expected`: rejected, like the request parser (#2190) and `net/http` (`message cannot contain multiple Content-Length headers`).
`Actual`: accepted, `ContentLength()` returns `20` (last value silently wins), which is a response desync vector for clients/proxies.

1. `ResponseHeader.parseHeaders` never tracked whether a `Content-Length` was already seen, so the second header overwrote the first.
2. `RequestHeader.parseHeaders` already returns `ErrDuplicateContentLength` for this case.

Mirrored the request-side check in the response parser and moved the existing duplicate-`Content-Length` case in `TestResponseHeaderReadError`.

### DIFF
--- a/header.go
+++ b/header.go
@@ -2993,6 +2993,7 @@ func (h *ResponseHeader) parseHeaders(buf []byte) (int, error) {
 	s.b = buf
 	var kv *argsKV
 	transferEncodingSeen := false
+	contentLengthSeen := false
 
 	for s.next() {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2295, opened 2026-06-18
strip cr/lf from header keys in normalizeHeaderKey

alhudz:
Repro: `h.Set("X-Foo\r\nInjected: true", "bar")` on a `ResponseHeader` (or `RequestHeader`), then read `h.Header()` — the output carries a second `Injected: true` line on the wire.
Cause: header values are sanitised against CR/LF in `initHeaderValueBytes`, but keys are not. `normalizeHeaderKey` only fixes case and returns early when the key holds a space or any other invalid byte, so an embedded `\r`/`\n` survives and `appendHeaderLine` writes the key verbatim, splitting the header block.
Fix: strip CR/LF in `normalizeHeaderKey`, the single point every key setter (`Set`/`Add`/`SetBytesKV`/`SetBytesV`/`AppendNormalizedHeaderKey`, both request and response) funnels through, mirroring the value path. Trailer keys (`isValidTrailerKey`) and cookie fields already reject/strip these bytes.

### DIFF
--- a/header.go
+++ b/header.go
@@ -3404,6 +3404,12 @@ func getHeaderKeyBytes(bufK []byte, key string, disableNormalizing bool) []byte 
 }
 
 func normalizeHeaderKey(b []byte, disableNormalizing bool) {
+	// Neutralise any CR/LF in the key so a header name can't break the
+	// message framing, mirroring the value handling in initHeaderValueBytes.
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2296, opened 2026-06-18
Avoid default content type for empty requests

### LINKED ISSUE #873: [Bug/Question] Hostclient sends content type header on DELETE request even if it is not set

nkgotcode:
Fixes #873.

This avoids adding the default `Content-Type` header when writing requests with no body, such as an empty `DELETE` request. Requests with a non-zero content length still keep the existing default content type behavior.

Validation:

```sh
go test ./ -run 'TestRequestContentType|TestRequestWriteDeleteNoDefaultContentType'
go test ./... -run 'TestRequestContentType|TestRequestWriteDeleteNoDefaultContentType'
go list ./... | rg -v '^github\.com/valyala/fasthttp$' | xargs go test
```

### DIFF
--- a/header.go
+++ b/header.go
@@ -2613,7 +2613,7 @@ func (h *RequestHeader) AppendBytes(dst []byte) []byte {
 	}
 
 	contentType := h.ContentType()
-	if !h.noDefaultContentType && len(contentType) == 0 && h.ContentLength() != 0 {
+	if !h.noDefaultContentType && len(contentType) == 0 && h.ContentLength() > 0 {
 		contentType = strDefaultContentType
 	}
 	if len(contentType) > 0 && !h.disableSpecialHeader {
--- a/header.go
+++ b/header.go
@@ -2613,7 +2613,7 @@ func (h *RequestHeader) AppendBytes(dst []byte) []byte {
 	}
 
 	contentType := h.ContentType()
-	if !h.noDefaultContentType && len(contentType) == 0 && !h.ignoreBody() {
+	if !h.noDefaultContentType && len(contentType) == 0 && h.ContentLength() != 0 {
 		contentType = strDefaultContentType
 	}
 	if len(contentType) > 0 && !h.disableSpecialHeader {
--- a/header_test.go
+++ b/header_test.go
@@ -1755,6 +1755,33 @@ func TestRequestContentTypeNoDefault(t *testing.T) {
 	}
 }
 
+func TestRequestWriteDeleteNoDefaultContentType(t *testing.T) {
+	t.Parallel()
+
+	var req Request
+	req.Header.SetMethod(MethodDelete)
+	req.Header.SetHost("example.com")
+
+	w := &bytes.Buffer{}
+	bw := bufio.NewWriter(w)
+	if err := req.Write(bw); err != nil {
+		t.Fatalf("Unexpected error: %v", err)
+	}
+	if err := bw.Flush(); err != nil {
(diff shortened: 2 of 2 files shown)


erikdubbelboer (changes requested):
You have to run all the tests...

nkgotcode:
Updated after the full test run.

What changed:
- Kept the no-default-Content-Type behavior for empty-body requests.
- Fixed the condition so unspecified/identity content length does not count as a body, while positive content lengths still get the existing default Content-Type behavior.

Validation:
```sh
go test ./ -run 'TestRequestHeaderAdd|TestRequestContentType|TestRequestWriteDeleteNoDefaultContentType'
go test ./...
```

Both pass locally.

erikdubbelboer:
Thanks!

PR #2298, opened 2026-06-20
strip semicolons from cookie setters to block attribute injection

alhudz:
Repro: set any response cookie field from untrusted input that contains a `;`, e.g. `c.SetValue("abc; Domain=evil.example; Path=/admin; Secure")`, then serialise with `c.Cookie()`. Reparsing the output gives `domain=evil.example`, `path=/admin` and `secure=true`, so the attacker-controlled value injected real attributes a browser honours.
Cause: the cookie setters sanitise CR/LF (`removeNewLines`, #2185) but not `;`, the attribute delimiter in a `Set-Cookie` header, so `SetKey`/`SetValue`/`SetDomain`/`SetPath` and the `*Bytes` variants store it verbatim.
Fix: neutralise `;` in the four user-set components, the same way newlines already are.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -163,6 +163,7 @@ func (c *Cookie) SetPath(path string) {
 	c.bufK = append(c.bufK[:0], path...)
 	c.path = normalizePath(c.path, c.bufK)
 	c.path = removeNewLines(c.path)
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2300, opened 2026-06-22
fix: prevent data race on FSCompressedFileSuffixes map

### LINKED ISSUE #2223: bug: FSCompressedFileSuffixes is a mutable exported map — data race risk
## Description

`FSCompressedFileSuffixes` (fs.go:456) is an exported mutable `map[string]string`. It is read during `FSHandler` initialization inside a `sync.Once`, but being exported, any caller can concurrently read/write this map while the `Once` executes.

```go
var FSCompressedFileSuffixes = map[string]string{
    "br":   ".br",
    "gzip": ".gz",
}
```

If a user modifies this map concurrently with `FS` initialization, it's a data race on the map (concurrent map read/write causes a fatal panic in Go).

## Impact

- Fatal crash (`concurrent map read and map write`) if modified after import while `FS` is being initialized concurrently.
- Low probability in typical usage but high severity when it occurs.

## Suggested Fix

Either:
1. Unexport the map and provide a `SetCompressedFileSuffixes()` function that copies.
2. Copy the map once at init time in the `sync.Once` block, so the original can be freely modified.

xbrxr03:
## Update

Per maintainer feedback, simplified this to documentation only. The original approach added `SetCompressedFileSuffix()` and a `sync.RWMutex`, but @erikdubbelboer pointed out that `FSCompressedFileSuffixes` is meant to be set once at startup and not modified concurrently.

The fix is now just a doc comment clarifying that the map should be set in `init()` or before the first `FS.NewRequestHandler` call, and not modified afterwards.

Fixes #2223

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -454,6 +454,11 @@ const FSCompressedFileSuffix = ".fasthttp.gz"
 // FSCompressedFileSuffixes is the suffixes FS adds to the original file names depending on encoding
 // when trying to store compressed file under the new file name.
 // See FS.Compress for details.
+//
+// This map is read during FSHandler initialization (which runs in a sync.Once).
+// It is not safe for concurrent modification. Set any custom suffixes in an
+// init() function or before the first call to FS.NewRequestHandler, and do not
+// modify the map afterwards.
 var FSCompressedFileSuffixes = map[string]string{
 	"gzip": ".fasthttp.gz",
 	"br":   ".fasthttp.br",


erikdubbelboer (changes requested):
I'm sorry but I don't really like changing the API like this. This is something that you're supposed to set in an `init()` or at the start of `main()` and then don't change anymore. Instead of all these changes lets just document that. Can you please change this.

xbrxr03:
Updated per your feedback — removed the API changes and just added documentation. The map should be set in `init()` and not modified afterwards. Thanks for the guidance!

erikdubbelboer:
Thanks!

PR #2301, opened 2026-06-23
validate trailer values in parseTrailer

alhudz:
`Repro:` send a chunked message ending in an undeclared trailer whose value holds a bare `\r`, e.g. `X-Sneaky: a\rInjected: yes`. After `Response.Read` the bare CR survives into `resp.Header.Header()` because an undeclared trailer is merged into the normal header set and re-emitted verbatim by `AppendBytes`. A regular header with the same byte is refused (`invalid header value`).

`Cause:` `parseTrailer` validated only the trailer key (and `isBadTrailer`), but never the value, whereas both `parseHeaders` paths run every value byte through `validHeaderValueByte`. So a trailer can smuggle control bytes (bare CR, NUL, other CTLs) past the validation the rest of the stack enforces.

`Fix:` run the same per-byte `validHeaderValueByte` check over the trailer value in `parseTrailer`, rejecting with `invalid trailer value`. Normal trailer values are unaffected.

### DIFF
--- a/header.go
+++ b/header.go
@@ -2729,6 +2729,11 @@ func parseTrailer(src []byte, dest []argsKV, disableNormalizing bool) ([]argsKV,
 		if isBadTrailer(s.key) {
 			return dest, 0, fmt.Errorf("forbidden trailer key %q", s.key)
 		}
+		for _, ch := range s.value {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2302, opened 2026-06-23
fix(pprofhandler): use exact path matching to prevent debug data exposure

### LINKED ISSUE #2258: security: pprofhandler imprecise path matching exposes debug data

xbrxr03:
## Summary

`PprofHandler` uses `bytes.HasPrefix` for routing, which means `/debug/pprof/cmdlineFoo` matches the `cmdline` handler. This leaks process command-line arguments (and other debug data) to any path that starts with a valid pprof endpoint prefix.

Fixes #2258.

## What changed

**`pprofhandler/pprof.go`**
- Replaced all `bytes.HasPrefix` route matches with `matchPprofPath`, which requires an exact path match or an exact match + trailing slash (matching `net/http/pprof`'s `DefaultServeMux` behaviour)
- The named-profile loop in the `default` branch also uses `matchPprofPath` instead of `HasPrefix`
- Added `matchPprofPath` helper function and pre-computed endpoint byte slices

**`pprofhandler/pprof_test.go`** (new)
- 20 tests covering exact matches, trailing slashes, prefix mismatches, and fallthrough behaviour
- Verifies that `/debug/pprof/cmdlineFoo` no longer serves the process command line

## Testing

All tests pass:
```
ok  github.com/valyala/fasthttp/pprofhandler  0.382s
```

Full repo test suite also passes — no regressions.

## Security impact

Before: `GET /debug/pprof/cmdlineFoo` → serves full command line
After: `GET /debug/pprof/cmdlineFoo` → "Unknown profile" (falls through to index handler)

This aligns fasthttp's pprof handler with the Go standard library's exact-match behaviour on `DefaultServeMux`.

### DIFF
--- a/pprofhandler/pprof.go
+++ b/pprofhandler/pprof.go
@@ -18,24 +18,50 @@ var (
 	index   = fasthttpadaptor.NewFastHTTPHandlerFunc(pprof.Index)
 )
 
+var (
+	cmdlinePath = []byte("/debug/pprof/cmdline")
+	profilePath = []byte("/debug/pprof/profile")
+	symbolPath  = []byte("/debug/pprof/symbol")
+	tracePath   = []byte("/debug/pprof/trace")
+)
+
+// pprofPrefix is the common prefix for all pprof paths.
+var pprofPrefix = []byte("/debug/pprof/")
+
+// matchPprofPath checks whether path exactly matches a pprof endpoint.
+// It accepts the exact path (e.g. /debug/pprof/heap) or the path with a
+// trailing slash (e.g. /debug/pprof/heap/), matching the behaviour of
+// net/http/pprof's DefaultServeMux registrations.
+func matchPprofPath(path, endpoint []byte) bool {
+	if bytes.Equal(path, endpoint) {
+		return true
+	}
+	// Allow trailing slash: /debug/pprof/heap/ matches /debug/pprof/heap
+	if len(path) == len(endpoint)+1 && path[len(path)-1] == '/' && bytes.Equal(path[:len(path)-1], endpoint) {
+		return true
+	}
+	return false
+}
+
 // PprofHandler serves server runtime profiling data in the format expected by the pprof visualization tool.
 //
 // See https://pkg.go.dev/net/http/pprof for details.
 func PprofHandler(ctx *fasthttp.RequestCtx) {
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Please fix the linting issues.

pprofhandler/pprof.go
```diff
@@ -18,24 +18,47 @@ var (
...
+// matchPprofPath checks whether path exactly matches a pprof endpoint.
+// It accepts the exact path (e.g. /debug/pprof/heap) or the path with a
+// trailing slash (e.g. /debug/pprof/heap/), matching the behaviour of
+// net/http/pprof's DefaultServeMux registrations.
+func matchPprofPath(path, endpoint []byte) bool {
+	if bytes.Equal(path, endpoint) {
+		return true
+	}
+	// Allow trailing slash: /debug/pprof/heap/ matches /debug/pprof/heap
+	if len(path) == len(endpoint)+1 && path[len(path)-1] == '/' && bytes.Equal(path[:len(path)-1], endpoint) {
```
erikdubbelboer: Let's match what `net/http` does, which returns a 404 for these paths ending in `/`

erikdubbelboer (changes requested):

xbrxr03:
Updated: matchPprofPath now uses strict equality only (removed trailing-slash matching). PprofHandler returns 404 for any path ending in /, matching net/http/pprof behavior. Also renamed the test from AcceptsTrailingSlash to RejectsTrailingSlash and updated the matchPprofPath test case to expect false for trailing slash.

All pprofhandler tests pass locally.

erikdubbelboer:
Thanks!

PR #2303, opened 2026-06-24
reject duplicate Transfer-Encoding in request header parser

alhudz:
`Repro`: a request carrying two `Transfer-Encoding` headers.

`POST / HTTP/1.1\r\nHost: x\r\nTransfer-Encoding: chunked\r\nTransfer-Encoding: chunked\r\n\r\n0\r\n\r\n`

`Expected`: rejected, like the response parser (`too many Transfer-Encoding headers`) and the request parser's own duplicate `Content-Length`/`Host` checks.
`Actual`: accepted and read as an ordinary chunked request, a `TE.TE` smuggling/desync vector when fasthttp sits behind or in front of a server that resolves the duplicate differently.

1. `RequestHeader.parseHeaders` never tracked whether a `Transfer-Encoding` header had already been seen, so a second one passed straight through.
2. `ResponseHeader.parseHeaders` already rejects this case.

Mirrored the response-side guard in the request parser and added the duplicate case to `TestRequestHeaderReadError`.

### DIFF
--- a/header.go
+++ b/header.go
@@ -3142,6 +3142,7 @@ func (h *RequestHeader) parseHeaders(buf []byte) (int, error) {
 	h.contentLength = -2
 
 	contentLengthSeen := false
+	transferEncodingSeen := false
 	hostSeen := false
 
 	var s headerScanner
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2304, opened 2026-06-25
fix(server): close listener on serve error in ListenAndServe helpers

fereidani:
Hi,

I've built a static analyzer for Go called [Ghoul](https://fereidani.com/ghoul), and I ran it against the fasthttp codebase. It flagged a few resource leaks in the ListenAndServe helpers, where a listener created with net.Listen is not closed when an error occurs before Serve takes ownership of it. It also flagged one piece of dead code.

Ghoul related reports:
```txt
server.go:1774:3: [resource-leak] resource from net.Listen allocated at server.go:1769 never closed (confidence: 6.4/10, CWE-404)
  --> server.go:1769:23: source:      resource allocated by net.Listen
server.go:1795:19: [resource-leak] resource from net.Listen passed to ServeTLS which may leak it on an error path (the helper owns it on some paths but not all) (confidence: 6.4/10, CWE-404)
  --> server.go:1791:23: source:      resource allocated by net.Listen
server.go:1814:24: [resource-leak] resource from net.Listen passed to ServeTLSEmbed which may leak it on an error path (the helper owns it on some paths but not all) (confidence: 6.4/10, CWE-404)
  --> server.go:1810:23: source:      resource allocated by net.Listen
peripconn.go:139:6: [dead-code] function uint322ip is never referenced; it is dead code (confidence: 7.4/10, CWE-563)
```

This change closes the listener on the error paths of `ListenAndServeUNIX` (chmod failure), `ListenAndServeTLS`, and `ListenAndServeTLSEmbed` (certificate load failure), so the file descriptor is released instead of leaked. The fix lives in the `ListenAndServe*` wrappers, which are the functions that allocate the listener, so the existing behavior of `ServeTLS` and `ServeTLSEmbed` (the caller owns the listener it passes in) is unchanged. I also added two regression tests that rebind the same address to confirm the listener is closed on the error path.

The `uint322ip` dead-code finding in `peripconn.go` is included above for awareness. I left it as-is because it is the inverse of the production-used `ip2uint32` and is covered by a round-trip test, but you may want to remove it if you consider it unused.

### DIFF
--- a/server.go
+++ b/server.go
@@ -1771,6 +1771,9 @@ func (s *Server) ListenAndServeUNIX(addr string, mode os.FileMode) error {
 		return err
 	}
 	if err = os.Chmod(addr, mode); err != nil {
+		// Close the listener so the unix socket file descriptor is not
+		// leaked when chmod fails before Serve takes ownership of it.
+		_ = ln.Close()
 		return fmt.Errorf("cannot chmod %#o for %q: %w", mode, addr, err)
 	}
 	return s.Serve(ln)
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

erikdubbelboer:
I'll remove `uint322ip` in a separate commit.

PR #2305, opened 2026-06-26
preserve duplicate request headers in ConvertRequest

alhudz:
1. `ConvertRequest` copies request headers with `r.Header.Set`, so a request that repeats a header (two `X-Forwarded-For`, several `Via`/`Forwarded`) reaches the `http.Handler` with only the last value.
2. The response path in `adaptor.go` already uses `Add`, and `net/http` keeps repeated headers as separate values, so middleware reading `Header.Values` behind the adaptor silently loses the earlier ones.
3. Switched the copy to `r.Header.Add`; added a regression test.

Repro: two `X-Forwarded-For` headers in; `Header.Values("X-Forwarded-For")` returned `["203.0.113.7"]` (expected both). With `Add` both survive.


erikdubbelboer:
Thanks!

PR #2306, opened 2026-06-29
Stackfull compressors

hablof:
Hello, why do we use stackless compressors? Is saving stack space really usefull? Any doc explaining?

I ran some benchmarks vs stackfull approach:
```
func BenchmarkStackfull(b *testing.B) {
	testdata := []byte(`hello world`)
	dst := make([]byte, 0, 1<<10)

	tests := []struct {
		name      string
		stackless func(dst []byte, src []byte) []byte
		stackfull func(dst []byte, src []byte) []byte
	}{
		{name: "brotli", stackless: AppendBrotliBytes, stackfull: AppendBrotliBytesStackfull},
		{name: "deflate", stackless: AppendDeflateBytes, stackfull: AppendDeflateBytesStackfull},
		{name: "gzip", stackless: AppendGzipBytes, stackfull: AppendGzipBytesStackfull},
	}

	for _, tt := range tests {
		b.Run(tt.name, func(b *testing.B) {
			b.Run("stackless", func(b *testing.B) {
				for b.Loop() {
					dst = tt.stackless(dst, testdata)[:0]
				}
			})
			b.Run("stackfull", func(b *testing.B) {
				for b.Loop() {
					dst = tt.stackfull(dst, testdata)[:0]
				}
			})
		})
	}
}
```
Here are the results:
```
goos: linux
goarch: amd64
cpu: AMD Ryzen 5 7500F 6-Core Processor             
BenchmarkStackfull/brotli/stackless-12         	  116010	     10878 ns/op	     132 B/op	       2 allocs/op
BenchmarkStackfull/brotli/stackfull-12         	  273409	      4419 ns/op	      24 B/op	       1 allocs/op
BenchmarkStackfull/deflate/stackless-12        	 2364337	       509.4 ns/op	     154 B/op	       2 allocs/op
BenchmarkStackfull/deflate/stackfull-12        	12102837	       101.1 ns/op	      29 B/op	       1 allocs/op
BenchmarkStackfull/gzip/stackless-12           	 2245543	       525.1 ns/op	     141 B/op	       2 allocs/op
BenchmarkStackfull/gzip/stackfull-12           	10197663	       115.5 ns/op	      28 B/op	       1 allocs/op
PASS
ok  	github.com/valyala/fasthttp	7.267s
```

In real apps that compress responses, it's even worse - chan send/recive inside stackless.Func triggers runtime scheduler, so you never konw when it will return to handler goroutine. While desireble workflow is process request completely before switching to the next one.

RFC:
* can we have Stackfull funcs as option?
* naming
* default behaviour (stackless vs stackfull)

### DIFF
--- a/brotli.go
+++ b/brotli.go
@@ -102,6 +102,21 @@ func AppendBrotliBytesLevel(dst, src []byte, level int) []byte {
 	return w.b
 }
 
+// AppendBrotliBytesLevelStackfull appends brotlied src to dst using the given
+// compression level and returns the resulting dst.
+//
+// Supported compression levels are:
+//
+//   - CompressBrotliNoCompression
+//   - CompressBrotliBestSpeed
+//   - CompressBrotliBestCompression
+//   - CompressBrotliDefaultCompression
+func AppendBrotliBytesLevelStackfull(dst, src []byte, level int) []byte {
+	w := &byteSliceWriter{b: dst}
+	WriteBrotliLevelStackfull(w, src, level) //nolint:errcheck
+	return w.b
+}
+
 // WriteBrotliLevel writes brotlied p to w using the given compression level
 // and returns the number of compressed bytes written to w.
 //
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
Thanks for digging into this and for the benchmark.

The current stackless behavior is intentional: it was added to bound memory usage under high concurrency. In particular, https://github.com/valyala/fasthttp/commit/30e92af08 moved non-streaming compression onto a small worker pool so we don't end up with one large compressor state / grown goroutine stack per concurrent response. So I don't think we should change the default behavior.

That said, I agree there are workloads where stackfull compression is a valid tradeoff: bounded concurrency, latency-sensitive handlers, or callers that prefer avoiding the stackless channel/scheduler overhead.

Could you extend this PR into a full opt-in configuration instead of only adding low-level `*Stackfull` helpers? The important path is `CompressHandler*`, since that is where most response compression users go through today.

Preferred shape:
- keep stackless as the default
- add a per-handler option, not a package-global switch
- document the tradeoff: stackfull may be faster per call, but can grow many request goroutine stacks and allow more concurrent compressor state

Something like `CompressHandlerLevelWithOptions` / `CompressHandlerBrotliLevelWithOptions` with a `CompressOptions` struct would be easier to extend later than adding many more handler function variants.

PR #2310, opened 2026-07-01
perf: zero-copy chunked body writes via io.WriterTo

sasgas:
## Motivation

`writeBodyChunked` reads the body stream through a fixed 4 KiB `copyBufPool`
buffer and frames every 4 KiB as a separate chunk. When the body stream is
already backed by large in-memory buffers (for example a proxy relaying cached
256 KiB blocks), this re-copies the whole body 4 KiB at a time and emits one
chunk header/flush per 4 KiB.

## Change

If the body stream implements `io.WriterTo`, frame the buffers it writes
directly through a small `chunkedBodyWriter` wrapper, avoiding the extra copy
and cutting the number of chunk frames and flushes. Streams that do not
implement `io.WriterTo` keep the existing 4 KiB read path unchanged.

This mirrors how `writeBodyFixedSize` already prefers `io.WriterTo` through
`copyZeroAlloc`.

## Notes

- Termination and trailer handling are unchanged: the `io.WriterTo` path ends
  with the same zero-length chunk (`writeChunk(w, nil)`) as the EOF path, so
  `writeTrailer` still runs afterwards.
- Empty writes are ignored so they are not mistaken for the end-of-stream chunk.

### DIFF
--- a/http.go
+++ b/http.go
@@ -2268,6 +2268,18 @@ type httpWriter interface {
 	Write(w *bufio.Writer) error
 }
 
+// ChunkedBodyWriterTo lets a body opt into zero-copy chunked framing via
+// WriteTo, skipping the copy through a pool buffer.
+//
+// The marker method is required rather than a bare io.WriterTo check: a
+// WriteTo promoted from an embedded reader would otherwise opt in by accident
+// and bypass an overridden Read. Implement only when WriteTo and Read emit
+// the same bytes.
+type ChunkedBodyWriterTo interface {
+	io.WriterTo
+	ChunkedBodyWriterTo()
+}
+
 type chunkedBodyWriter struct {
 	w   *bufio.Writer
 	err error
@@ -2285,9 +2297,8 @@ func (cw *chunkedBodyWriter) Write(p []byte) (int, error) {
 }
 
 func writeBodyChunked(w *bufio.Writer, r io.Reader) error {
-	// Frame buffers directly from an io.WriterTo body to avoid copying
-	// through copyBufPool.
-	if wt, ok := r.(io.WriterTo); ok {
+	// An opted-in body frames its WriteTo output directly, skipping copyBufPool.
+	if wt, ok := r.(ChunkedBodyWriterTo); ok {
 		cw := chunkedBodyWriter{w: w}
 		if _, err := wt.WriteTo(&cw); err != nil {
 			return err
--- a/http.go
+++ b/http.go
@@ -2268,7 +2268,36 @@ type httpWriter interface {
 	Write(w *bufio.Writer) error
 }
 
+type chunkedBodyWriter struct {
+	w   *bufio.Writer
+	err error
+}
+
+func (cw *chunkedBodyWriter) Write(p []byte) (int, error) {
+	if len(p) == 0 {
+		return 0, nil // an empty chunk marks end-of-stream
+	}
+	if err := writeChunk(cw.w, p); err != nil {
+		cw.err = err
+		return 0, err
+	}
+	return len(p), nil
+}
+
 func writeBodyChunked(w *bufio.Writer, r io.Reader) error {
+	// Frame buffers directly from an io.WriterTo body to avoid copying
+	// through copyBufPool.
+	if wt, ok := r.(io.WriterTo); ok {
+		cw := chunkedBodyWriter{w: w}
+		if _, err := wt.WriteTo(&cw); err != nil {
+			return err
+		}
+		if cw.err != nil {
+			return cw.err
+		}
+		return writeChunk(w, nil)
+	}
+
 	vbuf := copyBufPool.Get()
 	buf := vbuf.([]byte)
 
--- a/http.go
+++ b/http.go
@@ -2269,15 +2269,12 @@ type httpWriter interface {
 }
 
 // ChunkedBodyWriterTo lets a body opt into zero-copy chunked framing via
-// WriteTo, skipping the copy through a pool buffer.
-//
-// The marker method is required rather than a bare io.WriterTo check: a
-// WriteTo promoted from an embedded reader would otherwise opt in by accident
-// and bypass an overridden Read. Implement only when WriteTo and Read emit
-// the same bytes.
+// WriteTo. Return true only when WriteTo and Read emit the same bytes; a bare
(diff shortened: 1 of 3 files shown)


erikdubbelboer (changes requested):
I'm not sure about this one.

This changes semantics for body streams that override `Read` but also satisfy `io.WriterTo`, especially through an embedded reader.

There is already an in-repo example in `TestRevertPull1233`: `F` embeds `*strings.Reader` but overrides `Read` to limit each read to 10 bytes and sleep between reads. With this branch, `F` satisfies `io.WriterTo` because `strings.Reader.WriteTo` is promoted, so `wt.WriteTo(&cw)` bypasses `F.Read` entirely.

That removes the pacing provided by `F.Read` and makes `TestRevertPull1233` timing-dependent. On the exact merge base used by CI, the test passes without this change but fails 80/100 times locally with it:

    go test -run '^TestRevertPull1233$' -count=100

This was on macOS with Go 1.26.4. The current CI runs happened to take the passing side of the socket race.

I realize `copyZeroAlloc` already uses the same broad `io.WriterTo` optimization in other paths, but I don't think that is a reason to extend it to chunked body streams.

Could this optimization instead be restricted to concrete types whose behavior we know, such as `*bytes.Reader` and `*bytes.Buffer`?

sasgas:
You're right — a bare `io.WriterTo` check opts in a type whose `WriteTo` is merely promoted from an embedded reader, bypassing an overridden `Read` like `F`'s in `TestRevertPull1233`.

Rather than restrict to concrete types, I've gated the optimization behind an explicit opt-in marker interface:

```go
type ChunkedBodyWriterTo interface {
	io.WriterTo
	ChunkedBodyWriterTo()
}
```

`writeBodyChunked` now checks for `ChunkedBodyWriterTo`, not `io.WriterTo`. A promoted `WriteTo` can't satisfy this by accident, so a type has to explicitly declare that its `WriteTo` and `Read` are equivalent. This follows the existing `ReadCloserWithError` pattern already used for body streams in this file.

This keeps the door open for custom streaming bodies (e.g. a proxy handler that frames buffers without an intermediate copy) that a concrete-type whitelist would exclude, while a concrete type like `*bytes.Reader` simply falls back to the safe `Read` path.

I've added two tests: one asserting an opted-in body uses `WriteTo` and never `Read`, and one asserting a promoted-but-not-opted-in body (the `F` shape) uses `Read` and never its promoted `WriteTo`. `TestRevertPull1233` passes reliably again (`-count=100`).

http.go
```diff
@@ -2471,7 +2268,47 @@ type httpWriter interface {
...
+	if err := writeChunk(cw.w, p); err != nil {
+		cw.err = err
+		return 0, err
+	}
+	return len(p), nil
+}
+
 func writeBodyChunked(w *bufio.Writer, r io.Reader) error {
+	// An opted-in body frames its WriteTo output directly, skipping copyBufPool.
+	if wt, ok := r.(ChunkedBodyWriterTo); ok {
```
erikdubbelboer: Shouldn't this also check directly for types like `*bytes.Reader` and `*bytes.Buffer` that will never implement `ChunkedBodyWriterTo` but could still use this optimization?

http.go
```diff
@@ -2471,7 +2268,47 @@ type httpWriter interface {
...
+// ChunkedBodyWriterTo lets a body opt into zero-copy chunked framing via
+// WriteTo, skipping the copy through a pool buffer.
+//
+// The marker method is required rather than a bare io.WriterTo check: a
+// WriteTo promoted from an embedded reader would otherwise opt in by accident
+// and bypass an overridden Read. Implement only when WriteTo and Read emit
+// the same bytes.
+type ChunkedBodyWriterTo interface {
+	io.WriterTo
+	ChunkedBodyWriterTo()
```
erikdubbelboer: Can you change this to `SupportsChunkedBodyWriteTo() bool` so it's also possible to override a type that has it and return `false`.

erikdubbelboer (changes requested):

sasgas:
Addressed both review comments in `2fe748f`:

- The marker is now `SupportsChunkedBodyWriteTo() bool`, so a type that embeds another can opt back out by overriding it to return `false`.
- `*bytes.Reader` and `*bytes.Buffer` take the zero-copy path directly via a type switch, without implementing the interface.

Added tests for the opt-out override and the concrete-type path; `TestRevertPull1233` still passes with `-count=100`.

erikdubbelboer (changes requested):
The new `ChunkedBodyWriterTo` capability should be documented at the public APIs where callers decide to use an unknown-size body stream, otherwise it is difficult to discover at the point where it matters.

Could you add references in these places?

- `Request.SetBodyStream` and `Response.SetBodyStream`: under the `bodySize < 0` behavior, mention that fasthttp may use `WriteTo` instead of `Read` for `*bytes.Reader`, `*bytes.Buffer`, and streams implementing `ChunkedBodyWriterTo`.
- `RequestCtx.SetBodyStream`: add a short `See ChunkedBodyWriterTo` reference for custom streams writing unknown-size chunked bodies.
- `ChunkedBodyWriterTo` itself: state that it is consulted only for unknown-size bodies (`bodySize < 0`) written with chunked transfer encoding.

The interface contract should also be a little stronger than “emit the same bytes.” The original regression emitted the same eventual bytes but changed pacing. `SupportsChunkedBodyWriteTo` should return true only when `WriteTo` can safely replace `Read`, including any required pacing, accounting, transformations, or other observable side effects.

I do not think this needs to be added to `SetBodyStreamWriter`, `BodyStream`, `StreamBody`, or the README, since the capability does not apply directly at those surfaces.

sasgas:
Documented the capability at Request/Response/RequestCtx.SetBodyStream and strengthened the ChunkedBodyWriterTo contract to require WriteTo to preserve pacing, accounting, and other observable side effects — not just the eventual bytes. Left SetBodyStreamWriter/BodyStream/StreamBody/README untouched as suggested. Also fixed the failing lint (embeddedstructfieldcheck).

erikdubbelboer:
Can you please fix the lint issues, after that it can be merged.

sasgas:
Fixed the lint issue in 26a9579 — the unchecked type assertion is gone (verified locally with golangci-lint v2.5.0, config clean).

erikdubbelboer:
Thanks!

PR #2312, opened 2026-07-01
perf: avoid redundant scans when parsing request headers

ReneWerner87:
RequestHeader.parse walks the header block twice before the lines are actually parsed: readRawHeaders looks for the end of the block, then headerScanner searches for the CRLFCRLF terminator once more. The scanner now reuses the block end that readRawHeaders already found, guarded by a check that the block really ends in CRLFCRLF. Blocks with bare LF terminators fail that check and take the old search path, so those cases keep behaving exactly as before.

The second commit removes duplicated per-line work on top of that:

- readContinuedLineSlice already finds the colon in every header line; next() now reuses that position instead of running bytes.Cut over the line again. Lines handed out by readContinuedLineSlice can never start with a space or tab, so trimming never shifts the position.
- isValidHeaderKey also reports whether the key contains a space that survives trailing-whitespace trimming. The per-byte key loops in parseHeaders and parseTrailer only existed to derive exactly that flag, their reject branches were unreachable since the scanner validates first. The quirk that a space in a response header key closes the connection is preserved.
- readLine drops its searchStart loop; the scanner truncates the block at the terminator, so every line ends in \n.

Smaller cleanups along the way:

- trim header values with a plain loop, bytes.TrimLeft rebuilds its ASCII set on every call for the " \t" cutset
- drop the bytes.IndexByte(key, ' ') checks in front of normalizeHeaderKey; space is not a valid header field byte, so its validation loop refuses to canonicalize such keys either way, and removeNewLines runs before the early returns in both flows, including the #1917 case of a space before the colon

benchstat (Apple M2 Pro, n=10):

```
RequestHeaderRead-12     79.48n ± 1%   56.06n ± 1%   -29.48% (p=0.000)
ServerGet1ReqPerConn-12  1.693µ ± 3%   1.665µ ± 2%   ~ (p=0.052)
```

A new test covers mixed line endings: the fast path, the fallback, and the rejection of LF-only terminated blocks. The net diff removes more code than it adds.

### DIFF
--- a/header.go
+++ b/header.go
@@ -551,20 +551,29 @@ func validHeaderValueByte(c byte) bool {
 	return validHeaderValueByteTable[c] == 1
 }
 
-// isValidHeaderKey returns true if a is a valid header key.
-func isValidHeaderKey(a []byte) bool {
+// isValidHeaderKey returns whether a is a valid header key, and whether a
+// contains a space before its last non-space byte. Such a space survives
+// trailing-whitespace trimming, and a key carrying it is accepted but must
+// not be canonicalized. See https://go.dev/issue/34540 and
+// https://github.com/valyala/fasthttp/issues/1917.
+func isValidHeaderKey(a []byte) (valid, innerSpace bool) {
 	if len(a) == 0 {
-		return false
+		return false, false
 	}
+	seenSpace := false
 	for _, c := range a {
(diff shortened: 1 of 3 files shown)


ReneWerner87:
@erikdubbelboer

erikdubbelboer:
Nice improvement, thanks!

PR #2313, opened 2026-07-02
test: fix host comparison in FuzzURIParse

ReneWerner87:
The cifuzz job on #2312 hit this crasher in FuzzURIParse ([run](https://github.com/valyala/fasthttp/actions/runs/28555598522/job/84662236366)):

```
"http://[%255%2c%2c%2c%2c%4c%2c2c%2c%2c]": unexpected host: "[%5,,,,l,2c,,]". Expecting "[%5,,,,L,2c,,]"
panic: fatal
```
image

The crash is unrelated to that PR: uri.go is untouched there and the input reproduces on master.

FuzzURIParse lowercases the input to hide that fasthttp normalizes the host to lowercase while net/url does not. That trick cannot reach bytes hidden behind percent-encoding: "%4c" survives ToLower and only decodes to 'L' afterwards, so fasthttp returns "l" where net/url keeps "L". fasthttp's behavior is the documented one (Host is always lowercased), so this fixes the comparison instead: the expectation is lowered the same way fasthttp does it, byte-wise via lowercaseBytes. strings.ToLower would mangle non-UTF-8 bytes such as a decoded "%80", which the fuzzer promptly found while testing the fixed harness.

Both inputs are added as seeds; 120s of local fuzzing pass.

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -115,6 +115,10 @@ func FuzzURIParse(f *testing.F) {
 	f.Add(`http://foobar.com/aaa/bb?cc#dd`)
 	f.Add(`http://google.com?github.com`)
 	f.Add(`http://google.com#@github.com`)
+	// Hosts hiding an uppercase letter and a non-UTF-8 byte behind
+	// percent-encoding, see the host comparison below.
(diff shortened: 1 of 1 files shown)


ReneWerner87:
test fail is not related to my changes
image

erikdubbelboer:
Thanks!

PR #2315, opened 2026-07-03
validate domain and path cookie attribute values on parse

alhudz:
Repro: `c.Parse("sid=ok; Domain=victim.example\"; Path=/")` returns nil and `c.String()` re-emits the stray `"` inside `domain=`.
Cause: `Cookie.ParseBytes` runs `validCookieValue` on the primary value and on request-cookie values, but the `Domain`/`Path` attribute branches store the raw value via `initHeaderValueBytes` (CR/LF only), and it is round-tripped verbatim by `AppendBytes`. A `"` or `\` in an attribute therefore survives a parse then re-serialise (proxy/forwarding) at exactly the spot the primary value is now rejected.
Fix: apply `validCookieValue` to the `Domain` and `Path` branches, matching the primary-value guard added in `eb82c9a` and the existing `Max-Age`/`Expires` attribute error handling. Balanced quoted attributes are still unquoted during the trim pass and stay valid.

### DIFF
--- a/cookie.go
+++ b/cookie.go
@@ -428,11 +428,17 @@ func (c *Cookie) ParseBytes(src []byte) error {
 
 			case 'd': // "domain"
 				if caseInsensitiveCompare(strCookieDomain, k) {
+					if !validCookieValue(v) {
+						return ErrInvalidCookieValue
+					}
 					c.domain = initHeaderValueBytes(c.domain, v)
 				}
 
 			case 'p': // "path"
 				if caseInsensitiveCompare(strCookiePath, k) {
+					if !validCookiePathValue(v) {
+						return ErrInvalidCookieValue
+					}
 					c.path = initHeaderValueBytes(c.path, v)
 				}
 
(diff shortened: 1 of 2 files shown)


erikdubbelboer (changes requested):
This will cause this to fail:
```go
var c fasthttp.Cookie
var parsed fasthttp.Cookie

c.SetKey("sid");
c.SetValue("ok");
c.SetPath(`/a\b`)

if err := parsed.ParseBytes(c.Cookie()); err != nil {
  panic(err)
}
```

`SetPath` currently accepts and serializes backslashes, and that matches net/http’s `Cookie.Valid`/`String` behavior for `Path` values. net/http only rejects `;` and control bytes for cookie paths, not `\` or `"`. So using `validCookieValue` here makes `ParseBytes` stricter than fasthttp’s own setter/serializer contract and stricter than the path-value rules. I think `Path` needs path-specific validation instead of cookie-value validation.

alhudz:
Good catch. Switched `Path` to path-specific validation that only rejects control bytes and the `;` separator, so `\` and `"` stay valid and your `SetPath(`/a\b`)` round-trip parses fine now. `Domain` keeps `validCookieValue`, since `"` `;` `\` aren't legal in a hostname anyway. Test now covers the `\` path round-trip plus a control-byte path rejection.

cookie.go
```diff
@@ -659,6 +665,25 @@ func validCookieValue(value []byte) bool {
...
+// characters accepted by SetPath and net/http's Cookie.String. It rejects the
+// ';' attribute separator and control bytes, matching net/http's cookie path
+// rules. CR and LF are tolerated here because initHeaderValueBytes strips them
+// afterwards, the same as for the primary cookie value.
+func validCookiePathValue(value []byte) bool {
+	for _, b := range value {
+		if b == '\r' || b == '\n' {
+			continue
+		}
+		if b < 0x20 || b == 0x7f || b == ';' {
```
erikdubbelboer: This still accepts non-ASCII bytes. It still allows characters in the `0x80–0xff` range. `net/http` rejects everything `b >= 0x7f`. See: https://github.com/golang/go/blob/d90b98e65320778f3b1f99a6951ab20f04d218b3/src/net/http/cookie.go#L528-L530

erikdubbelboer (changes requested):

cookie.go
alhudz: Right, missed the high bytes. Changed the check to `b < 0x20 || b >= 0x7f || b == ';'` so anything outside `0x20-0x7e` is rejected, same range as `validCookiePathByte`. Added `0x80`, `0xff` and a UTF-8 path to the rejection cases; `go test .` is green.

erikdubbelboer:
Thanks!

PR #2316, opened 2026-07-03
re-enable forcetypeassert

### LINKED ISSUE #2232: safety: re-enable forcetypeassert linter to catch bare type assertions in CI
## Description

The `forcetypeassert` linter is explicitly disabled in `.golangci.yml` (line 15):

```yaml
linters:
  default: all
  disable:
    - forcetypeassert   # <-- disabled
```

This linter flags bare type assertions `v.(T)` that don't use comma-ok (`v, ok := x.(T)`). Bare assertions **panic at runtime** on type mismatch — they are Go's equivalent of an unchecked cast.

## Why This Matters

With `forcetypeassert` disabled, there is **no CI protection** against new bare type assertions introduced by contributors. A missing comma-ok in error handling, context value extraction, or interface dispatch becomes a latent crash that only triggers in production.

This is the exact category of bug that caused:
- #1205 — nil `RequestCtx` panic from pool
- #1987 — nil `perIPConnCounter` panic on shutdown
- #2026 — zero-value `RequestCtx` panic

## Current Bare Assertions in Codebase

There are **44 bare type assertions** from `sync.Pool.Get()` results (see #2231). These are controlled but unprotected by CI.

There are also **safe** comma-ok assertions in the codebase (e.g., `tcpdialer.go:432`, `http.go:1021`, `server.go:858`, `peripconn.go:125`) — showing the codebase already uses both patterns inconsistently.

## Suggested Fix

1. Remove `forcetypeassert` from the `disable` list in `.golangci.yml`
2. Add `//nolint:forcetypeassert` annotations to the known-safe pool assertions (the 44 instances from #2231)
3. All new code with bare assertions will fail CI, forcing developers to use comma-ok or justify the assertion

This follows the same pattern as the existing `//nolint:errcheck` and `// #nosec G115` annotations already in the codebase.

## Impact

- Medium severity — no immediate runtime risk, but **prevents future regressions** from reaching main
- Zero performance impact — lint-only change

Harshal96:
fixes #2232

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,6 @@ linters:
     - exhaustive
     - exhaustruct
     - forbidigo
-    - forcetypeassert
     - funcorder
     - funlen
     - gochecknoglobals
--- a/args.go
+++ b/args.go
@@ -19,7 +19,7 @@ const (
 // The returned Args may be returned to the pool with ReleaseArgs
 // when no longer needed. This allows reducing GC load.
 func AcquireArgs() *Args {
-	return argsPool.Get().(*Args)
+	return argsPool.Get().(*Args) //nolint:forcetypeassert
 }
 
 // ReleaseArgs returns the object acquired via AcquireArgs to the pool.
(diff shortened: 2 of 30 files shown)


erikdubbelboer:
Thanks!

PR #2318, opened 2026-07-03
refactor: improve internal interface names

### LINKED ISSUE #2229: fix: improve internal interface naming (headerInterface, connTLSer, connKeepAliveer)
## Summary

Three internal interface naming violations that make the code harder to read.

### 1. `headerInterface` — ends with "Interface"

| Current | File:Line |
|---|---|
| `headerInterface` | `streaming.go:12` |

Go convention never uses "Interface" as a suffix. Interface names should describe behavior, not the fact that something is an interface.

**Suggested:** `headerReader` or `headerAccessor` (depending on what the interface abstracts).

### 2. `connTLSer` — awkward `-er` suffix

| Current | File:Line |
|---|---|
| `connTLSer` | `server.go:839` |

Concatenating "conn" + "TLS" + "er" produces an unreadable name. The `-er` suffix should apply to a verb, not an acronym.

**Suggested:** `tlsHandshaker` or `tlsConn`.

### 3. `connKeepAliveer` — double-e from awkward `-er` suffix

| Current | File:Line |
|---|---|
| `connKeepAliveer` | `server.go:2054` |

"Alive" + "er" produces a double-e that reads as "Keep-Aliv-ee-er". This violates the spirit of the `-er` convention.

**Suggested:** `keepAliver` or `keepAliveConn`.

## Reference

- [Effective Go: Interface names](https://go.dev/doc/effective_go#interface-names)

Harshal96:
Fixes #2229


streaming.go
```diff
@@ -9,13 +9,13 @@ import (
 	"github.com/valyala/bytebufferpool"
 )
 
-type headerInterface interface {
+type requestStreamHeader interface {
```
erikdubbelboer: This could use a better name, this interface is used for response streams too. Maybe just `bodyStreamHeader`?

erikdubbelboer (changes requested):
Please rebase this on `master`.

streaming.go
Harshal96: done

erikdubbelboer:
Thanks!

PR #2321, opened 2026-07-04
avoid following a symlink when writing the FS compressed cache

alhudz:
Repro: enable `FS.Compress`, then (as a local user with write access to the served directory) plant a symlink at `<file>.fasthttp.gz.tmp` pointing at any file the server process can write, and request `<file>` with `Accept-Encoding: gzip`. The target file is truncated and overwritten with the compressed output.
Cause: `compressFileNolock` writes the cache to the predictable path `compressedFilePath + ".tmp"` via `os.Create`, which follows the symlink (and also leaves the cache file `0666`).
Fix: create the temp file with `os.CreateTemp` in the same directory (random name, `O_EXCL`, `0600`), so a pre-planted symlink cannot be followed; the atomic rename to the final path is unchanged.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1713,15 +1713,20 @@ func (h *fsHandler) compressFileNolock(
 
 	// Create temporary file, so concurrent goroutines don't use
 	// it until it is created.
-	tmpFilePath := compressedFilePath + ".tmp"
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks, great catch!

PR #2322, opened 2026-07-05
perf: reduce redundant scans and allocations in hot paths

ReneWerner87:
Five small, independent hot-path optimizations. Each is a semantic no-op verified against the full test suite, `go test -race`, and the URI/Cookie fuzz targets; every speedup is measured with `benchstat` (n>=10, 1s benchtime) on the relevant benchmark.

## Changes

**`normalizePath` (`uri.go`)** - dot-free paths skip the `bytes.Index` passes for `/./`, `/../` and `/..` (3 on unix, 7 on windows) behind a single `IndexByte('.')` guard, since every one of those tokens contains a dot.

**`parseHost` (`uri.go`)** and **`validCookieValue` (`cookie.go`)** - `bytes.ContainsAny` rebuilds an ASCII set from its constant needle on every call; a direct `IndexByte` / byte scan is equivalent and skips that setup. Both run per request (host / cookie).

**`serverDate` (`header.go`)** - `atomic.Value` only ever held a `[]byte`, so every default response paid an interface unbox plus a type assertion. `atomic.Pointer[[]byte]` removes both (and a `//nolint`).

**serve loop (`server.go`)** - two redundant `time.Now()` calls dropped by reusing `connTime` (per connection) and `ctx.time` (per request); `closeIdleConns` only requires the stored timestamp to be non-zero and in the past.

## Performance

Apple M2 Pro, go1.24, `benchstat` n>=10 at 1s benchtime. Times are the median ns/op.

| Optimization | Benchmark | Before | After | Change |
| --- | --- | --- | --- | --- |
| `normalizePath` dot-skip | `URIParsePath` | 19.8 ns | 17.4 ns | **-12%** |
| `parseHost` byte scan | `URIParseHostname` | 14.41 ns | 14.17 ns | **-1.7%** |
| `validCookieValue` byte scan | `CookieParseMin` | 35.70 ns | 27.64 ns | **-22.6%** |
| | `CookieParseNoExpires` | 82.20 ns | 73.03 ns | **-11.2%** |
| | `CookieParseFull` | 135.5 ns | 126.7 ns | **-6.5%** |
| serve loop timestamp reuse | `ServerGet1ReqPerConn` | 1672 ns | 1579 ns | **-5.6%** |
| `serverDate` atomic.Pointer | `ResponseHeaderWrite` | 10.21 ns | 10.19 ns | neutral (cleanup) |

All changes above are statistically significant (p < 0.05) except the `serverDate` row, which is a readability/type cleanup with no measurable cost. `allocs/op` is unchanged (0) across every benchmark.

## Verification

- `go test ./...` - green
- `go test -race ./...` - green, no data races
- `FuzzURIParse` and `FuzzCookieParse` - 10s each, no crashers

### DIFF
--- a/header.go
+++ b/header.go
@@ -2380,13 +2380,13 @@ func updateServerDate() {
 }
 
 var (
-	serverDate     atomic.Value
+	serverDate     atomic.Pointer[[]byte]
 	serverDateOnce sync.Once // serverDateOnce.Do(updateServerDate)
 )
 
 func refreshServerDate() {
 	b := AppendHTTPDate(nil, time.Now())
-	serverDate.Store(b)
+	serverDate.Store(&b)
 }
 
 // Write writes response header to w.
@@ -2465,7 +2465,7 @@ func (h *ResponseHeader) AppendBytes(dst []byte) []byte {
 
 	if !h.noDefaultDate {
 		serverDateOnce.Do(updateServerDate)
-		dst = appendHeaderLine(dst, strDate, serverDate.Load().([]byte)) //nolint:forcetypeassert
+		dst = appendHeaderLine(dst, strDate, *serverDate.Load())
 	}
 
 	// Append Content-Type only for non-zero responses
--- a/cookie.go
+++ b/cookie.go
@@ -656,7 +656,12 @@ func decodeCookieArg(dst, src []byte, skipQuotes bool) []byte {
 }
 
 func validCookieValue(value []byte) bool {
-	return !bytes.ContainsAny(value, "\";\\")
+	for _, c := range value {
+		if c == '"' || c == ';' || c == '\\' {
+			return false
+		}
+	}
+	return true
 }
 
 func trimCookieArgNoCopy(src []byte, skipQuotes bool) []byte {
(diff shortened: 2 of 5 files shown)


header.go
```diff
@@ -2380,13 +2380,13 @@ func updateServerDate() {
 }
 
 var (
-	serverDate     atomic.Value
+	serverDate     atomic.Pointer[[]byte]
```
erikdubbelboer, line 2383: This changes `serverDate.Load()` to return `*[]byte`, but the fast-error path in `server.go` still formats the load result directly:

```go
date = fmt.Sprintf("Date: %s\r\n", serverDate.Load())
```

Those early rejection responses now emit a malformed Date header such as `Date: &Sun, ... GMT` because `fmt` is formatting the pointer. `TestServerWriteFastError` still passes because it parses the response but does not assert the Date value. Please dereference in `writeFastError` too (`*serverDate.Load()` or append the bytes directly) and add coverage for that header.

ReneWerner87: Good catch, thanks! Fixed in b6bf306: `writeFastError` now dereferences the pointer, and `TestServerWriteFastError` asserts that the Date header parses as RFC1123, so a regression here fails the test now.

ReneWerner87:
@erikdubbelboer fixed

gaby (approved):
LGTM ⚡

erikdubbelboer:
Thanks!

PR #2323, opened 2026-07-08
fix: preserve pre-set status code in NewFastHTTPHandler

### LINKED ISSUE #1754: NewFastHTTPHandler does not set response status code
I recreated the issue because the reopen button is not on the [original issue](https://github.com/valyala/fasthttp/issues/1753).

The NewFastHTTPHandler() handler does not set the Status code for the response:

https://github.com/valyala/fasthttp/blob/master/fasthttpadaptor/adaptor.go#L59

I have pointed the line of code that needs to be updated to set the status from the ctx response.

The code in line 65
https://github.com/valyala/fasthttp/blob/d3a9c74c92588e83c465057c243e05b46d9118f7/fasthttpadaptor/adaptor.go#L65
Returns a nil value that eventually is changed to 200.

The fix needed is on line 59, the response
```
w := netHTTPResponseWriter{
	w:   ctx.Response.BodyWriter(),
	ctx: ctx,
}
```
needs to changed to:
```
w := netHTTPResponseWriter{
	w:   ctx.Response.BodyWriter(),
	statusCode: ctx.Response.StatusCode(), <---- this is needed
	ctx: ctx,
}
```
Line 65 needs to be removed since code is set by the caller, ctx already should contain the nil value or a status set by the caller.

xbrxr03:
Closes #1754

**Problem:** When a caller sets `ctx.Response.SetStatusCode()` before the handler returned by `NewFastHTTPHandler` runs, the pre-set status code is silently overwritten with 200. This is because `writer.status()` defaults to `http.StatusOK` when `WriteHeader` was never called by the net/http handler, ignoring whatever was already on `ctx.Response`.

This breaks frameworks like go.fiber + Templ that set a status code before delegating to a net/http handler through the adaptor.

**Fix:** In `writer.status()`, fall back to `ctx.Response.StatusCode()` before defaulting to `http.StatusOK`. Explicit `WriteHeader` calls from the net/http handler still take precedence (via `CompareAndSwap(0, code)`).

```go
// Before
func (w *writer) status() int {
    code := int(w.statusCode.Load())
    if code == 0 {
        return http.StatusOK
    }
    return code
}

// After
func (w *writer) status() int {
    code := int(w.statusCode.Load())
    if code == 0 {
        if ctxCode := w.ctx.Response.StatusCode(); ctxCode != 0 {
            return ctxCode
        }
        return http.StatusOK
    }
    return code
}
```

**Tests:** Two new test cases:
- `TestNewFastHTTPHandlerPreservesPresetStatusCode`: pre-set 403, handler calls `w.Write()` only → response status is 403
- `TestNewFastHTTPHandlerPresetStatusCodeOverriddenByHandler`: pre-set 403, handler calls `WriteHeader(404)` → response status is 404 (handler wins)

All existing tests pass.

AI-assisted: written with the help of an AI coding agent (OpenClaw/GlM-5.1).

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -347,6 +347,11 @@ func (w *writer) Close() error {
 func (w *writer) status() int {
 	code := int(w.statusCode.Load())
 	if code == 0 {
+		// No WriteHeader was called; check if ctx already has a status code set
+		// by a caller before NewFastHTTPHandler ran.
+		if ctxCode := w.ctx.Response.StatusCode(); ctxCode != 0 {
+			return ctxCode
+		}
 		return http.StatusOK
 	}
 	return code
--- a/fasthttpadaptor/adaptor_test.go
+++ b/fasthttpadaptor/adaptor_test.go
@@ -749,3 +749,55 @@ func TestWriterWriteRechecksStreamingReadyAfterLock(t *testing.T) {
 		t.Fatalf("timeout waiting for Write")
 	}
 }
+
+func TestNewFastHTTPHandlerPreservesPresetStatusCode(t *testing.T) {
+	t.Parallel()
+
+	// Verify that a status code set on ctx before calling NewFastHTTPHandler
+	// is preserved when the net/http handler does not call WriteHeader.
+	nethttpH := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Write([]byte("ok"))
+	})
+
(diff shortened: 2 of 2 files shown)


erikdubbelboer (changes requested):
Looks good but can you make sure the linter passes.

xbrxr03:
Lint fix pushed — addressed the two `errcheck` warnings by checking the return values of `w.Write` in the test file. All linters now pass cleanly.

erikdubbelboer:
Thanks!

Issue #1691, opened 2024-01-10
Path Traversal Attacks on Windows


egovorukhin:
Hello, I found another bug security on windows.

example - `curl http://localhost:8081/api/\../\../\../\../\../\../\../\../windows/win.ini -k`

SOLUTION

file **strings.go**

```golang
var (
   ...
  strBackSlashDotDotSlash = []byte(`\../`)
  ...
)
```
file **uri.go**
```golang
func normalizePath(dst, src []byte) []byte {
...
if filepath.Separator == '\\' {
...
        // remove /foo\../ parts
	for {
		n := bytes.Index(b, strBackSlashDotDotSlash)
		if n < 0 {
			break
		}
		nn := bytes.LastIndexByte(b[:n], '/')
		if nn < 0 {
			nn = 0
		}
		n += len(strBackSlashDotDotSlash) - 1
		copy(b[nn:], b[n:])
		b = b[:len(b)-n+nn]
	}
...
}
```

erikdubbelboer:
Can you make a pull request including a test that fails before and passes after your change?

egovorukhin:
https://github.com/valyala/fasthttp/pull/1698 - created pull request. 

How do you think about changing the function (normalizePath) more? Please check the changes

gaby:
@egovorukhin  @erikdubbelboer This was supposed to be reported using the process here: https://github.com/valyala/fasthttp/security

This could qualify for a CVE, and it was reported improperly.

erikdubbelboer:
I think we should just disable FS on windows. It's not safe, it's not properly tested and nobody should be using it in production.

egovorukhin:
Let's replace backslash with slash when used in windows, then a check for "path traversal" will occur. What do you think about this?

file - uri.go
func - normalizePath
```golang
if filepath.Separator == '\\' {
	dst = replaceSlashes(dst)
}
```
https://github.com/egovorukhin/fasthttp/blob/4d48887eb813b0f1caac0217751888897f601d50/uri.go#L579

file - uri_windows.go
```golang
func replaceSlashes(dst []byte) []byte {
	for i := range dst {
		if dst[i] == '\\' {
			dst[i] = '/'
		}
	}
	return dst
}
```
https://github.com/egovorukhin/fasthttp/blob/4d48887eb813b0f1caac0217751888897f601d50/uri_windows.go#L13

erikdubbelboer:
[...]

PR #2327, opened 2026-07-14
reject backslash '..' traversal in fs handler on windows

alhudz:
Repro: on Windows, `GET /foo/\..\..\windows/win.ini` against a default `FS`/`FSHandler` (no `PathRewrite`) serves a file outside `Root` (#1691).
Expected: rejected as a `..` traversal, the same as the `/`-separator case.
Actual: `ctx.Path()` normalisation only collapses `/`-delimited `..`, so `\..\` and `\../` pass through unchanged, and `handleRequest` only ran its `hasDotDotPathSegment` guard for rewritten paths. `pathToFilePath` then feeds the backslashes to `filepath.FromSlash`, producing a real parent-directory escape.
Fix: run the `..`-segment guard for every path on Windows, not just rewritten ones. `hasDotDotPathSegment` already treats `\` as a separator there, so any `..` segment is rejected whatever the separator. The regression test is Windows-only since `\` only separates path components there.

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1302,12 +1302,15 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 		ctx.Error("Are you a hacker?", StatusBadRequest)
 		return
 	}
-	if h.pathRewrite != nil {
-		// There is no need to check rewritten paths if path = ctx.Path(),
-		// since ctx.Path must normalize and sanitize the path.
-
+	// Rewritten paths bypass ctx.Path()'s normalization, so they must be
+	// checked for '..' segments here. On Windows every path is checked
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
The fix itself looks good, thanks for that.

But it seems like your test case is wrong and failing, can you have a look at that?

alhudz:
Good catch. Two of the three URIs used `\..\`, which `normalizePath` already collapses to a safe path on Windows before the handler runs, so they landed on the 404 branch rather than the 500 I was asserting. Only the `\../` form survives normalisation, so I've switched the cases to that. Pushed.

erikdubbelboer:
Thanks!

PR #2330, opened 2026-07-20
reject Transfer-Encoding on HTTP/1.0 requests

alhudz:
Repro: send `POST / HTTP/1.0\r\nHost: x\r\nTransfer-Encoding: chunked\r\n\r\n5\r\nHELLO\r\n0\r\n\r\n`. `RequestHeader.parseHeaders` frames it as chunked and swallows `HELLO` as the body.
Expected: per [RFC 9112 section 6.1](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.1), a server that receives an HTTP/1.0 message containing `Transfer-Encoding` must treat the framing as faulty, even if a `Content-Length` is present, and close the connection.
Actual: the request parser honours chunked regardless of version. A chunked-unaware 1.0 hop then treats the request as bodyless and forwards `5\r\nHELLO...` as the next request, while fasthttp consumes it as the first request's body (request smuggling).
Fix: reject the request in `RequestHeader.parseHeaders` when `noHTTP11` is set. The check sits before the `disableSpecialHeader` early-continue so that path is covered too, and `connectionClose` plus the usual header-error handling closes the connection.

`Transfer-Encoding` on HTTP/1.1 is unaffected; regression cases cover the 1.0 rejection with and without `Content-Length`, and the 1.1 chunked control.

### DIFF
--- a/header.go
+++ b/header.go
@@ -3180,6 +3180,14 @@ func (h *RequestHeader) parseHeaders(buf []byte, blockEnd int) (int, error) {
 		case 't':
 			if caseInsensitiveCompare(s.key, strTransferEncoding) {
 				isTransferEncoding = true
+				// RFC 9112 section 6.1 requires an HTTP/1.0 message carrying
+				// Transfer-Encoding to be treated as having faulty framing,
+				// even if Content-Length is present, and the connection to be
+				// closed afterwards.
+				if h.noHTTP11 {
+					h.connectionClose = true
+					return 0, ErrUnsupportedTransferEncoding
+				}
 				if transferEncodingSeen {
 					h.connectionClose = true
 					if h.secureErrorLogMessage {
--- a/header_test.go
+++ b/header_test.go
@@ -3428,6 +3428,10 @@ func TestRequestHeaderReadSuccess(t *testing.T) {
 		t.Fatalf("unexpected 'connection: close' for ancient http protocol")
 	}
 
+	// HTTP/1.1 still frames a chunked body.
+	testRequestHeaderReadSuccess(t, h, "POST /te HTTP/1.1\r\nHost: aa\r\nTransfer-Encoding: chunked\r\n\r\n",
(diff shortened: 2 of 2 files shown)


header.go
```diff
@@ -3235,6 +3235,15 @@ func (h *RequestHeader) parseHeaders(buf []byte, blockEnd int) (int, error) {
 			}
 		case 't':
 			if isTransferEncoding {
+				// Transfer-Encoding is undefined for HTTP/1.0, so an HTTP/1.0
+				// hop drops it and frames the body by Content-Length instead.
+				// Honouring chunked here would desync fasthttp from such a hop
+				// (request smuggling), so ignore it, mirroring the response
+				// parser above and net/http.
+				if h.noHTTP11 {
+					continue
```
erikdubbelboer: [RFC 9112 §6.1](https://www.rfc-editor.org/rfc/rfc9112.html#section-6.1) explicitly requires HTTP/1.0 messages containing `Transfer-Encoding` to be treated as faulty and the connection closed.

erikdubbelboer (changes requested):

header.go
alhudz: You're right, ignoring it was wrong - the RFC also says faulty "even if a Content-Length is present", which my fallback-to-`Content-Length` case got backwards.

Switched it to reject and close instead. The check now sits in the first `case 't'`, before the `disableSpecialHeader` early-continue, so that path is covered by the same guard and the `ContentLength` change I had is gone. Returns `ErrUnsupportedTransferEncoding` with `connectionClose` set, same shape as the duplicate transfer-encoding rejection just below it, so the connection closes via the usual header-error handling.

Tests moved to `TestRequestHeaderReadError` for both 1.0 cases (with and without `Content-Length`), 1.1 chunked control kept. Retitled the PR since it no longer ignores anything.

erikdubbelboer:
Thanks!

PR #2332, opened 2026-07-23
feat(server): deferred 100-continue via ExpectHandler

NeitrinoK:
Add an opt-in way for a request handler to decide whether to send the "100 Continue" interim response, instead of fasthttp always sending it eagerly before the handler runs.

Motivation
----------
The purpose of the Expect: 100-continue mechanism is to let a server reject a request BEFORE the client transmits a (potentially large) body. Today, when a request carries "Expect: 100-continue" and neither ExpectHandler nor ContinueHandler denies it, fasthttp writes the "100 Continue" response and begins reading the body before invoking the request handler. That means the client has already started (or completed) uploading the body by the time the handler gets to run, so a handler that rejects the request — e.g. after routing, authentication or an existence check that cannot be performed from request headers alone — cannot avoid the upload. Closing such a connection with an unread body in flight also risks a TCP RST that races the client's read of the response, surfacing as a transport error instead of the intended status code.

ExpectHandler already lets a server reject from the headers, but the decision often depends on state that is only available inside the request handler (after the routing/auth middleware has run). There was no way to defer the "100 Continue" decision that far.

What this adds
--------------
- StatusContinueDeferred: a sentinel that Server.ExpectHandler may return. Returning it invokes the request handler immediately WITHOUT sending "100 Continue" and WITHOUT reading the body. The value (-1) is not a valid HTTP status, so it cannot collide with a real rejection status.

- RequestCtx.ContinueRequestBody(): called by the handler when it wants the body. It sends the pending "100 Continue" response and then reads the body (honoring StreamRequestBody). It is a no-op when the request did not defer continue or when the body was already requested, so it is safe to call unconditionally.

Behavior:
- Accept: the handler calls ContinueRequestBody(); "100 Continue" is sent and the body is read, exactly as in the eager path.
- Reject: the handler responds without calling ContinueRequestBody(); "100 Continue" is never sent, so a well-behaved client never uploads the body. The connection is marked for close, which is graceful because there is no in-flight body to leave unread. This also guards against a client that begins uploading after its expect-continue timeout.

The feature is fully backward compatible: ExpectHandler returning StatusContinue (or any rejection status), ContinueHandler, and the default auto-accept behavior are all unchanged. The deferred path is only taken when ExpectHandler explicitly returns StatusContinueDeferred.

Includes TestServerExpectHandlerDeferredContinue covering both the accept (100-continue sent, body read and echoed) and reject (no 100-continue, 404 with Connection: close, body not consumed) paths.

### DIFF
--- a/server.go
+++ b/server.go
@@ -146,6 +146,28 @@ type ServeHandler func(c net.Conn) error
 // instead.
 //
 // It is safe to call Server methods from concurrently running goroutines.
+
+// StatusContinueDeferred may be returned by Server.ExpectHandler to defer the
+// "100 Continue" response instead of sending it eagerly before the request
+// handler runs.
+//
+// When ExpectHandler returns StatusContinueDeferred the request handler is
+// invoked immediately, but the request body is NOT read and no "100 Continue"
+// response is sent yet. The handler must call RequestCtx.ContinueRequestBody
+// before it reads the body: that call sends the pending "100 Continue" response
+// and reads the body. If the handler responds without calling
+// ContinueRequestBody (for example, to reject the request after authentication),
+// "100 Continue" is never sent, so a well-behaved client never uploads the body.
+//
+// This realizes the purpose of the Expect: 100-continue mechanism — letting the
+// server reject a request before the client transmits a (potentially large)
+// body — for decisions that can only be made inside the request handler (after
+// routing, authentication, etc.) rather than from request headers alone.
+//
+// The value is intentionally not a valid HTTP status code so it cannot collide
+// with a real status returned by ExpectHandler to reject the request.
+const StatusContinueDeferred = -1
+
 type Server struct {
 	noCopy noCopy
 
@@ -203,6 +225,11 @@ type Server struct {
 	// StatusContinue (100) to accept the request and proceed to read the body,
 	// or any other status code to reject it and close the connection since the
 	// client may have already started sending the request body.
+	//
+	// Returning StatusContinueDeferred defers the "100 Continue" response until
+	// the request handler asks for the body via RequestCtx.ContinueRequestBody,
+	// so a handler can reject a request after routing/authentication without the
+	// client ever uploading the body. See StatusContinueDeferred for details.
 	//
 	// The ctx provides access to request headers and connection metadata (e.g.
 	// RemoteAddr for IP-based filtering). The response must not be modified.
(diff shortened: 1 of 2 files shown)


server.go
```diff
@@ -146,6 +146,28 @@ type ServeHandler func(c net.Conn) error
 // instead.
 //
 // It is safe to call Server methods from concurrently running goroutines.
```
erikdubbelboer, line 148: This `Server` documentation is now not attached to `type Server` anymore.

erikdubbelboer (changes requested):
There's a race condition with `TimeoutHandler` as shown by this test (run with `-race`):
```go
func TestDeferredContinueTimeoutRace(t *testing.T) {
	serverConn, clientConn := net.Pipe()
	defer serverConn.Close()
	defer clientConn.Close()

	handlerStarted := make(chan struct{})
	handlerDone := make(chan error, 1)
	serverDone := make(chan error, 1)

	s := &Server{
		// ServeConn doesn't initialize this as Serve does.
		concurrencyCh: make(chan struct{}, 1),
		ExpectHandler: func(*RequestCtx) int {
			return StatusContinueDeferred
		},
		Handler: TimeoutHandler(func(ctx *RequestCtx) {
			close(handlerStarted)
			handlerDone <- ctx.ContinueRequestBody()
		}, 100*time.Millisecond, "timeout"),
	}

	go func() {
		serverDone <- s.ServeConn(serverConn)
	}()

	_, err := clientConn.Write([]byte(
		"POST / HTTP/1.1\r\n" +
			"Host: example.com\r\n" +
			"Expect: 100-continue\r\n" +
			"Content-Length: 5\r\n\r\n",
	))
	if err != nil {
		t.Fatal(err)
	}
	<-handlerStarted

	br := bufio.NewReader(clientConn)

	const wantContinue = "HTTP/1.1 100 Continue\r\n\r\n"
	gotContinue := make([]byte, len(wantContinue))
	if _, err := io.ReadFull(br, gotContinue); err != nil {
		t.Fatal(err)
	}
	if string(gotContinue) != wantContinue {
		t.Fatalf("unexpected interim response: %q", gotContinue)
	}

	// ContinueRequestBody is now blocked waiting for the body. TimeoutHandler
	// returns and serveConn starts using the same buffered reader and writer.
	var timeoutResponse Response
	if err := timeoutResponse.Read(br); err != nil {
		t.Fatal(err)
	}
	if timeoutResponse.StatusCode() != StatusRequestTimeout {
		t.Fatalf("status = %d; want %d",
			timeoutResponse.StatusCode(), StatusRequestTimeout)
	}

	// Wake both competing readers.
	_, _ = clientConn.Write([]byte("hello"))
	_ = clientConn.Close()

	select {
	case <-handlerDone:
	case <-time.After(time.Second):
		t.Fatal("ContinueRequestBody did not return")
	}
	select {
	case <-serverDone:
	case <-time.After(time.Second):
		t.Fatal("ServeConn did not return")
	}
}
```


And this test shows that body-read errors leave unread bytes on a reusable connection:
```go
func TestDeferredContinueReadErrorClosesConnection(t *testing.T) {
	var continueErr error

	s := &Server{
		MaxRequestBodySize: 4,
		ExpectHandler: func(*RequestCtx) int {
			return StatusContinueDeferred
		},
			continueErr = ctx.ContinueRequestBody()
			ctx.SetStatusCode(StatusRequestEntityTooLarge)
		},
	}

	rw := &readWriter{}
	rw.r.WriteString(
		"POST / HTTP/1.1\r\n" +
			"Host: example.com\r\n" +
			"Expect: 100-continue\r\n" +
			"Content-Length: 5\r\n\r\n" +
			"hello",
	)

	serveErr := s.ServeConn(rw)
	t.Logf("ServeConn error: %v", serveErr)

	if !errors.Is(continueErr, ErrBodyTooLarge) {
		t.Fatalf("ContinueRequestBody error = %v; want %v",
			continueErr, ErrBodyTooLarge)
	}

	const continueResponse = "HTTP/1.1 100 Continue\r\n\r\n"
	out := rw.w.String()
	if !strings.HasPrefix(out, continueResponse) {
		t.Fatalf("missing 100 Continue in %q", out)
	}

	br := bufio.NewReader(strings.NewReader(
		strings.TrimPrefix(out, continueResponse),
	))
	var finalResponse Response
	if err := finalResponse.Read(br); err != nil {
		t.Fatal(err)
	}
	if finalResponse.StatusCode() != StatusRequestEntityTooLarge {
		t.Fatalf("status = %d; want %d",
			finalResponse.StatusCode(), StatusRequestEntityTooLarge)
	}
	if !finalResponse.Header.ConnectionClose() {
		t.Fatalf(
			"body read failed but connection remained reusable; wire output: %q",
			out,
		)
	}
}
```

PR #2333, opened 2026-07-25
fix: consume informational responses before the final response

dev-willbird1936:
## Summary

`Response.ReadLimitBody` previously treated only `100 Continue` as interim. Other informational responses, such as `103 Early Hints`, were returned as final responses. With `PipelineClient`, this left the real final response unread and assigned it to the next queued request.

This change:

- consumes all `1xx` informational responses except `101 Switching Protocols` before reading the final response;
- keeps `101` terminal so upgraded-protocol bytes remain untouched;
- caps processing at 100 interim responses; and
- adds a focused two-request `PipelineClient` regression that proves each final response stays with its request.

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `go test -race ./...`

### DIFF
--- a/http.go
+++ b/http.go
@@ -1564,8 +1564,15 @@ func (resp *Response) Read(r *bufio.Reader) error {
 	return resp.ReadLimitBody(r, 0)
 }
 
+const maxInterimResponses = 100
+
+var errTooManyInterimResponses = errors.New("fasthttp: too many 1xx informational responses received")
+
 // ReadLimitBody reads response headers from the given r,
 // then reads the body using the ReadBody function and limiting the body size.
+//
+// Informational responses other than "101 Switching Protocols" are consumed
+// before the final response is read.
 //
 // If resp.SkipBody is true then it skips reading the response body.
 //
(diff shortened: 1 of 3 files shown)


http.go
```diff
@@ -1564,9 +1564,16 @@ func (resp *Response) Read(r *bufio.Reader) error {
 	return resp.ReadLimitBody(r, 0)
 }
 
+const maxInterimResponses = 100
```
erikdubbelboer, line 1569: Can you add a comment to describe what this configures.

http.go
```diff
@@ -1579,8 +1586,12 @@ func (resp *Response) ReadLimitBody(r *bufio.Reader, maxBodySize int) error {
 	if err != nil {
 		return err
 	}
-	if resp.Header.statusCode == StatusContinue {
-		// Read the next response according to http://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html .
+
+	for n := 0; resp.Header.statusCode >= 100 && resp.Header.statusCode <= 199 &&
+		resp.Header.statusCode != StatusSwitchingProtocols; n++ {
```
erikdubbelboer:
```suggestion
	for n := 0; ; n++ {
		if resp.Header.statusCode < 100 ||
			resp.Header.statusCode > 199 ||
			resp.Header.statusCode == StatusSwitchingProtocols {
			break
		}
```
Multi-line for statements are not very common.

erikdubbelboer (changes requested):
Thanks, just some minor changes and then it's perfect to merge.

erikdubbelboer:
Thanks!

PR #2334, opened 2026-07-25
fix(fasthttpproxy): select proxies from the request scheme

dev-willbird1936:
## Summary

`fasthttpproxy.Dialer.GetDialFunc` can only observe `host:port`, so it infers HTTPS from port 443. This selects the wrong configured proxy for HTTPS on a non-443 port and HTTP on port 443.

This change:

- adds `DialFuncWithScheme` and `DialScheme` hooks to `Client`, `HostClient`, and `PipelineClient`;
- adds `Dialer.GetDialFuncWithScheme`, which selects `HTTPProxy` or `HTTPSProxy` from the client's actual TLS decision;
- keeps `GetDialFunc` unchanged as a documented compatibility path; and
- adds focused tests for HTTPS on port 8443, HTTP on port 443, conventional scheme/port pairings, and `PipelineClient` propagation.

The new public API shape is intentionally small and can be adjusted during review if another name or shape is preferred.

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `go test -race ./...`

### DIFF
--- a/client.go
+++ b/client.go
@@ -197,6 +197,12 @@ type Client struct {
 	// If not set, DialTimeout is used.
 	Dial DialFunc
 
+	// Callback for establishing new connections to hosts.
+	//
+	// DialScheme receives the TLS requirement for the request. If set, it
+	// takes precedence over Dial and DialTimeout.
+	DialScheme DialFuncWithScheme
+
 	// TLS config for https connections.
 	//
 	// Default TLS config is used if not set.
@@ -554,6 +560,7 @@ func (c *Client) hostClient(host []byte, isTLS bool) (*HostClient, error) {
 		NoDefaultUserAgentHeader:      c.NoDefaultUserAgentHeader,
 		Dial:                          c.Dial,
 		DialTimeout:                   c.DialTimeout,
+		DialScheme:                    c.DialScheme,
 		DialDualStack:                 c.DialDualStack,
 		IsTLS:                         isTLS,
 		TLSConfig:                     c.TLSConfig,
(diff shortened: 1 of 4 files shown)


erikdubbelboer (changes requested):
Thanks for working on this. The underlying issue is real, but I'm not convinced we need to add `DialFuncWithScheme` and `DialScheme` to `Client`, `HostClient`, and `PipelineClient`. That's a lot of new API surface.

TLS is fixed for the lifetime of a `HostClient` or `PipelineClient`. `Client` also creates separate `HostClient`s for HTTP and HTTPS, and calls `ConfigureClient` after setting `IsTLS`. This means the proxy can already be selected once while configuring the host client:

```go
proxyConfig := httpproxy.FromEnvironment()
proxyFunc := proxyConfig.ProxyFunc()

client := &fasthttp.Client{
	ConfigureClient: func(hc *fasthttp.HostClient) error {
		scheme := "http"
		if hc.IsTLS {
			scheme = "https"
		}

		proxyURL, err := proxyFunc(&url.URL{
			Scheme: scheme,
			Host:   hc.Addr,
		})
		if err != nil {
			return err
		}
		if proxyURL == nil {
			// Use the default direct dialer.
			return nil
		}

		// The proxy was selected using the actual scheme. Pin both
		// fields so GetDialFunc's port heuristic is irrelevant.
		d := &fasthttpproxy.Dialer{
			Config: httpproxy.Config{
				HTTPProxy:  proxyURL.String(),
				HTTPSProxy: proxyURL.String(),
			},
			DialDualStack: hc.DialDualStack,
		}

		hc.Dial, err = d.GetDialFunc(false)
		return err
	},
}
```

This handles HTTPS on port 8443 and HTTP on port 443 correctly, while also evaluating `NO_PROXY` and the CGI check using the actual scheme.

The boilerplate could be hidden in a small `fasthttpproxy` helper such as:

```go
func (d *Dialer) GetDialFuncForTLS(useEnv, isTLS bool) (fasthttp.DialFunc, error)
```

That captures the stable TLS value during configuration and returns the existing `DialFunc`. The current proposal instead passes the same value on every dial, adds another field and precedence rule to three clients, and overrides `DialTimeout` without receiving its timeout. A follow-up API such as `DialSchemeTimeout` therefore seems almost inevitable.

Can you rework this so the scheme-aware selection stays inside `fasthttpproxy` and the core client API remains unchanged?

erikdubbelboer:
Thanks!

PR #2335, opened 2026-07-25
fix: reject Windows alternate data stream paths in FS

dev-willbird1936:
## Summary

On Windows, the default `osFS` passed request-path colons to `os.Open`. On NTFS, a path such as `public.txt:secret` opens a named alternate data stream instead of the file's normal content.

This change:

- rejects Windows-reserved colons before the default `osFS` is used;
- leaves caller-supplied filesystems unchanged;
- preserves the leading drive-letter colon required by `ServeFile` absolute paths; and
- adds a focused Windows test that creates a real NTFS alternate data stream and proves it is rejected.

The branch is rebased after the merged Windows backslash-traversal fix in PR #2327. Its adjacent traversal controls remain intact and pass.

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./...`
- `go test -race ./...`

### DIFF
--- a/fs.go
+++ b/fs.go
@@ -1302,6 +1302,14 @@ func (h *fsHandler) handleRequest(ctx *RequestCtx) {
 		ctx.Error("Are you a hacker?", StatusBadRequest)
 		return
 	}
+	// Prevent request paths from reaching NTFS alternate data streams through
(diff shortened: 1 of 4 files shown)


erikdubbelboer:
Thanks!

PR #2336, opened 2026-07-26
consume the rest of the declared body in readMultipartForm

alhudz:
`Repro`: a `multipart/form-data` POST whose declared body carries on past the closing boundary, with a second request line padded to land exactly where the parser stops reading.

`POST /first HTTP/1.1\r\nHost: x\r\nContent-Type: multipart/form-data; boundary=X\r\nContent-Length: 6100\r\n\r\n--X\r\nContent-Disposition: form-data; name="a"\r\n\r\n1\r\n--X--\r\n<pad>GET /smuggled HTTP/1.1\r\nHost: x\r\n\r\n`

`Expected`: one request. Everything inside `Content-Length` is body, not a message.

`Actual`: a `Server` with default settings runs the handler twice, for `/first` and then for `/smuggled`.

1. `readMultipartForm` gives `io.LimitReader(r, contentLength)` to `mime/multipart`, but `ReadForm` returns the moment it sees `--X--`, leaving the remaining declared body bytes in the connection reader. Nothing consumed them, so the serve loop parsed the next request straight out of the epilogue. The stop offset is deterministic, which is what makes the padding above line up.
2. The framing checks all sit in the header parsers, and the headers here are unambiguous: one `Content-Length`, no `Transfer-Encoding`. So a front end sees a single well-formed request while fasthttp executes two. Draining belongs in `readMultipartForm` because that is where the declared size is known.

Not reachable with `DisablePreParseMultipartForm`, or when `Content-Encoding` is set.

### DIFF
--- a/http.go
+++ b/http.go
@@ -71,6 +71,13 @@ type Request struct {
 
 	keepBodyBuffer bool
 
+	// Set when a streamed request body could not be fully drained, e.g. a
+	// malformed chunk was hit while discarding the multipart epilogue. The
+	// connection is then left mid-body, so the server must close it rather
+	// than reuse it, otherwise the leftover bytes are read back as the start
+	// of the next request.
+	bodyStreamDrainErr bool
+
 	// Used by Server to indicate the request was received on a HTTPS endpoint.
 	// Client/HostClient shouldn't use this field but should depend on the uri.scheme instead.
 	isTLS bool
@@ -1126,7 +1133,13 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 	// dynamic type set through SetBodyStream.
 	if req.bodyStream != nil {
 		defer func() {
-			_, _ = copyZeroAlloc(io.Discard, req.bodyStream)
+			if _, drainErr := copyZeroAlloc(io.Discard, req.bodyStream); drainErr != nil {
+				// The declared body could not be fully consumed, so the
+				// connection is left mid-body. Mark it so the server closes the
+				// connection instead of parsing the leftover as the next
+				// request.
+				req.bodyStreamDrainErr = true
+			}
 		}()
 	}
 
@@ -1304,6 +1317,7 @@ func (req *Request) resetSkipHeader() {
 	req.postArgs.Reset()
 	req.parsedPostArgs = false
 	req.isTLS = false
+	req.bodyStreamDrainErr = false
 }
 
 // RemoveMultipartFormFiles removes multipart/form-data temporary files
--- a/http.go
+++ b/http.go
@@ -1144,21 +1144,26 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 
 		mr := multipart.NewReader(bodyStream, req.multipartFormBoundary)
 		req.multipartForm, err = mr.ReadForm(8 * 1024)
-		if err != nil {
-			if lr != nil && lr.N <= 0 {
-				return nil, fmt.Errorf("cannot read multipart/form-data body: %w", ErrBodyTooLarge)
+
+		// ReadForm stops at the closing boundary, and on a parse error it stops
+		// wherever it gave up, so either way the rest of the declared body is
+		// still pending. Discard it through bodyStream so the leftover keeps
+		// counting towards maxBodySize, then off the raw stream so none of the
+		// declared body is left behind to be read back as the beginning of the
+		// next request on a keep-alive connection.
+		_, drainErr := copyZeroAlloc(io.Discard, bodyStream)
+		if bodyStream != req.bodyStream {
+			if _, rawErr := copyZeroAlloc(io.Discard, req.bodyStream); drainErr == nil {
+				drainErr = rawErr
 			}
-			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
+		}
+		if err == nil {
+			err = drainErr
 		}
 		if lr != nil && lr.N <= 0 {
-			req.RemoveMultipartFormFiles()
-			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", ErrBodyTooLarge)
+			err = ErrBodyTooLarge
 		}
-		// ReadForm stops at the closing boundary, so the rest of the declared
-		// body is still pending on the stream. Discard it, otherwise those
-		// bytes are read back as the beginning of the next request on a
-		// keep-alive connection.
-		if _, err = copyZeroAlloc(io.Discard, req.bodyStream); err != nil {
+		if err != nil {
 			req.RemoveMultipartFormFiles()
 			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
 		}
--- a/http.go
+++ b/http.go
@@ -1116,6 +1116,20 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 		return req.multipartForm, nil
 	}
 
+	// Every declared body byte has to come off req.bodyStream before we
+	// return, on the error paths as much as the success one. ReadForm stops at
+	// the closing boundary and the early errors (missing boundary, unsupported
+	// content-encoding, bad gzip header) stop even sooner, so without this the
+	// leftover bytes are read back as the start of the next request on a
+	// keep-alive connection. Draining req.bodyStream directly also avoids
+	// comparing reader interface values, which panics for an uncomparable
+	// dynamic type set through SetBodyStream.
+	if req.bodyStream != nil {
+		defer func() {
+			_, _ = copyZeroAlloc(io.Discard, req.bodyStream)
+		}()
+	}
+
 	req.multipartFormBoundary = string(req.Header.MultipartFormBoundary())
 	if req.multipartFormBoundary == "" {
 		return nil, ErrNoMultipartForm
@@ -1145,19 +1159,11 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 		mr := multipart.NewReader(bodyStream, req.multipartFormBoundary)
 		req.multipartForm, err = mr.ReadForm(8 * 1024)
 
-		// ReadForm stops at the closing boundary, and on a parse error it stops
-		// wherever it gave up, so either way the rest of the declared body is
-		// still pending. Discard it through bodyStream so the leftover keeps
-		// counting towards maxBodySize, then off the raw stream so none of the
-		// declared body is left behind to be read back as the beginning of the
-		// next request on a keep-alive connection.
-		_, drainErr := copyZeroAlloc(io.Discard, bodyStream)
-		if bodyStream != req.bodyStream {
-			if _, rawErr := copyZeroAlloc(io.Discard, req.bodyStream); drainErr == nil {
-				drainErr = rawErr
-			}
-		}
-		if err == nil {
+		// Drain the rest of the declared body through bodyStream so the
+		// leftover keeps counting towards maxBodySize; the deferred drain then
+		// clears whatever remains on the raw stream so the connection stays
+		// framed.
+		if _, drainErr := copyZeroAlloc(io.Discard, bodyStream); err == nil {
 			err = drainErr
 		}
 		if lr != nil && lr.N <= 0 {
--- a/http.go
+++ b/http.go
@@ -1176,8 +1176,17 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 		// leftover keeps counting towards maxBodySize; the deferred drain then
 		// clears whatever remains on the raw stream so the connection stays
 		// framed.
-		if _, drainErr := copyZeroAlloc(io.Discard, bodyStream); err == nil {
-			err = drainErr
+		if _, drainErr := copyZeroAlloc(io.Discard, bodyStream); drainErr != nil {
+			// The stream stopped short of the end of the declared body, so the
+			// connection is mid-body. Mark it sticky here rather than leaving it
+			// to the deferred drain: that retry can resume past the bad byte at
+			// a valid terminating chunk and return nil, which would otherwise
+			// leave the connection reusable and the leftover parsed as the next
+			// request.
+			req.bodyStreamDrainErr = true
+			if err == nil {
+				err = drainErr
+			}
 		}
 		if lr != nil && lr.N <= 0 {
 			err = ErrBodyTooLarge
--- a/http.go
+++ b/http.go
@@ -1249,6 +1249,14 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 	if err != nil {
 		return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
 	}
+	// ReadForm stops at the closing boundary, so anything the sender placed
+	// between it and Content-Length is still unread on r. Discard it, since
+	// on a keep-alive connection those bytes are otherwise read back as the
+	// beginning of the next request.
+	if _, err = copyZeroAlloc(io.Discard, lr); err != nil {
+		f.RemoveAll() //nolint:errcheck
+		return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
+	}
 	return f, nil
 }
 
--- a/http.go
+++ b/http.go
@@ -1154,6 +1154,14 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 			req.RemoveMultipartFormFiles()
 			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", ErrBodyTooLarge)
 		}
+		// ReadForm stops at the closing boundary, so the rest of the declared
+		// body is still pending on the stream. Discard it, otherwise those
+		// bytes are read back as the beginning of the next request on a
+		// keep-alive connection.
+		if _, err = copyZeroAlloc(io.Discard, req.bodyStream); err != nil {
+			req.RemoveMultipartFormFiles()
+			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
+		}
 	} else {
 		body := req.bodyBytes()
 		if bytes.Equal(ce, strGzip) {
@@ -1243,7 +1251,7 @@ func readMultipartForm(r io.Reader, boundary string, size, maxInMemoryFileSize i
 	if size <= 0 {
 		return nil, fmt.Errorf("form size must be greater than 0: given %d", size)
 	}
-	lr := io.LimitReader(r, int64(size))
+	lr := &io.LimitedReader{R: r, N: int64(size)}
 	mr := multipart.NewReader(lr, boundary)
 	f, err := mr.ReadForm(int64(maxInMemoryFileSize))
 	if err != nil {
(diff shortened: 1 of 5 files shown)


erikdubbelboer (changes requested):
This first test shows that it still isn't fixed for servers with `StreamRequestBody: true`:
```go
type oneByteReadWriter struct{ readWriter }

func (rw *oneByteReadWriter) Read(p []byte) (int, error) {
	if len(p) > 1 {
		p = p[:1]
	}
	return rw.readWriter.Read(p)
}

func TestServerDoesNotParseStreamedMultipartEpilogueAsRequest(t *testing.T) {
	form := "--x\r\nContent-Disposition: form-data; name=\"a\"\r\n\r\n1\r\n--x--\r\n"
	smuggled := "GET /smuggled HTTP/1.1\r\nHost: x\r\n\r\n"
	body := form + smuggled

	rw := &oneByteReadWriter{}
	fmt.Fprintf(&rw.r,
		"POST /first HTTP/1.1\r\n"+
			"Host: x\r\n"+
			"Content-Type: multipart/form-data; boundary=x\r\n"+
			"Content-Length: %d\r\n\r\n%s",
		len(body), body,
	)

	var paths []string
	s := Server{
		StreamRequestBody:            true,
		DisablePreParseMultipartForm: true,
		MaxRequestBodySize:           1, // Force RequestBodyStream.
			paths = append(paths, string(ctx.Path()))
			if string(ctx.Path()) == "/first" {
				if _, err := ctx.MultipartForm(); err != nil {
					t.Fatal(err)
				}
			}
		},
	}

	_ = s.ServeConn(rw)

	if len(paths) != 1 {
		t.Fatalf("handler paths = %q; want only [/first]", paths)
	}
}
```


This test shows that we now call the `Handler` for broken requests:
```go
func TestServerRejectsTruncatedMultipartBody(t *testing.T) {
	body := "--x\r\nContent-Disposition: form-data; name=\"a\"\r\n\r\n1\r\n--x--\r\n"

	rw := &readWriter{}
	fmt.Fprintf(&rw.r,
		"POST / HTTP/1.1\r\n"+
			"Host: x\r\n"+
			"Content-Type: multipart/form-data; boundary=x\r\n"+
			"Content-Length: %d\r\n\r\n%s",
		len(body)+1, body,
	)

	called := false
	s := Server{
		Handler: func(*RequestCtx) {
			called = true
		},
	}

	err := s.ServeConn(rw)

	if called {
		t.Error("handler was called for a truncated request body")
	}
	if !errors.Is(err, io.ErrUnexpectedEOF) {
		t.Errorf("ServeConn error = %v; want unexpected EOF", err)
	}
}
```

alhudz:
Both fixed, and your two tests are now in `server_test.go`.

Streamed case: `ctx.MultipartForm()` on a streamed body goes through the `req.bodyStream` branch of `MultipartFormWithLimit`, which never had the drain, only `readMultipartForm` did. It now discards the remainder of `req.bodyStream` after `ReadForm` returns, so the connection reader ends up at the end of the declared body either way.

Truncated case: `readMultipartForm` keeps the `*io.LimitedReader` now and checks `lr.N` after draining, so a body that ends short comes back as `io.ErrUnexpectedEOF` instead of a form. For what it's worth that one reproduces on master too, the multipart path just never enforced the declared length the way `readBody` does, but it's covered now.

Both tests fail on `f3f44b0` and pass on `39ae4e5`. `go test -race -shuffle=on ./...` and `golangci-lint` are clean.

One thing I deliberately left out: with `StreamRequestBody` a handler that never touches the body leaves it on the connection and the next request gets parsed out of those bytes, no multipart involved. That looks like a different fix so I did not want to fold it in here.

http.go
```diff
@@ -1154,6 +1154,14 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
 			req.RemoveMultipartFormFiles()
 			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", ErrBodyTooLarge)
 		}
+		// ReadForm stops at the closing boundary, so the rest of the declared
```
erikdubbelboer: This drain is only reached when `ReadForm` succeeds. A malformed multipart header returns earlier, after which server cleanup releases the `requestStream` without consuming it. With `StreamRequestBody=true`, `DisablePreParseMultipartForm=true`, `MaxRequestBodySize=1`, and `--x\r\nMalformed-Header\r\nGET /smuggled ...` inside the declared body, my server-level test observed handlers for both `/first` and `/smuggled`. Please drain the declared stream, or force connection close, on every parse and size-limit error path and add a malformed-multipart regression.

http.go
```diff
@@ -1154,6 +1154,14 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
+		// body is still pending on the stream. Discard it, otherwise those
+		// bytes are read back as the beginning of the next request on a
+		// keep-alive connection.
+		if _, err = copyZeroAlloc(io.Discard, req.bodyStream); err != nil {
```
erikdubbelboer: `copyZeroAlloc` cannot prove that a fixed-length `requestStream` reached `Content-Length`: when its underlying reader ends early, `requestStream.Read` returns ordinary `io.EOF`, which the copy treats as success. With a valid closing boundary and `Content-Length = len(body)+1`, streamed `ctx.MultipartForm()` returned a form with no error. Please surface `io.ErrUnexpectedEOF` for premature fixed-length EOF (or explicitly verify the stream's read count) and add the streaming counterpart of the truncated-body test.

erikdubbelboer (commented):
The two cases from my earlier review are fixed, but these three streamed-path cases still fail on the current head.

http.go
```diff
@@ -1154,6 +1154,14 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
+			req.RemoveMultipartFormFiles()
+			return nil, fmt.Errorf("cannot read multipart/form-data body: %w", err)
```
erikdubbelboer, line 1196: Draining `req.bodyStream` directly bypasses the `lr` used to enforce `maxBodySize`. A small valid form plus a 1 KiB epilogue passed `MultipartFormWithLimit(len(form)+1)` with no error. Since every byte inside `Content-Length` is request body—and the buffered path counts all of it—the streamed drain must preserve limit accounting while still consuming or closing the remainder for safe framing.

http.go
alhudz: Right, the drain sat behind the success path only. `ReadForm` now runs into the same drain whether it succeeded or gave up mid-parse, so a malformed part header ends up at the end of the declared body like any other. Added `TestServerDoesNotParseMalformedStreamedMultipartAsRequest` with your `--x\r\nMalformed-Header\r\n` body: `["/first" "/smuggled"]` on `39ae4e5`, only `/first` now.

http.go
alhudz: Good catch, the copy could not tell a complete body from a truncated one. Fixed it in `requestStream.Read` instead of at the call site: a fixed-length stream whose reader hits EOF short of `ContentLength` now returns `io.ErrUnexpectedEOF`, same as the chunked branch just above already does. `TestServerRejectsTruncatedStreamedMultipartBody` covers the streamed counterpart with `Content-Length = len(body)+1`.

http.go
alhudz: Agreed, every byte inside `Content-Length` is body and the limit has to see it. The drain now goes through `bodyStream` rather than `req.bodyStream`, so the epilogue is charged to `lr`, and if that trips the limit the remainder still gets discarded off the raw stream so the connection stays framed while the caller gets `ErrBodyTooLarge`. `TestServerStreamedMultipartEpilogueCountsTowardsLimit` checks both halves: your form plus a 1 KiB epilogue against `MultipartFormWithLimit(len(form)+1)` returns `ErrBodyTooLarge`, and a request appended after the epilogue is not executed.

alhudz:
All three streamed cases fixed in `d0f28ef`, replies inline.

Short version: the drain moved out from behind the success path so it also runs when `ReadForm` bails mid-parse, it now goes through `bodyStream` so the epilogue is charged to `lr`, and `requestStream.Read` reports `io.ErrUnexpectedEOF` when a fixed-length body ends short of `Content-Length` instead of a plain `io.EOF` the copy reads as success.

Three regressions in `server_test.go`: `TestServerDoesNotParseMalformedStreamedMultipartAsRequest`, `TestServerRejectsTruncatedStreamedMultipartBody`, `TestServerStreamedMultipartEpilogueCountsTowardsLimit`. All three fail on `39ae4e5` and pass on `d0f28ef`; the limit one also asserts the connection is still framed after `ErrBodyTooLarge`. `gofmt`, `go vet`, `golangci-lint` and `go test -race -shuffle=on ./...` are clean.

Still out of scope, same as last time: with `StreamRequestBody` a handler that never reads the body at all leaves it on the connection, no multipart involved. That one wants a fix in the serve loop rather than here.

http.go
```diff
@@ -1144,15 +1144,28 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
-			if lr != nil && lr.N <= 0 {
-				return nil, fmt.Errorf("cannot read multipart/form-data body: %w", ErrBodyTooLarge)
+
+		// ReadForm stops at the closing boundary, and on a parse error it stops
+		// wherever it gave up, so either way the rest of the declared body is
+		// still pending. Discard it through bodyStream so the leftover keeps
+		// counting towards maxBodySize, then off the raw stream so none of the
+		// declared body is left behind to be read back as the beginning of the
+		// next request on a keep-alive connection.
+		_, drainErr := copyZeroAlloc(io.Discard, bodyStream)
```
erikdubbelboer: These drains still sit after every setup-time return. With `StreamRequestBody=true`, `DisablePreParseMultipartForm=true`, and `MaxRequestBodySize=1`, a missing boundary, `Content-Encoding: br`, or an invalid gzip header returns before this block. In each case I padded the bytes consumed before `GET /smuggled ...` inside the declared `Content-Length`, and observed handlers for both `/first` and `/smuggled`. Please route every multipart exit through a drain of the declared stream, or make the server close the connection while unread body bytes remain, and add regressions for these early-error paths.

http.go
```diff
@@ -1144,15 +1144,28 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
+		if bodyStream != req.bodyStream {
```
erikdubbelboer: `Request.SetBodyStream` accepts any `io.Reader`, but comparing interface values here panics when the dynamic type is not comparable. A value reader such as `struct { data []byte; off *int }` parses successfully and then `MultipartForm()` panics with `runtime error: comparing uncomparable type`. Please track whether `bodyStream` was wrapped with a boolean (or equivalent) instead of comparing arbitrary reader interfaces, and add a regression.

erikdubbelboer (commented):
Two remaining issues on the current head.

http.go
alhudz: Right, those all returned before the drain block. Moved the raw-stream drain into a `defer` at the top of the streamed branch, so every exit routes through it now: missing boundary, `Content-Encoding: br`, and the bad gzip header included. The limit accounting stays where it was, the `bodyStream` drain still charges the epilogue to `lr`, the `defer` only clears whatever is left on the raw stream.

Three regressions in `server_test.go`: `TestServerDoesNotParseStreamedMultipartMissingBoundaryAsRequest`, `TestServerDoesNotParseStreamedMultipartBadContentEncodingAsRequest`, `TestServerDoesNotParseStreamedMultipartBadGzipAsRequest`. Each pads the smuggled request to the stop offset (0 for the first two, 10 for the gzip header read). All three give `["/first" "/smuggled"]` on `d0f28ef` and only `/first` on `34182dc`.

http.go
alhudz: Good catch. Dropped the interface comparison entirely: the `defer` drains `req.bodyStream` unconditionally, so there is nothing to compare against any more. `TestRequestMultipartFormUncomparableBodyStream` sets a value reader with a slice field through `SetBodyStream` and calls `MultipartForm`: `panic: comparing uncomparable type fasthttp.uncomparableReader` on `d0f28ef`, parses the form cleanly on `34182dc`.

alhudz:
Both fixed in `34182dc`, replies inline.

The raw-stream drain now sits in a `defer` at the top of the streamed branch, so every exit path routes through it, the three early errors (missing boundary, unsupported content-encoding, bad gzip header) included, and there is no reader-interface comparison left to panic on. Limit accounting is unchanged: the epilogue still drains through `lr` first.

Four regressions: `TestServerDoesNotParseStreamedMultipartMissingBoundaryAsRequest`, `TestServerDoesNotParseStreamedMultipartBadContentEncodingAsRequest`, `TestServerDoesNotParseStreamedMultipartBadGzipAsRequest` in `server_test.go`, and `TestRequestMultipartFormUncomparableBodyStream` in `http_test.go`. All four fail on `d0f28ef` (the first three run the handler for `/smuggled`, the last panics) and pass on `34182dc`. `gofmt`, `go vet`, `golangci-lint` and `go test -race -shuffle=on ./...` are clean.

http.go
```diff
@@ -1116,6 +1116,20 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
+	// return, on the error paths as much as the success one. ReadForm stops at
+	// the closing boundary and the early errors (missing boundary, unsupported
+	// content-encoding, bad gzip header) stop even sooner, so without this the
+	// leftover bytes are read back as the start of the next request on a
+	// keep-alive connection. Draining req.bodyStream directly also avoids
+	// comparing reader interface values, which panics for an uncomparable
+	// dynamic type set through SetBodyStream.
+	if req.bodyStream != nil {
+		defer func() {
+			_, _ = copyZeroAlloc(io.Discard, req.bodyStream)
```
erikdubbelboer: This drain error cannot be ignored. With StreamRequestBody=true, DisablePreParseMultipartForm=true, and a handler using ctx.FormValue("x"), a malformed chunked body beginning with \x00GET /smuggled ... makes requestStream.Read consume the invalid chunk-size byte and return an error. The defer drops that error, server cleanup releases the stream, and the next keep-alive iteration executes /smuggled. My regression observed handlers ["/first", "/smuggled"] on 34182dc; the identical test invokes only /first on d0f28ef. Please make any failed drain force the connection non-reusable or closed and add a malformed-chunk regression.

alhudz: Right, the defer swallowed the drain error and the connection got reused mid-body. A failed drain now sets `req.bodyStreamDrainErr`, and the serve loop folds that into `connectionClose`, so the connection is closed instead of the leftover being parsed as the next request. `TestServerDoesNotParseStreamedMultipartMalformedChunkAsRequest` uses a valid form chunk followed by your `\x00` chunk-size and `GET /smuggled`: `["/first" "/smuggled"]` on `34182dc`, only `/first` on `67f0ba2`. `gofmt`, `go vet`, `golangci-lint` and `go test -race -shuffle=on ./...` are clean.

http.go
```diff
@@ -1116,6 +1123,26 @@ func (req *Request) MultipartFormWithLimit(maxBodySize int) (*multipart.Form, er
...
+	// return, on the error paths as much as the success one. ReadForm stops at
+	// the closing boundary and the early errors (missing boundary, unsupported
+	// content-encoding, bad gzip header) stop even sooner, so without this the
+	// leftover bytes are read back as the start of the next request on a
+	// keep-alive connection. Draining req.bodyStream directly also avoids
+	// comparing reader interface values, which panics for an uncomparable
+	// dynamic type set through SetBodyStream.
+	if req.bodyStream != nil {
+		defer func() {
+			if _, drainErr := copyZeroAlloc(io.Discard, req.bodyStream); drainErr != nil {
```
erikdubbelboer, line 1136: The close bit only reflects the deferred retry's error, not an earlier framing error. If the drain below consumes one invalid chunk-size byte and fails, this defer can resume at a valid terminating chunk and return nil. `bodyStreamDrainErr` stays false and the bytes after the terminator are accepted as another request.

This regression fails on `67f0ba2` with handler paths `["/first" "/smuggled"]`:

```go
func TestServerClosesAfterAnyStreamedMultipartDrainError(t *testing.T) {
	form := "--x\r\nContent-Disposition: form-data; name=\"a\"\r\n\r\n1\r\n--x--\r\n"
	smuggled := "GET /smuggled HTTP/1.1\r\nHost: x\r\n\r\n"
	body := fmt.Sprintf("%x\r\n%s\r\n", len(form), form) +
		"\x00" + "0\r\n\r\n" + smuggled

	rw := &oneByteReadWriter{}
	fmt.Fprintf(&rw.r,
		"POST /first HTTP/1.1\r\nHost: x\r\n"+
			"Content-Type: multipart/form-data; boundary=x\r\n"+
			"Transfer-Encoding: chunked\r\n\r\n%s",
		body,
	)

	var paths []string
	s := Server{
		StreamRequestBody:            true,
		DisablePreParseMultipartForm: true,
		MaxRequestBodySize:           1,
			paths = append(paths, string(ctx.Path()))
			if string(ctx.Path()) == "/first" {
				_ = ctx.FormValue("x")
			}
		},
	}

	_ = s.ServeConn(rw)
	if len(paths) != 1 {
		t.Fatalf("handler paths = %q; want only [/first]", paths)
	}
}
```

The first framing error needs to remain sticky; a later retry must not make the connection reusable.

server.go
```diff
@@ -2653,6 +2653,7 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 		connectionClose = connectionClose ||
 			(s.MaxRequestsPerConn > 0 && connRequestNum >= uint64(s.MaxRequestsPerConn)) || // #nosec G115
 			ctx.Response.Header.ConnectionClose() ||
+			ctx.Request.bodyStreamDrainErr ||
```
erikdubbelboer: This reads `bodyStreamDrainErr` from the replacement context. The `timeoutResponse` branch above swaps out `ctx`, so `TimeoutErrorWithCode` (and therefore `TimeoutHandler`) leaves the close bit on the old request while this check sees a fresh false value.

This deterministic regression fails on `67f0ba2` with handler paths `["/first" "/smuggled"]`:

```go
func TestServerPreservesMultipartDrainErrorAcrossTimeoutResponse(t *testing.T) {
	form := "--x\r\nContent-Disposition: form-data; name=\"a\"\r\n\r\n1\r\n--x--\r\n"
	smuggled := "GET /smuggled HTTP/1.1\r\nHost: x\r\n\r\n"
	body := fmt.Sprintf("%x\r\n%s\r\n", len(form), form) + "\x00" + smuggled

	rw := &oneByteReadWriter{}
	fmt.Fprintf(&rw.r,
		"POST /first HTTP/1.1\r\nHost: x\r\n"+
			"Content-Type: multipart/form-data; boundary=x\r\n"+
			"Transfer-Encoding: chunked\r\n\r\n%s",
		body,
	)

	var paths []string
	s := Server{
		StreamRequestBody:            true,
		DisablePreParseMultipartForm: true,
		MaxRequestBodySize:           1,
			path := string(ctx.Path())
			paths = append(paths, path)
			if path == "/first" {
				_ = ctx.FormValue("x")
				ctx.TimeoutErrorWithCode("timeout", StatusRequestTimeout)
			}
		},
	}

	_ = s.ServeConn(rw)
	if len(paths) != 1 {
		t.Fatalf("handler paths = %q; want only [/first]", paths)
	}
}
```

`Request.Reset()` can similarly clear this bit before the serve loop observes it. Please keep the failure in server/connection-owned sticky state until the loop has made the close decision.

http.go
alhudz: Fixed in `c7802aa`. The inline drain now sets `bodyStreamDrainErr` the moment it fails, so the first framing error is sticky and a deferred drain that resumes past the bad byte at the terminating chunk can't clear it. `TestServerClosesAfterAnyStreamedMultipartDrainError` with your `\x00` + `0\r\n\r\n` body gives `["/first" "/smuggled"]` on `67f0ba2`, only `/first` now.

server.go
alhudz: Fixed in `c7802aa`. The serve loop now reads `bodyStreamDrainErr` off the original `ctx.Request` right after the handler and holds it in a connection-local, before `timeoutResponse` swaps `ctx` or `Reset` can clear it, so the close decision still sees it. `TestServerPreservesMultipartDrainErrorAcrossTimeoutResponse` gives `["/first" "/smuggled"]` on `67f0ba2`, only `/first` now.

PR #2340, opened 2026-07-28
fix: don't close the shutdown done channel twice

ReneWerner87:
`ShutdownWithContext` closes `s.done`, then waits for the open connections to drain. When it gives up on its context it returns without clearing `s.done`, so the channel stays closed and still in place.

`Serve` then keeps that closed channel, since it only creates one when `s.done` is nil, and the next `ShutdownWithContext` closes it a second time:

```
panic: close of closed channel
    (*Server).ShutdownWithContext  server.go:2072
    (*Server).Shutdown             server.go:2044
```

The `Serve` in between is required. Without it the retry hits the `s.ln == nil` early return and reports success while a connection is still running.

The fix tracks whether the channel was already closed, so the second close is skipped. The success path clears the flag along with the channel, so a shutdown that does drain leaves the server as it was. All of it sits under `s.mu`, which already guards every write to `s.done`.

`s.done` itself is deliberately left in place rather than replaced. The request that caused the timeout is still running, and `RequestCtx.Done()` reads `ctx.s.done`, so swapping the channel would hand that request a different one on its next call. The test pins both ends: it panics if the second close comes back, and it fails the `Done()` stability check if `Serve` swaps the channel.

Deliberately left open, all older than this change: a `Serve` after a timed out shutdown still hands new requests the already closed channel; the retry without a `Serve` returns nil while a connection is still running; and `s.done = nil` on the success path races with the post-release `Done()` that `RequestCtx.reset()` explicitly permits. Those need a done channel that is stable per request instead of read live from the server.

Full suite green under `-race -shuffle=on` and without `-race`, lint clean with v2.10.1, the new test 40x green under `-race -shuffle=on`.

### DIFF
--- a/server.go
+++ b/server.go
@@ -247,6 +247,10 @@ type Server struct {
 	idleConns map[net.Conn]*atomic.Int64
 	done      chan struct{}
 
+	// Whether done was already closed. A ShutdownWithContext that gives up on
+	// its context leaves it closed but in place, and it must not be closed twice.
+	doneClosed bool
+
 	// Server name for sending in response headers.
 	//
 	// Default server name is used if left blank.
@@ -2068,8 +2072,9 @@ func (s *Server) ShutdownWithContext(ctx context.Context) (err error) {
 
 	lnerr := s.closeListenersLocked()
 
-	if s.done != nil {
+	if s.done != nil && !s.doneClosed {
 		close(s.done)
+		s.doneClosed = true
 	}
 
 	// Closing the listener will make Serve() call Stop on the worker pool.
@@ -2084,6 +2089,7 @@ func (s *Server) ShutdownWithContext(ctx context.Context) (err error) {
 		if open := s.open.Load(); open == 0 {
 			// There may be a pending request to call ctx.Done(). Therefore, we only set it to nil when open == 0.
 			s.done = nil
+			s.doneClosed = false
 			return lnerr
 		}
 		// This is not an optimal solution but using a sync.WaitGroup
--- a/server_test.go
+++ b/server_test.go
@@ -4429,6 +4429,75 @@ func TestCloseIdleConnsKeepsTimestampOwnedByItsConn(t *testing.T) {
 	}
 }
 
+func TestShutdownWithContextTimeoutThenShutdownAgain(t *testing.T) {
+	t.Parallel()
+
+	ln := fasthttputil.NewInmemoryListener()
+	firstInHandler := make(chan struct{})
+	release := make(chan struct{})
+	doneChanged := make(chan bool, 1)
+
+	var served atomic.Int32
+	s := &Server{
+		Handler: func(ctx *RequestCtx) {
+			if served.Add(1) == 1 {
+				done := ctx.Done()
+				close(firstInHandler)
+				<-release
+				doneChanged <- done != ctx.Done()
+
+				return
+			}
+			ctx.Success("aaa/bbb", []byte("real response"))
+		},
+		Logger: &testLogger{},
+	}
+	go func() {
+		_ = s.Serve(ln)
+	}()
+
+	conn, err := ln.Dial()
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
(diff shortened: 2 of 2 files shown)


erikdubbelboer:
Thanks for the detailed repro. I think replacing `s.done` still leaves a correctness problem for the request that caused the shutdown timeout.

That request is necessarily still alive when `ShutdownWithContext` returns. A later `Serve` assigns a new channel to `s.done`, while `RequestCtx.Done()` reads `ctx.s.done`. If those happen concurrently this is a data race; even with synchronization, two `Done()` calls on the same live request can return different channels (the old closed channel, then the new open one). Besides violating the `context.Context` contract, that can make the old request stop observing the shutdown cancellation.

The existing test can expose the semantic problem with a small extension:

```diff
@@
 	ctxDone := make(chan bool, 1)
+	firstCtxDoneChanged := make(chan bool, 1)
@@
 		Handler: func(ctx *RequestCtx) {
 			if served.Add(1) == 1 {
+				done := ctx.Done()
 				close(firstInHandler)
 				<-release
+				firstCtxDoneChanged <- done != ctx.Done()
 				return
@@
-	close(release)
-
 	ln = fasthttputil.NewInmemoryListener()
@@
 	verifyResponse(t, bufio.NewReader(conn), StatusOK, "aaa/bbb", "real response")
+	close(release)
+	if <-firstCtxDoneChanged {
+		t.Fatal("RequestCtx.Done() changed for the request still running after shutdown timed out")
+	}
```

On `8cad55478b18b16233eb4f733ca1b611cbcd4e9a`, this deterministically fails with:

```
server_test.go:4431: RequestCtx.Done() changed for the request still running after shutdown timed out
```

The old handler probably needs a stable channel belonging to its request/serve generation. If reuse after a failed shutdown remains outside the contract instead, another option is to keep the close idempotent but reject restarting until the previous requests have drained.

ReneWerner87:
You're right, and thanks for the precise repro. Your extension fails deterministically here too, 5 runs out of 5 on `8cad554`:

```
server_test.go:4431: RequestCtx.Done() changed for the request still running after shutdown timed out
```

I dropped the `Serve` side of the change entirely. What is left is only the idempotent close, so `s.done` is never replaced and the request that caused the timeout keeps the channel it started with. Your assertion is now part of the test, and it fails if anyone reintroduces the swap.

Three things I found while checking your two suggestions, in case they are useful.

**The panic needs an intervening `Serve`.** Without one the retry hits `if s.ln == nil { return nil }` and reports success while a connection is still running. So these are two separate defects and this PR only removes the crash. I corrected the description accordingly.

**Rejecting the restart until the previous requests have drained does not hold as stated.** The natural drain check is `s.open.Load() == 0` under `s.mu`, but `ServeConn` increments `s.open` at `server.go:2212` without taking `s.mu`. `Serve` can therefore observe 0, install a fresh channel, and only then have `ServeConn` proceed and read `s.done` unsynchronized. Same race, just a narrower window.

**A stable per-generation channel collides with an existing decision.** `RequestCtx.reset()` says:

```go
// Don't reset ctx.s!
// We have a pool per server so the next time this ctx is used it
// will be assigned the same value again.
// ctx might still be in use for context.Done() and context.Err()
// which are safe to use as they only use ctx.s and no other value.
```

A snapshot field on `RequestCtx` breaks that as soon as the pooled ctx is recycled, so the stable channel would have to live somewhere the pool does not reuse.

**That contract also has a pre-existing race on master.** The success path's `s.done = nil` races with exactly the post-release `Done()` the comment permits:

```
WARNING: DATA RACE
Write at 0x00c0000fe3b0 by goroutine 11:
  (*Server).ShutdownWithContext   server.go:2086
Previous read at 0x00c0000fe3b0 by goroutine 15:
  (*RequestCtx).Done()            server.go:2998
```

This PR does not fix that, and neither would the swap. Whichever design you land on, the read side needs synchronizing.

So this is now scoped to "the second close must not panic", and leaves three things open: the race above, `Serve` after a timed out shutdown handing new requests an already fired `Done()`, and the retry without a `Serve` returning nil. Happy to take any of them on separately if you want to settle the direction first.

ReneWerner87:
@erikdubbelboer ready for the re-review

erikdubbelboer:
Thanks!

PR #2343, opened 2026-07-30
fix: reject out-of-range Content-Length instead of wrapping it

lenamonj:
`parseUintBuf` tests for overflow by comparing the product against the accumulator after computing it:

```go
vNew := 10*v + int(k)
// Test for overflow.
if vNew < v {
    return -1, i, errTooLongInt
}
```

`10*v` wraps modulo 2^64 and can land back above `v`, so the wrap goes unnoticed and a wrong value is returned in place of an error. `parseContentLength` calls `parseUintBuf` directly, so this reaches header parsing on both the request and the response side.

**Reproduction on master (`f83ac8c3`):**

```go
raw := "POST / HTTP/1.1\r\nHost: x\r\nContent-Length: 21000000000000000000\r\n\r\n"
var h RequestHeader
err := h.Read(bufio.NewReader(bytes.NewBufferString(raw)))
// err               = <nil>
// h.ContentLength() = 2553255926290448384
```

`strconv.ParseInt` rejects the same string as out of range.

The narrow band is why this survived: obvious values like `9223372036854775808` and `99999999999999999999` are already rejected correctly. It only shows up where `10*v` wraps back above `v`.

The fix moves the check before the multiplication, so the decimal strings `ParseUint` accepts are the ones `strconv.ParseInt` accepts for the platform's int size.

**Tests**, two added, both failing on master before the change:

- `TestParseUintMatchesStrconv` compares `ParseUint` against `strconv.ParseInt` across the wrap band. On master it fails at `ParseUint("21000000000000000000") = (2553255926290448384, <nil>)`.
- `TestParseContentLengthRejectsOverflow` pins the same thing at the header layer, request and response side.

`go test ./...` passes on all packages with the change applied.

### DIFF
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"net/http"
 	"strconv"
@@ -258,6 +259,10 @@ func AppendUint(dst []byte, n int) []byte {
 }
 
 // ParseUint parses uint from buf.
+//
+// A value too large for an int is an error rather than a wrapped result, so the
+// set of strings ParseUint accepts is the set strconv.ParseInt accepts for the
+// platform's int size.
 func ParseUint(buf []byte) (int, error) {
 	v, n, err := parseUintBuf(buf)
 	if n != len(buf) {
@@ -289,12 +294,18 @@ func parseUintBuf(b []byte) (int, int, error) {
 			}
 			return v, i, nil
 		}
-		vNew := 10*v + int(k)
-		// Test for overflow.
-		if vNew < v {
+		// Test for overflow before it happens. Comparing the product against
+		// the accumulator afterwards is not enough: 10*v wraps modulo 2^64 and
+		// can land back above v, in which case the overflow goes unnoticed and
+		// a wrong value is returned instead of an error.
+		if v > math.MaxInt/10 {
 			return -1, i, errTooLongInt
 		}
-		v = vNew
+		v *= 10
+		if v > math.MaxInt-int(k) {
+			return -1, i, errTooLongInt
+		}
+		v += int(k)
 	}
 	return v, n, nil
 }
--- a/bytesconv.go
+++ b/bytesconv.go
@@ -260,9 +260,9 @@ func AppendUint(dst []byte, n int) []byte {
 
 // ParseUint parses uint from buf.
 //
-// A value too large for an int is an error rather than a wrapped result, so the
-// set of strings ParseUint accepts is the set strconv.ParseInt accepts for the
-// platform's int size.
+// A value too large for an int is an error rather than a wrapped result, so
+// ParseUint accepts exactly the unsigned decimal strings whose value fits in an
+// int on the current platform.
 func ParseUint(buf []byte) (int, error) {
 	v, n, err := parseUintBuf(buf)
 	if n != len(buf) {
@@ -279,14 +279,25 @@ var (
 	errTooLongInt             = errors.New("too long int")
 )
 
+const (
+	// maxIntDiv10 is the largest accumulator that can still take another digit.
+	// Anything above it overflows an int when multiplied by 10.
+	maxIntDiv10 = math.MaxInt / 10
+
+	// maxSafeIntDigits is how many leading decimal digits can never overflow an
+	// int, whatever the word size: 10**18-1 fits a 64-bit int and 10**9-1 fits a
+	// 32-bit one. Go defines strconv.IntSize as 32 or 64 and nothing else.
+	// TestMaxSafeIntDigits checks both halves of that claim on the build's own
+	// int size.
+	maxSafeIntDigits = 9 * (strconv.IntSize / 32)
+)
+
 func parseUintBuf(b []byte) (int, int, error) {
-	n := len(b)
-	if n == 0 {
+	if len(b) == 0 {
 		return -1, 0, errEmptyInt
 	}
 	v := 0
-	for i := range n {
-		c := b[i]
+	for i, c := range b {
 		k := c - '0'
 		if k > 9 {
 			if i == 0 {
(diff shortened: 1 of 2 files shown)


erikdubbelboer (commented):
The overflow fix is directionally correct and the full suite passed locally when merged into current master. I found three issues to address; details are inline.

bytesconv.go
```diff
@@ -289,12 +294,18 @@ func parseUintBuf(b []byte) (int, int, error) {
...
-		if vNew < v {
+		// Test for overflow before it happens. Comparing the product against
+		// the accumulator afterwards is not enough: 10*v wraps modulo 2^64 and
+		// can land back above v, in which case the overflow goes unnoticed and
+		// a wrong value is returned instead of an error.
+		if v > math.MaxInt/10 {
+			return -1, i, errTooLongInt
+		}
+		v *= 10
+		if v > math.MaxInt-int(k) {
```
erikdubbelboer: This adds two dependent overflow comparisons for every digit in the common path. On an M2 Max with Go 1.26.4, repeated `BenchmarkParseUint` runs moved from about 5.8 ns/op on the base to 9.3 ns/op here (+60%); a dynamic-input benchmark confirmed roughly +56%. `parseUintBuf` is used by Content-Length and response status parsing. Since overflow is impossible before the 10th digit on 32-bit or the 19th on 64-bit, could the full max-value check be gated by the digit index? A local sketch using that shape passed the new tests and restored the benchmark to about 6 ns/op.

bytesconv_test.go
```diff
@@ -334,6 +335,66 @@ func TestParseUintError(t *testing.T) {
...
+// numbers is not enough: the overflow this pins only shows up in a narrow band
+// of values where 10*v wraps back above v.
+func TestParseUintMatchesStrconv(t *testing.T) {
+	t.Parallel()
+
+	values := []string{
+		"0", "1", "9", "10", "255", "1000", "0123", "00000000000000000001",
+		"9223372036854775806", "9223372036854775807", "9223372036854775808",
+		"18446744073709551615", "18446744073709551616",
+		"21000000000000000000", "22000000000000000000", "23000000000000000000",
```
erikdubbelboer, line 371: These hostile values exercise the 64-bit wrap band, but they do not pin the same bug on 32-bit: the old parser errors earlier for all of them. On 32-bit the old implementation incorrectly accepts `"5000000000"` as `705032704` without an error. Please add that value here or to `TestParseUintError32`; the `2^64` explanation in `bytesconv.go` should also be word-size-neutral.

bytesconv.go
```diff
@@ -258,6 +259,10 @@ func AppendUint(dst []byte, n int) []byte {
 }
 
 // ParseUint parses uint from buf.
+//
+// A value too large for an int is an error rather than a wrapped result, so the
+// set of strings ParseUint accepts is the set strconv.ParseInt accepts for the
```
erikdubbelboer: This acceptance-set claim is not exact: `strconv.ParseInt` accepts strings such as `"+1"` and `"-1"`, while `ParseUint` rejects both. Please describe this as unsigned decimal strings whose value fits in `int`, or use `strconv.ParseUint(s, 10, strconv.IntSize-1)` as the test oracle.

bytesconv.go
lenamonj: Fixed, though the cause turned out to be inlining rather than the arithmetic.

`parseUintBuf` sits at inline cost 79 against the budget of 80 on master. The version you reviewed comes to 84, so it stopped being inlined at all three call sites, including both Content-Length paths in `header.go`. That is most of what you measured. The index gate on its own does not recover it: gating while keeping the two-comparison body comes to cost 97, and I measured it no faster than the version you reviewed.

What fits is your gate plus a single combined test. Once `v` is known to be no larger than `math.MaxInt/10` the product cannot exceed `math.MaxInt+2`, so it either fits or wraps to `MinInt`/`MinInt+1`, and a sign test settles the only case left. Below `maxSafeIntDigits` neither part can fire, so the common path keeps one predictable comparison. Back to cost 79 and inlined at all three sites again.

linux/amd64, Go 1.26.2, `-cpu=1 -count=12`, benchstat, against current master:

```
              base       reviewed          final
ParseUint    5.151n   7.271n  +41%    6.229n  +21%
PULen3       2.919n   3.662n  +25%    3.169n   +9%
PULen6       4.392n   6.406n  +46%    6.405n  +46%
PULen7       4.853n   7.119n  +47%    6.240n  +29%
PULen9       6.220n   8.423n  +35%    7.407n  +19%
PULen19     11.72n   15.22n   +30%   13.77n   +17%
PUDynamic    9.300n  10.975n  +18%    9.150n   -2%
geomean      5.787n   7.769n  +34%    6.889n  +19%
```

`PULen*` are fixed-length serial parses I ran locally and did not commit. `PUDynamic` cycles digit lengths 1 through 18 and is the closest of these to a real mix of Content-Length values.

I have not got the fixed-length cases back to parity and I do not think it is reachable inside the inline budget: master spends one comparison per digit and gets the answer wrong, and any correct test spends at least one more. `PULen6` is the odd row, identical to four significant figures on both versions, which reads as code layout rather than the change.

If you would rather have parity than the fix in this shape, say so and I will look again.

bytesconv_test.go
lenamonj: You are right, and it is a band rather than one value. Simulating the old parser with a 32-bit int over round 10 and 11 digit values, these are all accepted with no error:

```
5000000000  6000000000  9999999999  10000000000
15000000000  48000000000  49000000000  53000000000  96000000000
```

I added a representative set here rather than to `TestParseUintError32`. The oracle already decides per platform what correct means, and every one of these is a valid int on 64-bit, so a single table pins the bug on 32-bit and checks the parsed value on 64-bit without needing a build tag or a second copy of the list.

Red on master both ways, green after the fix:

```
amd64  ParseUint("21000000000000000000") = (2553255926290448384, <nil>)
386    ParseUint("5000000000")           = (705032704, <nil>)
```

The `2^64` wording in `bytesconv.go` is gone: the comment now says the product wraps modulo the int size. `maxSafeIntDigits` is derived from `strconv.IntSize` rather than a build tag, and `TestMaxSafeIntDigits` checks both directions of that claim on whatever int size the build targets, so the gate cannot silently go wrong on a platform neither of us tested.

bytesconv.go
lenamonj: Agreed, the claim was not exact. Switched the oracle to `strconv.ParseUint(s, 10, strconv.IntSize-1)`, which accepts exactly the strings naming a non-negative int and rejects the leading sign that `ParseInt` allows, matching what `ParseUint` here actually does.

The doc comment now describes the set directly rather than pointing at another function:

```go
// A value too large for an int is an error rather than a wrapped result, so
// ParseUint accepts exactly the unsigned decimal strings whose value fits in an
// int on the current platform.
```

lenamonj:
Pushed as 6f7802f, replied inline to all three.

The short version: the benchmark regression was `parseUintBuf` falling out of inlining, not the added arithmetic. It sits at inline cost 79 of a budget of 80 on master, so the overflow test pushed it over and it stopped being inlined at all three call sites. Your digit-index gate plus a single combined overflow test brings it back to 79.

Full suite green on amd64, on `GOARCH=386`, and under `-race`.

erikdubbelboer:
Thanks!

PR #2345, opened 2026-07-31
only treat "//" in URI.Update as an authority after a scheme

alhudz:
Repro: `ctx.Redirect("https//evil.com/", 302)` on a request to `http://good.example/dir/page`.
Expected: `Location: http://good.example/dir/https/evil.com/`, which is what `net/url` `ResolveReference` gives for the same pair.
Actual: `Location: https://evil.com/`.

Cause: `updateBytes` calls a reference absolute whenever `bytes.Index(newURI, strSlashSlash)` matches anywhere in it, so a relative reference holding no colon at all is re-parsed as scheme plus authority. `/a//b` and `?q=//x` land in the same branch and come back as `http:///a/b` and `http:///`, with the base host dropped. `getRedirectURL` resolves `Location` headers through the same call, so the client sees it too.

Fix: only take the `//` as an authority delimiter at index 0 or directly after a valid scheme and a colon.

`splitHostURI` is left as is. Its callers hand it something they already assert is a URI, and its behaviour for `://host` and for a malformed scheme is pinned by existing tests.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -404,6 +404,24 @@ func isValidScheme(scheme []byte) bool {
 	return true
 }
 
+// isAuthorityDelimiter reports whether the "//" at index n in uri introduces an
+// authority. Per RFC 3986 that is only the case for a network-path reference,
+// which starts with "//", or when the "//" is preceded by a scheme and a colon.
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Nice find! Thanks!

Issue #79, opened 2016-04-16
Support for setting and reading basic auth headers


mark-rushakoff:
I'm already using fasthttp, and I have an upcoming requirement to support basic auth. The implementation is fairly simple, so I can manually set and parse request headers if I have to, but it would be more convenient if basic auth was part of the RequestHeader API.

kataras:
I have a different opinion, if that's ok, I think it's better to keep libraries like fasthttp as simple as possible, just do their job as best as they can, each lib has its own job...

Issue #72, opened 2016-03-22
ConnectionUpgrade and firefox.


Eepee1:
Hey,

Fasthttp's ConnectionUpgrade wont detect if http-request from firefox wants to upgrade, because firefox adds "keep-alive" to the Connection header. So firefox's Connection header looks like "keep-alive, Upgrade", while for others its just "Upgrade".

valyala:
@Eepee1 , thanks for the bug report. It should be fixed now.

Issue #69, opened 2016-03-15
Do we have equivalent to http.Header.Add?


vishr:
I am not able to find a function equivalent to `http.Header#Add(key, val string)`, do we have one?

valyala:
There is [ResponseHeader.Set](https://godoc.org/github.com/valyala/fasthttp#ResponseHeader.Set). But it cannot add multiple headers with the same name.
There are plans for adding `http.Header#Add()` equivalent to fasthttp later.

valyala:
Added [ResponseHeader.Add](https://godoc.org/github.com/valyala/fasthttp#ResponseHeader.Add) and [RequestHeader.Add](https://godoc.org/github.com/valyala/fasthttp#RequestHeader.Add) for setting multiple headers under the same key.

Issue #66, opened 2016-03-11
Handling connections after closing net.Listener


mtojek:
Hi @valyala ,

I am curious if it is possible (or will be in the future) to trace the worker pool in the context of connection pool size. After closing the net.Listener, the HTTP server does not accept any new connections and it allows earlier accepted ones to be finished (to be served until disconnection).
I need to find a way to monitor the connection pool for the case of graceful shutdown, just to be notified when the number of active connections decreases to zero. 

The reference implementation of HTTP server supporting such feature is available at:
https://github.com/tylerb/graceful
(it stores net.Conn in a hash map)

Kind regards,
Marcin

valyala:
Hi Marcin,
Yo can implement custom `net.Listener`, which will wrap accepted connections into a custom `net.Conn` with overridden `Close` method, which decrements the number of open connections. This way you may keep track on the number of open connections through custom listener and wait until this number drops to zero before returning from overridden `net.Listener.Close`. I'd suggest using non-zero timeout when waiting for pending connections, since certain clients may keep connections open for extended periods of time, so the graceful shutdown may never finish.

mtojek:
Thank you @valyala for the proposed solution, but it may be more difficult if I had already used another listener like e.g. https://github.com/kavu/go_reuseport.

valyala:
@mtojek , just wrap any listener into 'graceful shutdown' listener. Something like the following (I didn't test it, but you should get the idea):

```go
type GracefulListener struct {
    // inner listener
    ln net.Listener

    // maximum wait time for graceful shutdown
    maxWaitTime time.Duration

    // this channel is closed during graceful shutdown on zero open connections.
    done chan struct{}

    // the number of open connections
    connsCount uint64

    // becomes non-zero when graceful shutdown starts
    shutdown uint64
}

// NewGracefulListener wraps the given listener into 'graceful shutdown' listener.
func NewGracefulListener(ln net.Listener) net.Listener {
    return &GracefulListener{
        ln: ln,
        done: make(chan struct{}),
    }
}

func (ln *GracefulListener) Accept() (net.Conn, error) {
    c, err := ln.ln.Accept()
    if err != nil {
        return nil, err
    }
    atomic.AddUint64(&ln.connsCount, 1)
    return &gracefulConn{
        Conn: c,
        ln: ln,
    }, nil
}

// Close closes the inner listener and wait until all the pending open connections
// are closed before returning.
func (ln *GracefulListener) Close() error {
[...]

PR #2347, opened 2026-08-06
Add native HTTP/2 support

### LINKED ISSUE #144: HTTP/2 support
Is this planned at all? And if I was interested in implementing it myself, where would I start?

h2zi:
Nginx recently added HTTP/2 support for proxied upstreams, and I wanted to run fasthttp behind it. fasthttp only speaks HTTP/1, so I built native HTTP/2 support.

Close #144.

The `http2` package works directly with `Request`/`Response`/`RequestCtx`, there is no `net/http` conversion layer. Everything is opt-in: servers that don't call `ConfigureServer` keep the exact same HTTP/1 behavior as before.

```go
server := &fasthttp.Server{Handler: handler}
if err := http2.ConfigureServer(server, http2.ServerConfig{}); err != nil {
    log.Fatal(err)
}
log.Fatal(server.ListenAndServeTLS(":8443", "cert.pem", "key.pem"))
```

`ConfigureServer` enables h2 over TLS (ALPN), cleartext prior knowledge, and the h2c Upgrade handshake on the same listener, so both `curl --http2` modes and HTTP/2 reverse proxies work. `ConfigureHostClient` adds the client side, with HTTP/1 as fallback. Design notes are in `http2/DESIGN.md`.

The code was written and cross-audited by three AI models (GPT-5.6, Claude Opus 5, Claude Fable 5) over multiple rounds. Every bug fix has a regression test that was verified to fail without the fix.

## Conformance

- h2spec 2.6.0 strict mode: 147 tests, 146 passed, 1 skipped, 0 failed, on both CI configurations. The skipped case is timing-dependent inside h2spec itself (nginx skips it too); the package has a deterministic test for that behavior instead.
- Interop in both directions against `golang.org/x/net/http2`, plus curl (upgrade, prior knowledge, ALPN) and nghttp.
- nginx 1.31 proxying to fasthttp over HTTP/2 (`proxy_http_version 2`, the motivation for this PR): reads, JSON POSTs and 300KB assets are correct through the proxy, SSE stays unbuffered end to end, and the gRPC interop suite passes through `grpc_pass`.
- gRPC runs on plain fasthttp handlers (`StreamRequestBody` + `SetBodyStreamWriter` + trailers, no new API): the official grpc-go interop client passes 10/10 cases over both cleartext and TLS, `buf curl` (connect-go) passes on both listeners, and a grpc-go load client sustains ~100k unary calls/s with zero failures. `examples/grpcserver` is the runnable recipe.
- Fiber v3 builds against this branch and its whole suite passes with `-race`: 42 packages, no data races. The eight open functional issues on `dgrr/http2`, the HTTP/2 add-on Fiber used before, do not reproduce here ([#59](https://github.com/dgrr/http2/issues/59), [#69](https://github.com/dgrr/http2/issues/69), [#79](https://github.com/dgrr/http2/issues/79), [#72](https://github.com/dgrr/http2/issues/72), [#27](https://github.com/dgrr/http2/issues/27), [#64](https://github.com/dgrr/http2/issues/64), [#53](https://github.com/dgrr/http2/issues/53), [#58](https://github.com/dgrr/http2/issues/58), [#60](https://github.com/dgrr/http2/issues/60), [#66](https://github.com/dgrr/http2/issues/66)).
- An unconfigured HTTP/1 server, client and adaptor behave bit-for-bit like master: a behavior-diff harness (posted in the comments) compares callback sequences, public values, adaptor interface assertions and raw wire bytes across both revisions, and all 45 normalized records are byte-identical.
- A one-hour soak served **1,049,959,276 requests with 0 failures** across a realistic mix, rebuilding every connection each round. Goroutine count stayed at 10 from start to finish and heap in use ended within 0.6MB of where it started.
- 5 fuzz targets, golangci-lint clean, `go test -race -shuffle=on` passes, cross-compiles for linux/amd64, windows/amd64, linux/386, and the suite passes on Linux arm64.

## Performance

AlmaLinux 10.2 (kernel 6.12), AMD EPYC 9K65 (Zen 5), 32 cores, 61 GB RAM, no other load. 13-byte response, cleartext h2, one worker per core, servers alternated within each round, median of 5 rounds, zero failed requests.

| h2load, req/s | fasthttp | nghttpd 1.68 | h2o 2.3-dev | hyper 1.11 | node 22 cluster | nginx 1.26 | net/http |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| `-t 1 -c 1 -m 100 -n 200000` | 320k | 381k | **462k** | 113k | 72k | 74k | 54k |
| `-t 16 -c 64 -m 32 -n 2000000` | 3,279k | **5,245k** | 2,969k | 2,179k | 913k | 542k | 512k |
| `-t 16 -c 200 -m 100 -n 2000000` | 3,225k | **5,554k** | 3,286k | 2,364k | 973k | 1,047k | 534k |

The two C servers are ahead; against what a fasthttp user would actually be replacing this is **6.0-6.4x `net/http` + `x/net/http2`** and 3.4-3.6x node. Mean request time at `-c 200`: nghttpd 2.28ms, h2o 3.29ms, fasthttp 5.34ms, hyper 7.53ms, nginx 9.97ms, node 24.87ms, net/http 35.69ms.

Ranking depends on core count, because this implementation runs several goroutines per connection while the C servers run one event loop per core. Worker count varied on the same machine, `-t 16 -c 64 -m 32`:

| server cores | 1 | 2 | 4 | 8 | 16 | 32 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| fasthttp | 341k | 728k | 890k | 1,508k | 2,506k | 3,291k |
| h2o | 468k | 893k | 1,566k | 2,288k | 2,340k | 3,238k |
| nghttpd | 529k | 983k | 1,843k | 3,252k | 4,973k | 5,304k |

Reproducing these needs enough `-t`: at 32 server cores `-t 8` caps nghttpd at 4,570k while `-t 16` reaches 5,297k, so a low `-t` silently flatters this implementation, which is already server-bound at `-t 8`.

Compared with fasthttp's own HTTP/1 on the same handler (median of 8 runs): a single serialized request is slower over HTTP/2, 24.0µs vs 13.6µs, because of HPACK and framing overhead. From 10 concurrent requests up HTTP/2 is faster: 3.0µs vs 4.6µs at 100 streams, 4.1µs vs 5.9µs for 4KiB POSTs, while using between 1/10th and 1/1000th of the connections.

Both GET and POST run at 0 allocs/op at every concurrency level; the same harness measures x/net/http2 at 59 allocs/op. The two memory defaults (4MiB connection window, 128MB per-connection `RequestCtx` cache via `Server.MaxProtocolRequestCtxCacheBytes`) are justified with measurements in DESIGN.md. The HTTP/1 path is not affected: the hooks cost one nil check per connection and one per request.

### Where the remaining gap is, and an option that closes it

Both this implementation and nghttpd saturate their cores under load (1932% and 1903% of 20 pinned cores), so the gap is CPU cost per request: 176k req/s per core against nghttpd's 328k. A CPU profile puts **24% of that time in Go's scheduler and synchronization** (`selectgo`, `schedule`, `wakep`, `lock2`/`unlock2`, channel send/receive), because every request crosses reader → connection owner → stream worker → owner → writer and each hop is a park/wake. Raising in-flight requests does not help — throughput falls past 2048 in-flight, since more concurrency buys more scheduling rather than more parallelism.

Running handlers on the connection's own goroutine removes one hop. Measured on the same 20 pinned cores:

| | c=1 m=100 | c=64 m=32 | c=200 m=100 | mean latency at c=64 |
| --- | ---: | ---: | ---: | ---: |
| as submitted | 308k | 3,287k | 2,750k | 527µs |
| handlers inline | **456k** | **5,238k** | **5,845k** | **297µs** |
| | +48% | +59% | +113% | −44% |

That reaches nghttpd's level, but it is a real trade-off, not a free win, so it is not what this PR does:

- One slow handler stalls every other stream on the connection, which is the property the current design exists to avoid.
- It deadlocks any handler that waits on the connection owner. `StreamRequestBody` hangs immediately, verified: the handler blocks for DATA frames that only the owner can deliver, and the handler *is* the owner. Bidirectional streams via `RequestCtx.AcceptStream` have the same shape.

Making it safe needs the handler to hand ownership back at every blocking point, plus a watchdog for handlers that block in user code. That is a concurrency-core change with its own failure modes, so it belongs in a separate PR rather than this one. It could also ship as an opt-in `ServerConfig` flag documented as incompatible with streamed request bodies, if that is preferable.

## Pre-existing bugs found during testing

The test suites for this branch surfaced pre-existing bugs in master. Per review, this PR no longer changes observable HTTP/1 behavior, so those fixes were moved out and are now #2352, each with a regression test: the `RequestCtx.ID()` field overflow, `TimeoutHandler` rejecting every request under `ServeConn`, `IsTLS()`/`TLSConnectionState()` disagreeing when `MaxConnsPerIP` is set, `ConnState` marking a connection active before it sends anything (which also keeps `Shutdown` from closing it), `Add("Trailer", ...)` replacing the announced trailer set instead of extending it, `SetMaxConns` growth never serving queued waiters, and adaptor conformance gaps against `net/http`. Two fixes with no observable HTTP/1 effect (the pooled-timer panic below and a test data race) are #2351.

Two changes stay on this branch because HTTP/2 needs them; both were chosen for having no observable HTTP/1 effect:

- `AcquireTimer` no longer panics on a pooled reset that races an in-flight expiry. The race is a Go runtime bug, reported as golang/go#80760 with a fix under review (golang/go#80762); HTTP/2's timer reuse makes it reachable in normal operation. The fix is submitted separately in #2351 and the copy here goes away once that lands.
- `perIPConn` no longer returns itself to a `sync.Pool` on `Close`. HTTP/2 connection lifetimes routinely close the same connection from two owners, and the second `Close` could hand an unrelated client's live connection back to the pool.

## Notes

Response scheduling follows RFC 9218 (urgency levels, with `incremental` deciding whether same-urgency responses interleave). The protocol hooks expose no HTTP/2 types, so an HTTP/3 package could implement them with QUIC streams and QPACK directly.

One behavior change to be aware of: `Concurrency` and `MaxConnsPerIP` limit connections, which under HTTP/2 no longer indirectly limits concurrent handlers. This is documented rather than changed.

On size: the implementation is 7,885 lines, of which 5,502 are code once blank lines, comments and brace-only lines are set aside. Counted the same way, `golang.org/x/net/http2` is 8,690 including its HPACK or 7,540 without it — this PR reuses `x/net/http2/hpack` rather than reimplementing it. The rest of the diff is 7,295 lines of tests and 549 of docs and examples, and just nine existing files are touched, five of them source files.


http2/config.go
```diff
@@ -0,0 +1,485 @@
...
+	if value == nil {
+		return false
+	}
+	reflected := reflect.ValueOf(value)
+	switch reflected.Kind() {
+	case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Pointer, reflect.Slice:
+		return reflected.IsNil()
+	default:
+		return false
+	}
```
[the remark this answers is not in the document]
h2zi: `reflect.Pointer` is the Go 1.18+ name for `reflect.Ptr` (they are aliases), and the package builds on Go 1.24-1.26. No change needed.

http2/client.go
```diff
@@ -0,0 +1,658 @@
...
+	resp *fasthttp.Response,
+	openStream bool,
+	deadline time.Time,
+	now time.Time,
+) *clientStream {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.closed || c.goAway || c.nextStreamID > math.MaxInt32 || c.activeStreams >= c.peerMaxConcurrentStreams {
+		return nil
+	}
```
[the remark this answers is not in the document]
h2zi: Confirmed, fixed in 5a7be57: `reserveStream` now caps at min(configured, advertised), with a regression test. The same commit also stops counting server pushes against this limit on either end, which was a related bug this comment led to.

http2/server.go
```diff
@@ -0,0 +1,2183 @@
...
+	}
+	if event.truncated {
+		return c.rejectNewStream(event.streamID, xhttp2.ErrCodeEnhanceYourCalm)
+	}
+	if c.isGoingAway {
+		return c.rejectNewStream(event.streamID, xhttp2.ErrCodeRefusedStream)
+	}
+	if uint32(len(c.streams)) >= c.config.maxConcurrentStreams {
+		return c.rejectNewStream(event.streamID, xhttp2.ErrCodeRefusedStream)
+	}
```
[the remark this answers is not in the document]
h2zi: Confirmed, fixed in 5a7be57: admission now subtracts active pushes, so SETTINGS_MAX_CONCURRENT_STREAMS only limits peer-initiated streams. The client had the mirror-image bug (pushes counted against its own request allowance and against the push-admission limit), fixed in the same commit with an end-to-end test.

h2zi:
The four suppressed review findings were also checked. All four are real and fixed in 5a7be57, each with a regression test:

- `clientPool.remove` now also wakes waiters blocked on `MaxConns`, not just stream waiters.
- `releaseCachedRequestCtxs` releases body buffers before returning contexts to the server-wide pool, so closing a connection frees the cached memory as documented.
- Repeated trailer fields are preserved on both the server and client decode paths (`Add` instead of `Set`, with trailer-name deduplication since `AddTrailer` appends blindly).

gaby:
@h2zi You should run `/simplify` within claude after setting `/effort` to Max.

This PR is too big and probably has a lot of over-engineering.

h2zi:
@gaby Fair point. I ran `/simplify` with Claude Fable 5 — 242 lines gone from the non-test code.

On the size: 7,556 of the 17,639 lines are tests, another 549 are docs and examples. Only six existing files get touched at all (`server.go`, `client.go`, `header.go`, `bytesconv.go`, `peripconn.go`, `timer.go`), and they add 472 lines between them. The rest is all new files.

For a sense of scale, the implementation is 7,885 lines, or 5,502 once blank lines, comments and lines holding only a brace are set aside. Counted the same way `golang.org/x/net/http2` is 8,690 including its HPACK, or 7,540 without — this PR reuses `x/net/http2/hpack` rather than reimplementing it. So the implementation is smaller than the one it is measured against, and the bulk of the diff is the tests.

The cleanup didn't cost any speed — it picked some up. Same machine, same handler, median of 3:

| h2load | before | after | |
| --- | ---: | ---: | ---: |
| `-t 1 -c 1 -m 100` | 695k | **748k** | +7.7% |
| `-t 4 -c 32 -m 32` | 1,995k | **1,999k** | ~ |
| `-t 4 -c 200 -m 100` | 1,711k | **1,877k** | +9.7% |

All tests still pass.

server.go
```diff
@@ -2472,6 +2577,14 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 			}
 
 			if err == nil {
+				// The connection counts as active once a request has actually
+				// arrived, which is also what StateActive means. Marking it any
+				// earlier cleared the grace stamp taken when the connection was
+				// accepted, putting a peer that connects and then says nothing
+				// beyond the reach of Shutdown.
+				idleConnTime.Store(0)
+				s.setState(c, StateActive)
```
erikdubbelboer: Please keep the existing `ConnState` contract here. Before this PR, `StateActive` was emitted before the header read, so silent, EOF, and malformed connections reported `New → Active → Closed`; moving it behind a successful parse changes callback ordering and timing for every HTTP/1 server. The shutdown fix can retain the old `setState` location while delaying only `idleConnTime.Store(0)` (or tracking pending first requests separately).

header.go
```diff
@@ -1575,6 +1589,11 @@ func (h *ResponseHeader) AddBytesV(key string, value []byte) {
 // it will be sent after the chunked response body.
 func (h *ResponseHeader) AddBytesKV(key, value []byte) {
 	h.bufK, h.bufV = initHeaderKV(h.bufK, h.bufV, b2s(key), b2s(value), h.disableNormalizing)
+	// Adding to Trailer extends the announced set; only Set replaces it.
+	if caseInsensitiveCompare(strTrailer, h.bufK) {
+		_ = h.AddTrailerBytes(h.bufV)
```
erikdubbelboer: This changes the established behavior of `Add("Trailer", ...)` from replacement to accumulation for all HTTP/1 request and response headers. Please keep generic `Add` behavior unchanged and call `AddTrailer` explicitly where accumulation is required by the new implementation. Any global API semantics change needs a separate compatibility review.

client.go
```diff
@@ -1694,8 +1733,23 @@ var ErrTimeout = &timeoutError{}
 // SetMaxConns sets up the maximum number of connections which may be established to all hosts listed in Addr.
 func (c *HostClient) SetMaxConns(newMaxConns int) {
 	c.connsLock.Lock()
+	defer c.connsLock.Unlock()
 	c.MaxConns = newMaxConns
-	c.connsLock.Unlock()
+	// Grown capacity must reach requests already parked in the wait queue.
+	for c.connsCount < c.maxConnsLocked() {
```
erikdubbelboer: `SetMaxConns` used to update the limit only. It now immediately wakes queued requests and starts dials and, together with the `decConnsCount`/`ReleaseConn` changes, retires connections when the limit shrinks. That changes existing HTTP/1 pool timing, ordering, and connection counts. Please preserve the legacy HTTP/1 path and add an HTTP/2-specific capacity notification, or split live resizing into its own PR.

server.go
```diff
@@ -978,8 +1057,11 @@ func (ctx *RequestCtx) String() string {
 }
 
 // ID returns unique ID of the request.
+//
+// ConnID occupies the high 32 bits and ConnRequestNum the low 32, so the ID
+// repeats once either passes 2^32. Use those accessors to avoid the wrap.
 func (ctx *RequestCtx) ID() uint64 {
-	return (ctx.connID << 32) | ctx.connRequestNum
+	return (ctx.connID << 32) | (ctx.connRequestNum & 0xffffffff)
```
erikdubbelboer: This is an unrelated observable API change: after 2^32 requests on one connection, `ID()` now wraps the low bits rather than retaining the previous arithmetic result. Even if the old result violates the intended field layout, please split this fix so HTTP/2 support does not silently alter existing `RequestCtx.ID()` behavior.

fasthttpadaptor/request.go
```diff
@@ -16,7 +16,12 @@ import (
 // The http.Request must not be used after the fasthttp handler has returned!
 // Memory in use by the http.Request will be reused after your handler has returned!
 func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) error {
-	body := ctx.PostBody()
+	// Reading a streamed body here would defeat Server.StreamRequestBody.
```
erikdubbelboer: `ConvertRequest` is shared by existing HTTP/1 adaptor users. This rewrite changes when streamed handlers start, body materialization and error timing, HTTP/1.0 `ProtoMinor`, `ContentLength`, presence of `Host` in `Header`, and request trailers. Please keep the legacy HTTP/1 conversion behavior and add an HTTP/2-specific conversion path, or move these `net/http` conformance changes to a separate PR.

fasthttpadaptor/adaptor.go
```diff
@@ -254,8 +279,16 @@ func (w *writer) WriteHeader(code int) {
 }
 
 func (w *writer) Write(p []byte) (int, error) {
+	if w.hijacked.Load() {
+		return 0, http.ErrHijacked
+	}
+	// net/http fixes the status once the body starts. Taking whatever is in
+	// effect keeps a code preset on the context before the handler ran.
+	w.statusCode.CompareAndSwap(0, int64(w.status()))
```
erikdubbelboer: Calling `Write` now commits status 200, so `Write(...); WriteHeader(500)` changes from 500 to 200 for current adaptor users. The same shared writer also changes bodiless responses, trailers, post-hijack writes, `ErrAbortHandler`, and `CloseNotifier`. Please keep existing HTTP/1 writer semantics; HTTP/2 can use a separate concrete writer. A single runtime-gated type cannot preserve interface assertions such as `http.CloseNotifier`.

erikdubbelboer (changes requested):
Thanks for working on this. It's a super big PR so I'm not sure what to do with it yet. It's good that most of the code is separate. On my initial review I'll focus on backwards compatibility:

HTTP/2 being opt-in is the right boundary, but this PR currently changes observable behavior for users who never call `ConfigureServer` or `ConfigureHostClient`. That contradicts the PR promise of the exact same HTTP/1 behavior and is not acceptable for this merge.

Please preserve existing HTTP/1 and public API semantics, with HTTP/2-specific behavior kept behind protocol branches or separate concrete types. In particular, restore the existing `ConnState` timing, retain the current generic `Trailer` behavior, avoid live-resizing the HTTP/1 client pool, and preserve the legacy `fasthttpadaptor` request/response behavior.

Please also split the pre-existing bug fixes and `net/http` conformance changes into focused PRs unless they can be implemented without changing legacy behavior. Correctness fixes may be desirable, but they still require their own compatibility decision and should not silently expand this feature PR.

Before merging, please add compatibility tests demonstrating that an unconfigured HTTP/1 server/client produces the same callback sequences, public values, adaptor interface assertions, and wire responses as the base branch.

h2zi:
Thanks for the review — the boundary you're drawing is the right one, and the latest push restores it. An unconfigured HTTP/1 server or client now behaves exactly as the base branch: `StateActive` fires before the header read again, `Add("Trailer", ...)` replaces the announced set, `SetMaxConns` only updates the limit, `RequestCtx.ID()` keeps its arithmetic, and `fasthttpadaptor` is back to the base conversion and writer behavior.

None of these were intentional scope expansion — they surfaced while running h2spec, the gRPC interop matrix, and hour-long soaks against this branch, and I fixed them in place instead of splitting them out. I'll submit each as its own PR for its own compatibility review.

Two changes remain because HTTP/2 needs them for correctness. The timer.go reset-panic removal has a longer story than the diff suggests: the panic fired under this branch's stress and soak runs, and chasing it led past fasthttp into the runtime itself — a pooled timer can report Stop=true and then Reset=true for the same expiry when its callback is preempted between releasing the timer lock and taking the send lock. That's filed as golang/go#80760 with a fix under review in golang/go#80762. Under Go 1.23+ reset semantics the drain-and-panic pattern here guards a state that is no longer an error, so the change stays correct whether or not the upstream fix lands. The peripconn pool removal is allocation pattern only — a multiplexed connection holds the per-IP wrapper past Close, so pooling would alias it. Happy to handle either differently if you prefer.

server.go
h2zi: Restored: StateActive fires before the read again, with the grace stamp cleared alongside it. The silent-connection Shutdown fix will come back as its own PR.

header.go
h2zi: Restored; nothing internal relied on accumulation. The accumulation change will be its own PR.

client.go
h2zi: Restored: SetMaxConns only updates the limit for HTTP/1. The one remaining branch serves HTTP/2 slot waiters at the head of the queue, unreachable without a registered protocol.

server.go
h2zi: Restored; the mask will be its own PR.

fasthttpadaptor/request.go
h2zi: Restored: fasthttpadaptor is byte-for-byte the base branch again.

fasthttpadaptor/adaptor.go
h2zi: Restored together with request.go — the writer is the base type again, so the interface assertions are unchanged.

h2zi:
Added the requested compatibility tests. Since their point is to compare two revisions, they are a standalone harness rather than tree changes: the same test file runs against the base branch and against this PR, emits a normalized record line for each observable behavior, and the two record sets are diffed.

Coverage, all with an unconfigured `Server`/`HostClient`/adaptor:

- **Callback sequences** — `ConnState` order across keep-alive, `Connection: close`, and a connection that sends nothing (including which states fire before the first byte, and `Shutdown` while such a connection is pending), plus `ErrorHandler` and `ContinueHandler` firing.
- **Public values** — `ID()`/`ConnID()`/`ConnRequestNum()` relations across requests and connections, request getters, addresses, times, and `Add("Trailer", ...)` replacement semantics.
- **Adaptor interface assertions** — `http.Flusher`/`http.Hijacker`/`http.CloseNotifier`/`http.Pusher`/`io.ReaderFrom`/`io.StringWriter` on the `ResponseWriter`, `http.ResponseController` deadline/full-duplex errors, and `ConvertRequest` field mapping, plus adaptor wire bytes for buffered, streamed (Flush), hijacked, and 204 responses.
- **Wire responses** — raw response bytes for GET/HEAD/POST, chunked request and response bodies, response trailers, `Expect: 100-continue` accepted and rejected, Early Hints, bad request line, oversized headers, pipelined keep-alive, and HTTP/1.0; raw client request bytes plus `DoTimeout`, dial-error, and `ErrNoFreeConns` behavior, and `SetMaxConns` growth leaving an already-queued waiter parked.

Result: everything passes on both revisions, and the 45 record lines are **byte-identical** between the base branch and this PR head, stable across three consecutive runs. (`Date` values and ephemeral ports are normalized; chunk framing is decoded before comparison because chunk boundaries follow goroutine scheduling on both revisions.)

To reproduce, drop the two files below into the tree on each revision and run:

```sh
COMPAT_RECORD=/tmp/rec.txt go test -count=1 -p 1 -run 'TestCompatH1' . ./fasthttpadaptor/
```

then diff the record files. Happy to check the harness into the tree as well if you want it to stay.

<details>
<summary><code>compat_h1_test.go</code> (repository root)</summary>

```go
// Compatibility harness: run the same file against the base branch and the PR
// head, then diff the emitted RECORD lines. Every observable behavior of an
// unconfigured HTTP/1 server/client is captured as a normalized record:
// callback sequences, public values, and raw wire bytes.
//
// Usage (from the repository root, on each branch):
//
//	COMPAT_RECORD=/tmp/rec.txt go test -count=1 -run 'TestCompatH1' . ./fasthttpadaptor/
//
// Then diff the two record files.
package fasthttp_test

import (
	"bufio"
	"errors"
	"fmt"
	"io"
	"net"
	"os"
	"regexp"
	"strconv"
	"strings"
	"sync"
	"sync/atomic"
	"testing"
	"time"

	"github.com/valyala/fasthttp"
)

var compatRecordMu sync.Mutex

func record(t *testing.T, format string, args ...any) {
	t.Helper()
	line := t.Name() + " | " + fmt.Sprintf(format, args...)
	t.Logf("RECORD %s", line)
	path := os.Getenv("COMPAT_RECORD")
	if path == "" {
		return
	}
	compatRecordMu.Lock()
	defer compatRecordMu.Unlock()
	f, err := os.OpenFile(path, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
	if err != nil {
		t.Fatalf("opening record file: %v", err)
	}
	defer f.Close()
	if _, err := f.WriteString(line + "\n"); err != nil {
		t.Fatalf("writing record file: %v", err)
	}
}

var (
	compatDateRe = regexp.MustCompile(`Date: [A-Z][a-z]{2}, \d{2} [A-Z][a-z]{2} \d{4} \d{2}:\d{2}:\d{2} GMT`)
	compatPortRe = regexp.MustCompile(`127\.0\.0\.1:\d+`)
)

func normalizeWire(b []byte) string {
	s := compatDateRe.ReplaceAllString(string(b), "Date: <DATE>")
	return decodeChunked(compatPortRe.ReplaceAllString(s, "127.0.0.1:<PORT>"))
}

// decodeChunked collapses chunk framing: chunk boundaries depend on goroutine
// scheduling, not on any API contract, so records compare decoded bodies.
func decodeChunked(wire string) string {
	head, body, ok := strings.Cut(wire, "\r\n\r\n")
	if !ok || !strings.Contains(head, "Transfer-Encoding: chunked") {
		return wire
	}
	var decoded strings.Builder
	rest := body
	for {
		sizeLine, after, ok := strings.Cut(rest, "\r\n")
		if !ok {
			return wire
		}
		n, err := strconv.ParseInt(strings.TrimSpace(sizeLine), 16, 64)
		if err != nil {
			return wire
		}
		if n == 0 {
			return head + "\r\n\r\n<chunked>" + decoded.String() + "<trailers>" + after
		}
		if int64(len(after)) < n+2 {
			return wire
		}
		decoded.WriteString(after[:n])
		rest = after[n+2:]
	}
}

func startServer(t *testing.T, s *fasthttp.Server) (addr string, stop func()) {
	t.Helper()
	ln, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatalf("listening: %v", err)
	}
	done := make(chan struct{})
	go func() {
		defer close(done)
		_ = s.Serve(ln)
	}()
	return ln.Addr().String(), func() {
		_ = ln.Close()
		select {
		case <-done:
		case <-time.After(2 * time.Second):
		}
	}
}

// rawExchange writes raw request bytes and reads until EOF or deadline.
func rawExchange(t *testing.T, s *fasthttp.Server, raw string) string {
	t.Helper()
	addr, stop := startServer(t, s)
	defer stop()
	conn, err := net.Dial("tcp", addr)
	if err != nil {
		t.Fatalf("dialing: %v", err)
	}
	defer conn.Close()
	if _, err := conn.Write([]byte(raw)); err != nil {
		t.Fatalf("writing request: %v", err)
	}
	_ = conn.SetReadDeadline(time.Now().Add(1500 * time.Millisecond))
	b, _ := io.ReadAll(conn)
	return normalizeWire(b)
}

// readOneResponse reads a single fixed-length HTTP response off br.
func readOneResponse(t *testing.T, br *bufio.Reader) string {
	t.Helper()
	var sb strings.Builder
	contentLength := 0
	for {
		line, err := br.ReadString('\n')
		if err != nil {
			t.Fatalf("reading response line: %v", err)
		}
		sb.WriteString(line)
		if v, ok := strings.CutPrefix(strings.ToLower(line), "content-length:"); ok {
			n, err := strconv.Atoi(strings.TrimSpace(v))
			if err != nil {
				t.Fatalf("parsing Content-Length %q: %v", line, err)
			}
			contentLength = n
		}
		if line == "\r\n" {
			break
		}
	}
	if contentLength > 0 {
		body := make([]byte, contentLength)
		if _, err := io.ReadFull(br, body); err != nil {
			t.Fatalf("reading response body: %v", err)
		}
		sb.Write(body)
	}
	return sb.String()
}

func waitFor(t *testing.T, timeout time.Duration, cond func() bool, msg string) {
	t.Helper()
	deadline := time.Now().Add(timeout)
	for !cond() {
		if time.Now().After(deadline) {
			t.Fatal(msg)
		}
		time.Sleep(5 * time.Millisecond)
	}
}

type stateRecorder struct {
	mu  sync.Mutex
	seq []string
}

func (r *stateRecorder) hook(_ net.Conn, st fasthttp.ConnState) {
	r.mu.Lock()
	r.seq = append(r.seq, st.String())
	r.mu.Unlock()
}

func (r *stateRecorder) snapshot() string {
	r.mu.Lock()
	defer r.mu.Unlock()
	return strings.Join(r.seq, ",")
}

func (r *stateRecorder) sawClosed() bool {
	r.mu.Lock()
	defer r.mu.Unlock()
	for _, s := range r.seq {
		if s == "closed" {
			return true
		}
	}
	return false
}

func TestCompatH1ServerConnStateSequence(t *testing.T) {
	newServer := func(rec *stateRecorder) *fasthttp.Server {
		return &fasthttp.Server{
			Handler:   func(ctx *fasthttp.RequestCtx) { ctx.SetBodyString("ok") },
			ConnState: rec.hook,
		}
	}

	t.Run("KeepAliveTwoRequests", func(t *testing.T) {
		rec := &stateRecorder{}
		addr, stop := startServer(t, newServer(rec))
		defer stop()
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			t.Fatalf("dialing: %v", err)
		}
		br := bufio.NewReader(conn)
		for i := 0; i < 2; i++ {
			if _, err := conn.Write([]byte("GET / HTTP/1.1\r\nHost: a\r\n\r\n")); err != nil {
				t.Fatalf("writing request %d: %v", i, err)
			}
			readOneResponse(t, br)
		}
		_ = conn.Close()
		waitFor(t, 2*time.Second, rec.sawClosed, "connection never reached closed state")
		record(t, "states=%s", rec.snapshot())
	})

	t.Run("SilentThenRequest", func(t *testing.T) {
		rec := &stateRecorder{}
		addr, stop := startServer(t, newServer(rec))
		defer stop()
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			t.Fatalf("dialing: %v", err)
		}
		time.Sleep(300 * time.Millisecond)
		record(t, "states-before-first-byte=%s", rec.snapshot())
		br := bufio.NewReader(conn)
		if _, err := conn.Write([]byte("GET / HTTP/1.1\r\nHost: a\r\n\r\n")); err != nil {
			t.Fatalf("writing request: %v", err)
		}
		readOneResponse(t, br)
		_ = conn.Close()
		waitFor(t, 2*time.Second, rec.sawClosed, "connection never reached closed state")
		record(t, "states-final=%s", rec.snapshot())
	})

	t.Run("ConnectionClose", func(t *testing.T) {
		rec := &stateRecorder{}
		addr, stop := startServer(t, newServer(rec))
		defer stop()
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			t.Fatalf("dialing: %v", err)
		}
		if _, err := conn.Write([]byte("GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n")); err != nil {
			t.Fatalf("writing request: %v", err)
		}
		_ = conn.SetReadDeadline(time.Now().Add(time.Second))
		_, _ = io.ReadAll(conn)
		_ = conn.Close()
		waitFor(t, 2*time.Second, rec.sawClosed, "connection never reached closed state")
		record(t, "states=%s", rec.snapshot())
	})

	t.Run("ShutdownWithSilentConn", func(t *testing.T) {
		rec := &stateRecorder{}
		s := newServer(rec)
		addr, stop := startServer(t, s)
		defer stop()
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			t.Fatalf("dialing: %v", err)
		}
		time.Sleep(200 * time.Millisecond)
		shutdownDone := make(chan struct{})
		go func() {
			defer close(shutdownDone)
			_ = s.Shutdown()
		}()
		blocked := false
		select {
		case <-shutdownDone:
		case <-time.After(600 * time.Millisecond):
			blocked = true
		}
		record(t, "shutdown-blocked-by-silent-conn=%v states-at-check=%s", blocked, rec.snapshot())
		_ = conn.Close()
		select {
		case <-shutdownDone:
		case <-time.After(2 * time.Second):
			t.Fatal("Shutdown never returned after the silent conn closed")
		}
	})
}

func TestCompatH1ServerPublicValues(t *testing.T) {
	type reqValues struct {
		id, connID, reqNum     uint64
		host, path, ua         string
		remoteAddr, localAddr  string
		isTLS, timeZero        bool
		connTimeZero           bool
	}
	ch := make(chan reqValues, 4)
	s := &fasthttp.Server{
			ch <- reqValues{
				id:           ctx.ID(),
				connID:       ctx.ConnID(),
				reqNum:       ctx.ConnRequestNum(),
				host:         string(ctx.Host()),
				path:         string(ctx.Path()),
				ua:           string(ctx.UserAgent()),
				remoteAddr:   ctx.RemoteAddr().String(),
				localAddr:    ctx.LocalAddr().String(),
				isTLS:        ctx.IsTLS(),
				timeZero:     ctx.Time().IsZero(),
				connTimeZero: ctx.ConnTime().IsZero(),
			}
			ctx.SetBodyString("v")
		},
	}
	addr, stop := startServer(t, s)
	defer stop()

	doRequests := func(n int) {
		conn, err := net.Dial("tcp", addr)
		if err != nil {
			t.Fatalf("dialing: %v", err)
		}
		defer conn.Close()
		br := bufio.NewReader(conn)
		for i := 0; i < n; i++ {
			if _, err := conn.Write([]byte("GET /p?q=1 HTTP/1.1\r\nHost: values.test\r\nUser-Agent: compat-probe\r\n\r\n")); err != nil {
				t.Fatalf("writing request: %v", err)
			}
			readOneResponse(t, br)
		}
	}
	doRequests(2)
	doRequests(1)

	r1a, r1b, r2 := <-ch, <-ch, <-ch
	record(t, "conn1-req1: reqNum=%d idLow=%d idHighIsConnID=%v", r1a.reqNum, r1a.id&0xffffffff, r1a.id>>32 == r1a.connID)
	record(t, "conn1-req2: reqNum=%d idLow=%d idDelta=%d sameConnID=%v", r1b.reqNum, r1b.id&0xffffffff, r1b.id-r1a.id, r1b.connID == r1a.connID)
	record(t, "conn2-req1: reqNum=%d idLow=%d connIDDelta=%d", r2.reqNum, r2.id&0xffffffff, r2.connID-r1a.connID)
	record(t, "req: host=%q path=%q ua=%q isTLS=%v", r1a.host, r1a.path, r1a.ua, r1a.isTLS)
	record(t, "addrs: remote=%q local=%q", compatPortRe.ReplaceAllString(r1a.remoteAddr, "127.0.0.1:<PORT>"), compatPortRe.ReplaceAllString(r1a.localAddr, "127.0.0.1:<PORT>"))
	record(t, "time: timeZero=%v connTimeZero=%v", r1a.timeZero, r1a.connTimeZero)
}

func TestCompatH1TrailerHeaderSemantics(t *testing.T) {
	var resp fasthttp.ResponseHeader
	_ = resp.AddTrailer("Foo")
	_ = resp.AddTrailer("Bar")
	record(t, "resp AddTrailer x2 -> header=%q", normalizeWire([]byte(resp.String())))

	var respAdd fasthttp.ResponseHeader
	respAdd.Add("Trailer", "Foo")
	respAdd.Add("Trailer", "Bar")
	record(t, "resp Add(Trailer) x2 -> header=%q", normalizeWire([]byte(respAdd.String())))

	var respMix fasthttp.ResponseHeader
	_ = respMix.SetTrailer("Alpha, Beta")
	respMix.Add("Trailer", "Gamma")
	record(t, "resp SetTrailer then Add(Trailer) -> header=%q", normalizeWire([]byte(respMix.String())))

	var req fasthttp.RequestHeader
	req.Add("Trailer", "Foo")
	req.Add("Trailer", "Bar")
	record(t, "req Add(Trailer) x2 -> header=%q", req.String())
}

func TestCompatH1ServerWire(t *testing.T) {
	t.Run("PlainGet", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { ctx.SetBodyString("hello, world!") }}
		record(t, "%q", rawExchange(t, s, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("Head", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { ctx.SetBodyString("hello, world!") }}
		record(t, "%q", rawExchange(t, s, "HEAD / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("PostEcho", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { _, _ = ctx.Write(ctx.PostBody()) }}
		record(t, "%q", rawExchange(t, s, "POST /e HTTP/1.1\r\nHost: a\r\nContent-Length: 5\r\nConnection: close\r\n\r\nabcde"))
	})

	t.Run("ChunkedRequestEcho", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { _, _ = ctx.Write(ctx.PostBody()) }}
		raw := "POST /e HTTP/1.1\r\nHost: a\r\nTransfer-Encoding: chunked\r\nConnection: close\r\n\r\n3\r\nabc\r\n2\r\nde\r\n0\r\n\r\n"
		record(t, "%q", rawExchange(t, s, raw))
	})

	t.Run("StreamedChunkedResponse", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) {
			ctx.SetBodyStreamWriter(func(w *bufio.Writer) {
				_, _ = w.WriteString("part1,")
				_ = w.Flush()
				_, _ = w.WriteString("part2")
			})
		}}
		record(t, "%q", rawExchange(t, s, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("ResponseTrailer", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) {
			if err := ctx.Response.Header.SetTrailer("X-Checksum"); err != nil {
				record(t, "SetTrailer err=%v", err)
				return
			}
			ctx.Response.Header.Set("X-Checksum", "abc123")
			ctx.SetBodyStreamWriter(func(w *bufio.Writer) {
				_, _ = w.WriteString("payload")
			})
		}}
		record(t, "%q", rawExchange(t, s, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("BadRequestLine", func(t *testing.T) {
		var handlerErr atomic.Value
		s := &fasthttp.Server{
			Handler:      func(ctx *fasthttp.RequestCtx) {},
			ErrorHandler: func(_ *fasthttp.RequestCtx, err error) { handlerErr.Store(err.Error()) },
		}
		record(t, "%q", rawExchange(t, s, "BOGUS\r\n\r\n"))
		record(t, "errorHandler=%q", handlerErr.Load())
	})

	t.Run("OversizedHeader", func(t *testing.T) {
		var handlerErr atomic.Value
		s := &fasthttp.Server{
			Handler:      func(ctx *fasthttp.RequestCtx) {},
			ErrorHandler: func(_ *fasthttp.RequestCtx, err error) { handlerErr.Store(err.Error()) },
		}
		raw := "GET / HTTP/1.1\r\nHost: a\r\nX-Big: " + strings.Repeat("x", 5000) + "\r\nConnection: close\r\n\r\n"
		record(t, "%q", rawExchange(t, s, raw))
		errStr, _ := handlerErr.Load().(string)
		if len(errStr) > 100 {
			errStr = errStr[:100]
		}
		record(t, "errorHandler-prefix=%q", errStr)
	})

	t.Run("Expect100Default", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { _, _ = ctx.Write(ctx.PostBody()) }}
		raw := "POST /e HTTP/1.1\r\nHost: a\r\nExpect: 100-continue\r\nContent-Length: 5\r\nConnection: close\r\n\r\nabcde"
		record(t, "%q", rawExchange(t, s, raw))
	})

	t.Run("Expect100Rejected", func(t *testing.T) {
		s := &fasthttp.Server{
			Handler:         func(ctx *fasthttp.RequestCtx) { _, _ = ctx.Write(ctx.PostBody()) },
			ContinueHandler: func(_ *fasthttp.RequestHeader) bool { return false },
		}
		raw := "POST /e HTTP/1.1\r\nHost: a\r\nExpect: 100-continue\r\nContent-Length: 5\r\nConnection: close\r\n\r\nabcde"
		record(t, "%q", rawExchange(t, s, raw))
	})

	t.Run("CtxError", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { ctx.Error("not found", fasthttp.StatusNotFound) }}
		record(t, "%q", rawExchange(t, s, "GET /missing HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("EarlyHints", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) {
			ctx.Response.Header.Add("Link", "</style.css>; rel=preload; as=style")
			ctx.Response.Header.Add("Link", "</script.js>; rel=preload; as=script")
			if err := ctx.EarlyHints(); err != nil {
				record(t, "EarlyHints err=%v", err)
				return
			}
			ctx.SetBodyString("<html>done</html>")
		}}
		record(t, "%q", rawExchange(t, s, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("KeepAlivePipelined", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { ctx.SetBodyString("r") }}
		raw := "GET /1 HTTP/1.1\r\nHost: a\r\n\r\nGET /2 HTTP/1.1\r\nHost: a\r\n\r\n"
		record(t, "%q", rawExchange(t, s, raw))
	})

	t.Run("HTTP10", func(t *testing.T) {
		s := &fasthttp.Server{Handler: func(ctx *fasthttp.RequestCtx) { ctx.SetBodyString("ten") }}
		record(t, "%q", rawExchange(t, s, "GET / HTTP/1.0\r\nHost: a\r\n\r\n"))
	})
}

// pipeDial returns a DialFunc handing each dial's server end to script.
func pipeDial(script func(net.Conn)) fasthttp.DialFunc {
	return func(string) (net.Conn, error) {
		serverConn, clientConn := net.Pipe()
		go script(serverConn)
		return clientConn, nil
	}
}

// readRawRequest reads request bytes until the header terminator, then the
// declared Content-Length worth of body.
func readRawRequest(conn net.Conn) ([]byte, error) {
	br := bufio.NewReader(conn)
	var raw []byte
	contentLength := 0
	for {
		line, err := br.ReadString('\n')
		if err != nil {
			return raw, err
		}
		raw = append(raw, line...)
		if v, ok := strings.CutPrefix(strings.ToLower(line), "content-length:"); ok {
			if n, err := strconv.Atoi(strings.TrimSpace(v)); err == nil {
				contentLength = n
			}
		}
		if line == "\r\n" {
			break
		}
	}
	if contentLength > 0 {
		body := make([]byte, contentLength)
		if _, err := io.ReadFull(br, body); err != nil {
			return raw, err
		}
		raw = append(raw, body...)
	}
	return raw, nil
}

func TestCompatH1ClientWireAndValues(t *testing.T) {
	t.Run("Get", func(t *testing.T) {
		captured := make(chan []byte, 1)
		hc := &fasthttp.HostClient{
			Addr: "example.com:80",
			Dial: pipeDial(func(conn net.Conn) {
				raw, err := readRawRequest(conn)
				if err != nil {
					captured <- []byte(fmt.Sprintf("read error: %v", err))
					return
				}
				captured <- raw
				_, _ = io.WriteString(conn, "HTTP/1.1 200 OK\r\nContent-Length: 2\r\nX-Resp: yes\r\nConnection: close\r\n\r\nhi")
				_ = conn.Close()
			}),
		}
		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/path?x=1")
		if err := hc.Do(req, resp); err != nil {
			t.Fatalf("Do() error: %v", err)
		}
		record(t, "request=%q", <-captured)
		record(t, "status=%d body=%q respHeader=%q", resp.StatusCode(), resp.Body(), normalizeWire([]byte(resp.Header.String())))
	})

	t.Run("PostBody", func(t *testing.T) {
		captured := make(chan []byte, 1)
		hc := &fasthttp.HostClient{
			Addr: "example.com:80",
			Dial: pipeDial(func(conn net.Conn) {
				raw, _ := readRawRequest(conn)
				captured <- raw
				_, _ = io.WriteString(conn, "HTTP/1.1 200 OK\r\nContent-Length: 0\r\nConnection: close\r\n\r\n")
				_ = conn.Close()
			}),
		}
		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/submit")
		req.Header.SetMethod(fasthttp.MethodPost)
		req.Header.Set("X-Custom", "cv")
		req.SetBodyString("k=v&a=b")
		if err := hc.Do(req, resp); err != nil {
			t.Fatalf("Do() error: %v", err)
		}
		record(t, "request=%q", <-captured)
	})

	t.Run("DoTimeout", func(t *testing.T) {
		hc := &fasthttp.HostClient{
			Addr: "example.com:80",
			Dial: pipeDial(func(conn net.Conn) {
				_, _ = readRawRequest(conn)
				time.Sleep(2 * time.Second)
				_ = conn.Close()
			}),
		}
		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/")
		err := hc.DoTimeout(req, resp, 300*time.Millisecond)
		record(t, "isErrTimeout=%v err=%q", errors.Is(err, fasthttp.ErrTimeout), fmt.Sprint(err))
	})

	t.Run("DialError", func(t *testing.T) {
		hc := &fasthttp.HostClient{
			Addr: "example.com:80",
			Dial: func(string) (net.Conn, error) { return nil, errors.New("boom") },
		}
		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/")
		err := hc.Do(req, resp)
		record(t, "err=%q", fmt.Sprint(err))
	})

	t.Run("NoFreeConnsImmediate", func(t *testing.T) {
		release := make(chan struct{})
		hc := &fasthttp.HostClient{
			Addr:     "example.com:80",
			MaxConns: 1,
			Dial: pipeDial(func(conn net.Conn) {
				_, _ = readRawRequest(conn)
				<-release
				_, _ = io.WriteString(conn, "HTTP/1.1 200 OK\r\nContent-Length: 0\r\nConnection: close\r\n\r\n")
				_ = conn.Close()
			}),
		}
		firstDone := make(chan error, 1)
		go func() {
			defer fasthttp.ReleaseRequest(req)
			req.SetRequestURI("http://example.com/")
			firstDone <- hc.Do(req, resp)
		}()
		waitFor(t, 2*time.Second, func() bool { return hc.ConnsCount() == 1 }, "first request never acquired the conn")

		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/")
		err := hc.Do(req, resp)
		record(t, "second-do isErrNoFreeConns=%v err=%q", errors.Is(err, fasthttp.ErrNoFreeConns), fmt.Sprint(err))

		close(release)
		if err := <-firstDone; err != nil {
			t.Fatalf("first request error: %v", err)
		}
	})
}

func TestCompatH1SetMaxConnsKeepsQueuedWaiterParked(t *testing.T) {
	release := make(chan struct{})
	hc := &fasthttp.HostClient{
		Addr:               "example.com:80",
		MaxConns:           1,
		MaxConnWaitTimeout: 500 * time.Millisecond,
		Dial: pipeDial(func(conn net.Conn) {
			_, _ = readRawRequest(conn)
			<-release
			_, _ = io.WriteString(conn, "HTTP/1.1 200 OK\r\nContent-Length: 0\r\nConnection: close\r\n\r\n")
			_ = conn.Close()
		}),
	}
	do := func() error {
		defer fasthttp.ReleaseRequest(req)
		req.SetRequestURI("http://example.com/")
		return hc.Do(req, resp)
	}

	firstDone := make(chan error, 1)
	go func() { firstDone <- do() }()
	waitFor(t, 2*time.Second, func() bool { return hc.ConnsCount() == 1 }, "first request never acquired the conn")

	secondDone := make(chan error, 1)
	secondStart := time.Now()
	go func() { secondDone <- do() }()
	time.Sleep(150 * time.Millisecond)

	// Base behavior: growing MaxConns must not wake an already-queued waiter;
	// it waits for a released conn or times out.
	hc.SetMaxConns(2)
	select {
	case err := <-secondDone:
		waited := time.Since(secondStart)
		record(t, "second-do isErrNoFreeConns=%v waitedFullTimeout=%v", errors.Is(err, fasthttp.ErrNoFreeConns), waited >= 400*time.Millisecond)
	case <-time.After(2 * time.Second):
		t.Fatal("second request never finished")
	}

	close(release)
	if err := <-firstDone; err != nil {
		t.Fatalf("first request error: %v", err)
	}
}
```

</details>

<details>
<summary><code>fasthttpadaptor/compat_h1_test.go</code></summary>

```go
// Compatibility harness for the adaptor: run the same file against the base
// branch and the PR head, then diff the emitted RECORD lines. See the root
// compat_h1_test.go for usage.
package fasthttpadaptor_test

import (
	"fmt"
	"io"
	"net"
	"net/http"
	"os"
	"regexp"
	"sort"
	"strconv"
	"strings"
	"sync"
	"testing"
	"time"

	"github.com/valyala/fasthttp"
	"github.com/valyala/fasthttp/fasthttpadaptor"
)

var compatRecordMu sync.Mutex

func record(t *testing.T, format string, args ...any) {
	t.Helper()
	line := t.Name() + " | " + fmt.Sprintf(format, args...)
	t.Logf("RECORD %s", line)
	path := os.Getenv("COMPAT_RECORD")
	if path == "" {
		return
	}
	compatRecordMu.Lock()
	defer compatRecordMu.Unlock()
	f, err := os.OpenFile(path, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
	if err != nil {
		t.Fatalf("opening record file: %v", err)
	}
	defer f.Close()
	if _, err := f.WriteString(line + "\n"); err != nil {
		t.Fatalf("writing record file: %v", err)
	}
}

var (
	compatDateRe = regexp.MustCompile(`Date: [A-Z][a-z]{2}, \d{2} [A-Z][a-z]{2} \d{4} \d{2}:\d{2}:\d{2} GMT`)
	compatPortRe = regexp.MustCompile(`127\.0\.0\.1:\d+`)
)

func normalizeWire(b []byte) string {
	s := compatDateRe.ReplaceAllString(string(b), "Date: <DATE>")
	return decodeChunked(compatPortRe.ReplaceAllString(s, "127.0.0.1:<PORT>"))
}

// decodeChunked collapses chunk framing: chunk boundaries depend on goroutine
// scheduling, not on any API contract, so records compare decoded bodies.
func decodeChunked(wire string) string {
	head, body, ok := strings.Cut(wire, "\r\n\r\n")
	if !ok || !strings.Contains(head, "Transfer-Encoding: chunked") {
		return wire
	}
	var decoded strings.Builder
	rest := body
	for {
		sizeLine, after, ok := strings.Cut(rest, "\r\n")
		if !ok {
			return wire
		}
		n, err := strconv.ParseInt(strings.TrimSpace(sizeLine), 16, 64)
		if err != nil {
			return wire
		}
		if n == 0 {
			return head + "\r\n\r\n<chunked>" + decoded.String() + "<trailers>" + after
		}
		if int64(len(after)) < n+2 {
			return wire
		}
		decoded.WriteString(after[:n])
		rest = after[n+2:]
	}
}

func startServer(t *testing.T, s *fasthttp.Server) (addr string, stop func()) {
	t.Helper()
	ln, err := net.Listen("tcp", "127.0.0.1:0")
	if err != nil {
		t.Fatalf("listening: %v", err)
	}
	done := make(chan struct{})
	go func() {
		defer close(done)
		_ = s.Serve(ln)
	}()
	return ln.Addr().String(), func() {
		_ = ln.Close()
		select {
		case <-done:
		case <-time.After(2 * time.Second):
		}
	}
}

func rawExchange(t *testing.T, h http.Handler, raw string) string {
	t.Helper()
	s := &fasthttp.Server{Handler: fasthttpadaptor.NewFastHTTPHandler(h)}
	addr, stop := startServer(t, s)
	defer stop()
	conn, err := net.Dial("tcp", addr)
	if err != nil {
		t.Fatalf("dialing: %v", err)
	}
	defer conn.Close()
	if _, err := conn.Write([]byte(raw)); err != nil {
		t.Fatalf("writing request: %v", err)
	}
	_ = conn.SetReadDeadline(time.Now().Add(1500 * time.Millisecond))
	b, _ := io.ReadAll(conn)
	return normalizeWire(b)
}

func TestCompatH1AdaptorInterfaces(t *testing.T) {
	report := make(chan string, 1)
	h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		_, isFlusher := w.(http.Flusher)
		_, isHijacker := w.(http.Hijacker)
		_, isCloseNotifier := w.(http.CloseNotifier) //nolint:staticcheck // asserting the legacy surface
		_, isPusher := w.(http.Pusher)
		_, isReaderFrom := w.(io.ReaderFrom)
		_, isStringWriter := w.(io.StringWriter)
		rc := http.NewResponseController(w)
		errRead := rc.SetReadDeadline(time.Now().Add(time.Second))
		errWrite := rc.SetWriteDeadline(time.Now().Add(time.Second))
		errDuplex := rc.EnableFullDuplex()
		report <- fmt.Sprintf(
			"flusher=%v hijacker=%v closeNotifier=%v pusher=%v readerFrom=%v stringWriter=%v setReadDeadline=%q setWriteDeadline=%q enableFullDuplex=%q",
			isFlusher, isHijacker, isCloseNotifier, isPusher, isReaderFrom, isStringWriter,
			fmt.Sprint(errRead), fmt.Sprint(errWrite), fmt.Sprint(errDuplex))
		w.WriteHeader(http.StatusOK)
	})
	rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n")
	record(t, "%s", <-report)
}

func TestCompatH1AdaptorRequestFields(t *testing.T) {
	report := make(chan string, 1)
	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		body, _ := io.ReadAll(r.Body)
		var headerKeys []string
		for k := range r.Header {
			headerKeys = append(headerKeys, k)
		}
		sort.Strings(headerKeys)
		report <- fmt.Sprintf(
			"method=%s url=%q requestURI=%q proto=%s host=%q contentLength=%d transferEncoding=%v headerKeys=%v xOne=%v remoteAddr=%q tlsNil=%v body=%q",
			r.Method, r.URL.String(), r.RequestURI, r.Proto,
			compatPortRe.ReplaceAllString(r.Host, "127.0.0.1:<PORT>"),
			r.ContentLength, r.TransferEncoding, headerKeys, r.Header["X-One"],
			compatPortRe.ReplaceAllString(r.RemoteAddr, "127.0.0.1:<PORT>"),
			r.TLS == nil, body)
		w.WriteHeader(http.StatusOK)
	})
	raw := "POST /pa?q=2 HTTP/1.1\r\nHost: a\r\nX-One: 1\r\nX-One: 2\r\nContent-Length: 5\r\nConnection: close\r\n\r\nhello"
	rawExchange(t, h, raw)
	record(t, "%s", <-report)
}

func TestCompatH1AdaptorWire(t *testing.T) {
	t.Run("BufferedDetectContentType", func(t *testing.T) {
		h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
			_, _ = w.Write([]byte("<html><body>hi</body></html>"))
		})
		record(t, "%q", rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("ExplicitStatusAndHeader", func(t *testing.T) {
		// A single custom key: with several keys the emission order would
		// follow Go's randomized map iteration, on both branches alike.
		h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
			w.Header().Add("X-Multi", "m1")
			w.Header().Add("X-Multi", "m2")
			w.WriteHeader(http.StatusCreated)
			_, _ = w.Write([]byte("ok"))
		})
		record(t, "%q", rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("FlushStreams", func(t *testing.T) {
		h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
			_, _ = w.Write([]byte("first,"))
			w.(http.Flusher).Flush()
			_, _ = w.Write([]byte("second"))
		})
		record(t, "%q", rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("Hijack", func(t *testing.T) {
		h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
			conn, bufrw, err := w.(http.Hijacker).Hijack()
			if err != nil {
				record(t, "hijack err=%v", err)
				return
			}
			_, _ = bufrw.WriteString("HTTP/1.1 200 OK\r\nContent-Length: 3\r\n\r\nraw")
			_ = bufrw.Flush()
			_ = conn.Close()
		})
		record(t, "%q", rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})

	t.Run("NoContent204", func(t *testing.T) {
		h := http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
			w.WriteHeader(http.StatusNoContent)
		})
		record(t, "%q", rawExchange(t, h, "GET / HTTP/1.1\r\nHost: a\r\nConnection: close\r\n\r\n"))
	})
}
```

</details>

thylong:
Hey @h2zi @erikdubbelboer !
We're very much interested in your work on this topic as lack of native HTTP/2 support is the sole thing that currently prevents us from adopting FastHTTP for our project.

Aside from cheering out loud for your awesome work, let us know if help and active support can be provided, happy to give a hand ! 🙏

PR #2350, opened 2026-08-10
fix(request): warn when Request.Body() is called on streamed requests

### LINKED ISSUE #1765: It is not safe to read all stream body to memery without max size limit

VedantMadane:
Fixes #1765.

Add a warning message via the Server's Logger when Request.Body() is called on a request backed by a stream in a server context. This alerts developers about potential out-of-memory (OOM) vectors and prompts them to use streaming alternatives (Request.BodyStream() or Request.WriteBodyTo()).


erikdubbelboer (commented):
Two remaining issues from the re-review:

server.go
```diff
@@ -2380,6 +2380,7 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 
 	ctx := s.acquireCtx(c)
 	ctx.connTime = connTime
+	ctx.Request.logger = s.logger()
```
erikdubbelboer, line 2383: Please attach the logger for every request after any context reacquisition, rather than only once per connection. Request.Reset() clears req.logger at the end of each keep-alive request, so second and later streamed requests never warn. With ReduceMemoryUsage=true, acquireByteReader releases and reacquires the context before the first request, so even request one has no logger. A server-level regression with two pipelined streamed POSTs produced 1 warning normally and 0 with ReduceMemoryUsage (expected 2 in both cases).

http.go
```diff
@@ -916,6 +918,10 @@ func (req *Request) SwapBody(body []byte) []byte {
 // If the body is backed by a stream, Body reads the entire stream into memory.
 // Use BodyStream to read it incrementally.
 func (req *Request) Body() []byte {
+	if req.bodyStream != nil && req.logger != nil {
+		req.logger.Printf("WARNING: Request.Body() reads the entire stream into memory. " +
+			"Use Request.BodyStream() or Request.WriteBodyTo() instead to avoid out-of-memory errors.")
```
erikdubbelboer, line 923: Request.WriteBodyTo() does not exist; the public method is Request.BodyWriteTo(io.Writer). Please correct the warning and assert the complete suggestion in the test, since the current prefix-only assertion does not catch this.

http_test.go
```diff
@@ -4332,3 +4332,39 @@ func TestWriteBodyChunkedConcreteTypes(t *testing.T) {
 		}
 	}
 }
+
+type mockLogger struct {
```
erikdubbelboer: Please reuse `testLogger` from `server_test.go` in both tests instead of adding `mockLogger`. It strips the first word, but stripping `WARNING:` is fine here. Just adjust the assertions to check the remaining message. After that this can be merged!

erikdubbelboer (changes requested):

_pr_meta.json
```diff
@@ -0,0 +1,8 @@
+{
+  "repo": "valyala/fasthttp",
+  "number": 2350,
+  "base": "master",
+  "head": "fix-stream-body-warning",
+  "fork": "VedantMadane/fasthttp",
+  "title": "fix(request): warn when Request.Body() is called on streamed requests"
+}
```
erikdubbelboer, line 8: Please remove this file from the PR.

erikdubbelboer (changes requested):

server.go
VedantMadane: Logger is attached each serveConn loop iteration after acquireByteReader (and on Init/Init2), so keep-alive and ReduceMemoryUsage reacquisition paths get a fresh logger. Covered by TestServerRequestBodyStreamWarningPipelined for both ReduceMemoryUsage values.

http.go
VedantMadane: Corrected — warning text uses Request.BodyWriteTo(); test asserts the full message.

http_test.go
VedantMadane: Done — both tests use testLogger from server_test.go and assert the message after the first-word strip.

_pr_meta.json
VedantMadane: Removed _pr_meta.json from the PR in 26e6e67.

VedantMadane:
Thanks @erikdubbelboer — addressed the remaining notes:

- Logger attached after every context reacquisition in the serveConn loop (+ Init/Init2); pipelined streamed POST test covers ReduceMemoryUsage true/false
- Warning text uses `Request.BodyWriteTo()` with a full-string assertion
- Tests reuse `testLogger`
- `_pr_meta.json` removed
- Also restored `DisableRedirectPathNormalizing` in `copyToSkipBody` (it had been accidentally overwritten when adding logger copy)

PR #2351, opened 2026-08-11
Fix a pooled timer panic and a test data race

h2zi:
Two fixes split out of #2347 per review. Neither changes observable HTTP/1 behavior; both come with a regression test that fails without the fix.

**`AcquireTimer` can panic with "active timer trapped into initTimer()" for correct callers.** The root cause is a Go runtime race: a pooled timer's expiry can race the owner's `select`, and the same in-flight send gets counted as pending twice — `Stop` voids it and reports true, then `Reset` counts that already-voided send again and reports true as well. Reproduced with a standalone program (no fasthttp involved) and reported as golang/go#80760, with a fix under review (golang/go#80762). Under Go 1.23+ semantics `Reset` on such a timer is safe — the pre-1.23 drain-and-panic pattern here guards a state that is no longer an error — so the panic is removed and the reset handles it. The regression test covers the return-value path that triggered the panic: an active timer makes `Reset` report true deterministically, which is the value the race can also produce for a stopped one.

**`TestRequestCtxInit` and `TestServerLogger` reset `globalConnID` and then assert a fixed connection ID.** A connection left over from an earlier test can call `nextConnID` in between, which both races that write and shifts the ID the context actually receives. Both tests now leave the counter alone and build their expectation from the ID that was generated.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -3413,19 +3413,19 @@ func TestRequestCtxNoHijackNoResponse(t *testing.T) {
 }
 
 func TestRequestCtxInit(t *testing.T) {
-	// This test can't run parallel as it modifies globalConnID.
-
 	var ctx RequestCtx
 	var logger testLogger
-	globalConnID = 0x123456
 	ctx.Init(&ctx.Request, zeroTCPAddr, &logger)
 	ip := ctx.RemoteIP()
 	if !ip.IsUnspecified() {
 		t.Fatalf("unexpected ip for bare RequestCtx: %q. Expected 0.0.0.0", ip)
 	}
+	if requestNum := ctx.ID() & 0xffffffff; requestNum != 0 {
+		t.Fatalf("unexpected request number in ID: %d. Expected 0", requestNum)
+	}
 	ctx.Logger().Printf("foo bar %d", 10)
 
-	expectedLog := "#0012345700000000 - 0.0.0.0:0<->0.0.0.0:0 - GET http:/// - foo bar 10\n"
+	expectedLog := fmt.Sprintf("#%016X - 0.0.0.0:0<->0.0.0.0:0 - GET http:/// - foo bar 10\n", ctx.ID())
 	if logger.out != expectedLog {
 		t.Fatalf("Unexpected log output: %q. Expected %q", logger.out, expectedLog)
 	}
@@ -3939,11 +3939,11 @@ func TestServerEmptyResponse(t *testing.T) {
 }
 
 func TestServerLogger(t *testing.T) {
-	// This test can't run parallel as it modifies globalConnID.
-
 	cl := &testLogger{}
+	var connID uint64
 	s := &Server{
 		Handler: func(ctx *RequestCtx) {
+			connID = ctx.ConnID()
 			logger := ctx.Logger()
 			h := &ctx.Request.Header
 			logger.Printf("begin")
@@ -3966,8 +3966,6 @@ func TestServerLogger(t *testing.T) {
 		},
 	}
 
-	globalConnID = 0
-
 	if err := s.ServeConn(rwx); err != nil {
 		t.Fatalf("Unexpected error from serveConn: %v", err)
 	}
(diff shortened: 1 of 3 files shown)


erikdubbelboer (commented):
I found one functional test-stability issue and one minor accuracy issue in the timer-race explanation.

server_test.go
```diff
@@ -3413,11 +3413,12 @@ func TestRequestCtxNoHijackNoResponse(t *testing.T) {
...
 
 func TestRequestCtxInit(t *testing.T) {
-	// This test can't run parallel as it modifies globalConnID.
+	// This test can't run parallel as it modifies globalConnID. The store is
+	// atomic because a leftover connection may still be calling nextConnID.
 
 	var ctx RequestCtx
 	var logger testLogger
-	globalConnID = 0x123456
+	atomic.StoreUint64(&globalConnID, 0x123456)
```
erikdubbelboer: **[P2] Atomic store leaves the connection-ID assertions flaky**

Making this store atomic removes the detector warning, but it does not stabilize the following ID. As the new comment acknowledges, a leftover goroutine can call `nextConnID` after this `Store` and before `ctx.Init`; the context then receives a different ID and the fixed expected log fails. The same issue applies to the store in `TestServerLogger`.

I reproduced this with a race-free stress probe under `-race`: a concurrent `nextConnID` caused the context to receive `0x123894` instead of `0x123457`. Please avoid resetting the global here—for example, capture the generated connection ID and build the expected output from it, so these assertions are independent of concurrent users of the counter.

timer.go
```diff
@@ -9,10 +9,10 @@ func initTimer(t *time.Timer, timeout time.Duration) *time.Timer {
 	if t == nil {
 		return time.NewTimer(timeout)
 	}
-	if t.Reset(timeout) {
-		// developer sanity-check
-		panic("BUG: active timer trapped into initTimer()")
-	}
+	// Reset may report the timer active: an expiry can race the owner's select
+	// choosing another case, and then ReleaseTimer's drain runs before the
+	// in-flight delivery. Since Go 1.23 Reset discards that delivery.
```
erikdubbelboer: **[P3] Update the timer-race explanation**

The [upstream diagnosis](https://github.com/golang/go/issues/80760) was corrected: the observed sequence is `Stop` returning true, followed by `Reset` incorrectly returning true for the same voided in-flight send—not `Stop` returning false followed by a missed drain. Ignoring `Reset`'s result remains correct because Go 1.23+ guarantees that `Reset` discards prior timer delivery, but this comment and the PR description should reflect the corrected interleaving.

server_test.go
h2zi: Rewritten: both tests leave `globalConnID` alone and build their expectation from the ID that was generated — `TestRequestCtxInit` from `ctx.ID()`, `TestServerLogger` from the `ConnID()` its handler saw. The request-number half is still asserted explicitly, so the ID layout stays covered without depending on the counter.

timer.go
h2zi: Corrected in the comment, the commit message and the PR description: `Stop` voids the in-flight send and reports true, then `Reset` counts that same voided send as still pending and reports true as well. Thanks for catching that the text still carried the original diagnosis.

timer_test.go
```diff
@@ -0,0 +1,20 @@
+package fasthttp
+
+import (
+	"testing"
+	"time"
+)
+
+func TestInitTimerToleratesActiveTimer(t *testing.T) {
```
erikdubbelboer: Could we rename this to `TestInitTimerDoesNotPanicWhenResetReportsActive` and add a comment explaining why the test deliberately starts with an active timer? An active timer makes `Reset` return `true` deterministically, so this covers the return-value path that triggered the panic, but it does not exercise the `Stop`/pool/expiry race itself. For example:

```go
// An active timer makes Reset return true deterministically. The runtime race
// can return the same value for a stopped timer, and initTimer must ignore it.
```

That keeps the regression guard while avoiding implying that active timers are the expected pooled state. The PR description’s “drives the race directly” wording should be adjusted accordingly.

erikdubbelboer (commented):
One small clarification to make the regression test intent precise.

timer_test.go
h2zi: Renamed to `TestInitTimerDoesNotPanicWhenResetReportsActive` with your comment, and reworded the PR description — it now says the test covers the return-value path rather than the race itself.

h2zi: done

erikdubbelboer:
Thanks!

PR #2352, opened 2026-08-11
HTTP/1 behavior fixes split from the HTTP/2 PR

h2zi:
Seven fixes split out of #2347 per review. Each one changes an observable HTTP/1 behavior — that is why they were moved out of the HTTP/2 PR into their own compatibility review — and each comes with a regression test that fails without the fix. They are independent commits; happy to split further into one PR per change if preferred.

Per review, everything trailer-specific was moved out of this PR into a follow-up covering request and response trailer parsing/serialization together (same-name upfront/trailer provenance included), with size/allocation benchmarks.

Current behavior → fixed behavior:

- **`RequestCtx.ID()`** — the request counter is ORed into the connection-ID half of the value, so past 2³² requests on one connection it corrupts the connection ID. Not reachable in practice, but the invariant shouldn't depend on that arithmetic. The counter is now masked to its 32-bit field.
- **`ConnState`** — a connection was marked `StateActive` on accept, before it sent a single byte. It is now marked active on its first byte — before the header completes, so `Shutdown` never cuts a request in flight, and whether or not that request then parses — matching net/http. This also means `Shutdown` can close a connection that connected and never sent anything, instead of waiting for it.
- **`SetMaxConns`** — growing the limit did nothing for requests already queued waiting for a connection slot; they kept waiting for a release or timed out with `ErrNoFreeConns`. Queued waiters are now served as soon as the limit grows.
- **`fasthttpadaptor`** — handlers written against net/http met an adaptor that broke under contracts their server honored: `panic(http.ErrAbortHandler)` surfaced as a crash, `Flush` after `Hijack` blocked forever, and HTTP/1.0 and HTTP/2 requests both reported minor version 1. The flushed body now feeds fasthttp's chunked writer directly, so an abnormal handler exit travels in-band: the server truncates the response and drops the connection, flushed bytes always reach the peer first, a pre-write abort sends nothing and hands the connection back to the server for teardown, and nothing touches the request after the server released it. Where the two servers merely differ in style — a status set after the body started, `Host` in the header map, `ContentLength` for buffered bodies, writes after `Hijack`, bodies on 204/304 — the adaptor deliberately keeps fasthttp's own semantics.
- **`TimeoutHandler`** — on a server driven by `ServeConn` every request was rejected with 429, because the admission channel is only allocated by `Serve` and a nil channel in a select always takes the default. The gate is now allocated on demand.
- **`TLSConnectionState()`** — returns nil when `MaxConnsPerIP` is set even though `IsTLS()` reports true, because only `IsTLS` unwraps the per-IP wrapper. Both now unwrap it.

### DIFF
--- a/fasthttpadaptor/adaptor.go
+++ b/fasthttpadaptor/adaptor.go
@@ -64,10 +64,23 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 		go func() {
 			defer func() {
 				if rec := recover(); rec != nil {
-					ctx.Logger().Printf("panic in net/http handler: %v", rec)
+					// ErrAbortHandler is net/http's quiet give-up.
+					mode := modePanicked
+					if rec == http.ErrAbortHandler {
+						mode = modeAborted
+					}
+					if w.pw != nil {
+						// The stream is the server's now: the exit is recorded
+						// while the request is still ours, then the pipe error
+						// cuts the connection.
+						w.stream.fail(ctx, rec)
+						_ = w.pw.CloseWithError(io.ErrUnexpectedEOF)
+					} else if mode == modePanicked {
+						ctx.Logger().Printf("panic in net/http handler: %v", rec)
+					}
 
 					select {
-					case w.modeCh <- modePanicked:
+					case w.modeCh <- mode:
 					default:
 					}
 				} else {
@@ -115,8 +128,9 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 			releaseWriter(w)
 
 		case modeFlushed:
-			// Streaming: send headers and start SetBodyStreamWriter.
+			// Streaming: send the headers now and hand the body to SetBodyStream.
 			ctx.SetStatusCode(w.status())
+			ctx.Response.ImmediateHeaderFlush = true
 
 			haveContentType := false
 			for k, vv := range w.Header() {
@@ -140,35 +154,11 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 				w.mu.Unlock()
 			}
 
-			ctx.SetBodyStreamWriter(func(bw *bufio.Writer) {
-				// Ensure cleanup only after the stream completes.
-				defer releaseWriter(w)
-
-				// Send pre-flush bytes.
-				if b := w.consumePreflush(); len(b) > 0 {
-					_, _ = bw.Write(b)
-					_ = bw.Flush()
-				}
-
-				// Stream subsequent writes from the pipe until EOF.
-				buf := bufferPool.Get().(*[]byte) //nolint:forcetypeassert
-				defer bufferPool.Put(buf)
-
-				for {
-					n, err := w.pr.Read(*buf)
-					if n > 0 {
-						if _, e := bw.Write((*buf)[:n]); e != nil {
-							return
-						}
-						if e := bw.Flush(); e != nil {
-							return
-						}
-					}
-					if err != nil {
-						return
-					}
-				}
-			})
+			// The pipe feeds fasthttp's chunked writer directly, so the
+			// server truncates the response and drops the connection when a
+			// handler exit closes the pipe with an error.
+			w.stream = &streamBody{pre: w.consumePreflush(), w: w}
+			ctx.Response.SetBodyStream(w.stream, -1)
 
 			// Signal the writer that streaming is ready so Flush() can return.
 			close(w.streamReady)
@@ -177,10 +167,79 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 			releaseWriter(w)
 			return
 
+		case modeAborted:
+			releaseWriter(w)
+			// Hijacking skips the response write and leaves teardown to the
+			// server, so a pooled per-IP wrapper is closed exactly once; the
+			// handler's Close only acts when KeepHijackedConns is set.
+			ctx.HijackSetNoResponse(true)
+			ctx.Hijack(func(c net.Conn) { _ = c.Close() })
+
 		case modePanicked:
 			panic("net/http handler panicked")
 		}
 	}
+}
+
+// streamBody feeds a flushed response through fasthttp's chunked writer and
+// recycles the writer once the server is done with it.
+type streamBody struct {
+	mu     sync.Mutex
+	pre    []byte // pooled pre-flush bytes, recycled by Close
+	w      *writer
+	err    error
+	closed bool
+}
+
+func (b *streamBody) Read(p []byte) (int, error) {
+	b.mu.Lock()
+	if b.closed {
+		b.mu.Unlock()
+		return 0, io.ErrClosedPipe
+	}
+	if len(b.pre) > 0 {
+		n := copy(p, b.pre)
+		b.pre = b.pre[n:]
+		b.mu.Unlock()
+		return n, nil
+	}
+	b.mu.Unlock()
+	// The pipe is read without the lock, so Close and fail never wait on it.
+	return b.w.pr.Read(p)
+}
+
+// fail records a handler exit while the server still owns the request; a
+// closed stream may belong to a request already released or reused.
+func (b *streamBody) fail(ctx *fasthttp.RequestCtx, rec any) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.closed {
+		return
+	}
+	b.err = io.ErrUnexpectedEOF
+	if rec != http.ErrAbortHandler {
+		ctx.Logger().Printf("panic in net/http handler: %v", rec)
+	}
+}
+
+// Close fails further reads before the pre-flush buffer is recycled, since a
+// discarded body may still be read by a compression goroutine, and reports
+// a handler exit the server has not read.
+func (b *streamBody) Close() error {
+	_ = b.w.pr.Close()
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if !b.closed {
+		b.closed = true
+		releaseWriter(b.w)
+	}
+	return b.err
+}
+
+// CloseWithError reports the handler's exit when CompressHandler's goroutine,
+// not the server, closes the stream.
+func (b *streamBody) CloseWithError(error) error {
+	return b.Close()
 }
 
 var bufferPool = sync.Pool{
@@ -195,6 +254,7 @@ const (
 	modeFlushed
 	modeHijacked
 	modePanicked
+	modeAborted
 )
 
 // Writer implements http.ResponseWriter + http.Flusher + http.Hijacker for the adaptor.
@@ -207,8 +267,9 @@ type writer struct {
 	responseBody []byte
 	bufPool      *[]byte
 
-	pr *io.PipeReader
-	pw *io.PipeWriter
+	pr     *io.PipeReader
+	stream *streamBody
+	pw     *io.PipeWriter
 
 	hijacked atomic.Bool
 
@@ -221,15 +282,11 @@ type writer struct {
 }
 
 func acquireWriter(ctx *fasthttp.RequestCtx) *writer {
-	pr, pw := io.Pipe()
 	return &writer{
-		ctx:          ctx,
-		h:            make(http.Header),
-		responseBody: nil,
-		pr:           pr,
-		pw:           pw,
-		modeCh:       make(chan int, 1),
-		streamReady:  make(chan struct{}),
+		ctx:         ctx,
+		h:           make(http.Header),
+		modeCh:      make(chan int, 1),
+		streamReady: make(chan struct{}),
 	}
 }
 
@@ -278,7 +335,11 @@ func (w *writer) Write(p []byte) (int, error) {
 }
 
 func (w *writer) Flush() {
+	if w.hijacked.Load() {
+		return
+	}
 	w.flushOnce.Do(func() {
+		w.pr, w.pw = io.Pipe()
 		select {
 		case w.modeCh <- modeFlushed:
 		default:
@@ -337,8 +398,9 @@ func (w *writer) Hijack() (net.Conn, *bufio.ReadWriter, error) {
 
 func (w *writer) Close() error {
 	w.closeOnce.Do(func() {
-		_ = w.pw.Close()
-		_ = w.pr.Close()
+		if w.pw != nil {
+			_ = w.pw.Close()
+		}
 	})
 	return nil
 }
--- a/fasthttpadaptor/request.go
+++ b/fasthttpadaptor/request.go
@@ -25,13 +25,20 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 	}
 
 	r.Method = b2s(ctx.Method())
-	r.Proto = b2s(ctx.Request.Header.Protocol())
-	if r.Proto == "HTTP/2" {
-		r.ProtoMajor = 2
-	} else {
-		r.ProtoMajor = 1
+	// net/http spells the HTTP/2 version "HTTP/2.0", and its minor version is 0.
+	switch r.Proto = b2s(ctx.Request.Header.Protocol()); r.Proto {
+	case "HTTP/1.1":
+		r.ProtoMajor, r.ProtoMinor = 1, 1
+	case "HTTP/1.0":
+		r.ProtoMajor, r.ProtoMinor = 1, 0
+	case "HTTP/2", "HTTP/2.0":
+		r.Proto, r.ProtoMajor, r.ProtoMinor = "HTTP/2.0", 2, 0
+	default:
+		var ok bool
+		if r.ProtoMajor, r.ProtoMinor, ok = http.ParseHTTPVersion(r.Proto); !ok {
+			r.ProtoMajor, r.ProtoMinor = 1, 1
+		}
 	}
-	r.ProtoMinor = 1
 	r.ContentLength = int64(len(body))
 	r.RemoteAddr = ctx.RemoteAddr().String()
 	r.Host = b2s(ctx.Host())
--- a/server.go
+++ b/server.go
@@ -852,26 +852,30 @@ type tlsConn interface {
 	ConnectionState() tls.ConnectionState
 }
 
+// tlsConnection unwraps ctx.c to its TLS connection, if it is one.
+// The cast is to (tlsConn) instead of (*tls.Conn), since it catches
+// cases with overridden tls.Conn such as:
+//
+//	type customConn struct {
+//	    *tls.Conn
+//
+//	    // other custom fields here
+//	}
+func (ctx *RequestCtx) tlsConnection() (tlsConn, bool) {
+	conn := ctx.c
+	// perIPConn wraps the net.Conn in the Conn field.
+	if pic, ok := conn.(*perIPConn); ok {
+		conn = pic.Conn
+	}
+	tc, ok := conn.(tlsConn)
+	return tc, ok
+}
+
 // IsTLS returns true if the underlying connection is tls.Conn.
 //
 // tls.Conn is an encrypted connection (aka SSL, HTTPS).
 func (ctx *RequestCtx) IsTLS() bool {
-	// cast to (tlsConn) instead of (*tls.Conn), since it catches
-	// cases with overridden tls.Conn such as:
-	//
-	// type customConn struct {
-	//     *tls.Conn
-	//
-	//     // other custom fields here
-	// }
-
-	// perIPConn wraps the net.Conn in the Conn field
-	if pic, ok := ctx.c.(*perIPConn); ok {
-		_, ok := pic.Conn.(tlsConn)
-		return ok
-	}
-
-	_, ok := ctx.c.(tlsConn)
+	_, ok := ctx.tlsConnection()
 	return ok
 }
 
@@ -882,7 +886,7 @@ func (ctx *RequestCtx) IsTLS() bool {
 // The returned state may be used for verifying TLS version, client certificates,
 // etc.
 func (ctx *RequestCtx) TLSConnectionState() *tls.ConnectionState {
-	tc, ok := ctx.c.(tlsConn)
+	tc, ok := ctx.tlsConnection()
 	if !ok {
 		return nil
 	}
--- a/server.go
+++ b/server.go
@@ -988,8 +988,11 @@ func (ctx *RequestCtx) String() string {
 }
 
 // ID returns unique ID of the request.
+//
+// ConnID occupies the high 32 bits and ConnRequestNum the low 32, so the ID
+// repeats once either passes 2^32. Use those accessors to avoid the wrap.
 func (ctx *RequestCtx) ID() uint64 {
-	return (ctx.connID << 32) | ctx.connRequestNum
+	return (ctx.connID << 32) | (ctx.connRequestNum & 0xffffffff)
 }
 
 // ConnID returns unique connection ID.
--- a/server.go
+++ b/server.go
@@ -242,7 +242,8 @@ type Server struct {
 
 	nextProtos map[string]ServeHandler
 
-	concurrencyCh chan struct{}
+	concurrencyCh     chan struct{}
+	concurrencyChOnce sync.Once
 
 	idleConns map[net.Conn]*atomic.Int64
 	done      chan struct{}
@@ -498,7 +499,7 @@ func TimeoutWithCodeHandler(h RequestHandler, timeout time.Duration, msg string,
 	}
 
 	return func(ctx *RequestCtx) {
-		concurrencyCh := ctx.s.concurrencyCh
+		concurrencyCh := ctx.s.getConcurrencyCh()
 		select {
 		case concurrencyCh <- struct{}{}:
 		default:
@@ -1992,9 +1993,7 @@ func (s *Server) Serve(ln net.Listener) error {
 	if s.done == nil {
 		s.done = make(chan struct{})
 	}
-	if s.concurrencyCh == nil {
-		s.concurrencyCh = make(chan struct{}, maxWorkersCount)
-	}
+	s.getConcurrencyCh()
 	s.mu.Unlock()
 
 	wp := &workerPool{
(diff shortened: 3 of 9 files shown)


fasthttpadaptor/request.go
```diff
@@ -25,18 +30,31 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+		r.ContentLength = int64(len(body))
 	}
-	r.ProtoMinor = 1
-	r.ContentLength = int64(len(body))
 	r.RemoteAddr = ctx.RemoteAddr().String()
 	r.Host = b2s(ctx.Host())
 	r.TLS = ctx.TLSConnectionState()
-	r.Body = io.NopCloser(bytes.NewReader(body))
+	if streaming {
+		r.Body = io.NopCloser(ctx.RequestBodyStream())
```
erikdubbelboer: **[P1] Drain or close unread streamed request bodies before reusing the connection.** Passing `RequestBodyStream()` through is useful, but neither the adaptor nor this `io.NopCloser` drains the remainder when `ServeHTTP` returns. With `StreamRequestBody=true`, `MaxRequestBodySize=1`, a `POST` body of `DATA`, and a pipelined `GET /second`, an adaptor handler that ignored the body received the second method as `ATAGET`: the unread `ATA` became part of the next request. Calling `r.Body.Close()` does not help because this wrapper's close is a no-op. Please drain the remaining body after the handler returns, or mark/close the connection when the body remains unread.

fasthttpadaptor/adaptor.go
```diff
@@ -64,6 +65,14 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 		go func() {
 			defer func() {
 				if rec := recover(); rec != nil {
+					// ErrAbortHandler is net/http's quiet give-up.
+					if rec == http.ErrAbortHandler {
+						select {
+						case w.modeCh <- modeAborted:
+						default:
+						}
+						return
```
erikdubbelboer: **[P1] Ensure `ErrAbortHandler` closes an already-flushed stream.** This `return` skips the common `w.Close()` below. If the handler has already called `Flush`, `modeFlushed` has won and the later `modeAborted` send has no receiver; the pipe writer never closes, so response-body consumption blocks indefinitely. A direct `Flush(); panic(http.ErrAbortHandler)` probe reproduced the hang. Cleanup must always close the pipe, while still aborting the connection quietly.

fasthttpadaptor/adaptor.go
```diff
@@ -277,8 +304,44 @@ func (w *writer) Write(p []byte) (int, error) {
...
 }
 
+// CloseNotify implements http.CloseNotifier. The interface is deprecated in
+// net/http in favour of Request.Context, but handlers written against it still
+// expect a ResponseWriter to provide it.
+func (w *writer) CloseNotify() <-chan bool {
+	w.closeNotifyOnce.Do(func() {
+		ch := make(chan bool, 1)
+		w.closeNotifyCh = ch
+		if done := w.contextDone(); done != nil {
```
erikdubbelboer: **[P1] `RequestCtx.Done()` is not a client-disconnect signal.** Fasthttp documents and implements this channel as closing only during server shutdown, so `CloseNotify` never fires when the client connection goes away. A client-close probe timed out, and on a normal long-running `Serve` every call here also leaves one goroutine waiting until shutdown. This needs a real connection-lifetime signal (or the adaptor should avoid claiming `http.CloseNotifier`) rather than proxying the server-wide shutdown channel.

fasthttpadaptor/adaptor.go
```diff
@@ -177,6 +196,11 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 			releaseWriter(w)
 			return
 
+		case modeAborted:
+			releaseWriter(w)
+			ctx.HijackSetNoResponse(true)
+			ctx.Hijack(func(net.Conn) {})
```
erikdubbelboer: **[P2] Explicitly close aborted connections when hijacked connections are retained.** This empty callback relies on fasthttp's automatic post-hijack close, but that close is intentionally disabled when `Server.KeepHijackedConns` is true. In that configuration an `ErrAbortHandler` probe left the client connection open until its read deadline. The abort callback should close the connection it receives.

header.go
```diff
@@ -1806,6 +1811,11 @@ func (h *RequestHeader) AddBytesV(key string, value []byte) {
 // it will be sent after the chunked request body.
 func (h *RequestHeader) AddBytesKV(key, value []byte) {
 	h.bufK, h.bufV = initHeaderKV(h.bufK, h.bufV, b2s(key), b2s(value), h.disableNormalizing)
+	// Adding to Trailer extends the announced set; only Set replaces it.
+	if caseInsensitiveCompare(strTrailer, h.bufK) {
```
erikdubbelboer: **[P2] Preserve `RequestHeader.DisableSpecialHeader`.** This unconditional special-case runs before `setSpecialHeader`, bypassing that function's `disableSpecialHeader` guard. In raw-header mode, adding lowercase `trailer` no longer remains a raw header (and its case/order/lookup behavior changes), contrary to the API contract that special headers are not processed. Please gate this branch with `!h.disableSpecialHeader`.

erikdubbelboer (commented):
I found five behavior and compatibility issues that should be addressed before merging. Details are inline.

fasthttpadaptor/request.go
h2zi: Dropped the streamed-body pass-through; `ConvertRequest` buffers with `PostBody()` again, as on master, so an ignored body cannot bleed into the next request.

Worth flagging separately: the same corruption reproduces without the adaptor. A plain fasthttp handler that ignores the body, with `StreamRequestBody: true` and `MaxRequestBodySize: 1`, sees the pipelined `GET /second` arrive as method `ATAGET` — `serveConn` returns the `requestStream` to its pool without draining it. `SetConnectionClose()` contains it. The old adaptor never hit this only because `PostBody()` materializes the stream, which is the drain. Handing streamed bodies to net/http handlers seems worth doing once the core decides between draining and closing; happy to open that as its own PR.

fasthttpadaptor/adaptor.go
h2zi: Fixed: the recover path now picks the mode and falls through to the shared `w.Close()`, so the pipe closes whichever mode won. `TestHandlerAbortAfterFlush` covers it and hangs without the fix.

fasthttpadaptor/adaptor.go
h2zi: Dropped `CloseNotify` entirely. You're right that `Done()` is the shutdown channel, so it never fired on client disconnect and left a goroutine parked per call. A real signal needs connection-lifetime plumbing that doesn't belong in this PR, and claiming the interface without it is worse than not claiming it.

fasthttpadaptor/adaptor.go
h2zi: Fixed: the abort callback now closes the connection it receives, so `KeepHijackedConns` no longer leaves it open until the read deadline.

header.go
h2zi: Fixed: the branch is gated with `!h.disableSpecialHeader`, so raw-header mode keeps a lowercase `trailer` as an ordinary header. Covered by `TestAddTrailerKeepsRawHeaderMode`.

fasthttpadaptor/adaptor.go
erikdubbelboer: This still reproduces at the current head (`919fcde`). The shared `w.Close()` fixes the hang, but once `modeFlushed` has won, the later `modeAborted` signal is never observed by the completed mode switch. The pipe therefore reaches a clean EOF and fasthttp keeps the connection reusable. In a pipelined probe, `/first` did `Write("partial"); Flush(); panic(http.ErrAbortHandler)`, followed by `GET /second` on the same connection: the first body completed with a nil error and `/second` received a response. `net/http` aborts the connection in this case. Please make a post-Flush abort close/reset the underlying connection too, and add a regression asserting that the pipelined second request is not served.

fasthttpadaptor/request.go
```diff
@@ -68,3 +75,28 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+			}
+			out[name] = nil
+		}
+	}
+	return out
+}
+
+func fillTrailer(h *fasthttp.RequestHeader, dst http.Header) {
+	for name := range dst {
+		if v := h.Peek(name); len(v) > 0 {
```
erikdubbelboer: **[P2] Preserve every request-trailer value and keep trailers out of `Request.Header`.** `Peek` returns only the first value for an announced field. With `Trailer: X-Foo` followed by trailer lines `X-Foo: one` and `X-Foo: two`, this adaptor produces `r.Trailer.Values("X-Foo") == ["one"]`; the later `Header.All()` loop also leaves `Trailer: X-Foo` and both `X-Foo` values in `r.Header`. The equivalent `net/http` request has both values in `r.Trailer` and neither trailer field in `r.Header`. This can lose legitimate duplicate metadata and make middleware treat trailers as ordinary upfront headers. Please copy all values (for example via `PeekAll`) and exclude the trailer announcement plus announced trailer keys from the subsequent header copy.

fasthttpadaptor/adaptor.go
h2zi: Fixed: an abort now flags the writer, and the stream reader closes the underlying connection when it sees the flag, so a post-Flush abort can't end in a clean EOF and the connection never serves a pipelined request. `TestHandlerAbortAfterFlush` now runs your probe — it reads the first chunk, releases the abort, and asserts both that the body errors and that no second response arrives (plus that the handler never ran for it). The flag (rather than `CloseWithError` alone) matters because the writer's cleanup also closes the read end, and that race could otherwise mask the abort error. Also rebased onto current master.

fasthttpadaptor/request.go
h2zi: Fixed: `fillTrailer` copies every value via `PeekAll`, and the header copy skips the `Trailer` announcement plus all announced keys, so they live only in `r.Trailer`. `TestConvertRequestCarriesTrailers` now sends `Foo: one` and `Foo: two` and asserts both values in `r.Trailer` and neither field in `r.Header`.

fasthttpadaptor/adaptor.go
h2zi: Reworked more deeply after auditing the flag approach: closing the connection from the stream callback raced the response writer (a stress probe lost the flushed prefix entirely) and kept a reference to the ctx after the server could release it. The flushed body now goes straight into `Response.SetBodyStream`, so the abort rides fasthttp's own body-stream error path: the server truncates without the terminating chunk and drops the connection, flushed bytes always reach the peer first (per-chunk flush), and any handler panic after Flush — not just `ErrAbortHandler` — truncates the same way. Pre-write aborts now end in `StateClosed` instead of a fake `StateHijacked`. `TestHandlerAbortAfterFlush` runs your pipelined probe with no timing coordination, and a new disconnect test covers mid-stream client loss under `-race`.

fasthttpadaptor/request.go
h2zi: Also reworked while fixing this: the announced set now comes from fasthttp's already-parsed trailer list (`PeekTrailerKeys`) instead of re-splitting the header, all values are copied, and announced names plus the `Trailer` line stay out of `r.Header`. The no-trailer path costs the same as master again (`BenchmarkConvertRequest` is back to baseline). On the response side the announced set dedupes, and a buffered response with trailers switches to chunked so they actually arrive — the trailer test now asserts the wire, not the header map.

fasthttpadaptor/request.go
```diff
@@ -66,5 +75,20 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 		}
 	}
 
+	// net/http surfaces trailers in Request.Trailer, not in the header map,
+	// and drops one nobody announced. fasthttp already parsed the announced
+	// set; the values sit in the header map copied above.
+	if announced := ctx.Request.Header.PeekTrailerKeys(); len(announced) > 0 {
```
erikdubbelboer: **[P1] Clear Trailer when reusing the destination request.** ConvertRequest clears r.Header on every call, but r.Trailer is assigned only when the new request announces trailers. Reusing one http.Request for a trailer-bearing request and then a trailer-free request leaves the first Trailer map attached to the second request; I reproduced X-Result: first leaking across that sequence. Clear or replace r.Trailer before this conditional so a no-trailer request cannot inherit metadata from its predecessor.

fasthttpadaptor/request.go
```diff
@@ -66,5 +75,20 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
 		}
 	}
 
+	// net/http surfaces trailers in Request.Trailer, not in the header map,
```
erikdubbelboer: **[P2] Distinguish received trailers from the announcement list.** PeekTrailerKeys reports only names from the Trailer header, not every field received in the trailer section. HTTP permits an unannounced trailer, and net/http exposes it in Request.Trailer after EOF. With a raw chunked request whose zero chunk is followed by X-Late: value and no Trailer announcement, this conversion leaves r.Trailer empty and exposes X-Late in r.Header as if it were an upfront header. Preserve trailer-section provenance and move every received trailer value into r.Trailer.

fasthttpadaptor/adaptor.go
```diff
@@ -368,3 +434,49 @@ func (w *writer) consumePreflush() []byte {
...
+			name = http.CanonicalHeaderKey(after)
+		} else if !announced[key] {
+			continue
+		}
+		if err := ctx.Response.Header.AddTrailer(name); err != nil {
+			continue
+		}
+		// The trailer section carries one value per name.
+		if len(values) > 0 {
+			ctx.Response.Header.Set(name, values[0])
```
erikdubbelboer: **[P2] Preserve every response-trailer value.** http.Header supports multiple values for one trailer name, but values[0] silently discards every later Header.Add. A handler that announces X-Result and adds one then two produces only Trailer.Values("X-Result") == ["one"] through this adaptor, whereas net/http carries both. Please serialize all values; the fasthttp TrailerHeader path currently peeks only one value, so the storage and wire path both need to retain duplicates.

erikdubbelboer (commented):
Re-reviewed at d04e8a4e90180f5678974b6814cf6d1e2355bf6a. The two previously open discussions are fixed and can be resolved. I found three remaining trailer compatibility issues, detailed inline.

fasthttpadaptor/request.go
h2zi: Fixed: `ConvertRequest` resets `r.Trailer` unconditionally before the announced-set check, so a reused request cannot inherit its predecessor's trailers. `TestConvertRequestResetsTrailerOnReuse` runs your sequence.

fasthttpadaptor/request.go
h2zi: Fixed at the parsing layer: `parseTrailer` now records every field received in the trailer section into the trailer key set, so provenance survives — `PeekTrailerKeys` reports announced and received names alike, and the adaptor moves all of them into `r.Trailer` and out of `r.Header`. Your raw chunked probe (`X-Late` with no announcement) is `TestConvertRequestCarriesUnannouncedTrailer`, plus a core-level `TestReadTrailerRecordsUnannouncedKeys`. A side effect worth noting: a proxied request now re-announces what it actually received.

fasthttpadaptor/adaptor.go
h2zi: Fixed in both layers: the storage keeps every `Add` and `TrailerHeader` now emits one line per stored value (it also no longer emits an empty line for an announced key without a value). `Trailer.Values(\"X-Result\")` carries `[one two]` through the adaptor and both lines reach the wire — `TestHandlerWritesTrailers` asserts the wire, `TestTrailerHeaderKeepsEveryValue` the serialization.

I also ran a net/http-vs-adaptor differential over the trailer matrix (announced/unannounced × single/multi × buffered/flushed × explicit Content-Length × empty body, both directions) to shake out the rest in one pass: it caught that a buffered response with an explicit `Content-Length` should keep it and drop the trailers, and that a bare `http.TrailerPrefix` field on a buffered response is dropped rather than forcing chunked — both now match net/http.

fasthttpadaptor/request.go
```diff
@@ -66,5 +75,21 @@ func ConvertRequest(ctx *fasthttp.RequestCtx, r *http.Request, forServer bool) e
...
+	// net/http surfaces trailers in Request.Trailer, not in the header map.
+	// fasthttp already parsed the received key set; the values sit in the
+	// header map copied above. A reused request must not keep old trailers.
+	r.Trailer = nil
+	if announced := ctx.Request.Header.PeekTrailerKeys(); len(announced) > 0 {
+		r.Trailer = make(http.Header, len(announced))
+		for _, k := range announced {
+			name := http.CanonicalHeaderKey(string(k))
+			r.Trailer[name] = nil
+			for _, v := range ctx.Request.Header.PeekAll(name) {
```
erikdubbelboer: **[P1] Preserve header/trailer value provenance.** `PeekAll(name)` returns every stored value for the key because `RequestHeader` currently co-mingles values from the initial header section and the trailer section. For a raw chunked request with upfront `X-Mixed: upfront` and a later unannounced `X-Mixed: late`, `net/http` yields `Header.Values("X-Mixed") == ["upfront"]` and `Trailer.Values("X-Mixed") == ["late"]`; this conversion instead deletes the header and produces `Trailer == ["upfront", "late"]`. A trailer can therefore retroactively reclassify an ordinary upfront field. Please preserve per-value trailer provenance (or otherwise copy only values parsed from the trailer section) before deleting the key from `r.Header`.

erikdubbelboer (commented):
Re-reviewed at 6a75e5efd4e82d8bf1db54e7d413fdd40f8c0c0a. The three previous trailer findings are fixed and their discussions remain resolved. I found one remaining trailer-provenance issue, detailed inline.

fasthttpadaptor/request.go
h2zi: Right — the root cause is that the header store co-mingles both sections, so the adaptor cannot tell an upfront value from a trailer value through any existing API. Fixing it properly needs a small amount of core support, and I'd rather ask before adding surface: my preference would be to count trailer-section values at parse time (they always land at the tail of the store) and expose them, e.g. `ReceivedTrailers() iter.Seq2[[]byte, []byte]` next to `Trailers()`/`PeekTrailerKeys()` — read-side only, zero cost when unused. An alternative is keeping it unexported and giving the adaptor an internal hook, but that couples the packages. Would you take the small public accessor, or do you prefer a different shape?

erikdubbelboer: I would prefer not to add a public `ReceivedTrailers` API here. Please move all trailer-specific work into a separate follow-up PR and keep the received header/trailer provenance private in core. That PR can make the existing `Header()`/`TrailerHeader()` representations section-correct and let the adaptor consume those, with focused object-size/allocation benchmarks.

For #2352, please remove the trailer-only changes:

- the complete `8ddecc86` trailer commit: `Add("Trailer", ...)` accumulation/deduplication, recording unannounced received keys, and multi-value `TrailerHeader` serialization;
- request-trailer conversion in `ConvertRequest` and its reset/announced/unannounced tests;
- response-trailer handling in `NewFastHTTPHandler`, including `announcedTrailers`, `isTrailerField`, `writeTrailers`, the trailer-only `doneBody` path, and the trailer state on `streamBody`;
- the response trailer wire tests (and incidental trailer setup in non-trailer streaming tests).

The separate trailer PR should cover request and response parsing/serialization together, including same-name upfront/trailer values, announced and unannounced fields, duplicates, copy/reset/reuse, outgoing programmatic trailers, and adaptor round trips. It can be based on the remaining adaptor streaming work after this PR lands.

That leaves this PR focused on: `RequestCtx.ID()` field masking; ConnState activation timing; waking queued `HostClient` waiters when `SetMaxConns` grows (and convergence when it shrinks); the non-trailer `fasthttpadaptor` contracts around `ErrAbortHandler`, panic/flush truncation, Flush-after-Hijack, and protocol version reporting; lazy `TimeoutHandler` admission for `ServeConn`; and TLS state through `perIPConn`.

h2zi: Done: every trailer-specific change is out — the trailer commit, the `ConvertRequest` conversion and its tests, the adaptor response-trailer machinery (`announcedTrailers`/`isTrailerField`/`writeTrailers`, the trailer `doneBody` path, `streamBody`'s trailer state), the wire tests, and the incidental setup in the streaming disconnect test. `header.go` is untouched by this PR again, and the adaptor's trailer behavior is back to master's. The PR now carries exactly the scope you listed. I'll open the trailer follow-up with the coverage you described — both directions, same-name upfront/trailer provenance kept private in core, duplicates, reuse, programmatic trailers, adaptor round trips, and size/allocation benchmarks — once this lands.

fasthttpadaptor/adaptor.go
```diff
@@ -177,12 +171,39 @@ func NewFastHTTPHandler(h http.Handler) fasthttp.RequestHandler {
 			releaseWriter(w)
 			return
 
+		case modeAborted:
+			releaseWriter(w)
+			// Closing the connection here makes the server's response write
+			// fail: nothing reaches the peer and ConnState reports Closed.
+			if c := ctx.Conn(); c != nil {
+				_ = c.Close()
```
erikdubbelboer: With `MaxConnsPerIP` enabled, closing this connection clears `perIPConn.Conn` and returns the wrapper to its pool. The server still tries to write the response afterwards, so this panics instead of returning a write error. I reproduced it over TCP with `MaxConnsPerIP: 1` and a handler that calls `panic(http.ErrAbortHandler)` before flushing. Can you make the abort skip the response write and leave connection cleanup to the server, and add a regression test with the per-IP limit enabled?

server.go
```diff
@@ -2492,6 +2505,12 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 			}
 
 			if err == nil {
+				// Active only once bytes arrive; an idle peer must stay
+				// reclaimable by Shutdown.
+				idleConnTime.Store(0)
```
erikdubbelboer: This now marks the connection active only after the complete header has arrived. If a keep-alive client starts its next request but the header is still incomplete, the connection remains idle and `Shutdown` closes it mid-request. I reproduced this by sending the second request without its final CRLF, starting shutdown, then sending the remainder. The request completes on the PR base but gets disconnected here. Can you mark the connection active as soon as the first byte arrives and add coverage for shutdown during a partial keep-alive header?

erikdubbelboer (changes requested):
I found some new issues.

fasthttpadaptor/adaptor.go
h2zi: Fixed: a pre-write abort now goes through `HijackSetNoResponse` + `Hijack`, so the response write is skipped and the server owns the teardown — the pooled per-IP wrapper is closed exactly once, since `hijackConn.Close` only acts under `KeepHijackedConns` (which keeps the earlier point that the callback closes what it receives). `TestHandlerAbortWithPerIPLimit` runs your TCP repro with `MaxConnsPerIP: 1` for both `KeepHijackedConns` settings, then checks the slot is free and the server still serves; it hits the nil-pointer panic without the fix.

server.go
h2zi: Fixed: the connection is marked active on its first byte, before the header completes. Keep-alive requests already `Peek(1)`; the first request now peeks the same way purely to observe arrival — the error path is unchanged, `Header.Read` still reports whatever went wrong — and the `ReduceMemoryUsage` path activates right after `acquireByteReader`. That also removed the parse-failure special case. `TestServerShutdownWaitsForPartialKeepAliveRequest` runs your sequence (second request minus its final CRLF, Shutdown, then the remainder) and expects the 200; without the fix the connection never goes active on the partial header.

server.go
```diff
@@ -2434,11 +2450,23 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 						err = ErrNothingRead{error: err}
 					}
 				}
+				firstByte = len(b) != 0
+			} else {
+				// Only the arrival matters here; Header.Read reports any error.
+				b, _ := br.Peek(1)
```
erikdubbelboer: Ignoring the error from `Peek(1)` loses a read timeout, since `bufio.Reader` clears it when returning it. We then reset the deadline and try reading again. With `ReadTimeout: 100ms`, I waited for the first read to time out before sending the request: base returns 408, but this branch accepts it and returns 200. Can you keep the error and go through the existing error path, and add a regression test for a request arriving after the first-byte timeout?

erikdubbelboer (changes requested):

server.go
h2zi: Fixed: a new connection now waits for its first byte with the same `Peek(1)` as a keep-alive one, error included, so a timeout there takes the existing path — 408 on a new connection, silent close on keep-alive. The deadline armed when the connection opened stays in force for that request; it is only re-armed after a keep-alive peek. `TestServerFirstRequestAfterReadTimeout` runs your sequence over TCP and gets a 200 without the fix.

h2zi:
A self-review pass over this PR's own code turned up a few more issues; all are fixed with regression tests, and the ReadTimeout deadline matrix now matches master in every mode:

- Shutdown closing a silent connection with `MaxConnsPerIP` set: the pooled per-IP wrapper was recycled under the serving goroutine, which then wrote the error response through a nil `Conn`. The wrapper is no longer recycled and its Close is idempotent (new first commit), and a read that fails because the server closed the connection gets no error response, as in net/http.
- `ReduceMemoryUsage`: the first request's deadline is re-armed after the byte reader's read, as on master.
- `SetMaxConns`: a connection dialed for a waiter now starts the idle cleaner like `AcquireConn` does; without one it stayed in the pool for good.
- fasthttpadaptor: `Flush()` puts the headers on the wire at once (a handler flushing before its first write left the client waiting for them). After a Flush, a panic is reported through the body stream and the server log rather than through a request the server may already have released, and `ErrAbortHandler` stays quiet. Closing a flushed body no longer recycles its pre-flush buffer while `CompressHandler`'s goroutine may still be reading it, a data race on HEAD responses.

h2zi:
@erikdubbelboer Please take a look at this when you have a moment.

h2zi:
Rebased onto current master; the only conflict was `fakeServer`, now built around the atomic `done` from #2402. Another pass over the diff also:

- lets the first-byte change reuse the existing `err == nil` branch for `StateActive`, since the peek now sets `err` on every request;
- has the silent-connection shutdown test check that nothing is logged for a connection the server closed itself;
- sends the headers together with any bytes buffered before a `Flush`, rather than in a write of their own.

The description now also lists the per-IP wrapper fix.

PR #2353, opened 2026-08-11
fix: wait for streaming reads before releasing pooled resources

### LINKED ISSUE #2359: Data race: streamed response body close releases pooled *bufio.Reader and *requestStream while Read is in flight
### Summary

The close closure installed on a streamed response body in `transport.RoundTrip` returns two shared pooled objects to their pools with no synchronization against an in-flight `(*requestStream).Read`. Any caller that closes a streamed response body from a cancellation goroutine — the normal way to abort a hung SSE/chunked stream — races the reader.

### Where

`client.go` (v1.71.0), the closure installed at the end of `transport.RoundTrip`:

```go
resp.bodyStream = newCloseReaderWithError(rbs, func(wErr error) error {
    hc.ReleaseReader(br)                    // *bufio.Reader -> c.readerPool
        releaseRequestStream(r)             // *requestStream -> requestStreamPool
    }
    ...
})
```

`releaseRequestStream` (`streaming.go:97-104`) zeroes `prefetchedBytes`, `totalBytesRead`, `chunkLeft`, `reader`, `header` and `Put`s the struct, concurrently with `(*requestStream).Read` reading `rs.chunkLeft` (`streaming.go:31`), writing it, and dereferencing `rs.reader` (`streaming.go:32`).

### Race detector output

Observed frames (full stacks in the linked downstream issue):

```
WARNING: DATA RACE
  fasthttp/streaming.go:100  releaseRequestStream()
vs
  fasthttp/streaming.go:31   (*requestStream).Read()
```

### Why it looks like more than a benign warning

1. `rs.reader = nil` (`streaming.go:101`) vs `parseChunkSize(rs.reader)` (`streaming.go:32`) — nil dereference.
2. `requestStreamPool.Put` while the pointer is still live — a later request can share the struct; a torn/negative `chunkLeft` reaches the `p[:bytesToRead]` slicing in `Read`.
3. `hc.ReleaseReader(br)` is unconditional, and `readerPool` is a per-`HostClient` `sync.Pool` shared across every request to that host (`client.go:181`, `AcquireReader` `:1975`, `ReleaseReader` `:1995`). Another request can acquire and `Reset` that `*bufio.Reader` onto a different connection while the victim goroutine is still reading through it. For a proxy or gateway that is a cross-request data-bleed shape, not just a crash.

I have **not** observed (3) end to end — I am reporting it as a mechanism implied by the shared pool, not as a reproduced bleed.

### Versions

Reproduced on fasthttp **v1.71.0** and **v1.73.0** (`go test -race`, macOS, Go 1.25.6). Upgrading fasthttp does not change the outcome.

### Reproducer

I do not have a minimal standalone fasthttp reproducer — this was found through a downstream consumer, `maximhq/bifrost`, whose `SetupStreamCancellation` calls `CloseWithError` from a separate goroutine on context cancellation. That downstream issue has the full race-detector stacks:

https://github.com/maximhq/bifrost/issues/6143

The general shape to reproduce: start a chunked/SSE response, have one goroutine blocked in `resp.BodyStream().Read`, and close the body from another goroutine.

### Question for maintainers

Is there a supported, concurrency-safe way to abort an in-flight streamed body read? As far as I can tell only `net.Conn` close and `SetReadDeadline` are safe for concurrent use, and neither is reachable from a caller holding only the response. If closing from another goroutine is unsupported, documenting that on `Response.BodyStream` / `StreamResponseBody` would help downstream libraries a lot — several of them appear to assume `net/http` semantics, where `Body.Close` is safe against a concurrent `Read`.

tbphp:
## Problem

When `StreamResponseBody` is enabled, `CloseWithError` currently releases the pooled `bufio.Reader` and `requestStream` before closing the underlying connection. If another goroutine is blocked in `requestStream.Read`, the release path clears and pools the same object while it is still being read.

This produces a real data race during stream cancellation and timeout handling:

```text
Write: fasthttp.releaseRequestStream
Previous read: fasthttp.(*requestStream).Read
```

PR #2211 prevents a second close from releasing the stream twice, but the first close can still race with an in-flight read.

## Fix

- close the underlying connection first when a streamed body must be interrupted
- wait for all in-flight reads to return before releasing reader, stream, and connection pool objects
- make concurrent closes wait for the first cleanup to finish
- preserve connection reuse for response bodies that were already fully buffered

## Tests

Added deterministic regressions for:

- cancellation while a network read is in flight
- an active read after EOF
- connection reuse for an unread, fully buffered response

Local non-race validation:

- `go test . -count=1`
- `go vet ./...`

Race and multi-platform validation are left to this repository CI.

Fixes: https://github.com/valyala/fasthttp/issues/2359

### DIFF
--- a/client.go
+++ b/client.go
@@ -3233,6 +3233,62 @@ var DefaultTransport RoundTripper = &transport{}
 
 type transport struct{}
 
+// clientStreamBody keeps pooled response resources alive until every in-flight
+// Read has returned. interrupt must unblock network reads without releasing the
+// connection wrapper or reader pools; release performs that cleanup afterward.
+type clientStreamBody struct {
+	reader      io.Reader
+	interrupt   func()
+	release     func(bool)
+	forceClose  bool
+	closed      atomic.Bool
+	fullyRead   atomic.Bool
+	activeReads atomic.Int32
+	readersLock sync.RWMutex
+	closeDone   chan struct{}
+}
+
+func (s *clientStreamBody) Read(p []byte) (int, error) {
+	if s.closed.Load() {
+		return 0, io.ErrClosedPipe
+	}
+
+	s.readersLock.RLock()
+	s.activeReads.Add(1)
+	defer func() {
+		s.activeReads.Add(-1)
+		s.readersLock.RUnlock()
+	}()
+	if s.closed.Load() {
+		return 0, io.ErrClosedPipe
+	}
+
+	n, err := s.reader.Read(p)
+	if errors.Is(err, io.EOF) {
+		s.fullyRead.Store(true)
+	}
+	return n, err
+}
+
+func (s *clientStreamBody) CloseWithError(err error) error {
+	if !s.closed.CompareAndSwap(false, true) {
+		<-s.closeDone
+		return nil
+	}
+	defer close(s.closeDone)
+
+	closeConnection := s.forceClose || err != nil || !s.fullyRead.Load() || s.activeReads.Load() > 0
+	if closeConnection {
+		// Interrupt before waiting on readersLock so a blocked network Read can exit.
+		s.interrupt()
+	}
+
+	s.readersLock.Lock()
+	defer s.readersLock.Unlock()
+	s.release(closeConnection)
+	return nil
+}
+
 func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (retry bool, err error) {
 	customSkipBody := resp.SkipBody
 	customStreamBody := resp.StreamBody
@@ -3324,22 +3380,31 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
 	closeConn := resetConnection || req.ConnectionClose() || resp.ConnectionClose()
 	if customStreamBody && resp.bodyStream != nil {
 		rbs := resp.bodyStream
-		var closed atomic.Bool
-		resp.bodyStream = newCloseReaderWithError(rbs, func(wErr error) error {
-			if !closed.CompareAndSwap(false, true) {
-				return nil
-			}
-			hc.ReleaseReader(br)
-			if r, ok := rbs.(*requestStream); ok {
-				releaseRequestStream(r)
-			}
-			if closeConn || resp.ConnectionClose() || wErr != nil {
-				hc.CloseConn(cc)
-			} else {
-				hc.ReleaseConn(cc)
-			}
-			return nil
-		})
+		bodyStream := &clientStreamBody{
+			reader:     rbs,
+			forceClose: closeConn,
+			closeDone:  make(chan struct{}),
+			interrupt: func() {
+				_ = conn.Close()
+			},
+			release: func(closeConnection bool) {
+				hc.ReleaseReader(br)
+				if r, ok := rbs.(*requestStream); ok {
+					releaseRequestStream(r)
+				}
+				if closeConnection {
+					hc.CloseConn(cc)
+				} else {
+					hc.ReleaseConn(cc)
+				}
+			},
+		}
+		if _, ok := rbs.(*requestStream); !ok {
+			// The network body is already buffered, so closing this in-memory stream
+			// does not prevent the connection from being reused.
+			bodyStream.fullyRead.Store(true)
+		}
+		resp.bodyStream = bodyStream
 		return false, nil
 	}
 	hc.ReleaseReader(br)
--- a/client.go
+++ b/client.go
@@ -3233,18 +3233,18 @@ var DefaultTransport RoundTripper = &transport{}
 
 type transport struct{}
 
-// clientStreamBody keeps pooled response resources alive until every in-flight
-// Read has returned. interrupt must unblock network reads without releasing the
-// connection wrapper or reader pools; release performs that cleanup afterward.
+// clientStreamBody serializes reads and keeps pooled response resources alive
+// until an in-flight Read has returned. interrupt must unblock network reads
+// without releasing the connection wrapper or reader pools; release performs
+// that cleanup afterward.
 type clientStreamBody struct {
-	reader      io.Reader
-	interrupt   func()
-	release     func(bool)
-	forceClose  bool
-	closed      atomic.Bool
-	fullyRead   atomic.Bool
-	readersLock sync.RWMutex
-	closeOnce   sync.Once
+	reader    io.Reader
+	interrupt func()
+	release   func(bool)
+	closed    atomic.Bool
+	fullyRead bool
+	readLock  sync.Mutex
+	closeOnce sync.Once
 }
 
 func (s *clientStreamBody) Read(p []byte) (int, error) {
@@ -3252,15 +3252,15 @@ func (s *clientStreamBody) Read(p []byte) (int, error) {
 		return 0, io.ErrClosedPipe
 	}
 
-	s.readersLock.RLock()
-	defer s.readersLock.RUnlock()
+	s.readLock.Lock()
+	defer s.readLock.Unlock()
 	if s.closed.Load() {
 		return 0, io.ErrClosedPipe
 	}
 
 	n, err := s.reader.Read(p)
 	if errors.Is(err, io.EOF) {
-		s.fullyRead.Store(true)
+		s.fullyRead = true
 	}
 	return n, err
 }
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -3233,6 +3233,62 @@ var DefaultTransport RoundTripper = &transport{}
...
+	if s.closed.Load() {
+		return 0, io.ErrClosedPipe
+	}
+
+	s.readersLock.RLock()
+	s.activeReads.Add(1)
+	defer func() {
+		s.activeReads.Add(-1)
+		s.readersLock.RUnlock()
+	}()
```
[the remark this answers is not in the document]
tbphp: The second `closed.Load()` after incrementing `activeReads` closes this window. If Close wins between `RLock` and `Add`, it sets `closed` and waits for the write lock; the reader then increments, observes `closed`, returns without entering the underlying `Read`, and releases the read lock. If the second check observes false, `activeReads` is already visible before the underlying read can begin, so a later Close interrupts the connection. Moving the increment before `RLock` would count a read that has not acquired the lock or entered the underlying reader and can unnecessarily close a reusable connection.

client.go
```diff
@@ -3233,6 +3233,62 @@ var DefaultTransport RoundTripper = &transport{}
...
+	if closeConnection {
+		// Interrupt before waiting on readersLock so a blocked network Read can exit.
+		s.interrupt()
+	}
+
+	s.readersLock.Lock()
+	defer s.readersLock.Unlock()
+	s.release(closeConnection)
+	return nil
+}
```
erikdubbelboer: Could we simplify this by using `sync.Once` for close completion and `RWMutex.TryLock` to detect an in-flight read? A failed `TryLock` means a `Read` owns the read lock, so we force-close and interrupt before waiting. This removes both the manual active-read counter and the close-completion channel while preserving clean-EOF/buffered connection reuse.

```suggestion
// clientStreamBody keeps pooled response resources alive until every in-flight
// Read has returned. interrupt must unblock network reads without releasing the
// connection wrapper or reader pools; release performs that cleanup afterward.
type clientStreamBody struct {
	reader      io.Reader
	interrupt   func()
	release     func(bool)
	forceClose  bool
	closed      atomic.Bool
	fullyRead   atomic.Bool
	readersLock sync.RWMutex
	closeOnce   sync.Once
}

func (s *clientStreamBody) Read(p []byte) (int, error) {
	if s.closed.Load() {
		return 0, io.ErrClosedPipe
	}

	s.readersLock.RLock()
	defer s.readersLock.RUnlock()
	if s.closed.Load() {
		return 0, io.ErrClosedPipe
	}

	n, err := s.reader.Read(p)
	if errors.Is(err, io.EOF) {
		s.fullyRead.Store(true)
	}
	return n, err
}

func (s *clientStreamBody) CloseWithError(err error) error {
	s.closeOnce.Do(func() {
		s.closed.Store(true)

		closeConnection := s.forceClose || err != nil || !s.fullyRead.Load()
		locked := s.readersLock.TryLock()
		if !locked {
			closeConnection = true
		}
		if closeConnection {
			s.interrupt()
		}
		if !locked {
			// The interrupt lets a blocked network Read release readersLock.
			s.readersLock.Lock()
		}
		defer s.readersLock.Unlock()
		s.release(closeConnection)
	})
	return nil
}
```

erikdubbelboer (commented):
Non-blocking simplification suggestion. The current implementation looks correct; this version expresses the same synchronization with fewer moving parts. I validated it with the PR race regressions x100, the root package under `-race`, `go test ./...`, `go vet ./...`, and `golangci-lint run`.

client.go
```diff
@@ -3324,22 +3380,31 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
...
-				hc.CloseConn(cc)
-			} else {
-				hc.ReleaseConn(cc)
-			}
-			return nil
-		})
+		bodyStream := &clientStreamBody{
+			reader:     rbs,
+			forceClose: closeConn,
+			closeDone:  make(chan struct{}),
```
erikdubbelboer: With `sync.Once`, this per-instance completion channel is no longer needed.

```suggestion
```

client_test.go
```diff
@@ -120,6 +121,253 @@ func TestClientConnectionCounts(t *testing.T) {
...
+		unblockOnce.Do(func() {
+			close(reader.unblock)
+		})
+	}
+	t.Cleanup(unblock)
+
+	released := make(chan bool, 1)
+	stream := &clientStreamBody{
+		reader:    reader,
+		closeDone: make(chan struct{}),
```
erikdubbelboer: This initializer can go away with the `sync.Once` variant.

```suggestion
```

client.go
tbphp: Thanks, this is cleaner. sync.Once and TryLock remove the extra bookkeeping while keeping the interrupt-before-wait behavior. I’ll update the PR tomorrow and rerun the relevant checks before pushing.

tbphp: Updated in 12d1764. The implementation now uses sync.Once and RWMutex.TryLock, with the activeReads and per-instance closeDone bookkeeping removed.
Thanks again — this is much cleaner.

client.go
```diff
@@ -3233,6 +3233,59 @@ var DefaultTransport RoundTripper = &transport{}
...
+// Read has returned. interrupt must unblock network reads without releasing the
+// connection wrapper or reader pools; release performs that cleanup afterward.
+type clientStreamBody struct {
+	reader      io.Reader
+	interrupt   func()
+	release     func(bool)
+	forceClose  bool
+	closed      atomic.Bool
+	fullyRead   atomic.Bool
+	readersLock sync.RWMutex
```
erikdubbelboer: I think this can be simplified one step further. Some wrapper is still required: `conn.Close()` unblocks an active `Read`, but does not wait for it to return before the pooled reader/stream and response state may be released. However, `requestStream` itself is not safe for concurrent `Read` calls, so allowing multiple readers through an `RWMutex` does not preserve a supported behavior.

Could we instead:

- handle the non-`*requestStream` case before constructing this wrapper, releasing `br` and the already-buffered connection immediately;
- use a single `sync.Mutex`/`TryLock`, which serializes reads and makes a failed `TryLock` mean that a read is active;
- keep `fullyRead` as a normal boolean protected by that mutex;
- let the `release` closure capture `closeConn` (`closeConn || discard`), removing `forceClose` from this type?

The close path would still store `closed` first, try the mutex, interrupt before waiting when the stream is unread, errored, or actively being read, and use `sync.Once` so concurrent closes wait for cleanup. That leaves one simple invariant—either one read or cleanup owns the mutex—while retaining clean-EOF connection reuse. Non-blocking; the current version looks correct.

erikdubbelboer (commented):
One additional non-blocking simplification after re-review; the current implementation is correct.

client.go
tbphp: I made another pass based on your feedback. Could you take another look?

client.go
```diff
@@ -3322,25 +3373,25 @@ func (t *transport) RoundTrip(hc *HostClient, req *Request, resp *Response) (ret
...
+		if rbs, ok := resp.bodyStream.(*requestStream); ok {
+			resp.bodyStream = &clientStreamBody{
+				reader: rbs,
+				interrupt: func() {
+					_ = conn.Close()
+				},
+				release: func(discard bool) {
+					hc.ReleaseReader(br)
+					releaseRequestStream(rbs)
+					if closeConn || discard {
```
erikdubbelboer: Could we keep honoring a late `resp.SetConnectionClose()` decision here?

`closeConn` is captured before `RoundTrip` returns, so this condition ignores a flag set by the caller while handling the streamed body. The non-`*requestStream` path below also releases the connection before the caller can set that flag. This drops behavior deliberately added in #1603 for proxies that need to mark the target connection non-reusable after downstream streaming fails.

This focused regression passes at the base (`63901c3d`) and fails at this head with `got 1 dial, want 2`:

```go
func TestClientStreamSetConnectionCloseAfterDo(t *testing.T) {
	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
		w.Header().Set("Content-Length", "11")
		_, _ = io.WriteString(w, "hello world")
	}))
	defer server.Close()

	var dials atomic.Int32
	client := HostClient{
		Addr:                strings.TrimPrefix(server.URL, "http://"),
		MaxResponseBodySize: 1, // force a network-backed requestStream
		StreamResponseBody:  true,
			dials.Add(1)
			return net.Dial("tcp", addr)
		},
	}
	defer client.CloseIdleConnections()

	for i := range 2 {
		var req Request
		req.SetRequestURI(server.URL)
		var resp Response
		if err := client.Do(&req, &resp); err != nil {
			t.Fatal(err)
		}
		if _, err := io.ReadAll(resp.BodyStream()); err != nil {
			t.Fatal(err)
		}
		if i == 0 {
			resp.SetConnectionClose()
		}
		if err := resp.CloseBodyStream(); err != nil {
			t.Fatal(err)
		}
	}

	if got := dials.Load(); got != 2 {
		t.Fatalf("got %d dials, want 2", got)
	}
}
```

The fully buffered case fails the same way with `MaxResponseBodySize` left at zero. Could the cleanup re-check `resp.ConnectionClose()` and retain a compatible close hook for buffered streamed bodies?

tbphp: Done — I’ve updated it in 1c6c09a. Could you check it again?

erikdubbelboer:
Thanks!

PR #2354, opened 2026-08-11
Add canonical header lookup

### LINKED ISSUE #872: [Bug/Question] SetCanonical does not let you Peek the value
I can't seem to Peek the header value after using `SetCanonical` with a second capital letter like `ETag` or `WWW-`, I'm wondering if this is intended behaviour without using `ctx.Response.Header.DisableNormalizing()`.
```go
func main() {
	ctx := &fasthttp.RequestCtx{}

	//ctx.Response.Header.Set("ETag", "13-1831710635")
	ctx.Response.Header.SetCanonical([]byte("ETag"), []byte("13-1831710635"))

	fmt.Println(string(ctx.Response.Header.Header()))
	// HTTP/1.1 200 OK
	// Date: Fri, 04 Sep 2020 19:24:43 GMT
	// ETag: 13-1831710635

	fmt.Println(string(ctx.Response.Header.Peek("ETag")))
	// ""
}
```

fzlzjerry:
## Summary

- add `PeekCanonical` to request and response headers for keys stored with `SetCanonical`
- bypass key normalization without mutating the caller-provided key
- cover exact `ETag` lookups for both header types

## Validation

- `go test ./... -count=1` (Go 1.25)
- `go test -race -shuffle=on ./...` (Go 1.25)
- `golangci-lint run --verbose` (v2.10.1, 0 issues)
- `gofmt` on changed files
- `git diff --check`

Fixes #872

### DIFF
--- a/header.go
+++ b/header.go
@@ -1907,6 +1907,16 @@ func (h *ResponseHeader) PeekBytes(key []byte) []byte {
 	return h.peek(h.bufK)
 }
 
+// PeekCanonical returns header value for the given key without normalizing it.
+// The key must match the canonical form used with SetCanonical.
+//
+// The returned value is valid until the response is released,
+// either though ReleaseResponse or your request handler returning.
+// Do not store references to the returned value. Make copies instead.
+func (h *ResponseHeader) PeekCanonical(key []byte) []byte {
+	return h.peek(key)
+}
+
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2356, opened 2026-08-11
fix: small read buffer error doesn't match the ignored string anymore

Pedr0Rocha:
After [this](https://github.com/valyala/fasthttp/commit/968cd3ce27d09a56d3ac9a3d4f231492b8be6e73), the sentinel errors contain a `fasthttp:` prefix.


erikdubbelboer:
Thanks!

PR #2357, opened 2026-08-13
perf: try to avoid reallocations while reading zstd compressed requests

ethercrow:
One of my applications is spending noticeable amount of CPU in `runtime.growslice` under zstd decompression, so I've tried to reduce the amount of reallocations there a bit.

The added tests are entirely vibe coded, but the main changes are not.

Screenshot 2026-08-13 at 14 48 21

### DIFF
--- a/zstd_test.go
+++ b/zstd_test.go
@@ -114,47 +114,66 @@ func TestEstimateUnzstdSize(t *testing.T) {
 	}
 }
 
-func TestWriteUnzstdPresizesDestination(t *testing.T) {
+func TestWriteUnzstdOptimizedDestinations(t *testing.T) {
 	t.Parallel()
 
-	body := moderatelyCompressibleBody(110000)
+	body := bytes.Repeat([]byte("payload"), 1000)
 	compressedBody := zstdEncodeWithFCS(t, body)
+	prefix := []byte("prefix")
 
-	for _, maxBodySize := range []int{0, len(body), 2 * len(body)} {
-		var bb bytebufferpool.ByteBuffer
-		n, err := writeUnzstd(&bb, compressedBody, maxBodySize)
-		if err != nil {
-			t.Fatalf("unexpected error with maxBodySize=%d: %v", maxBodySize, err)
-		}
-		if n != len(body) {
-			t.Fatalf("unexpected decompressed size %d with maxBodySize=%d. Expecting %d", n, maxBodySize, len(body))
-		}
-		if !bytes.Equal(bb.B, body) {
-			t.Fatalf("unexpected decompressed body with maxBodySize=%d", maxBodySize)
-		}
-		// The destination has been pre-sized from the frame header, so
-		// decompression must not have grown it beyond the initial reservation.
-		if cap(bb.B) > 2*len(body) {
-			t.Fatalf("destination buffer was reallocated with maxBodySize=%d: cap=%d, decompressed size=%d",
-				maxBodySize, cap(bb.B), len(body))
-		}
+	testCases := []struct {
+		name      string
+		newWriter func() (io.Writer, func() []byte)
+	}{
+		{name: "byte slice writer", newWriter: func() (io.Writer, func() []byte) {
+			w := &byteSliceWriter{b: append([]byte(nil), prefix...)}
+			return w, func() []byte { return w.b }
+		}},
+		{name: "byte buffer pool", newWriter: func() (io.Writer, func() []byte) {
+			w := &bytebufferpool.ByteBuffer{B: append([]byte(nil), prefix...)}
+			return w, func() []byte { return w.B }
+		}},
+		{name: "bytes buffer", newWriter: func() (io.Writer, func() []byte) {
+			w := bytes.NewBuffer(append([]byte(nil), prefix...))
+			return w, w.Bytes
+		}},
+	}
+
+	want := append(append([]byte(nil), prefix...), body...)
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			w, result := testCase.newWriter()
+			n, err := writeUnzstd(w, compressedBody, 0)
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if n != len(body) {
+				t.Fatalf("unexpected decompressed size %d. Expecting %d", n, len(body))
+			}
+			if got := result(); !bytes.Equal(got, want) {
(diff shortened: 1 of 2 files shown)


zstd_test.go
```diff
@@ -53,6 +56,138 @@ func testZstdBytesSingleCase(s string) error {
...
+	expectedSize = 4_000_000
+	if size := estimateUnzstdSize(forgedHeader); size != expectedSize {
+		t.Fatalf("unexpected estimate %d for a forged header. Expecting %d", size, expectedSize)
+	}
+}
+
+func zstdFrameHasContentSize(p []byte) bool {
+	var header zstd.Header
+	return header.Decode(p) == nil && header.HasFCS
+}
```
erikdubbelboer: This can be a compact table without the pseudo-random body or multi-megabyte streaming fixture. Compression ratio does not affect whether an FCS below the cap is used, the forged header covers the cap, and this adds coverage for the skippable-frame loop. A tiny literal valid frame makes the no-FCS case deterministic.

```suggestion
func TestEstimateUnzstdSize(t *testing.T) {
	t.Parallel()

	body := bytes.Repeat([]byte("a"), 11_000)
	withFCS := zstdEncodeWithFCS(t, body)
	withoutFCS := []byte{
		0x28, 0xb5, 0x2f, 0xfd, // zstd magic
		0x00,       // no frame content size and not a single segment
		0x00,       // 1 KiB window
		0x09, 0, 0, // last raw block containing one byte
		'x',
	}
	skippablePrefix := []byte{
		0x50, 0x2a, 0x4d, 0x18, // skippable frame magic
		0x03, 0, 0, 0, // payload size
		'f', 'o', 'o',
	}
	forgedFCS := []byte{
		0x28, 0xb5, 0x2f, 0xfd,
		0xe0,
		0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff, 0xff,
	}
	invalid := []byte("not zstd")

	testCases := []struct {
		name string
		src  []byte
		want int
	}{
		{name: "frame content size", src: withFCS, want: len(body)},
		{name: "leading skippable frame", src: append(skippablePrefix, withFCS...), want: len(body)},
		{name: "without frame content size", src: withoutFCS, want: 2 * len(withoutFCS)},
		{name: "forged frame content size is clamped", src: forgedFCS, want: 4_000_000},
		{name: "invalid input", src: invalid, want: 2 * len(invalid)},
		{name: "empty input", src: nil, want: 0},
	}

	for _, testCase := range testCases {
		t.Run(testCase.name, func(t *testing.T) {
			if got := estimateUnzstdSize(testCase.src); got != testCase.want {
				t.Fatalf("unexpected estimate %d. Expecting %d", got, testCase.want)
			}
		})
	}
}
```

erikdubbelboer (commented):
Thanks for the changes, the code looks good but the tests can use some changes:

Two non-blocking suggestions to make the tests smaller and more precise. I applied both suggestions together and verified `go test ./...` and `golangci-lint run`.

zstd_test.go
```diff
@@ -53,6 +56,138 @@ func testZstdBytesSingleCase(s string) error {
...
+	// The pre-allocation must not exceed maxBodySize, and the body must still
+	// be rejected as too large.
+	var bb bytebufferpool.ByteBuffer
+	if _, err := writeUnzstd(&bb, compressedBody, 100); err != ErrBodyTooLarge {
+		t.Fatalf("unexpected error %v. Expecting %v", err, ErrBodyTooLarge)
+	}
+	if cap(bb.B) > 2*len(body) {
+		t.Fatalf("unexpected destination buffer capacity %d", cap(bb.B))
+	}
+}
```
erikdubbelboer: The final-capacity checks do not prove that pre-sizing happened: the first one also passes on the base commit (`len=110000`, `cap=114688`, threshold `220000`), while the size-limit behavior is already covered by `TestBodyDecodeWithLimitTooLarge`. I would test the observable behavior of all three optimized destinations and put the allocation claim in a benchmark.

```suggestion
func TestWriteUnzstdOptimizedDestinations(t *testing.T) {
	t.Parallel()

	body := bytes.Repeat([]byte("payload"), 1000)
	compressedBody := zstdEncodeWithFCS(t, body)
	prefix := []byte("prefix")

	testCases := []struct {
		name      string
		newWriter func() (io.Writer, func() []byte)
	}{
		{name: "byte slice writer", newWriter: func() (io.Writer, func() []byte) {
			w := &byteSliceWriter{b: append([]byte(nil), prefix...)}
			return w, func() []byte { return w.b }
		}},
		{name: "byte buffer pool", newWriter: func() (io.Writer, func() []byte) {
			w := &bytebufferpool.ByteBuffer{B: append([]byte(nil), prefix...)}
			return w, func() []byte { return w.B }
		}},
		{name: "bytes buffer", newWriter: func() (io.Writer, func() []byte) {
			w := bytes.NewBuffer(append([]byte(nil), prefix...))
			return w, w.Bytes
		}},
	}

	want := append(append([]byte(nil), prefix...), body...)
	for _, testCase := range testCases {
		t.Run(testCase.name, func(t *testing.T) {
			w, result := testCase.newWriter()
			n, err := writeUnzstd(w, compressedBody, 0)
			if err != nil {
				t.Fatalf("unexpected error: %v", err)
			}
			if n != len(body) {
				t.Fatalf("unexpected decompressed size %d. Expecting %d", n, len(body))
			}
			if got := result(); !bytes.Equal(got, want) {
				t.Fatalf("unexpected destination %q. Expecting %q", got, want)
			}
		})
	}
}

func BenchmarkWriteUnzstdPresized(b *testing.B) {
	body := bytes.Repeat([]byte("payload"), 16*1024)
	encoder, err := zstd.NewWriter(nil)
	if err != nil {
		b.Fatal(err)
	}
	compressedBody := encoder.EncodeAll(body, nil)
	encoder.Close()

	b.SetBytes(int64(len(body)))
	b.ReportAllocs()
	b.ResetTimer()
	for b.Loop() {
		var dst bytebufferpool.ByteBuffer
		if _, err := WriteUnzstd(&dst, compressedBody); err != nil {
			b.Fatal(err)
		}
	}
}
```

erikdubbelboer:
Please fix the gosec security issue from the CI. After that it can be merged.

ethercrow:
@erikdubbelboer looks like a false positive to me, so I silenced it with a comment.

erikdubbelboer:
@ethercrow I'm sorry but now the linting is failing.

erikdubbelboer:
Thanks!

PR #2358, opened 2026-08-15
feat: add the QUERY method (RFC 10008)

h2zi:
[RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html), published in June 2026, registers QUERY: a safe, idempotent method that carries its query in the request content instead of the URI.

Most of it already worked here. I probed the current behavior over both HTTP/1 and HTTP/2 before writing any code:

- method parsing is a character-table check rather than a whitelist, so `QUERY /contacts HTTP/1.1` already parses
- `ignoreBody()` only excludes GET and HEAD, so the request content is read and written normally
- `parsePostArgs` keys off Content-Type rather than the method, so `ctx.PostArgs()` works for a form-urlencoded QUERY

Redirects with a buffered body were already correct too, which section 2.5 is specific about, since the POST exception for 301/302 must not apply to QUERY:

|       | 301   | 302   | 303                | 307   | 308   |
| ----- | ----- | ----- | ------------------ | ----- | ----- |
| POST  | →GET  | →GET  | →GET, body dropped | POST  | POST  |
| QUERY | QUERY | QUERY | →GET, body dropped | QUERY | QUERY |

That falls out of the 301/302 downgrade being limited to `IsPost()` while the 303 branch is written for any method. The added cases lock it in.

### QUERY is retried like the other idempotent methods

`isIdempotent` did not know the method, so the client would not retry a QUERY after a connection failure:

```
isIdempotent(GET) = true    isIdempotent(POST)  = false
isIdempotent(PUT) = true    isIdempotent(QUERY) = false   <-
```

Section 5.1 registers QUERY with safe=yes and idempotent=yes, and repeating a request automatically is the point of the method over POST. `TestClientIdempotentRequest` covers it and was verified to fail without the change.

### A redirect no longer replays a consumed body stream

This one is a pre-existing bug that has nothing to do with QUERY; it came out of review here because QUERY keeps its body on four statuses instead of two. Verified on a separate worktree at an unmodified `upstream/master`, sending a streamed `hello` body:

| streamed body | 301       | 302       | 307      | 308      | 303          |
| ------------- | --------- | --------- | -------- | -------- | ------------ |
| POST          | **hangs** | **hangs** | `POST\|` | `POST\|` | `GET\|` (ok) |
| QUERY         | `QUERY\|` | `QUERY\|` | `QUERY\|`| `QUERY\|`| `GET\|` (ok) |

Writing the first hop consumes the stream, so a redirect that keeps the body was followed with an empty one. On 301/302 the POST switch to GET is worse than an empty body: the request still announces a `Content-Length` the drained stream can no longer fill, and the next hop blocks forever. PUT behaves like POST.

`HostClient.do` already captures `hasBodyStream` before the first write and refuses to retry for this exact reason. The redirect loop now captures the same flag up front (by the time a 3xx arrives, `req.IsBodyStream()` is already false, since `writeBodyStream` closes the stream) and fails with `ErrRedirectBodyStream` on any body-preserving status. The flag tracks the body it guards: a 303 drops the body and clears it, so once the request has become a body-less GET, later body-preserving statuses in the chain are followed normally — same as master. `TestClientRedirectBodyStream` covers all four statuses for both methods, the two 303 cases, and the 303-then-307 chain; with the guard removed it hangs and the test binary times out.

Happy to move this part into its own PR if it should be reviewed separately.

## The rest

`MethodQuery`, `RequestHeader.IsQuery`, `RequestCtx.IsQuery` and `HeaderAcceptQuery` are names with no behavior of their own.

Left to the application rather than the framework: rejecting a missing Content-Type (section 2), choosing between 415/422/406 (section 2.1), serializing `Accept-Query` as a Structured Field (section 3), the cache key (section 2.7, there is no cache layer here), and CORS preflight (section 4).

31 lines of source and 152 of tests. `go test -race`, `go vet` and golangci-lint are clean.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1294,6 +1294,9 @@ func doRequestFollowRedirects(
 			req.ResetBody()
 			req.postArgs.Reset()
 			req.parsedPostArgs = false
+			// The body is gone for the rest of the chain, so later
+			// body-preserving redirects have nothing left to replay.
+			hasBodyStream = false
 		case req.Header.IsPost() && (statusCode == StatusMovedPermanently || statusCode == StatusFound):
 			// RFC 9110 sections 15.4.2/15.4.3 Note: for historical reasons a
 			// user agent MAY change the request method from POST to GET for a
--- a/client.go
+++ b/client.go
@@ -1622,7 +1622,7 @@ func (c *HostClient) PendingRequests() int {
 }
 
 func isIdempotent(req *Request) bool {
-	return req.Header.IsGet() || req.Header.IsHead() || req.Header.IsPut()
+	return req.Header.IsGet() || req.Header.IsHead() || req.Header.IsPut() || req.Header.IsQuery()
 }
 
 func (c *HostClient) do(req *Request, resp *Response) (bool, error) {
--- a/client.go
+++ b/client.go
@@ -1200,6 +1200,10 @@ var (
 	// ErrTooManyRedirects is returned by clients when the number of redirects followed
 	// exceed the max count.
 	ErrTooManyRedirects = errors.New("fasthttp: too many redirects detected when doing the request")
+	// ErrRedirectBodyStream is returned by clients when a redirect that keeps the request
+	// body is received for a request whose body is a stream. The hop that produced the
+	// redirect consumed the stream, so the body cannot be sent again.
+	ErrRedirectBodyStream = errors.New("fasthttp: cannot follow a body-preserving redirect for a request with a body stream")
 
 	// ErrHostClientRedirectToDifferentScheme is returned when a HostClient follows a redirect to a different protocol.
 	ErrHostClientRedirectToDifferentScheme = errors.New("fasthttp: hostclient can't follow redirects to a different protocol," +
@@ -1230,6 +1234,9 @@ func doRequestFollowRedirects(
 ) (statusCode int, body []byte, err error) {
 	redirectsCount := 0
 	initialHost := hostnameFromURLString(url)
+	// Writing the request consumes a body stream, so remember it here: by the
+	// time a redirect arrives req.IsBodyStream() is already false.
+	hasBodyStream := req.IsBodyStream()
 
 	for {
 		req.SetRequestURI(url)
@@ -1259,6 +1266,14 @@ func doRequestFollowRedirects(
 		url = getRedirectURL(url, location, req.DisableRedirectPathNormalizing, redirectURI)
 		stripSensitiveHeadersOnRedirect(req, initialHost, redirectURI)
 		ReleaseURI(redirectURI)
+
+		// Every redirect but 303 keeps the body, and a consumed stream cannot
+		// produce it again. Fail instead of following with an empty body, the
+		// same reason the retry path refuses to retry such a request.
+		if hasBodyStream && statusCode != StatusSeeOther {
+			err = ErrRedirectBodyStream
+			break
+		}
 
 		switch {
 		case statusCode == StatusSeeOther:
--- a/client_test.go
+++ b/client_test.go
@@ -2221,6 +2221,8 @@ func TestClientRedirectBodyStream(t *testing.T) {
 				ctx.Redirect("/landing", StatusTemporaryRedirect)
 			case "/redirect-308":
 				ctx.Redirect("/landing", StatusPermanentRedirect)
+			case "/redirect-303-then-307":
+				ctx.Redirect("/redirect-307", StatusSeeOther)
 			case "/landing":
 				ctx.SetBodyString(string(ctx.Method()) + "|" + string(ctx.PostBody()))
 			default:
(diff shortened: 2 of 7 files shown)


client_test.go
```diff
@@ -2164,6 +2164,13 @@ func TestClientRedirectMethodSwitch(t *testing.T) {
 		{MethodPost, "/redirect-302", "GET|hello"},
 		// 307 must preserve both the method and the body.
 		{MethodPost, "/redirect-307", "POST|hello"},
+		// RFC 10008 section 2.5: the POST exception for 301/302 does not
+		// apply to QUERY, so the method and body survive all three.
+		{MethodQuery, "/redirect-301", "QUERY|hello"},
```
erikdubbelboer, line 2172: These cases cover buffered bodies, but `DoRedirects` also accepts requests using `SetBodyStream`. The first hop consumes and clears that stream, so a body-preserving 301/302/307/308 is followed with an empty QUERY body. I reproduced this on this head: a streamed `hello` body through a 307 arrived as `QUERY|` instead of `QUERY|hello`. Since [RFC 10008 section 2.5](https://www.rfc-editor.org/rfc/rfc10008.html#section-2.5) calls for a similar QUERY request at the redirected URI, could we either make streamed bodies replayable or return an error instead of silently following when replay is impossible? Please also add a streamed-body regression case; including 308 in the matrix would cover all four body-preserving statuses.

erikdubbelboer (commented):
Thank you for the contribution and for the careful RFC analysis and focused tests. I found one redirect edge case worth addressing.

client_test.go
h2zi: Good catch, thank you. Fixed in 46d13d1 by returning an error.

Before writing the fix I reproduced it on a separate worktree at an unmodified `upstream/master` (918acd2, none of this PR's commits) to see how far it reached, and it turns out to predate QUERY. Streamed `hello` body, echoed by the landing handler as `method|body`:

| streamed body | 301 | 302 | 307 | 308 | 303 |
| --- | --- | --- | --- | --- | --- |
| POST | **hangs** | **hangs** | `POST\|` | `POST\|` | `GET\|` (ok) |
| QUERY | `QUERY\|` | `QUERY\|` | `QUERY\|` | `QUERY\|` | `GET\|` (ok) |

So POST already loses a streamed body on 307/308 today, and on 301/302 it does something worse than arriving empty: the switch to GET keeps the announced `Content-Length` while the stream is already drained, so the next hop blocks forever on a body that never comes. PUT behaves like POST on 307/308. QUERY only made this easy to notice, since it keeps its body on all four statuses instead of just two.

I went with the error rather than making streams replayable: an `io.Reader` cannot be rewound in general, and buffering it would defeat the point of `SetBodyStream`. It also matches what the client already does elsewhere — `HostClient.do` captures `hasBodyStream` before the first write and refuses to retry for exactly this reason. The redirect loop now captures the same flag up front (by the time a 3xx arrives `req.IsBodyStream()` is already false, since `writeBodyStream` closes the stream) and fails with `ErrRedirectBodyStream` on any body-preserving status. 303 is untouched and still works, as it drops the body regardless.

`TestClientRedirectBodyStream` covers all four body-preserving statuses for both POST and QUERY, plus the two 303 cases; with the guard removed it hangs and the test binary times out. `TestClientRedirectMethodSwitch` now includes 308 as you suggested.

Since this part is a pre-existing bug that has nothing to do with QUERY, I am happy to move it into its own PR if you would rather review it separately from the method addition — just say the word.

client.go
```diff
@@ -1260,6 +1267,14 @@ func doRequestFollowRedirects(
 		stripSensitiveHeadersOnRedirect(req, initialHost, redirectURI)
 		ReleaseURI(redirectURI)
 
+		// Every redirect but 303 keeps the body, and a consumed stream cannot
+		// produce it again. Fail instead of following with an empty body, the
+		// same reason the retry path refuses to retry such a request.
+		if hasBodyStream && statusCode != StatusSeeOther {
```
erikdubbelboer, line 1273: `hasBodyStream` records only the initial request state and is never cleared. After a streamed request receives a 303, the switch below correctly drops the body and converts it to a body-less GET. If that GET then receives a 301/302/307/308, this stale flag still returns `ErrRedirectBodyStream` even though there is no body left to replay.

This can be reproduced by extending `TestClientRedirectBodyStream`:

```go
case "/redirect-303-then-307":
	ctx.Redirect("/redirect-307", StatusSeeOther)
```

and adding:

```go
{nil, MethodPost, "/redirect-303-then-307", "GET|"},
```

It fails on this head with `ErrRedirectBodyStream`; the same chain succeeds on `master`. Could the replay-required state be cleared after 303 and this regression case added?

h2zi: You are right, and thank you for the exact repro — it failed here exactly as described. Fixed in abd5371.

The mistake was that `hasBodyStream` recorded the entry state while the thing it guards has a lifecycle: the body can be legitimately dropped mid-chain, and 303 is the only transition that does that (nothing else in the loop calls `ResetBody`, and no path can install a new stream mid-chain). So the flag is now cleared in the 303 branch, right where the body is dropped, and afterwards any body-preserving status is followed as the body-less GET it now is — same as master.

I went through the remaining chain shapes to convince myself this closes the state space: a body-preserving status *before* the 303 already returns the error on the first hop, so "stream → 307 → 303" is unreachable; after a 303 the flag can never be set again, so longer chains add no new states; and a second 303 in a row is a no-op. Both of your regression cases (POST and QUERY through `/redirect-303-then-307`) are added, verified to fail on the previous head and pass now, and I re-checked that the original eight body-preserving cases still hang the test binary when the guard is removed.

erikdubbelboer:
Thanks!

Issue #1764, opened 2024-04-22
Feature request : update and show all examples for doing zero allocation when using fasthttp. 100% zero alloc for each req / resp. 


kolinfluence:
Would it be more accurate to reconsider the representation of "doing nothing" when displaying 0 alloc/op? Could we perhaps offer a benchmark showcasing alloc/op for a basic "hello world" scenario?

I was taken aback by the alloc/op for the "hello world" scenario. Is there a chance I missed something in my approach? Could we illustrate a scenario with significantly reduced alloc/op?

For instance, could we explore employing the HTTP parser for header parsing?

Alternatively, could we examine a scenario where we could utilize the package from uptrace.dev, which boasts zero alloc for parsing HTTP headers, among other features?

P.S.: I'm expressing this as an avid supporter of fasthttp and eagerly anticipating enhancements, as i am adopting this package across various applications.

is there a hello world http parsing example with lesser garbage generated ?

erikdubbelboer:
It doesn't generate garbage. That's exactly why we have these tests: https://github.com/valyala/fasthttp/blob/master/allocation_test.go

What it does do is allocate memory that will be reused over time. And of course it will also generate garbage if you create and destroy structs yourself instead of using the fasthttp Acquire/Release functions.

kolinfluence:
@erikdubbelboer ok i get it. thx

kolinfluence:
opening this issue because 0 alloc / op should be used as example instead. i change the title from 
"Surprised that using routing / http header parsing will generate some garbage."
to feature request to have zero alloc examples.

aabishkaryal:
I have created a PR that shows an example on zero allocation when using fasthttp. https://github.com/valyala/fasthttp/pull/2362

PR #2362, opened 2026-08-18
feat(examples): add zero-allocation server example

aabishkaryal:
#1764 asked for examples that demonstrate zero-allocation usage. helloworldserver allocated on every request through `fmt.Fprintf` and a zero-value `Cookie` instead of `AcquireCookie`/`ReleaseCookie`.

This rewrites helloworldserver so the handler runs at 0 allocs/op. It uses `Response.AppendBody`/`AppendBodyString` with the `Append*` helpers (`AppendHTTPDate`, `AppendIPv4`, `strconv.AppendUint`) and a pooled scratch buffer for the values that aren't already `[]byte`.

`TestZeroAllocation` checks the claim, mirroring `TestAllocationServeConn` in allocation_test.go. Run against the old `fmt`-based handler, the same test reports 22 allocations.

Two notes on behaviour:

* The `-compress` flag still works. Compression allocates, so the 0 allocs/op claim covers the uncompressed path.
* The raw request dump is dropped. It was debug output rather than something specific to fasthttp.

Timestamps now print as HTTP dates instead of Go's default `time.Time` format, which is a side effect of using `AppendHTTPDate`.

Also added a short note on writing zero-allocation handlers under "Fasthttp best practices" in the README.

Refs #1764


erikdubbelboer (changes requested):
I like it!

But I think, since things like this is the main use case for fasthttp, this should replace the `helloworldserver` example. Can you change that please?

aabishkaryal:
> I like it!
> 
> But I think, since things like this is the main use case for fasthttp, this should replace the `helloworldserver` example. Can you change that please?

@erikdubbelboer i have replaced the helloworld example as you requested.

examples/helloworldserver/helloworldserver.go
```diff
@@ -26,30 +31,66 @@ func main() {
...
+	resp.AppendBodyString("Request method is \"")
+	resp.AppendBody(ctx.Method())
+	resp.AppendBodyString("\"\nRequestURI is \"")
+	resp.AppendBody(ctx.RequestURI())
+	resp.AppendBodyString("\"\nRequested path is \"")
+	resp.AppendBody(ctx.Path())
+	resp.AppendBodyString("\"\nHost is \"")
+	resp.AppendBody(ctx.Host())
+	resp.AppendBodyString("\"\nQuery string is \"")
+	resp.AppendBody(ctx.QueryArgs().AppendBytes((*buf)[:0]))
```
erikdubbelboer: This still allocates on every request when the query string is longer than 64 bytes. `AppendBytes` grows the buffer, but the new slice isn't saved in `*buf`, so the pool keeps returning the original small buffer.

Can you keep the grown buffer and add a longer query string to the allocation test please?

```suggestion
	*buf = ctx.QueryArgs().AppendBytes((*buf)[:0])
	resp.AppendBody(*buf)
```

erikdubbelboer (changes requested):
Thanks for updating the example! One small thing left.

examples/helloworldserver/helloworldserver.go
aabishkaryal: @erikdubbelboer this has been addressed. Thank you.

erikdubbelboer:
Thanks!

PR #2365, opened 2026-08-18
Release empty hijack readers before handoff

### LINKED ISSUE #2363: Release empty request reader before hijack to avoid per-connection retention
## Problem

`Server.serveConn` always hands `br` to `hijackConnHandler` when the request is hijacked, even when `br.Buffered() == 0`. The reader is therefore returned to `readerPool` only after the hijack handler exits. For long-lived WebSockets this retains the full server `ReadBufferSize` for the socket lifetime.

This is especially expensive when `ReadBufferSize` must be 32 KiB to accept proxy/CDN forwarded headers and cookies: a 10,000-WebSocket profile attributed about 393 MiB (71.55% of sampled in-use heap) to `bufio.NewReaderSize`. Lowering the header buffer would reintroduce HTTP 431 failures.

Current upstream `server.go:2713-2718` still does:

```go
var hjr io.Reader = c
if br != nil {
    hjr = br
    br = nil
}
```

## Suggested behavior

After the HTTP request is fully consumed and before the hijack handoff:

- when `br.Buffered() == 0`, release the reader to the server pool and pass the raw `net.Conn` to the hijack handler;
- when bytes are buffered, preserve the current reader handoff so a WebSocket frame coalesced with the upgrade request is not lost.

Bytes still in the kernel socket buffer remain readable through the raw connection; only a non-empty userspace buffer requires the wrapper.

## Required regression coverage

1. zero buffered bytes takes the released-reader path;
2. upgrade request + first WebSocket frame coalesced in one write preserves the buffered frame;
3. a partial first frame spanning the buffered reader and raw socket remains intact;
4. TLS coverage for changed record/read boundaries;
5. the 101 response is flushed before the hijack handler proceeds, preserving existing ordering.

I can send a PR with these tests if maintainers agree with the approach.

qtnx:
Fixes #2363.

Release the server bufio.Reader before hijack when it has no buffered bytes, so long-lived hijacked sockets do not retain ReadBufferSize. Keep the existing reader handoff when bytes are buffered, preserving coalesced and partial WebSocket frames.

Coverage includes empty, coalesced, partial, TLS, 101 flush ordering, and a retention benchmark.

### DIFF
--- a/server_hijack_reader_test.go
+++ b/server_hijack_reader_test.go
@@ -1,0 +1,284 @@
+package fasthttp
+
+import (
+	"bufio"
+	"crypto/tls"
+	"io"
+	"net"
+	"runtime"
+	"sync"
+	"testing"
+	"time"
+)
+
+const hijackUpgradeRequest = "GET / HTTP/1.1\r\nHost: example.test\r\nConnection: Upgrade\r\nUpgrade: websocket\r\n\r\n"
+
+func newHijackServer(h HijackHandler) *Server {
+	return &Server{
+		Handler: func(ctx *RequestCtx) {
+			ctx.SetStatusCode(StatusSwitchingProtocols)
+			ctx.Response.Header.Set("Connection", "Upgrade")
+			ctx.Response.Header.Set("Upgrade", "websocket")
+			ctx.Hijack(h)
+		},
+	}
+}
+
+func readHijackResponse(t *testing.T, c net.Conn) *bufio.Reader {
+	t.Helper()
+
+	br := bufio.NewReader(c)
+	var resp Response
+	if err := resp.Read(br); err != nil {
+		t.Fatalf("cannot read hijack response: %v", err)
+	}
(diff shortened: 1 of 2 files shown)


server_hijack_reader_test.go
```diff
@@ -0,0 +1,284 @@
...
+	got := make([]byte, len(payload))
+	if _, err := io.ReadFull(br, got); err != nil {
+		t.Fatalf("cannot read hijack payload: %v", err)
+	}
+	if string(got) != string(payload) {
+		t.Fatalf("unexpected hijack payload %q; want %q", got, payload)
+	}
+}
+
+func BenchmarkServerHijackEmptyReaderRetention(b *testing.B) {
```
erikdubbelboer, line 221: Please remove this, I don't think we need a specific benchmark for this.

erikdubbelboer (changes requested):
Thanks for the changes! Please fix the lint errors and remove the benchmark and then it can be merged.

PR #2366, opened 2026-08-19
Replace github.com/andybalholm/brotli with github.com/molecule-man/go-brrr

### LINKED ISSUE #2364: [Improvement]: replace dependency "github.com/andybalholm/brotli" with "github.com/molecule-man/go-brrr"
Use a better/faster more maintained dependency of fiber.

Fiber user here, just noticed it depends indirectly on `github.com/andybalholm/brotli`. The dependency comes from `github.com/valyala/fasthttp` (this package).

The maintainer of `github.com/andybalholm/brotli` himself recommends [#1](https://github.com/andybalholm/brotli/issues/63#issuecomment-4422788654) [#2](https://github.com/andybalholm/brotli/issues/38#issuecomment-4422811153) using the new library for performance and features.

This dependency-replacement makes sense because:

1. "`github.com/molecule-man/go-brrr`" is basically a drop-in replacement for "github.com/andybalholm/brotli"
2. "`github.com/molecule-man/go-brrr`" is actively maintained
3. "`github.com/molecule-man/go-brrr`" uses SIMD and ist more efficient and faster

My own Benchmarks with 3 CSV files can be seen here: https://claude.ai/code/artifact/e6cd2e0d-6e00-493d-8121-21493aad2e08
For me this is a net positive

@ReneWerner87 
@erikdubbelboer 

Thanks in advance :)

gaby:
Fixes valyala/fasthttp#2364

`andybalholm/brotli` is no longer actively developed and its author points users at `go-brrr` as the successor. `go-brrr` is a pure-Go brotli (RFC 7932) implementation with SIMD-accelerated hot paths, so it is both faster and maintained.

Net diff: **+17 / −25 across 5 files**.

## Public API impact: none

`CompressBrotliBestSpeed` and `CompressBrotliBestCompression` keep the same values (0 and 11) — those lines don't even appear in the diff. No exported fasthttp symbol references the library's types; only the unexported pool helpers `acquireBrotliReader` / `acquireRealBrotliWriter` did.

## Import alias

The library declares `package brrr` while its import path ends in `go-brrr`, so Go requires an explicit alias either way. It's aliased to `brotli`:

```go
brotli "github.com/molecule-man/go-brrr"
```

This keeps the diff limited to lines that genuinely changed, and matches the surrounding identifiers (`CompressBrotliBestSpeed`, `acquireBrotliReader`, `normalizeBrotliCompressLevel`) in a file named `brotli.go`. Tradeoff worth naming: the alias disagrees with the package's declared name, which can mildly surprise someone jumping to the definition.

## API differences handled

- `Reader.Reset` no longer returns an error, so `acquireBrotliReader` drops its error result. Both call sites (`writeUnbrotli`, `readFileHeader`) simplify accordingly.
- `brotli.NewWriter(w, level)` replaces `brotli.NewWriterLevel(w, level)` and validates the level rather than clamping it internally. `acquireRealBrotliWriter` now passes the **normalized** level. This also fixes a latent inconsistency: previously a level like `20` produced a writer at the library's internally-clamped quality but was filed in pool slot 4, so a pooled writer's real level could disagree with the slot holding it. For any in-range level (0–11) behavior is identical.

`*Writer` satisfies the existing `stackless.Writer` interface (`Write`/`Flush`/`Close`/`Reset`), so the stackless compression path is untouched.

## Verification

**Wire-format compatibility, both directions.** Levels 0/1/4/5/6/9/11 × empty, 1-byte, 165 KB and 1 MB payloads. Old-encoder → new-decoder and new-encoder → old-decoder both round-trip byte-exactly, with compressed sizes within a few bytes at every level. This matters because `fs.go` serves precompressed `.br` assets that may have been produced by the old encoder.

**CI.** All 11 checks green on `c2bd73c` — lint, Fuzzing, and test across Go 1.25.x/1.26.x on ubuntu, macOS and Windows.

Locally: `golangci-lint run ./...` → 0 issues, and `go test -race -count=1 .` passes.

**Benchmarks.** 256 KiB HTML payload, linux/amd64, Xeon @ 2.10 GHz, same harness before and after:

| | before | after | speedup |
|---|---|---|---|
| compress lvl 0 | 120.7 µs | 60.1 µs | 2.01× |
| compress lvl 4 | 595.3 µs | 515.4 µs | 1.16× |
| compress lvl 6 | 618.6 µs | 507.9 µs | 1.22× |
| decompress | 396.1 µs | 307.4 µs | 1.29× |

Compression ratios were unchanged in the compatibility runs above, so this is throughput gained without giving up ratio.


brotli.go
```diff
@@ -6,7 +6,7 @@ import (
 	"io"
 	"sync"
 
-	"github.com/andybalholm/brotli"
+	brotli "github.com/molecule-man/go-brrr"
```
erikdubbelboer: Can you put this named import in a separate block below please.

fs.go
```diff
@@ -19,9 +19,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/andybalholm/brotli"
 	"github.com/klauspost/compress/gzip"
 	"github.com/klauspost/compress/zstd"
+	brotli "github.com/molecule-man/go-brrr"
```
erikdubbelboer: Can you put this named import in a separate block below please.

erikdubbelboer:
This dependency change is currently blocked because it changes malformed-input validation.

`github.com/andybalholm/brotli` returns `brotli: excessive input` when a complete Brotli stream has buffered trailing bytes. `github.com/molecule-man/go-brrr` v1.0.1 instead returns the decoded payload with a nil error. I reproduced this through fasthttp by appending `de ad be ef` to a valid Brotli stream: the current dependency rejects it, while the PR version of `AppendUnbrotliBytes` accepts it.

This affects `WriteUnbrotli`, `AppendUnbrotliBytes`, and the request/response `BodyUnbrotli*` / `BodyUncompressed*` paths. Before we can merge, `go-brrr` needs to reject excessive input (or fasthttp must preserve the old validation another way), and this PR needs a regression test proving trailing bytes are rejected.

gaby:
@erikdubbelboer Fixed

brotli.go
```diff
@@ -192,6 +193,54 @@ func AppendUnbrotliBytes(dst, src []byte) ([]byte, error) {
...
+
+// brotliSliceReader hands the decoder everything but the final byte of b,
+// releasing that byte only once the decoder asks for more input. A brotli
+// stream is self-terminating and its final byte always carries stream bits, so
+// a decoder that succeeds without asking for the held back byte ended before
+// the end of b: the leftover is excessive input.
+type brotliSliceReader struct {
+	b        []byte
+	tail     byte
+	tailRead bool
```
erikdubbelboer: I don't think you need this extra state. This should work:
```go
// brotliSliceReader hands the decoder everything but the final byte of b,
// releasing that byte only once the decoder asks for more input. A brotli
// stream is self-terminating and its final byte always carries stream bits, so
// a decoder that succeeds without asking for the held back byte ended before
// the end of b: the leftover is excessive input.
type brotliSliceReader struct {
    b []byte
}

func newBrotliSliceReader(b []byte) *brotliSliceReader {
    return &brotliSliceReader{b: b}
}

func (r *brotliSliceReader) Read(p []byte) (int, error) {
    if len(r.b) > 1 {
        // Always withhold the final byte.
        n := copy(p, r.b[:len(r.b)-1])
        r.b = r.b[n:]
        return n, nil
    }
    if len(r.b) == 1 {
        if len(p) == 0 {
            return 0, nil
        }
        p[0] = r.b[0]
        r.b = r.b[1:]
        return 1, nil
    }
    return 0, io.EOF
}

func (r *brotliSliceReader) excessiveInput() bool {
    return len(r.b) > 0
}
```

erikdubbelboer (changes requested):

erikdubbelboer:
Great, thanks!

PR #2368, opened 2026-08-20
discard unread streamed request body before the next request

alhudz:
Repro: run a `StreamRequestBody: true` server whose handler returns without reading the body, then send `POST /first` with a body of 8 KiB padding followed by `GET /smuggled HTTP/1.1\r\nHost: x\r\n\r\n`. Expected: the handler runs once. Actual: it runs a second time for `/smuggled`, and with a chunked body the next real request on the connection gets a 400 instead.

Cause: `serveConn` released the `requestStream` after the handler returned without consuming what was left of it, so the unread body bytes stayed in `br` and were parsed as the next request (only the first 8 KiB of a fixed-length body is prefetched, and none of a chunked one). `net/http` discards up to 256 KiB here and otherwise closes the connection. #2336 only covers the multipart reader; this is the general case for any handler that does not drain the stream.

Fix: after the handler returns, discard what is left of the stream up to 256 KiB, otherwise answer with `Connection: close`. Same for a request that timed out through `TimeoutHandler`, since its goroutine still owns the stream. `requestStream.Read` now also stays at `io.EOF` after the last chunk instead of reading the next request as a chunk size. Hijacked requests are left alone. The added tests fail on master and pass with this change.

### DIFF
--- a/server.go
+++ b/server.go
@@ -2637,6 +2637,12 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 		ctx.connRequestNum = connRequestNum
 		ctx.time = time.Now()
 
+		// Capture the streamed request body reader, if any, before running the
+		// handler. The handler may return without reading all of it, or abandon
+		// it via CloseBodyStream; either way it is the server, not the handler,
+		// that owns draining it and returning it to the pool afterwards.
+		reqStream, _ := ctx.Request.bodyStream.(*requestStream)
+
 		// If a client denies a request the handler should not be called
 		if continueReadingRequest {
 			s.Handler(ctx)
@@ -2644,13 +2650,16 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 
 		timeoutResponse = ctx.timeoutResponse
 		if timeoutResponse != nil {
-			// The timed out handler may still be reading the streamed body,
-			// so the rest of it cannot be discarded here: close the connection
-			// instead of parsing the unread body as the next request.
-			if ctx.Request.bodyStream != nil {
+			// The timed out handler runs in its own goroutine and may still be
+			// reading the streamed body from br, so it cannot be drained here.
+			// Close the connection, and keep br out of the reader pool while
+			// that goroutine still owns it by dropping our reference to it.
+			if reqStream != nil {
 				connectionClose = true
+				br = nil
 			}
-			// Acquire a new ctx because the old one will still be in use by the timeout out handler.
+			reqStream = nil
+			// Acquire a new ctx because the old one will still be in use by the timed out handler.
 			ctx = s.acquireCtx(c)
 			timeoutResponse.CopyTo(&ctx.Response)
 		}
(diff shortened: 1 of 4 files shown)


server.go
```diff
@@ -2652,6 +2664,16 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 		hijackNoResponse = ctx.hijackNoResponse && hijackHandler != nil
 		ctx.hijackNoResponse = false
 
+		// The handler may have returned without reading the whole streamed
+		// body. What is left of it would be parsed as the next request on
+		// this connection, so discard it, or close the connection when there
+		// is too much left to read.
+		if !connectionClose && hijackHandler == nil {
```
erikdubbelboer: Please determine all connection-closing conditions before attempting this drain. `ctx.Response.Header.ConnectionClose()` (as well as `MaxRequestsPerConn` and `CloseOnShutdown`) is not folded into `connectionClose` until line 2690. With `StreamRequestBody` and an unfinished chunked upload, a handler that deliberately calls `ctx.Response.Header.SetConnectionClose()` now blocks here waiting for more body bytes instead of sending its response and closing. I reproduced this with only the chunked headers on the wire: this PR timed out waiting for a response, while current `master` returned immediately. Please move the final close decision before this block and skip the discard whenever the connection will close.

server.go
```diff
@@ -2652,6 +2664,16 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
...
+			if rs, ok := ctx.Request.bodyStream.(*requestStream); ok && !rs.discard(maxUnreadStreamBodySize) {
```
erikdubbelboer: This check relies on `bodyStream` still being installed after the handler returns. `ctx.Request.CloseBodyStream()`—the exported cleanup API referenced by the `BodyStream` documentation—releases the `requestStream` and clears this field without consuming the wire bytes. A handler that closes an unread stream therefore skips this block; with a body containing 8 KiB of padding followed by `GET /smuggled`, I observed handler paths `[/first /smuggled /next]`. Please track the original stream completion or abandonment independently, and do not return it to the pool until the server has drained it or committed to closing the connection.

server.go
```diff
@@ -2638,6 +2644,12 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 
 		timeoutResponse = ctx.timeoutResponse
 		if timeoutResponse != nil {
+			// The timed out handler may still be reading the streamed body,
+			// so the rest of it cannot be discarded here: close the connection
+			// instead of parsing the unread body as the next request.
+			if ctx.Request.bodyStream != nil {
```
erikdubbelboer: Closing the connection is necessary here, but it is not sufficient while the `TimeoutHandler` goroutine still owns the stream. After `ctx` is replaced, `serveConnCounted` exits and unconditionally returns `br` to `s.readerPool`; the original `requestStream` still points at that reader. A later connection can retrieve and reset it while the timed-out handler reads. In a focused reproduction, I reset the pooled reader tail to `replaced`, resumed the timed-out handler, and it read `replaced` instead of the original tail. Please keep `br` out of the pool and coordinate cleanup until the timed-out handler relinquishes the stream/context.

erikdubbelboer (changes requested):
Requesting changes for three request-stream lifecycle issues: final connection-close conditions are evaluated after draining, CloseBodyStream can bypass unread-body handling, and a timed-out handler can retain a reader that is returned to the pool.

server.go
alhudz: Moved the full close decision up so `MaxRequestsPerConn`, `CloseOnShutdown` and the handler's own `Connection: close` are all folded into `connectionClose` before the drain, and the drain is skipped whenever the connection is closing. Added `TestServerSkipsDiscardOnHandlerConnectionClose`: it sends only the chunked header with a handler that sets `Connection: close`, blocks on `master`, and returns straight away with this change.

server.go
alhudz: Right, `CloseBodyStream` dropped the wire bytes. The server now captures the `*requestStream` before the handler runs and owns its lifecycle: `req.closeBodyStream` no longer returns it to the pool, so a handler that abandons an unread body still gets drained (or the connection closed) afterwards, and the captured reference stays valid. Added `TestServerDiscardsStreamedBodyClosedByHandler`, which reproduced `[/first /smuggled /next]` before the change and now gives `[/first /next]`.

server.go
alhudz: Good catch on the pooled reader. On the timeout path, when a stream is still installed I now drop the server's own reference to `br` (set it to nil) so it isn't returned to `readerPool` while the timed-out handler goroutine still reads from it; the connection still closes as before. The `requestStream` is left for GC once that goroutine relinquishes it rather than being pooled, so a later connection can't reset the reader out from under it.

alhudz:
Pushed `f915a09` covering all three: the close decision is finalised before the drain, the server owns the `requestStream` past a handler's `CloseBodyStream`, and `br` is kept out of the pool while a timed-out handler still reads it. Two new tests (`TestServerDiscardsStreamedBodyClosedByHandler`, `TestServerSkipsDiscardOnHandlerConnectionClose`) both fail on the previous commit and pass now; `go test -race` is green.

PR #2369, opened 2026-08-24
docs: show how to read a bounded response prefix

### LINKED ISSUE #1724: Question: Add client.MaxResponseBody*Read*Size?
Hi, thanks for the great lib.

I'm encountering the same issue as described in issue https://github.com/valyala/fasthttp/issues/686. By setting `Client.MaxResponseBodySize`, it returns `ErrBodyTooLarge` when the response body length exceeds this value. 

However, I would still like to read part of the response body instead of receiving an error, and also prevent potential DOS attacks caused by large responses.

Would it be possible to provide a configuration similar to `client.MaxResponseBody*Read*Size`? This would be similar to what's implemented in [retryablehttp-go](https://github.com/projectdiscovery/retryablehttp-go/blob/97b0628cb1d65f3d2636e8a4262b0faf01ba1d70/client.go#L49-L51).

Thanks!

fzlzjerry:
## What does this PR do?

Documents the existing response-streaming pattern for retaining a bounded body prefix without adding another client option or precedence rule.

- Adds an executable `Response.BodyStream` example that reads `limit + 1` bytes, returns `ErrBodyTooLarge` when truncated, and closes the stream with the error so the partially consumed connection is discarded.
- Expands the `Client.StreamResponseBody` and `HostClient.StreamResponseBody` documentation with the required stream-closing semantics.
- Documents `ReadCloserWithError` and how it is used when a caller stops before EOF.

The pattern can be enabled per response with `resp.StreamBody = true`, or for every response from a client with `StreamResponseBody = true`.

Fixes #1724.

## Tests

- `go test ./...`
- `go test -race .`
- `go vet ./...`
- `golangci-lint run --new-from-rev=origin/master`

### DIFF
--- a/client_example_test.go
+++ b/client_example_test.go
@@ -1,9 +1,14 @@
 package fasthttp_test
 
 import (
+	"errors"
+	"fmt"
+	"io"
 	"log"
+	"net"
 
 	"github.com/valyala/fasthttp"
+	"github.com/valyala/fasthttp/fasthttputil"
 )
 
 func ExampleHostClient() {
@@ -37,4 +42,69 @@ func ExampleHostClient() {
 
 func useResponseBody(body []byte) {
 	// Do something with body :)
+}
+
+func ExampleResponse_BodyStream() {
+	listener := fasthttputil.NewInmemoryListener()
+	server := &fasthttp.Server{
+		Handler: func(ctx *fasthttp.RequestCtx) {
+			ctx.SetBodyString("hello world")
+		},
+	}
+	go func() {
+		_ = server.Serve(listener)
+	}()
+	defer func() {
+		_ = server.Shutdown()
+	}()
+
+	readBodyPrefix := func(resp *fasthttp.Response, limit int) ([]byte, error) {
+		stream := resp.BodyStream()
+		closer, ok := stream.(fasthttp.ReadCloserWithError)
+		if !ok {
+			_ = resp.CloseBodyStream()
+			return nil, fmt.Errorf("unexpected body stream type %T", stream)
+		}
+
+		// Read one extra byte so an exact-size body can be distinguished from
+		// a truncated body.
+		body, err := io.ReadAll(io.LimitReader(stream, int64(limit)+1))
+		if err != nil {
+			_ = closer.CloseWithError(err)
+			return body, err
+		}
+		if len(body) > limit {
+			body = body[:limit]
+			_ = closer.CloseWithError(fasthttp.ErrBodyTooLarge)
+			return body, fasthttp.ErrBodyTooLarge
+		}
+
+		return body, resp.CloseBodyStream()
+	}
+
+	client := &fasthttp.Client{
+		Dial: func(string) (net.Conn, error) {
+			return listener.Dial()
+		},
+	}
+	defer client.CloseIdleConnections()
+
+	req := fasthttp.AcquireRequest()
+	defer fasthttp.ReleaseRequest(req)
+	req.SetRequestURI("http://example.com/")
+
+	resp := fasthttp.AcquireResponse()
+	defer fasthttp.ReleaseResponse(resp)
+	resp.StreamBody = true
+
+	if err := client.Do(req, resp); err != nil {
+		fmt.Printf("request error: %v\n", err)
+		return
+	}
+	body, err := readBodyPrefix(resp, 5)
+	fmt.Printf("%s\n%t\n", body, errors.Is(err, fasthttp.ErrBodyTooLarge))
+
+	// Output:
+	// hello
+	// true
 }
--- a/client_example_test.go
+++ b/client_example_test.go
@@ -60,6 +60,11 @@ func ExampleResponse_BodyStream() {
 
 	readBodyPrefix := func(resp *fasthttp.Response, limit int) ([]byte, error) {
 		stream := resp.BodyStream()
+		// BodyStream is nil for responses without a body, such as HEAD, 204,
+		// and 304 responses.
+		if stream == nil {
+			return nil, nil
+		}
 		closer, ok := stream.(fasthttp.ReadCloserWithError)
 		if !ok {
 			_ = resp.CloseBodyStream()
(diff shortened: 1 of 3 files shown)


erikdubbelboer:
Thanks for working on this. Is another option really needed here? This can already be done with response streaming:

```go
func readBodyPrefix(resp *fasthttp.Response, limit int) ([]byte, error) {
	stream := resp.BodyStream()
	closer := stream.(fasthttp.ReadCloserWithError)

	// Read one extra byte so an exact-size body can be distinguished from
	// a truncated body.
	body, err := io.ReadAll(io.LimitReader(stream, int64(limit)+1))
	if err != nil {
		_ = closer.CloseWithError(err)
		return body, err
	}
	if len(body) > limit {
		body = body[:limit]
		_ = closer.CloseWithError(fasthttp.ErrBodyTooLarge)
		return body, fasthttp.ErrBodyTooLarge
	}

	return body, resp.CloseBodyStream()
}
```

It can be enabled per response:

```go
resp.StreamBody = true
if err := client.Do(req, resp); err != nil {
	return err
}
body, err := readBodyPrefix(resp, 4*1024)
```

Or for every response from a client:

```go
client.StreamResponseBody = true
```

This retains at most `limit` bytes and closes the connection when the remaining response wasn't consumed, which is the behavior added by this PR. I would rather document this existing pattern than add `MaxResponseBodyReadSize` to both `Client` and `HostClient`, together with another set of precedence rules with `MaxResponseBodySize`.

fzlzjerry:
That makes sense. I overlooked that response streaming already provides the bounded-prefix behavior without adding another limit and precedence path. I’ll rework this PR into documentation/example coverage for the existing `StreamBody` / `StreamResponseBody` pattern and remove the new fields and parser changes.

fzlzjerry:
Reworked as suggested. I dropped the `MaxResponseBodyReadSize` fields and parser changes, rebuilt the branch on the latest `master`, and replaced them with an executable `Response.BodyStream` example using the `limit + 1` / `CloseWithError` pattern. I also clarified the stream-closing contract on `Client`, `HostClient`, `Response.BodyStream`, and `ReadCloserWithError`. The full test suite, root-package race tests, `go vet`, and diff-scoped lint all pass.

client_example_test.go
```diff
@@ -38,3 +43,68 @@ func ExampleHostClient() {
...
+	go func() {
+		_ = server.Serve(listener)
+	}()
+	defer func() {
+		_ = server.Shutdown()
+	}()
+
+	readBodyPrefix := func(resp *fasthttp.Response, limit int) ([]byte, error) {
+		stream := resp.BodyStream()
+		closer, ok := stream.(fasthttp.ReadCloserWithError)
```
erikdubbelboer, line 68: One small edge case: Response.BodyStream() is nil for responses that cannot have a body, such as HEAD, 204, and 304, even when StreamBody is true. This helper will return an unexpected body stream type error for those valid responses. Could we handle a nil stream as an empty body before asserting ReadCloserWithError?

fzlzjerry: Good catch. `readBodyPrefix` now treats a nil `BodyStream` as an empty body before the `ReadCloserWithError` assertion, covering HEAD, 204, and 304 responses. This is in `5cb4ab3`; I re-ran the full test suite, root-package race tests, `go vet`, and diff-scoped lint.

erikdubbelboer:
PR #2373 has now merged. It changes streamed `Client` / `HostClient` responses so fixed-length bodies are exposed as network-backed streams immediately, instead of sometimes being fully buffered before `Do` returns. That resolves the underlying issue I found with this example and makes the bounded-prefix pattern valid.

Could you rebase this PR onto the current `master` and keep it as a focused documentation/example follow-up?

- Keep the executable `ExampleResponse_BodyStream`, including the nil-stream handling.
- Preserve the more complete `Client.StreamResponseBody` and `HostClient.StreamResponseBody` documentation added by #2373, and add the example reference without replacing those details.
- Keep the `Response.BodyStream` / `ReadCloserWithError` documentation, but narrow the interface claim to streamed responses returned by `Client` and `HostClient` `Do` methods. `PipelineClient` deliberately exposes a buffered `BodyStream` and that reader does not implement `ReadCloserWithError`.
- Please retain `Fixes #1724`; that issue is still open and this PR remains its closing documentation follow-up.

The current heads conflict in `client.go`, so the rebase is required. Once the branch is updated, I’ll re-review it.

fzlzjerry:
Updated the branch on top of #2373 and narrowed the `ReadCloserWithError` documentation to streamed bodies returned by `Client.Do` and `HostClient.Do`, so it no longer implies support from `PipelineClient`. The branch retains the nil-stream handling in `ExampleResponse_BodyStream`, the fuller `StreamResponseBody` documentation, and `Fixes #1724`.

Locally, `go test .`, the targeted example test, and `go vet ./...` pass. Ready for another look.

http.go
```diff
@@ -355,9 +355,12 @@ func (req *Request) CloseBodyStream() error {
 	return req.closeBodyStream()
 }
 
-// BodyStream returns io.Reader.
+// BodyStream returns the response body stream.
 //
-// You must CloseBodyStream or ReleaseResponse after you use it.
+// When response streaming is enabled, the caller must close the stream with
+// [Response.CloseBodyStream] after reading it. To report why a read stopped
+// before EOF, assert the returned reader to [ReadCloserWithError] and call
```
erikdubbelboer, line 362: Could we narrow this wording to streamed responses returned by `Client.Do` and `HostClient.Do` as well? `PipelineClient` returns a buffered reader that doesn't implement `ReadCloserWithError`, so following this advice there would panic. The interface documentation below already makes this distinction.

erikdubbelboer (changes requested):

http.go
fzlzjerry: Updated in edd6d04. `Response.BodyStream` now explicitly limits the `ReadCloserWithError` assertion to streamed response bodies returned by `Client.Do` and `HostClient.Do`, matching the interface documentation. The general stream-closing requirement remains separate.

The bounded-prefix example, both PipelineClient streaming regression tests, `go test .`, and `go vet ./...` pass locally.

erikdubbelboer:
Thanks!

PR #2371, opened 2026-08-26
fix: send / as the path when the request-target is query-only

### LINKED ISSUE #1659: Issue: Default Path Not Set in FastHTTP Client for URIs with Only Query Parameters

official-burak:
Fixes #1659

RFC 7230 5.3.1: if the path is empty, the client must send `/` as the origin-form path.

`http://example.com?foo=bar` was serialized as `GET ?foo=bar` when path normalizing was off, and `RequestHeader.SetRequestURI("?foo=bar")` did the same. `net/http` and many servers reject that request-target.

Empty `RequestURI` already became `/`. Do the same when it starts with `?`.

### DIFF
--- a/header.go
+++ b/header.go
@@ -786,6 +786,12 @@ func (h *RequestHeader) RequestURI() []byte {
 	requestURI := h.requestURI
 	if len(requestURI) == 0 {
 		requestURI = strSlash
+	} else if requestURI[0] == '?' {
+		// Origin-form requires a path. An empty path is "/".
+		h.requestURI = append(h.requestURI, 0)
+		copy(h.requestURI[1:], h.requestURI[:len(h.requestURI)-1])
+		h.requestURI[0] = '/'
+		requestURI = h.requestURI
 	}
 	return requestURI
 }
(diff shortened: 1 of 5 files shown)


erikdubbelboer:
Thanks!

PR #2372, opened 2026-08-29
fix: let request timeouts override client timeouts

### LINKED ISSUE #2370: request overrides  req.timeout  cannot extend shorter client timeout
client.ReadTimeout shorter timout always take precedence over per request overrides

this means , a client cannot make a network call which may require longer timeouts

related code block
```

> func (t *transport) RoundTrip(.......

readDeadline := deadline
	if hc.ReadTimeout > 0 {
		tmpReadDeadline := time.Now().Add(hc.ReadTimeout)
		if readDeadline.IsZero() || tmpReadDeadline.Before(readDeadline) {
			readDeadline = tmpReadDeadline
		}
	}

writeDeadline := deadline
	if hc.WriteTimeout > 0 {
		tmpWriteDeadline := time.Now().Add(hc.WriteTimeout)
		if writeDeadline.IsZero() || tmpWriteDeadline.Before(writeDeadline) {
			writeDeadline = tmpWriteDeadline
		}
	}

```

Harshal96:
Fixes #2370 by allowing explicit per-request timeouts to override shorter HostClient read and write timeouts. Adds regression coverage that verifies both connection deadlines use the request timeout.

```
go test -shuffle=on ./...
go test -race -run '^TestHostClientRequestTimeoutOverridesClientTimeout$' .
```

### DIFF
--- a/client.go
+++ b/client.go
@@ -265,10 +265,14 @@ type Client struct {
 
 	// Maximum duration for full response reading (including body).
 	//
+	// If a request timeout is set, the shorter timeout applies.
+	//
 	// By default response read timeout is unlimited.
 	ReadTimeout time.Duration
 
 	// Maximum duration for full request writing (including body).
+	//
+	// If a request timeout is set, the shorter timeout applies.
 	//
 	// By default request write timeout is unlimited.
 	WriteTimeout time.Duration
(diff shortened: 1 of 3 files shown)


erikdubbelboer (changes requested):
Thanks for the PR. I don't think we should change this behavior.

`ReadTimeout` and `WriteTimeout` are maximum durations. A request deadline is an additional overall deadline, so the effective deadline should remain:

```
min(request deadline, client read/write deadline)
```

This is the existing behavior introduced in #1336. With this change, `DoTimeout(req, 60*time.Second)` would silently bypass a client's 5-second I/O safety limit. That would be surprising and potentially breaking.

Please keep the current implementation and instead update the documentation to make the precedence explicit, with regression tests covering both cases: the client timeout wins when it is shorter, and the request timeout wins when it is shorter. No additional API is needed.

Harshal96:
@valyala done

erikdubbelboer:
Thanks!

PR #2374, opened 2026-08-30
Give TestClientHeadWithBody a CI-proof read timeout

h2zi:
`TestClientHeadWithBody` fails intermittently on the windows CI runners (seen twice on #2352's runs, both `1.25.x, windows-latest`): its 10ms `ReadTimeout` has to cover the whole first exchange, and a loaded runner misses that window, so the first request times out before the test reaches the sequence it means to check. The second request only needs the timeout to fire eventually, so a larger value keeps the test's intent while surviving slow runners.


erikdubbelboer:
Thanks!

PR #2376, opened 2026-08-31
Do not start the idle-conn cleaner after a failed dial

AshSgDe29071999:
Fixes #2171

`AcquireConn` started `connsCleaner` before `dialHostHard`. When dial failed and no connections remained, the cleaner still slept for `MaxIdleConnDuration`, leaking a goroutine per failed host (worse with a large idle timeout).

Dial first. Start the cleaner after a successful connection, or if other in-use connections still need it. Reset `connsCleanerRun` when the failed dial leaves the pool empty.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1912,14 +1912,27 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
 		}
 	}
 
-	if startCleaner {
-		go c.connsCleaner()
-	}
-
 	conn, err := c.dialHostHard(reqTimeout)
 	if err != nil {
 		c.decConnsCount()
+		if startCleaner {
+			c.connsLock.Lock()
+			if c.connsCount == 0 {
+				c.connsCleanerRun = false
+				startCleaner = false
+			}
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -1912,15 +1912,28 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
...
+		}
+		if !startCleaner {
+			return nil, err
+		}
+		// Other in-use connections still need the cleaner.
+		go c.connsCleaner()
 		return nil, err
 	}
+
+	if startCleaner {
```
erikdubbelboer, line 1934: If the first dial is slow, other dials can succeed and return their connections to the pool, but none of them will start the cleaner because `connsCleanerRun` is already true. Those connections stay open past `MaxIdleConnDuration` until the first dial returns, potentially forever with a custom `Dial`. Can we let a successful concurrent dial start the cleaner? Please also add a test where the first dial stays blocked while a second connection succeeds and becomes idle.

erikdubbelboer (changes requested):

PR #2377, opened 2026-09-01
client: add context-aware request execution

### LINKED ISSUE #2198: Client: add DoContext method for context.Context-aware request execution

RRXXZZYY:
Fixes #2198.

## What

- add package-level, `Client`, and `HostClient` `DoContext` entry points
- add optional context-aware dialer and transport interfaces while preserving legacy `Do` dispatch
- propagate cancellation through connection-pool waits, default DNS/dial, TLS, request writes, closable request streams, response reads, and streamed responses
- preserve the earlier of the request timeout and context deadline, including existing `ErrTimeout` semantics
- keep ordinary buffered `DoTimeout` / `DoDeadline` requests on the existing no-watcher hot path

## Why

Copying a context deadline into the request timeout is not enough to interrupt pool waits, dialing, TLS negotiation, blocked I/O, or streamed response reads. This change carries cancellation through those blocking boundaries while retaining compatibility for callers that use the existing dialer and transport interfaces.

## Verification

- `go vet ./...` with Go 1.25.0
- `go test -count=1 -shuffle=on ./...` with Go 1.25.0 and Go 1.26.7
- `go test -race -count=1 -shuffle=on ./...` with Go 1.25.0 and Go 1.26.7
- focused context tests under `-race -count=10`
- focused context and stream-close tests under `-count=20` with Go 1.25.0 on Windows/amd64
- request-body cancellation/timeout/early-exit regressions under `-race -count=100`
- golangci-lint v2.10.1 in a clean LF checkout: 0 issues
- `BenchmarkClientGetTimeoutFastServer`, 5 runs at 2s with `-benchmem`: HEAD and this branch both report 230 B/op and 7 allocs/op; timing distributions overlap
- `git diff --check`

## Compatibility and boundaries

- legacy `Dial`, `DialTimeout`, and custom `RoundTripper` implementations cannot be forcibly interrupted; callers that need cancellation should use `DialContext` and `RoundTripperWithContext`
- cancellation calls `Close` on a request body stream when it implements `io.Closer`; a non-interruptible reader must return on its own
- full and race validation was run on Linux/amd64; Windows/amd64 received the focused 20-run context suite, while the full Windows suite, Windows race mode, and macOS remain unverified locally and are left to CI

## AI assistance

I used AI tools to help trace blocking paths, draft implementation and tests, and review the final diff. I manually reviewed every change, reproduced the identified edge cases, and ran all verification listed above.

### DIFF
--- a/client.go
+++ b/client.go
@@ -3,6 +3,7 @@ package fasthttp
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -38,6 +39,15 @@ import (
 // positive MaxResponseBodySize when requesting untrusted servers.
 func Do(req *Request, resp *Response) error {
 	return defaultClient.Do(req, resp)
+}
+
+// DoContext performs the given request and fills the given response, honoring
+// cancellation and deadlines from ctx.
+//
+// The request's existing timeout is also honored. If both it and ctx have a
+// deadline, the earlier deadline is used.
+func DoContext(ctx context.Context, req *Request, resp *Response) error {
+	return defaultClient.DoContext(ctx, req, resp)
 }
 
 // DoTimeout performs the given request and waits for response during
@@ -207,6 +217,10 @@ type Client struct {
 
 	// Transport defines a transport-like mechanism that wraps every request/response.
 	Transport RoundTripper
+
+	// Callback for establishing new connections to hosts while honoring a
+	// request context. DialContext takes precedence over DialTimeout and Dial.
+	DialContext DialFuncWithContext
 
 	// Callback for establishing new connections to hosts.
 	//
@@ -492,6 +506,22 @@ func (c *Client) DoDeadline(req *Request, resp *Response, deadline time.Time) er
 	return c.Do(req, resp)
 }
 
+// DoContext performs the given request and fills the given response, honoring
+// cancellation and deadlines from ctx.
+//
+// The request's existing timeout is also honored. If both it and ctx have a
+// deadline, the earlier deadline is used. A nil context causes a panic.
+// If the request body stream implements io.Closer, cancellation calls Close;
+// the stream should use it to unblock an in-progress Read. A body reader that
+// cannot be interrupted must return on its own before cancellation can complete.
+// Custom dialers and transports must use DialContext and
+// RoundTripperWithContext respectively to be interrupted by cancellation.
+func (c *Client) DoContext(ctx context.Context, req *Request, resp *Response) error {
+	return doRequestContext(ctx, req, resp, func() error {
+		return c.do(ctx, req, resp, true)
+	})
+}
+
 // DoRedirects performs the given http request and fills the given http response,
 // following up to maxRedirectsCount redirects. When the redirect count exceeds
 // maxRedirectsCount, ErrTooManyRedirects is returned.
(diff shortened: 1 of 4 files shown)


erikdubbelboer (changes requested):
To be honest, I'm not so sure about this one. It adds a lot of complexity and makes things slower. `fasthttp` is build for the highest performance and it's API is designed around that. That's why `context` is not a very important thing in `fasthttp`. If you need context support just use `net/http`.

An example of the extra overhead is this test:
```go
func TestDoTimeoutClosableBodyAllocations(t *testing.T) {
	client := &HostClient{
		Addr: "example.com:80",
		Dial: func(string) (net.Conn, error) {
			return &fakeClientConn{
				ch: make(chan struct{}, 1),
				s:  []byte("HTTP/1.1 200 OK\r\nContent-Length: 2\r\n\r\nok"),
			}, nil
		},
	}
	defer client.CloseIdleConnections()

	var req Request
	var resp Response
	defer req.Reset()
	defer resp.Reset()
	req.SetRequestURI("http://example.com/")
	req.Header.SetMethod(MethodPost)

	reader := strings.NewReader("body")
	body := io.NopCloser(reader)
	allocs := func(timeout time.Duration) float64 {
		return testing.AllocsPerRun(100, func() {
			reader.Reset("body")
			req.SetBodyStream(body, 4)
			var err error
			if timeout == 0 {
				err = client.Do(&req, &resp)
			} else {
				err = client.DoTimeout(&req, &resp, timeout)
			}
			if err != nil {
				t.Fatal(err)
			}
		})
	}

	// Compare with untimed requests so allocations from the fake connection
	// cancel out. AllocsPerRun warms up each path before measuring it.
	withoutTimeout := allocs(0)
	withTimeout := allocs(time.Second)
	if withTimeout > withoutTimeout {
		t.Fatalf("timeout adds %.0f allocations per request (without: %.0f, with: %.0f)",
			withTimeout-withoutTimeout, withoutTimeout, withTimeout)
	}
}
```

It passes on `master` but on your branch it shows that the code causes 10 extra heap allocations.

RRXXZZYY:
Thanks for the concrete allocation test. The extra allocations came from routing the default transport's legacy RoundTrip through the context watcher. I've split that path so Do/DoTimeout keep the original fast path; only DoContext installs cancellation machinery. I also added the allocation check as a regression test. The allocation test, related context/timeout tests, and go vet ./... pass.

PR #2378, opened 2026-09-01
fix: write trailers for buffered bodies in Request.Write and Response.Write

aaydin-tr:
@erikdubbelboer `writeTrailer` is only called from the streamed body paths, so writing a request/response with a buffered body and registered trailers produces `Content-Length` framing with a dangling `Trailer:` header and the values are never written.

Found while working on #2104, split out as requested there.

[RFC 9112 7.1.2](https://datatracker.ietf.org/doc/html/rfc9112#name-chunked-trailer-section)

### DIFF
--- a/http.go
+++ b/http.go
@@ -1856,6 +1856,9 @@ func (req *Request) Write(w *bufio.Writer) error {
 		req.Header.SetContentLength(len(body))
 	}
 	if hasBody && len(req.Header.trailer) > 0 && req.Header.IsHTTP11() {
+		if len(body) > 0 && !req.Header.noDefaultContentType && len(req.Header.ContentType()) == 0 {
+			req.Header.SetContentTypeBytes(strDefaultContentType)
+		}
 		req.Header.SetContentLength(-1)
 		if err = req.Header.Write(w); err != nil {
 			return err
--- a/http.go
+++ b/http.go
@@ -1855,12 +1855,12 @@ func (req *Request) Write(w *bufio.Writer) error {
 		hasBody = true
 		req.Header.SetContentLength(len(body))
 	}
-	if hasBody && len(req.Header.trailer) > 0 {
+	if hasBody && len(req.Header.trailer) > 0 && req.Header.IsHTTP11() {
 		req.Header.SetContentLength(-1)
 		if err = req.Header.Write(w); err != nil {
 			return err
 		}
-		if err = writeBodyChunked(w, bytes.NewReader(body)); err != nil {
+		if err = writeBufferedBodyChunked(w, body); err != nil {
 			return err
 		}
 		return req.Header.writeTrailer(w)
@@ -2300,12 +2300,12 @@ func (resp *Response) Write(w *bufio.Writer) error {
 	if sendBody || bodyLen > 0 {
 		resp.Header.SetContentLength(bodyLen)
 	}
-	if sendBody && len(resp.Header.trailer) > 0 {
+	if sendBody && len(resp.Header.trailer) > 0 && resp.Header.IsHTTP11() {
 		resp.Header.SetContentLength(-1)
 		if err := resp.Header.Write(w); err != nil {
 			return err
 		}
-		if err := writeBodyChunked(w, bytes.NewReader(body)); err != nil {
+		if err := writeBufferedBodyChunked(w, body); err != nil {
 			return err
 		}
 		return resp.Header.writeTrailer(w)
@@ -2803,6 +2803,13 @@ var copyBufPool = sync.Pool{
 }
 
 func writeChunk(w *bufio.Writer, b []byte) error {
+	if err := writeChunkNoFlush(w, b); err != nil {
+		return err
+	}
+	return w.Flush()
+}
+
+func writeChunkNoFlush(w *bufio.Writer, b []byte) error {
 	n := len(b)
 	if err := writeHexInt(w, n); err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


http.go
```diff
@@ -2290,6 +2300,16 @@ func (resp *Response) Write(w *bufio.Writer) error {
 	if sendBody || bodyLen > 0 {
 		resp.Header.SetContentLength(bodyLen)
 	}
+	if sendBody && len(resp.Header.trailer) > 0 {
+		resp.Header.SetContentLength(-1)
```
erikdubbelboer, line 2307: This also switches buffered responses to chunked encoding for HTTP/1.0 requests, which don't support it. The request branch has the same problem: an HTTP/1.0 POST with a buffered body and a registered trailer now produces a request that our own parser rejects with `ErrUnsupportedTransferEncoding`. Both cases work before this change. Can we keep the existing framing when chunked encoding isn't supported?

http.go
```diff
@@ -1855,6 +1855,16 @@ func (req *Request) Write(w *bufio.Writer) error {
 		hasBody = true
 		req.Header.SetContentLength(len(body))
 	}
+	if hasBody && len(req.Header.trailer) > 0 {
+		req.Header.SetContentLength(-1)
+		if err = req.Header.Write(w); err != nil {
+			return err
+		}
+		if err = writeBodyChunked(w, bytes.NewReader(body)); err != nil {
```
erikdubbelboer: `writeBodyChunked` calls `writeChunk`, which flushes the writer after every chunk. So even a four-byte buffered body now causes two flushes before the trailers are written, while `Write` is documented not to flush. This also happens in the response branch. Can we write the chunk framing without explicitly flushing for buffered bodies, so callers can still batch their writes?

erikdubbelboer (changes requested):

http.go
aaydin-tr: Yes, missed that. Now only using chunked when the header is HTTP/1.1, HTTP/1.0 keeps the old Content-Length framing. The response header on the server side doesn't know the request version though, so I set noHTTP11 on it in serveConn for HTTP/1.0 requests. Added tests for request, parsed HTTP/1.0 response and ServeConn with both versions.

http.go
aaydin-tr: Split writeChunk into writeChunkNoFlush + flush, buffered bodies now go through writeBufferedBodyChunked which doesn't flush at all. Streaming path is unchanged. Added a test with a counting writer under the bufio.Writer to make sure nothing gets written during Write.

http.go
```diff
@@ -1855,6 +1855,16 @@ func (req *Request) Write(w *bufio.Writer) error {
 		hasBody = true
 		req.Header.SetContentLength(len(body))
 	}
+	if hasBody && len(req.Header.trailer) > 0 && req.Header.IsHTTP11() {
+		req.Header.SetContentLength(-1)
```
erikdubbelboer, line 1862: Setting the content length to -1 here also removes the default Content-Type. The header writer only adds `application/octet-stream` when the content length is positive, so a buffered POST with trailers now sends no Content-Type unless the caller sets one explicitly. Can we preserve the existing default when switching to chunked encoding?

server.go
```diff
@@ -2682,6 +2682,11 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 			// connections are keep-alive by default.
 			ctx.Response.Header.setNonSpecial(strConnection, strKeepAlive)
 		}
+		if !ctx.Request.Header.IsHTTP11() {
```
erikdubbelboer: This still misses `TimeoutErrorWithResponse`: above this point we replace `ctx` with a fresh one, whose request defaults to HTTP/1.1. An HTTP/1.0 request receiving a buffered timeout response with trailers therefore still gets chunked encoding. Can we save the original request version before replacing the context and use that here?

erikdubbelboer (changes requested):

server.go
aaydin-tr: Missed that one. Now saving IsHTTP11 from the original request before the ctx swap and using it for the noHTTP11 flag. Used the same value for the keep-alive branch right above it too, since it had the same problem for HTTP/1.0 timeout responses. Added a ServeConn test with TimeoutErrorWithResponse for both versions.

http.go
aaydin-tr: Right, the request writer only adds the default for a positive Content-Length. Now setting application/octet-stream before switching to chunked, under the same conditions the writer uses (non-empty body, no explicit Content-Type, default not disabled), so an empty POST still gets none like before. Added a test covering default, explicit, no-default and empty body.

server.go
```diff
@@ -2641,6 +2642,10 @@ func (s *Server) serveConnCounted(c net.Conn, countConcurrency bool) error {
 			s.Handler(ctx)
 		}
 
+		// Remember the request version before ctx may be replaced with a
+		// fresh one below, whose request defaults to HTTP/1.1.
+		isHTTP11 = ctx.Request.Header.IsHTTP11()
```
erikdubbelboer, line 2647: Can we save this before calling `s.Handler(ctx)`? `TimeoutHandler` can return while the handler goroutine is still modifying the request, so reading `IsHTTP11()` here races with those changes. I reproduced this with a handler calling `SetProtocol`: the race detector reports a race on this line with the PR, while the same test passes on master.

erikdubbelboer (changes requested):

server.go
aaydin-tr: Good catch, moved it up next to where connectionClose is saved so it's read before the handler runs. Added a test with TimeoutHandler where the handler calls SetProtocol and keeps going after the timeout, for both 1.0 and 1.1.

erikdubbelboer:
Thanks!

PR #2379, opened 2026-09-03
avoid race condition for closing the streaming connection at EOF

akshaydeo:
`requestStream` has no terminal state, so after it returns io.EOF for a chunked body the next Read re-enters parseChunkSize and waits for a chunk header that never comes, because a keep-alive connection stays open after the body ends. Any caller that reads a streamed body to EOF and then reads again, which is what draining before release does, parks a goroutine and never releases the connection, leaking one per affected response until the process restarts. This is separate from the pooled-resource race fixed in #2353: that made closing safe while a read is in flight, this is about a read issued after the body has legitimately ended. The fix records EOF on the stream and returns io.EOF immediately on subsequent reads, clearing the flag in releaseRequestStream so pooled instances start clean; the fixed-length paths were already re-checking totalBytesRead against contentLength and are unaffected, and strictEOF still converts a short read to io.ErrUnexpectedEOF without marking the stream terminal. TestRequestStreamStaysAtEOF reads a chunked body to EOF over a deliberately open pipe and then reads again, failing before with Read after EOF blocked waiting for another chunk header and passing after. Full suite green with and without -race (1171 tests, 16 packages), no existing test changed or removed.

### DIFF
--- a/streaming.go
+++ b/streaming.go
@@ -22,11 +22,22 @@ type requestStream struct {
 	totalBytesRead  int
 	chunkLeft       int
 	strictEOF       bool
+	eof             bool
 }
 
 func (rs *requestStream) Read(p []byte) (int, error) {
 	if rs.reader == nil {
 		panic("BUG: reading released body stream")
+	}
+
+	// The stream is terminal once the body has ended. Without this, a chunked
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks!

PR #2380, opened 2026-09-06
fix: prevent stale query string after clearing query args

cuishuang:
When a URI is parsed, `QueryArgs()` sets `parsedQueryArgs` to true while the original raw query remains in `queryString`.

If all query arguments are subsequently removed with `QueryArgs().Del()` or `QueryArgs().Reset()`, `queryArgs.Len()` becomes zero. `URI.RequestURI()` then incorrectly falls back to the stale `queryString`, causing the removed query to reappear.

For example:

```go
var u URI
if err := u.Parse(nil, []byte("http://example.com/path?a=1")); err != nil {
    panic(err)
}

u.QueryArgs().Del("a")
fmt.Println(string(u.RequestURI()))
```

Before this change:

```text
/path?a=1
```

Expected:

```text
/path
```

Treat `parsedQueryArgs` as the source of truth once query arguments have been parsed. If the parsed arguments are empty, omit the query string instead of falling back to the original raw query.

### DIFF
--- a/uri.go
+++ b/uri.go
@@ -762,9 +762,11 @@ func (u *URI) RequestURI() []byte {
 	} else {
 		dst = appendQuotedPath(u.requestURI[:0], u.Path())
 	}
-	if u.parsedQueryArgs && u.queryArgs.Len() > 0 {
-		dst = append(dst, '?')
-		dst = u.queryArgs.AppendBytes(dst)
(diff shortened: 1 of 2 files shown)


erikdubbelboer:
Thanks! Nice fix!

PR #2387, opened 2026-09-07
perf: reduce per-request CPU in header, URI and connection handling

gaby:
## Summary

Cuts per-request CPU across the hot paths: header serialization and parsing, URI parsing, HTTP date formatting, and the server and client request loops. Parse results, accepted byte sets and serialized output are identical to master (verified by differential fuzzing, see below), and every zero-allocation path stays at zero allocations.

## Changes

**Header serialization** (0174a81)
- `appendHeaderLine`, `AppendBytes` and `formatStatusLine` append `": "` and CRLF as byte constants instead of copying from package-level slices, which removes two `memmove` calls per header line.
- Default `HTTP/1.1 <code> <message>\r\n` status lines are precomputed once and appended with a single copy.
- `RequestHeader.Write` and `ResponseHeader.Write` build the header directly in the `bufio.Writer` buffer when it has room, instead of building it in `bufV` and copying it over. `Header()` is unchanged.

**Header parsing** (0174a81, 84de019)
- Header values are validated eight bytes at a time; only words that may contain a tab or an invalid byte fall back to the per-byte table lookup. The accepted set is unchanged.
- `readRawHeaders` records the line terminators it finds and hands them to the scanner, so the request header block is no longer scanned for `\n` twice.
- The scanner validates the key while searching for its colon, one pass instead of `IndexByte` plus a validation loop.
- `validHeaderFieldByteTable` now covers all 256 byte values, removing the range check per byte.
- `RequestHeader.Method()` fills the default `GET` into the header's own buffer instead of allocating a new slice on every call (05ef200; the result stays request-local).

**URI parsing** (b5365d2)
- The control-byte scan of the request target runs eight bytes at a time.
- `IndexByte` gates the backward `LastIndexByte` scans for `@` and `:` in the host.
- Host bytes in `unescape` are validated through a generated table (`hostShouldEscapeTable`) instead of calling `shouldEscape` for every byte.
- `lowercaseBytes` lowercases eight bytes at a time.
- `appendQuotedPath` and `AppendQuotedArg` copy the prefix that needs no escaping in bulk.

**Date formatting** (acdfd03)
- `AppendHTTPDate` formats the RFC1123 fields directly and only falls back to `time.AppendFormat` for years outside 0..9999. This is used for `Last-Modified`, `Set-Cookie` expiry and the cached `Date` header.

**Client** (87861cd, 05ef200)
- `clientConn` tracks whether a read or write deadline is currently set, so requests without timeouts skip two deadline updates per request instead of clearing deadlines that were never set. The public `ReleaseConn` marks both deadlines as set, so a deadline changed through `cc.Conn()` by external code is still cleared before the next request.

**Server** (8cbdf1d)
- The `Expect: 100-continue` check is reported by `readLimitBody`/`readBodyStream` instead of peeking the header a second time per request.
- The `Server` header value is sanitized once per `Server` and copied into each response instead of being scanned for CR/LF on every response.
- The line terminator offsets a pooled `RequestHeader` keeps are capped at 512; the scanner searches for the rest itself.

## Performance

Methodology: prebuilt test binaries of master (8bb4f7a) and this branch, run interleaved (5 rounds of `-count=2`, n=10 per row) and compared with benchstat. Intel Xeon 2.8 GHz, 4 vCPUs, Linux, Go 1.25.0. The end-to-end and header rows use `-cpu 1` because multi-CPU runs in this container vary by ±30% between runs; the remaining rows use the default parallel mode. Every row shown has p < 0.05.

**Server and client, end to end**

| Benchmark | master | branch | Δ |
|---|---|---|---|
| ServerGet1ReqPerConn | 2.258 µs | 2.071 µs | -8.3% |
| ServerGet10ReqPerConn | 1.229 µs | 1.074 µs | -12.6% |
| ServerGet10KReqPerConn | 1123 ns | 929 ns | -17.3% |
| ServerPost10KReqPerConn | 1.343 µs | 1.154 µs | -14.1% |
| ClientGetEndToEnd10TCP | 13.47 µs | 12.76 µs | -5.3% |
| ClientGetEndToEnd10Inmemory | 4.310 µs | 3.648 µs | -15.4% |
| ClientGetTimeoutFastServer | 3.780 µs | 3.663 µs | -3.1% |
| PipelineClient10 | 3.364 µs | 3.138 µs | -6.7% |

**Headers, URIs and helpers**

| Benchmark | master | branch | Δ |
|---|---|---|---|
| RequestHeaderRead | 775.9 ns | 675.2 ns | -13.0% |
| ResponseHeaderRead | 684.5 ns | 611.7 ns | -10.6% |
| RequestHeaderWrite | 66.56 ns | 49.07 ns | -26.3% |
| ResponseHeaderWrite | 102.8 ns | 59.69 ns | -41.9% |
| NormalizeHeaderKeyCommonCase | 21.76 ns | 19.91 ns | -8.5% |
| URIParsePath | 49.02 ns | 42.61 ns | -13.1% |
| URIParsePathQueryString | 32.60 ns | 25.38 ns | -22.2% |
| URIParseHostname | 50.56 ns | 40.04 ns | -20.8% |
| URIFullURI | 8.584 ns | 7.959 ns | -7.3% |
| CookieParseMin | 39.30 ns | 37.75 ns | -3.9% |
| AppendHTTPDate (ad hoc benchmark, no repo benchmark exists) | 149.1 ns | 37.16 ns | -75.1% |

Unchanged within noise: ArgsParse, CookieParseFull, RequestHeaderPeekBytesNonSpecialHeader, RequestCtxRedirect. Allocations are unchanged: all listed paths remain at 0 allocs/op, and PipelineClient10 drops from 6 to 5 B/op.

Tried and dropped because they did not measure as improvements: a word-at-a-time `removeNewLines` (faster for short values, slower for long ones), a single-pass host delimiter scan, `IndexByte`-based args and cookie scanning, and precomputed `Date`/`Content-Type` header lines.

## Behaviour notes

- Parse results, accepted byte sets and serialized output are identical to master. The one observable text change is that the "malformed mime header line" error quotes the offending physical line instead of the line joined with its obs-fold continuation.
- `RequestHeader.Method()` on a header with no method set stores the default `GET` in the header's own buffer and returns it, the same way `RequestURI()` normalizes in place. Nothing is shared between headers.

## Verification

- `go test ./...` and `go test -race .` pass; golangci-lint v2.13.2 (the CI version) reports no issues.
- All ten fuzz targets ran for 45 s each without findings.
- Differential fuzzing against a renamed copy of master covered request and response header parsing, full request and response reads, header writes through `bufio.Writer`s of random sizes and fill levels, URI parsing, query args, quoting, cookies, dates and header set/peek: 2.5 M random inputs per category (weighted toward CR, LF, tabs, colons, spaces, DEL and high bytes) with zero mismatches.
- New unit tests compare every word-at-a-time helper, the generated host table, the fused key scanner and the date formatter against the previous implementations for all byte values at every position, and cover the two review findings (request-local default method, deadlines cleared after `ReleaseConn`).


gaby:
Findings were found during a performance audit using `Claude Fable 5.1`

gaby:
@erikdubbelboer Let me know of any changes or suggestions.

erikdubbelboer (changes requested):
Thanks for the PR. But I'm sorry, this PR mixes too many independent changes, making it big and hard to review. Can you split it up into smaller pull requests that are easier to review please?

ReneWerner87:
https://github.com/gaby/fasthttp/pull/154 and https://github.com/gaby/fasthttp/pull/153 are the most important PRs because they restore performance to its previous level 
But they should be posted to the upstream fasthttp repo, not the fork.

Issue #2385, opened 2026-09-07
VisitHeaderParams stops at an empty parameter, dropping every parameter after it


ReneWerner87:
`VisitHeaderParams` returns instead of skipping when a `;`-separated element is empty or carries no `=`, so the callback never sees any parameter that follows. RFC 9110 §5.6.6 makes the parameter after a `;` optional:

```
parameters      = *( OWS ";" OWS [ parameter ] )
parameter       = parameter-name "=" parameter-value
```

so `a/b;; c=d` and a trailing `;` are well formed, and an empty element has to be skipped rather than end the scan.

### Reproduce

```go
func params(header string) []string {
	var out []string
	fasthttp.VisitHeaderParams([]byte(header), func(k, v []byte) bool {
		out = append(out, fmt.Sprintf("%s=%s", k, v))
		return true
	})
	return out
}
```

fasthttp v1.73.0, go 1.26:

| header | got | expected |
| --- | --- | --- |
| `text/plain; charset=utf-8` | `[charset=utf-8]` | same |
| `text/plain;; charset=utf-8` | `[]` | `[charset=utf-8]` |
| `text/plain; ; charset=utf-8` | `[]` | `[charset=utf-8]` |
| `text/plain; foo; charset=utf-8` | `[]` | `[charset=utf-8]` |
| `text/plain; foo=; charset=utf-8` | `[]` | `[charset=utf-8]` |
| `multipart/form-data; ; boundary=x` | `[]` | `[boundary=x]` |
| `text/plain; charset=utf-8; foo` | `[charset=utf-8]` | same |

The last row is the same code path terminating harmlessly because the malformed element is last, which is why this is easy to miss.

### Where

`header.go` in v1.73.0, two returns inside the loop end the whole scan where only the current element is at fault:

- L602 `if len(b) == 0 || !validHeaderFieldByte(b[n]) { return }`, reached by an empty element
- L610 `if n >= len(b)-1 || b[n] != '=' { return }`, reached by an element without a value

### Why it matters

The signature gives the caller no way to tell "no more parameters" from "gave up here". `mime.ParseMediaType` is equally strict about these inputs but returns `mime: invalid media parameter`, so a caller can react. A `VisitHeaderParams` caller silently sees a truncated list, and on `multipart/form-data` that means `boundary` disappears.

Advancing to the next `;` in both places instead of returning would match the grammar and stop a strict case from hiding valid parameters behind it. Happy to send a PR if that is the direction you want.

[...]

PR #2388, opened 2026-09-08
Skip RFC-valid empty elements in VisitHeaderParams

vzer200:
Partially addresses #2385: RFC-valid empty parameter elements.

`VisitHeaderParams` currently stops at the second semicolon in `text/plain;; charset=utf-8`, so the callback never receives `charset`. Skip empty elements and accept SP/HTAB as optional whitespace after semicolons, following [RFC 9110 sections 5.6.3 and 5.6.6](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.6).

The documented stop-on-invalid behavior remains: malformed `foo` and `foo=` still stop parsing. Quoted values, callback cancellation and input bytes are preserved.

Validation:

- Added 15 table cases and updated the existing empty-element expectation and cancellation coverage; relevant tests fail before the fix and pass afterward.
- Full `go test -shuffle=on -count=1 ./...` passes. The existing fuzz target passed 1,625,982 executions in 15 seconds.
- Independent review re-ran the full normal suite, targeted race tests and 4,361 additional parameter/boundary/input-preservation checks; all pass.
- The first full race run hit a 200 ms timeout in `TestHostClientMaxConnsWithDeadline`, without a race warning. The same seed passes on both the clean baseline and the patched retry; the timeout was not reproduced.
- gofmt and diff whitespace checks pass; the existing benchmark still reports zero allocations.

AI-assisted implementation, independently reviewed and tested by another coding agent.


erikdubbelboer:
Thanks!

PR #2389, opened 2026-09-08
fix: preserve redirect path normalization setting in Request.CopyTo

cuishuang:
`Request.CopyTo` does not preserve `DisableRedirectPathNormalizing`.

This field controls redirect path normalization in `DoRedirects`. Since `copyToSkipBody()` copies the other request-level behavior flags, omitting this field causes a copied request to behave differently from the source request during redirects.

For example:

```go
var src, dst Request

src.DisableRedirectPathNormalizing = true
src.CopyTo(&dst)
```

Before this change, `dst.DisableRedirectPathNormalizing` remained `false`.

Copy `DisableRedirectPathNormalizing` in `copyToSkipBody()` so `Request.CopyTo` preserves the request's redirect behavior.

### DIFF
--- a/http.go
+++ b/http.go
@@ -1015,6 +1015,7 @@ func (req *Request) copyToSkipBody(dst *Request) {
 	dst.isTLS = req.isTLS
 
 	dst.UseHostHeader = req.UseHostHeader
+	dst.DisableRedirectPathNormalizing = req.DisableRedirectPathNormalizing
 
 	// do not copy multipartForm - it will be automatically
 	// re-created on the first call to MultipartForm.
--- a/http_test.go
+++ b/http_test.go
@@ -147,6 +147,18 @@ func TestRequestCopyTo(t *testing.T) {
 	testRequestCopyTo(t, &req)
 }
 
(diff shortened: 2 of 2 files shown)


erikdubbelboer (changes requested):
Than you for this fix! Can you add this to the existing `TestRequestCopyTo` instead of adding a separate test? Set `DisableRedirectPathNormalizing` to `true` on the populated request and compare the source and destination values in `testRequestCopyTo`. That covers both `false` and `true`. The helper currently only compares the serialized request and body, so it misses this flag.

cuishuang:
> Than you for this fix! Can you add this to the existing `TestRequestCopyTo` instead of adding a separate test? Set `DisableRedirectPathNormalizing` to `true` on the populated request and compare the source and destination values in `testRequestCopyTo`. That covers both `false` and `true`. The helper currently only compares the serialized request and body, so it misses this flag.

@erikdubbelboer Done, thanks! I’ve moved the coverage into TestRequestCopyTo, updated testRequestCopyTo to compare the flag, and squashed everything into a single commit.

erikdubbelboer:
Thanks!

PR #2392, opened 2026-09-10
fix: normalize a trailing /. path segment

sachhg:
## Summary

`normalizePath` does not remove a trailing `/.` segment, so `URI.Path()` returns `/foo/.` where RFC 3986 requires `/foo/`.

RFC 3986 section 5.2.4 step 2B: "if the input buffer begins with a prefix of `/./` or `/.`, where `.` is a complete path segment, then replace that prefix with `/` in the input buffer".

The function already handles `/./` in the middle of a path, `/../` in the middle, and a trailing `/..`. Only the trailing `/.` case is missing.

## Reproduction

```go
var u fasthttp.URI
u.SetPath("/foo/bar/.")
fmt.Printf("%q\n", u.Path()) // "/foo/bar/." , want "/foo/bar/"
```

| input | `URI.Path()` before | RFC 3986 |
| --- | --- | --- |
| `/.` | `/.` | `/` |
| `/aaa/.` | `/aaa/.` | `/aaa/` |
| `/aaa/./.` | `/aaa/.` | `/aaa/` |
| `/aaa/../.` | `/.` | `/` |
| `/aaa%2F.` | `/aaa/.` | `/aaa/` |

I found this by implementing RFC 3986 section 5.2.4 directly and diffing it against `URI.Path()` over generated paths built from the segments `a`, `.`, `..`, `...`, `a.b` and `.hidden` up to depth 4, skipping inputs with `//` so that the duplicate-slash collapsing did not confuse the comparison. Every one of the 31 differences was a path ending in `/.`, with no other class of difference. After the change the same sweep reports 0 differences across 1814 inputs.

## Changes

A trailing `/.` is now dropped, keeping the slash that preceded it, which is what step 2B asks for. It is placed after the existing trailing `/..` handling. That ordering matters: `/aaa/../.` has to lose `/aaa/..` first, and removing a `.` segment can never produce a new `..` segment, so a single pass is enough.

The `.` has to be a complete segment, so the check requires the match to sit at the very end, exactly like the existing trailing `/..` check. `/aaa/...` and `/aaa/.b` are ordinary segments and are left alone.

## Testing

Extended `TestURIPathNormalize` with the trailing dot cases above, plus `/aaa/...` and `/aaa/.b` to pin down that longer dot runs and dotfiles are not touched.

Without the change the test fails with `Unexpected path "/.". Expected "/"`. With it the whole suite passes: all 11 packages, `go vet` clean, `gofmt` clean, and `golangci-lint v2.13.2`, the version pinned in `.github/workflows/lint.yml`, reports 0 issues.

Windows backslash handling is untouched, and `DisablePathNormalizing` still bypasses this path entirely.


uri.go
```diff
@@ -698,6 +698,13 @@ func normalizePath(dst, src []byte) []byte {
 		b = b[:nn+1]
 	}
 
+	// remove trailing /. , which denotes the current directory and so leaves
+	// the trailing slash behind. See RFC 3986 section 5.2.4 step 2B.
+	n = bytes.LastIndex(b, strSlashDot)
```
erikdubbelboer: Can you use `bytes.HasSuffix(b, strSlashDot)` here and then trim the last byte? `bytes.LastIndex` scans the whole path for ordinary filenames containing a dot, even though we only need to check the last two bytes.

I benchmarked `normalizePath` on Go 1.26.4 / M2 Max. `/index.html` goes from about 42 ns to 54 ns, and `/assets/scripts/application.min.js` from 86 ns to 121 ns. With `HasSuffix` these are about 45 ns and 88 ns.

erikdubbelboer (changes requested):

uri.go
sachhg: Done in c4e1850. It now checks `bytes.HasSuffix(b, strSlashDot)` and trims the last byte, so an ordinary path only compares its final two bytes. Thanks for benchmarking it.

erikdubbelboer:
Thanks!

PR #2393, opened 2026-09-11
Strip semicolons in RequestHeader.SetCookie

youdie006:
All eight `Cookie` setters run `removeSemicolons` (`cookie.go:166,174,188,194,246,252,266,272`). `RequestHeader.SetCookie` (`header.go:1679`) is the one cookie setter that does not — `header.go` never mentions the helper.

```go
h.SetCookie("sid", "abc; admin=1")   // emits:  Cookie: sid=abc; admin=1
```

That is two cookies. The caller set one, and the client then disagrees with itself: `h.Cookie("sid")` returns `"abc; admin=1"` with no `admin` cookie, while the same bytes parsed back by `RequestHeader.Read` give `sid="abc"` **plus `admin="1"`**.

`SetCookieBytesK`/`SetCookieBytesKV` delegate here, so this is the only site. `ResponseHeader.SetCookie` takes an already-sanitised `*Cookie` and is fine.

This looks like half of an existing job: #2185 (CRLF) changed `cookie.go` **and** `header.go` — it is what put `initHeaderValueString` into this function — while #2298 (semicolons) changed `cookie.go` only.

<details>
<summary>Verification</summary>

| row | result |
|---|---|
| new test on `master` | FAIL, `unexpected 1 cookie separators` |
| with the fix | PASS |
| sanitise the key only | FAIL |
| sanitise the value only | FAIL |

Run locally on go1.26.3: `go test -shuffle=on ./...`, `go test -race -shuffle=on ./...`, `gofmt -l`, `go vet ./...`, and `golangci-lint run` pinned at the v2.13.2 the lint workflow uses — 0 issues.

`fasthttpproxy/TestDialerGetDialFunc` fails on my machine with `bind: address already in use`; it fails identically on pristine `master` here and that package contains no reference to `SetCookie`.

The test sits beside `TestRequestHeaderSetCookieSanitizesNewLines` and mirrors it, plus a round-trip through `RequestHeader.Read` so it asserts the injected cookie is gone rather than only that the separator is absent.

`removeSemicolons` replaces `;` with a space, so the value becomes `abc  admin=1`. I reused the repo's helper rather than inventing a stricter one; happy to match `net/http` and drop the byte instead if you prefer.

I filed this publicly rather than by the `SECURITY.md` email route because the threat model is already published here — #2298 is a public PR describing the same attack and `cookie.go:490-495` spells it out. Say the word if you would rather it had gone to email.
</details>

---

Disclosure: I used Claude (an AI assistant) while preparing this change. Every result above I ran and verified myself.


erikdubbelboer:
Thanks!

PR #2394, opened 2026-09-11
fix: recheck idle connections before queuing waiters

### LINKED ISSUE #2391: HostClient: lost wakeup between AcquireConn pool check and waiter registration
## Summary

`HostClient.AcquireConn` can miss a returned connection between checking the pool and registering its waiter when `MaxConnWaitTimeout > 0`. The request can then wait until its timeout and return `ErrNoFreeConns` (or `ErrTimeout` when the request timeout overrides the wait timeout), even though a connection is idle.

Verified against current master, commit `8bb4f7a4fdd42017d2d1a7b24d2e0f19df71b8c4`.

## Interleaving

For example, with `MaxConns: 1` and the sole connection currently in use:

1. Goroutine A calls `AcquireConn`, observes no idle connections and a full pool, and releases `connsLock`.
2. Before A calls `queueForIdle`, goroutine B calls `ReleaseConn`. There is no registered waiter yet, so B appends the connection to `c.conns`.
3. A calls `queueForIdle`, which acquires the lock but only appends the waiter; it does not recheck idle connections.
4. A waits on `w.ready` while the connection is already idle. Without a subsequent pool event that delivers a connection, A times out.

The relevant [pool check, unlock, and subsequent waiter registration](https://github.com/valyala/fasthttp/blob/8bb4f7a4fdd42017d2d1a7b24d2e0f19df71b8c4/client.go#L1836-L1945) are separate critical sections. [ReleaseConn](https://github.com/valyala/fasthttp/blob/8bb4f7a4fdd42017d2d1a7b24d2e0f19df71b8c4/client.go#L2103-L2140) can run between them.

There is an analogous capacity case: if an in-use connection closes and `decConnsCount` frees capacity during this gap, the subsequently registered waiter does not initiate a replacement dial despite capacity being available.

This is a synchronization/lost-wakeup race, not necessarily a data race detectable by `go test -race`.

## Deterministic reproducer

Place the following test in the repository root, in package `fasthttp`. It models the legal interleaving at the internal waiter-registration boundary, avoiding timing-dependent goroutine scheduling. It does not drive a complete HTTP request.

```go
package fasthttp

import (
	"testing"
	"time"
)

func TestWaiterRegistrationRechecksIdle(t *testing.T) {
	c := &HostClient{
		MaxConns: 1,
		MaxConnWaitTimeout: time.Second,
		connsCount: 1,
	}
	// AcquireConn has observed no idle connections and a full pool,
	// then unlocked connsLock, but has not called queueForIdle yet.
	cc := &clientConn{}
	c.ReleaseConn(cc)

	// Resume the pending AcquireConn at its waiter-registration step.
	w := &wantConn{ready: make(chan struct{}, 1)}
	c.queueForIdle(w)
	select {
	case <-w.ready:
		if w.conn != cc || w.err != nil {
			t.Fatalf("unexpected delivery: conn=%p err=%v", w.conn, w.err)
		}
	default:
		t.Fatalf("waiter remains blocked despite %d idle connection(s)", len(c.conns))
	}
}
```

Run:

```sh
go test -run '^TestWaiterRegistrationRechecksIdle$' -count=1
```

Observed on the master commit above:

```text
--- FAIL: TestWaiterRegistrationRechecksIdle (0.00s)
    waiter_registration_test.go:28: waiter remains blocked despite 1 idle connection(s)
```

Expected: the pending acquisition should consume the available connection instead of remaining queued.

## Possible fix

Make the decision to wait atomic with the pool check, or recheck both idle connections and available connection capacity under `connsLock` immediately before enqueueing. The latter would deliver an idle connection, reserve capacity and start a dial, or enqueue only when neither is available. Pool strategy and cancellation behavior need to be preserved.

The waiter-registration gap appears to date back to #764. The fixes in #1831 (delivery past a cancelled waiter) and #2266 (failed-waiter bookkeeping) address different races and leave this gap in current master.

HarveyBase:
## Description

`HostClient.AcquireConn` could miss a connection returned between its idle-pool check and waiter registration. The waiter would then remain blocked until `MaxConnWaitTimeout`, even though an idle connection was available.

`queueForIdle` now rechecks the idle pool while holding `connsLock` and delivers an available connection directly to the waiter before queueing it. This closes the lost-wakeup window without expanding the lock around network dialing.

## Testing

- Added a deterministic regression test for the returned-connection/waiter-registration interleaving.
- Repeated the focused test 100 times with `-race`.
- `go test ./...` passed.
- `go vet ./...` passed.
- `gofmt` and `git diff --check` passed.

Fixes #2391


client.go
```diff
@@ -1937,6 +1937,26 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
 func (c *HostClient) queueForIdle(w *wantConn) {
 	c.connsLock.Lock()
 	defer c.connsLock.Unlock()
+	if n := len(c.conns); n > 0 {
```
erikdubbelboer, line 1939: Can we also recheck the available connection capacity here? If the last connection closes between the pool check in `AcquireConn` and waiter registration, `decConnsCount` frees its slot, but there is no idle connection for this check to find. We still queue the waiter without starting a replacement dial, so it can time out even with `connsCount == 0` and `MaxConns == 1`. I reproduced this on both the base and this branch; it is the other case described in #2391. Please reserve the available slot under `connsLock`, start the dial outside the lock, and add a regression test for this case.

erikdubbelboer (changes requested):

client.go
HarveyBase: Addressed in f43dbd0: the slot is now reserved under connsLock and the dial is started outside the lock, with a regression test for the MaxConns=1 case

client.go
```diff
@@ -1936,12 +1936,47 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
...
+	// the connection count, freeing capacity without adding an idle connection.
+	// Reserve the freed slot and dial a replacement connection for w.
+	maxConns := c.MaxConns
+	if maxConns <= 0 {
+		maxConns = DefaultMaxConnsPerHost
+	}
+	if c.connsCount < maxConns {
+		c.connsCount++
+		c.connsLock.Unlock()
+		go c.dialConnFor(w)
```
erikdubbelboer, line 1971: Can we also restart the connection cleaner here if it has stopped? If the last connection closes between the pool check in `AcquireConn` and waiter registration, the cleaner can observe `connsCount == 0` and exit before we reserve this slot. We then dial a replacement without restarting the cleaner, so once returned to the pool it can stay open indefinitely despite `MaxIdleConnDuration`. I reproduced this three times with `-race`. Please set `connsCleanerRun` under `connsLock` and start the cleaner outside the lock, as in `AcquireConn`, and add a regression test for this case.

erikdubbelboer (changes requested):

client.go
HarveyBase: Addressed in 5fb9251: connsCleanerRun is now set under connsLock (with the same connectionClose guard as in AcquireConn, now passed through to queueForIdle) and the cleaner is started outside the lock. Added TestHostClientQueueForIdleRestartsCleaner, which fails on the previous code with the idle connection never closed and passes with -race.

client.go
```diff
@@ -1934,14 +1934,59 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
...
+	// cleaner may have exited after observing connsCount == 0 before the slot
+	// was reserved, so restart it as well.
+	maxConns := c.MaxConns
+	if maxConns <= 0 {
+		maxConns = DefaultMaxConnsPerHost
+	}
+	if c.connsCount < maxConns {
+		c.connsCount++
+		startCleaner := false
+		if !c.connsCleanerRun && !connectionClose {
```
erikdubbelboer, line 1973: Can we also handle the `Connection: close` cancellation case here? Since this replacement dial is asynchronous, the waiter can time out before it completes. `wantConn.cancel` or `dialConnFor` then returns the connection through `ReleaseConn`, but this guard leaves no cleaner running, so the idle connection and `connsCount` can remain indefinitely. I reproduced this consistently under `-race`: after five `MaxIdleConnDuration` intervals the client still had `idle=1` and `cleaner=false`. Please either start the cleaner for this asynchronous path regardless of `connectionClose`, or ensure cancellation closes rather than pools the connection, and add a regression test for this case.

erikdubbelboer (changes requested):
Thanks for addressing the previous cleaner restart issue.

client.go
HarveyBase: Addressed in 6a68def: ReleaseConn now restarts the cleaner whenever a connection is pooled while it is not running, covering this case and the transfer-dial path. 072c941 extends the cancellation regression to connectionClose=true.

PR #2399, opened 2026-09-14
Add opt-in ETag support to FS

anandghegde:
## Summary

- add `FS.GenerateETag` (off by default, like `Compress` and `AcceptByteRange`); `ServeFile`, `ServeFS` and `FSHandler` are unchanged
- when enabled, `FS` sends `ETag: W/"<hex mtime>-<hex size>"` on 200, 206 and HEAD responses
- `If-None-Match` is checked with weak comparison (RFC 9110 §13.1.2). Lists, `*` and both `W/"x"` and `"x"` are supported. A match returns `304 Not Modified`, and the 304 keeps the `ETag` header
- when `If-None-Match` is present, `If-Modified-Since` is ignored (RFC 9110 §13.1.3). If `GenerateETag` is off, requests are handled exactly as before

The main use is static assets behind CDNs and browser caches that revalidate with `If-None-Match`. nginx, Caddy and Go's `http.ServeContent` all support it; `FS` currently only has `Last-Modified`.

## Design notes

- **Weak, per-representation tag.** The size in the tag is the size of the body actually sent. When `Compress` is on, the gzip, br and zstd versions each get their own tag. A variant is only kept if it is smaller than the original, so its tag always differs from the uncompressed one. Two compressed variants could still have the same size and mtime, and mtime only has one-second resolution. So the tag can't promise byte-for-byte equality and has to be weak. nginx makes the same call: it weakens its `"mtime-size"` ETag when it gzips on the fly.
- **No tag when the modification time is unknown.** `embed.FS` and other `fs.FS` implementations with a zero mtime get no `ETag`. A size-only tag would return false 304s after a same-size change.
- The tag is computed once when the file is opened and cached with the `fsFile`, the same way `lastModifiedStr` is, so serving a request adds no allocations.
- The `If-None-Match` parser follows `net/http`'s `checkIfNoneMatch`/`scanETag`. Malformed input stops the scan and counts as no match.

## Usage

```go
fs := &fasthttp.FS{
	Root:         "/var/www",
	Compress:     true,
	GenerateETag: true,
}
h := fs.NewRequestHandler()
```

## Tests

New tests:
- `TestFSGenerateETag`: tag format; exact, strong-form, list and `*` matches; mismatched, malformed and unterminated values; `If-None-Match` taking precedence over `If-Modified-Since` both ways; `ETag` on 304 and 206 responses; a `Range` request with a matching tag gets 304
- `TestFSGenerateETagDisabled`: no `ETag`, `If-None-Match` ignored, `If-Modified-Since` still works
- `TestFSGenerateETagCompress`: the gzip variant gets its own tag, and revalidating it returns 304
- `TestFSGenerateETagFS`: `fs.FS` with an mtime gets a tag; with a zero mtime it gets none

Commands run:
- `go test -count=1 ./...` (Go 1.27.1): all packages pass
- `go test -count=1 -race -run 'TestFS|TestServeFile' .`: pass
- `golangci-lint run ./...` (v2.13.2, same version as CI): 0 issues
- `go vet .`, `gofmt -l .`: clean

Claude Code helped write this change. I reviewed it.

https://claude.ai/code/session_01QBWFWqjbUQGS1c9LsM2bxX


erikdubbelboer (changes requested):
Thanks for the pull request! I found two cases that need fixing:

- A matching `If-None-Match` returns 304 for every request method. I tested POST with a matching tag and PUT with `*`, and both return 304. Only GET and HEAD should return 304; other methods should return 412.
- Only the first `If-None-Match` header is checked. Sending `"other"` in the first header and the current ETag in a second header returns 200, while putting the same values in one comma-separated header returns 304. Please check all occurrences of the header.

Could you add tests for both cases as well?

anandghegde:
Thanks for catching both of these. A matching If-None-Match now returns 304 only for GET and HEAD, and 412 Precondition Failed for any other method. All If-None-Match header lines are checked now, so two separate headers behave the same as one comma-separated header. Added tests for both (POST with a matching tag, PUT with `*`, and the ETag in a second header).

erikdubbelboer (changes requested):
Thanks for fixing the previous two cases! Both now behave correctly.

I found one more parser edge case. `If-None-Match: "bad tag", W/"<current>"` currently matches the current tag and returns 304 for GET/HEAD (or 412 for other methods). The space makes the first entity-tag invalid, so parsing should stop and the condition should count as no match, as described in the PR.

Could you validate the characters inside each entity-tag and make malformed input stop processing the entire combined field, including across repeated `If-None-Match` header lines? Please add a regression test for this case as well.

anandghegde:
Good catch, thanks. Each entity-tag is now checked against the etagc rule from RFC 9110 §8.8.3, and the first malformed element stops the scan of the whole combined field value, including the later `If-None-Match` lines, so the condition counts as no match and the request is served normally.

Tests cover your case (`"bad tag", W/"<current>"`), a bad tag after a matching one, control and DEL characters, missing and unterminated quotes, and a malformed tag in a second header line; the valid list, `*` and multi-header cases still return 304.

PR #2401, opened 2026-09-15
Return no values from PeekAll for absent special headers

youdie006:
`peekAll` guards `Host`, `Content-Type` and `User-Agent` with `len(...) > 0` (`header.go:2037-2048`), and the response side guards `Content-Type`, `Content-Encoding` and `Server` the same way (`:2085-2096`). The `Content-Length`, `Cookie`/`Set-Cookie` and `Trailer` arms appended unconditionally (`:2055`, `:2059`, `:2064`, `:2103`, `:2106`, `:2108`), so `PeekAll` returned one empty value for a header the message never carried:

```go
var req fasthttp.RequestHeader
req.SetHost("example.com")

req.PeekAll("Content-Length")  // [""]  len 1
req.Peek("Content-Length")     // ""
// All() does not yield it at all
```

`All()` already guards those same fields (`header.go:1099`, `:1120`, `:1258`, `:1270`), and so does the write path (`:2526`, `:2546`, `:2660`, `:2679`), so `Peek`, `PeekAll` and `All` gave three different answers about whether the header was there. `len(h.PeekAll(k)) > 0` as a presence test is wrong for exactly these headers, and a loop that re-emits `PeekAll` results writes a bogus `Content-Length: ` or `Set-Cookie: ` line.

This adds the same `len(...) > 0` guard at the six sites.

<details>
<summary>Verification</summary>

Go 1.26.3. `md5sum header.go` on every row; the test file is byte-identical across rows.

`go test -count=1 -run 'TestRequestHeaderPeekAll|TestResponseHeaderPeekAll' .`

| `header.go` | md5 | result |
| --- | --- | --- |
| this PR | `9a06db9c` | pass |
| pristine `HEAD` | `0d6420d7` | `Unexpected size for key "Content-Length": 1. Expected 0` (both tests) |
| revert request `Content-Length` guard | `005947c0` | fails on `"Content-Length"` |
| revert request `Trailer` guard | `66dbcb2b` | fails on `"Trailer"` |
| revert request `Cookie` guard | `2f06ef61` | fails on `"Cookie"` |
| revert response `Content-Length` guard | `e94f29f2` | fails on `"Content-Length"` |
| revert response `Set-Cookie` guard | `4b65d559` | fails on `"Set-Cookie"` |
| revert response `Trailer` guard | `e495a10e` | fails on `"Trailer"` |
| request `Content-Length` `> 0` to `> 1` | `042ad3e3` | `size 0. Expected 1` |
| response `Trailer` `> 0` to `> 1` | `74be824c` | `size 0. Expected 1` |

Each revert row removes one guard and fails on that header alone, so the six sites are independently required. The two over-correction rows fail in the opposite direction, which is what the single-element rows (`Content-Length: 0`, a one-entry `Trailer`) are there for.

The existing tests already assert an empty `PeekAll` after `Del` for `Content-Type`, `Host` and `Content-Encoding` (`header_test.go:3910-3915`, `:3966-3969`); this extends that to the three headers it stopped short of.

Also run, as `.github/workflows` runs them: `go test -shuffle=on ./...` and `go test -race -shuffle=on ./...` (10 packages ok), `gofmt -l` clean, `go vet ./...` clean, `golangci-lint v2.13.2` (the pinned version) reports 0 issues.

Two things I could not do here: `fasthttpproxy TestDialerGetDialFunc` fails with `listen tcp :8001: bind: address already in use` because something else on this machine holds that port; it fails identically on pristine `header.go` and `header_test.go`, and this change does not touch that package. And my local toolchain is 1.26.3 only, so I did not run the `1.25.x` row of the matrix; the diff is `len()` and `if` only.

</details>

Disclosure: this change was written with AI assistance (Claude). The measurements and mutants above were reproduced and checked by me before opening this PR.


erikdubbelboer:
Nice find. Thanks!

Issue #2234, opened 2026-05-27
bug: Shutdown sets s.done = nil, making RequestCtx.Err() block forever


pageton:
## Description

After `ShutdownWithContext` completes and all connections drain (`open == 0`), it sets `s.done = nil` at `server.go:2039`. Any surviving `RequestCtx` (e.g., from hijacked connections, or contexts captured in background goroutines) that calls `Err()` will hit:

```go
func (ctx *RequestCtx) Err() error {
    select {
    case <-ctx.Done():  // Done() returns nil channel → blocks forever
        return context.Canceled
    default:
        return nil
    }
}
```

`ctx.Done()` returns `ctx.s.done` which is now `nil`. Receiving from a nil channel blocks forever.

Similarly, `Done()` itself returns a nil channel, which violates the `context.Context` contract (should return a closed channel after cancellation).

## Impact

- Goroutines holding a reference to a `RequestCtx` after shutdown will block forever in `Err()` or `Done()`
- Hijacked connections that call `ctx.Err()` after server shutdown will hang
- Any code using `RequestCtx` as a `context.Context` after shutdown is broken

## Suggested Fix

Instead of setting `s.done = nil`, close it and leave it non-nil:

```go
if open := s.open.Load(); open == 0 {
    close(s.done)
    return lnerr
}
```

Or guard `Done()` and `Err()` against nil:

```go
func (ctx *RequestCtx) Done() <-chan struct{} {
    if ctx.s == nil || ctx.s.done == nil {
        ch := make(chan struct{})
        close(ch)
        return ch
    }
    return ctx.s.done
}
```

## File

`server.go:2039`, `server.go:2932-2952`

erikdubbelboer:
Closing as not a blocking bug. `Err()` uses a `select` with `default`, so a nil `Done()` channel does not block; `s.done = nil` also supports server reuse after shutdown completes.

PR #1908, opened 2024-11-25
fix: ShutdownWithContext and ctx.Done() exist race.

r27153733:
fix #1907 #1906

PR #2402, opened 2026-09-16
fix: synchronize RequestCtx.Done() with ShutdownWithContext

cpsc:
`RequestCtx.Done()` reads `Server.done` without holding `s.mu`, and `ShutdownWithContext` sets it to nil once the last connection has drained (`server.go:2105`). #1908 made that write wait for `open == 0`, which covers reads made while a handler is running. It does not cover reads made after the handler returns, which `RequestCtx.reset()` explicitly allows for `Done()` and `Err()`.

Those reads are common. Passing the `RequestCtx` as a `context.Context` to `database/sql` starts `Rows.awaitDone`, which selects on `ctx.Done()` and calls `ctx.Err()`; `context.WithCancel(ctx)` starts a goroutine that does the same. Neither goroutine is part of the connection count, so nothing orders its read against the shutdown write. #2340 lists this race under "deliberately left open".

The new test in `server_race_test.go` hands `ctx.Done()` to a goroutine that waits for it and then calls `ctx.Err()`, returns from the handler, and shuts the server down. The goroutine wakes up from the close, but its `Err()` has no ordering against the nil write that follows, so the race is there regardless of scheduling. On master it fails in every run (30 of 30 separate processes; the detector reports a given race once per process, so `-count` does not help):

    WARNING: DATA RACE
    Write at 0x00c0001fe830 by goroutine 12:
      github.com/valyala/fasthttp.(*Server).ShutdownWithContext()
          server.go:2105
      github.com/valyala/fasthttp.(*Server).Shutdown()
          server.go:2062
    Previous read at 0x00c0001fe830 by goroutine 16:
      github.com/valyala/fasthttp.(*RequestCtx).Done()
          server.go:3028
      github.com/valyala/fasthttp.(*RequestCtx).Err()
          server.go:3042

The fix stores `done` in an `atomic.Pointer`, like `serverDate` in `header.go`. Writes still happen under `s.mu`. `Done()` has to stay lock-free: `ShutdownWithContext` holds `s.mu` while it waits for handlers, so a handler that checked `ctx.Done()` would block on the lock and `Shutdown()` would never return.

The values `Done()` returns are unchanged: the open channel while serving, the closed channel during shutdown, nil after a shutdown that drained. The nil stays, since #2234 kept it to support reusing the server. Not setting it to nil would move the write into the next `Serve`, and replacing the channel there is what the review of #2340 objected to. The other items #2340 left open need a done channel per request or per `Serve`, and this does not attempt that. `RequestCtx.Err()` benchmarks the same before and after (about 5.3 ns/op, 0 allocs).

We ran into this through Bifrost, which passes the `RequestCtx` to GORM in its HTTP handlers and gets this report from the race detector when the server shuts down.

`go test -race -shuffle=on ./...` and `go test ./...` pass, golangci-lint v2.13.2 reports 0 issues, and the new test passes in 30 of 30 processes with the fix.


erikdubbelboer:
Thanks!

PR #2403, opened 2026-09-16
fix: fill in the request time where it is left at zero

ReneWerner87:
## Why

Two paths leave `RequestCtx.time` at the zero value, so `Time()` reports year 1:

- `Init2`, and with it `Init`, sets `connTime` but not `time`. A ctx built by hand, the way `fasthttpadaptor` and tests build one, reports year 1.
- When a `TimeoutHandler` fires, the request loop swaps in a fresh ctx because the old one still belongs to the handler goroutine. That fresh ctx keeps a zero `time` and `connTime`, so `idleConnTime.Store(ctx.time.Unix())` stores year 1 for the connection and `ConnTime()` of every later request on it is year 1 as well.

The idle timestamp is the one that bites: `closeIdleConns` closes a connection whose stored second is in the past (`t != 0 && now-t >= 0`), so a graceful shutdown drops a connection that finished a request a millisecond ago.

## What changes

- `Init2` sets `time` next to `connTime`, from the same clock read.
- The request loop keeps the request time in a local, hands it and `connTime` to the replacement ctx after a timeout, and stores that local in the idle bookkeeping instead of reading it back from a ctx that may have been replaced.

`TestRequestCtxInit` gains a time assertion and `TestTimeoutHandlerKeepsRequestTime` covers the timeout path. Both fail on master.


ReneWerner87:
Folded into #2404, which carries this commit plus the opt in Server.LazyRequestTime and one description for both. Closing here to keep the review in one place.

PR #2404, opened 2026-09-16
fix: request time left at zero, and an opt in Server.LazyRequestTime

ReneWerner87:
Two commits: a fix, and an opt in option on top of it.

Both come out of the work behind https://gofiber.github.io/benchmarks/, where we track the per request cost of Fiber v2 and v3. The CPU profile of a small keep alive request sent us into the server loop, `time.runtimeNow` turned out to be about 9% of the request, and reading the code around it turned up two spots where the request time is never filled in.

## Fix: the request time is left at zero

Two paths leave `RequestCtx.time` at the zero value, so `Time()` reports year 1:

- `Init2`, and with it `Init`, sets `connTime` but not `time`. A ctx built by hand, the way `fasthttpadaptor` and tests build one, reports year 1.
- When a `TimeoutHandler` fires, the request loop swaps in a fresh ctx because the old one still belongs to the handler goroutine. That fresh ctx keeps a zero `time` and `connTime`, so `idleConnTime.Store(ctx.time.Unix())` stores year 1 for the connection and `ConnTime()` of every later request on it is year 1 as well.

The idle timestamp is the one that bites: `closeIdleConns` closes a connection whose stored second is in the past (`t != 0 && now-t >= 0`), so a graceful shutdown drops a connection that finished a request a millisecond ago.

What changes:

- `Init2` sets `time` next to `connTime`, from the same clock read.
- The request loop keeps the request time in a local, hands it and `connTime` to the replacement ctx after a timeout, and stores that local in the idle bookkeeping instead of reading it back from a ctx that may have been replaced.

`TestRequestCtxInit` gains a time assertion and `TestTimeoutHandlerKeepsRequestTime` covers the timeout path. Both fail on master. The fix itself is free and costs nothing: 357.8 against 358.5 ns, -0.7 [-6.9, +5.6].

## Option: Server.LazyRequestTime

`serveConnCounted` fills `ctx.time` from `time.Now()` for every request. Two places use it: `RequestCtx.Time()`, a convenience for handlers, and the idle bookkeeping, which only needs whole seconds. A handler that never asks for the time pays the clock read anyway.

A new opt in field, `Server.LazyRequestTime`:

- `Time()` reads the clock on its first call within a request and keeps that value.
- The idle bookkeeping takes its second from the goroutine that already refreshes the server date once per second. The server starts that refresher when the option is set, so the coarse second is never zero.
- The request loop leaves `ctx.time` empty instead of filling it from the clock.

The default path is untouched, `Time()` included: the lazy read only happens for a ctx whose server has the option set.

With the option, `Time()` reports the moment of its first call rather than the moment the server started reading the request, and it must not be called from several goroutines at once. `TestServerLazyRequestTime` covers the value, the idle timestamp and the untouched default.

## Numbers

`Server.ServeConn`, one small keep alive request per iteration, go1.27. The same code is linked in many function layouts (`-ldflags=-randlayout`), because the layout alone moves this benchmark by tens of nanoseconds; every figure is the mean over layouts with a Welch 95% interval. `Request.Read` alone is the control, nothing in this change can reach it.

| base | machine | serve_conn | control |
| --- | --- | --- | --- |
| master | AMD EPYC, 40 layouts | 334.9 to 312.1 ns, -22.9 [-26.6, -19.1], -6.8% | +1.5 [-0.4, +3.4] |
| #2387 | AMD EPYC, 40 layouts | 305.7 to 269.6 ns, -36.1 [-43.9, -28.3], -11.8% | +1.4 [-3.5, +6.3] |
| #2387 | Apple M2 Pro, 8 layouts | 304.5 to 275.2 ns, -29.2 [-31.7, -26.8], -9.6% | -0.4 [-3.6, +2.8] |

The benchmark handler never calls `Time()`, which is the case the option is for. A handler that does call it pays one clock read, as today.


server.go
```diff
@@ -378,6 +378,17 @@ type Server struct {
...
 
+	// Reads the wall clock only when a handler asks for the request time
+	// instead of once per request, if set to true.
+	//
+	// RequestCtx.Time() then reports the moment of its first call rather
+	// than the moment the server started reading the request, and it must
+	// not be called from several goroutines at once. The server itself
+	// keeps its idle bookkeeping on a clock refreshed once per second.
+	//
+	// Reading the clock per request is the default.
```
erikdubbelboer, line 389: Could we use this wording to match the surrounding field comments more closely? The default timestamp is recorded immediately before calling the handler, not when the server starts reading the request. Also, connection setup and deadlines still read the clock, and RequestCtx already documents its concurrency restriction.

```suggestion
	// Defers recording the request time until the first call to
	// RequestCtx.Time() if set to true.
	//
	// RequestCtx.Time() returns the time of that first call and returns
	// the same value on subsequent calls during the request.
	// Idle connection tracking uses a clock refreshed once per second.
	//
	// By default the request time is recorded immediately before
	// calling the request handler.
```

erikdubbelboer (changes requested):
thanks for the pull! Could you adjust the LazyRequestTime documentation as suggested below?

server.go
ReneWerner87: ok sure, thanks for this hint, will do it in a minute

ReneWerner87: Applied your wording as is in e2b51e7.

ReneWerner87:
@erikdubbelboer fixed

erikdubbelboer:
Thanks!

PR #2405, opened 2026-09-17
fix: preserve zero Max-Age when parsing cookies

jakezwang:
Parsing `Max-Age=0` currently stores the same value used for an absent attribute, so serializing the cookie drops the deletion instruction and may fall back to `Expires`.

Store an explicit zero as the existing negative MaxAge sentinel. This preserves `max-age=0` across parsing and serialization and keeps its precedence over `Expires`, as specified in [RFC 6265](https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2).

Tests cover zero, attribute casing, Expires, duplicate attributes and reuse without Max-Age.

Validation: `go test -shuffle=on ./...`, `go test -race -shuffle=on ./...`, `go vet ./...`, and `golangci-lint run` (v2.13.2).


erikdubbelboer:
Thanks!

PR #2406, opened 2026-09-19
client: recheck idle pool before queueing waiters

### LINKED ISSUE #2391: HostClient: lost wakeup between AcquireConn pool check and waiter registration

team-humaki:
AcquireConn dropped `connsLock` after seeing an empty/full pool, then called `queueForIdle`. A `ReleaseConn` (or `decConnsCount`) in that gap left the waiter blocked until timeout even though a connection was idle or capacity was free.

`queueForIdle` now rechecks under the lock: deliver an idle conn, reserve capacity and dial, or enqueue only when neither is available.

Fixes #2391


client.go
```diff
@@ -1945,6 +1945,40 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
...
+			return
+		}
+	}
+	maxConns := c.MaxConns
+	if maxConns <= 0 {
+		maxConns = DefaultMaxConnsPerHost
+	}
+	if c.connsCount < maxConns {
+		c.connsCount++
+		go c.dialConnFor(w)
```
erikdubbelboer, line 1991: Can you also restart the connection cleaner here when needed, like `AcquireConn` does?

During the gap before `queueForIdle`, the last connection can close and the cleaner can see `connsCount == 0` and exit. This path then creates a replacement connection without starting the cleaner again. Once released, that connection can stay idle indefinitely despite `MaxIdleConnDuration`.

I reproduced this by closing the last connection, letting the cleaner exit, then calling `queueForIdle` and releasing the replacement connection. It remained in the idle pool after its idle timeout.

erikdubbelboer (changes requested):
Thanks for the pull request! There is one case where the new dial path needs to restart the idle connection cleaner.

client.go
team-humaki: Thanks — queueForIdle now restarts connsCleaner on the replacement dial path the same way AcquireConn does, with a regression test that closes the last conn, lets the cleaner exit, dials a replacement, and checks the idle conn is collected after MaxIdleConnDuration.

client.go
```diff
@@ -1945,6 +1945,51 @@ func (c *HostClient) AcquireConn(reqTimeout time.Duration, connectionClose bool)
...
 	c.connsLock.Lock()
 	defer c.connsLock.Unlock()
+
+	// Recheck under the lock. AcquireConn drops connsLock after seeing an
+	// empty/full pool; ReleaseConn or decConnsCount can run in that gap
+	// and leave an idle conn (or free capacity) that a waiter would
+	// otherwise miss until timeout.
+	if n := len(c.conns); n > 0 {
+		var cc *clientConn
+		switch c.ConnPoolStrategy {
```
erikdubbelboer, line 1955: Could you preserve the existing invalid-strategy error here? If an idle connection appears during the registration gap while `ConnPoolStrategy` is unsupported, this switch leaves `cc` nil and falls through. Since `connsCount` is still at the limit, the waiter is queued and eventually returns a timeout instead of `ErrConnPoolStrategyNotImpl`. I reproduced this deterministically with `ConnPoolStrategyType(100)`, one idle connection, and `connsCount == MaxConns`. A `default` case that delivers `ErrConnPoolStrategyNotImpl` to the waiter and returns would keep the behavior consistent with `AcquireConn`.

erikdubbelboer (changes requested):
Thanks for addressing the cleaner lifecycle issue. One edge case remains in the new idle-pool recheck.

client.go
team-humaki: Thanks — queueForIdle now has a default case that delivers ErrConnPoolStrategyNotImpl to the waiter (same as AcquireConn) instead of falling through to a wait/timeout. Added a regression test with ConnPoolStrategyType(100), one idle conn, and a full pool.

team-humaki:
Follow-up for the invalid-strategy path: queueForIdle now returns ErrConnPoolStrategyNotImpl when an idle conn appears during the registration gap and ConnPoolStrategy is unsupported, matching AcquireConn. Test covers ConnPoolStrategyType(100) + idle conn + MaxConns full.

PR #2411, opened 2026-09-21
perf: memoize authority parsing and fast-path the root path

h2zi:
`URI.parse` validates and unescapes the authority on every request, although on a keep-alive connection it is the same `Host` bytes each time, and it sends the path `/` through the general normalizer.

- The URI keeps the last authority it parsed and the result. A repeat skips `parseHost`; anything else is parsed as before and replaces the memo. The key is the bytes as passed, taken before `parseHost` rewrites its input in place (the argument may alias `u.Host()`), dropped again on a failed parse, and not kept at all above 256 bytes, so an implausibly long `Host` is not retained twice.
- A request-target of `/` skips path normalization.
- A target with a single leading slash is origin-form whatever it contains, so the `://` search that decides whether to split a scheme and host off the target only runs for the other forms; `//host/path` still splits as before.

| benchmark | master | this PR | Δ |
| --- | ---: | ---: | ---: |
| MeasURIParseRootKeepAlive | 57.200n ± 11% | 5.649n ± 5% | -90.12% (p=0.002) |
| URIParseHostname | 16.86n ± 23% | 12.58n ± 72% | ~ |
| URIParsePath | 11.550n ± 27% | 7.955n ± 50% | -31.13% (p=0.026) |
| URIParsePathQueryString | 12.675n ± 22% | 7.463n ± 56% | -41.12% (p=0.004) |
| ServeWrkPipeline64 | 270.3n ± 3% | 211.8n ± 1% | -21.64% (p=0.002) |
| ServeCurlPipeline16 | 669.0n ± 2% | 624.0n ± 4% | -6.73% (p=0.002) |
| ServerGet10KReqPerConn | 225.7n ± 9% | 218.1n ± 8% | ~ |
| ServerGet1ReqPerConn | 693.8n ± 3% | 698.5n ± 6% | ~ |

Geomean over the twelve upstream `BenchmarkServerGet*`/`BenchmarkServerPost*` benchmarks: 388.5n → 378.7n (-2.52%).

Apple M4, Go 1.27, `go test -c` binaries of master and this branch run alternately for 6 rounds, `benchstat`; ~ means no significant difference at p < 0.05. `BenchmarkServe*` are the in-memory serve benchmarks from #2409; `Meas*` are one-off microbenchmarks described below the table.

`MeasURIParseRootKeepAlive` calls `Parse(host, "/")` on one `URI` in a loop, the keep-alive steady state.

No API or behavior change: every `(host, uri)` pair in a 440-case corpus and every request in a 6,700-request corpus parses to the same fields as on master, and a differential fuzzer over both parsers ran clean. Tests cover the memo bound, the aliased argument and the origin-form decision.


ReneWerner87:
nice improvement

PR #2416, opened 2026-09-21
fix: accept a negative Max-Age when parsing cookies

SulimanAbdulrazzaq:
`Cookie.ParseBytes` reads `Max-Age` with `ParseUint`, which rejects a leading `-`, so a `Set-Cookie` value such as `sid=; Max-Age=-1; Path=/; HttpOnly` fails with `unexpected trailing char found: expecting 0-9`.

`ResponseHeader.Cookie` ignores that error and still returns true, so a client gets the cookie back without its Max-Age (a session cookie instead of a deletion) and without the attributes that follow it. Output of a small program that reads a response carrying that header:

```
before: found=true MaxAge()=0 Path()="" HTTPOnly()=false String()="sid="
after:  found=true MaxAge()=-1 Path()="/" HTTPOnly()=true String()="sid=; max-age=0; path=/; HttpOnly"
```

[RFC 6265, section 5.2.2](https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2) allows the value to start with `-` and says a Max-Age of zero or less expires the cookie immediately. Since #2405 a parsed `Max-Age=0` is stored as the `-1` "delete now" value; this does the same for negative values, as net/http's `readSetCookies` does for any value `<= 0`. Such a cookie serializes as `max-age=0`, like `SetMaxAge(-1)` does since #1902. The parser is otherwise as strict as before: `-`, `--1` and `-1s` are still rejected.

`TestCookieParseNegativeMaxAge` fails on master (all seven cases return the error above) and passes with this change. `go test -shuffle=on ./...` and `go vet ./...` pass with Go 1.27.1, and golangci-lint v2.13.2 reports nothing for the changed files (its two ST1000 findings in the Windows-only `reuseport` files appear on master as well; the checks were run on Windows).


SulimanAbdulrazzaq:
@erikdubbelboer when you have a moment, could you approve the workflow runs and take a look? `Cookie.ParseBytes` rejects a negative `Max-Age` such as `Max-Age=-1`, so the client gets the cookie back without its Max-Age and without the attributes after it, which turns a deletion into a session cookie. The Lint, Test and CIFuzz workflows are waiting for approval, and the branch has no conflicts.

